PR #106, opened 2017-08-24
build: Lower minVersion to 6.10

mattdsteele:
Hi folks!

I'd love to try this out in our company, but we're stuck on Node 6.10.2. And it takes about a month for us to upgrade runtimes due to the way we package, test, and distribute things.

On a whim I tried lowering the minimum version to 6.10, and it seems to be building the component fine.

### DIFF
--- a/bin/stencil
+++ b/bin/stencil
@@ -10,7 +10,7 @@ process.env.IONIC_CLI_BIN = __filename;
 var util = require('./util');
 var path = require('path');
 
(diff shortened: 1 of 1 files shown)


adamdbradley:
Thanks for the PR! I instead added an option to just skip the node version check entirely. I imagine others may run into this issue at various versions, but would still like to lock it down to support LTS, even though it probably works in versions before that.

PR #118, opened 2017-08-30
Fixed event name: onDoubleClick -> onDblClick, fixes #114

### LINKED ISSUE #114: onDoubleClick does not trigger event
Event onDoubleClick does not seem to fire.

To reproduce this I created a simple component with two buttons - one with `onClick` one with `onDoubleClick` - the former event fires, the latter doesn't.

Here is the code:

    @Component({
        tag: 'my-button'
    })
    export class MyButton {

        private onClick(event: MouseEvent) {
            console.log('onClick', event);
        }

        private onDoubleClick(event: MouseEvent) {
            console.log('onDoubleClick', event);
        }

        render() {
            return <div>
                <button onClick={(ev) => this.onClick(ev)}>Click</button>
                <button onDoubleClick={(ev) => this.onDoubleClick(ev)}>DoubleClick</button>
            </div>
        }
    }

derkoe:
The event name is "dblclick" not "doubleclick" - this should fix #114

### DIFF
--- a/src/util/jsx-interfaces.ts
+++ b/src/util/jsx-interfaces.ts
@@ -895,8 +895,8 @@ declare global {
       onClickCapture?: EventHandler<MouseEvent>;
       onContextMenu?: EventHandler<MouseEvent>;
       onContextMenuCapture?: EventHandler<MouseEvent>;
-      onDoubleClick?: EventHandler<MouseEvent>;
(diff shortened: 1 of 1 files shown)


adamdbradley:
Awesome, good catch! Thanks for the PR!

PR #123, opened 2017-09-02
feat(testing): layout the basic testing API

kensodemann:
This is just the start of the utilities. I am still relying too heavily on the existing mock* code from test/index.ts. That should probably get moved and cleaned up, but for now I am using a lot of it.

Two functions:
register() - creates a platform, registers the defined components
render() - can be used to test the DOM or (via $instance) the component class

I removed the idea of a compile() function for now since render() returns an object that defines the component class instance on $instance.

Next steps:

1. determine how to use with component defined in tsx file
2. refactor more test/index.ts code into here

### DIFF
--- a/src/core/testing/test/util.spec.ts
+++ b/src/core/testing/test/util.spec.ts
@@ -1,0 +1,101 @@
+import { h } from '../../../core/renderer/h';
+import * as util from '../util';
+
+class IonTest {
+  render() {
+    return [
+      h('div', 0, 'hi')
+    ];
+  }
+}
+
+class IonThingOne {
+  render() {
+    return [
+      h('div', 0, 'thing')
+    ];
+  }
+}
+
+class IonThingTwo {
+  x = 18;
+  y = 24;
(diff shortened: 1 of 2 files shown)


src/core/testing/test/util.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+      const plt = util.register();
+      expect(plt).toBeTruthy();
+      expect(typeof plt.defineComponent).toEqual('function');
+      expect(typeof plt.getComponentMeta).toEqual('function');
+      expect(typeof plt.getContextItem).toEqual('function');
+    });
+
+    it('registers the components', () => {
+      const plt = util.register(
+        [{
```
kensodemann: I am hoping to not have to specify the `tagNameMeta` in cases where that is defined in the `@Component` decorator in a `.tsx` file, but I still need to look into how that is done.

adamdbradley:
awesome thanks!

Issue #124, opened 2017-09-03
autoFocus and autoComplete are not rendered as attribute


derkoe:
When you create an input element with autoComplete or autoFocus the attributes are not added to the DOM.

The following example:

    <input class="new-todo" placeholder="What needs to be done?" autoComplete="on" autoFocus
       onKeyUp={event => this.onKeyUp(event)} />

results with this in the DOM:

    <input placeholder="What needs to be done?" class="new-todo" ssrc="0.1.">

jgw96:
Thanks for opening a PR with us! We will review.

adamdbradley:
Thanks for the PR!!!

PR #125, opened 2017-09-03
Handle autoComplete and autoFocus as attribute

derkoe:
This fixes the issue #124 

I'm not sure if this is the right way to fix it. There's also the possibility to change the logic in isAttr in jsx-to-[vnode.ts](https://github.com/ionic-team/stencil/blob/master/src/compiler/transpile/transformers/jsx-to-vnode.ts#L283)


adamdbradley:
awesome thanks!!

PR #132, opened 2017-09-05
fix(init): check for activelyLoadingChildren before trying to splice it

jgw96:
Not entirely sure this is the correct fix as I am not familiar with this part of the codebase but it does make sense to me that we might have an ancestorHostElement but it might not have `_activelyLoadingChildren`, like when all the components have loaded.


adamdbradley:
Had to fix this another way, thanks

PR #142, opened 2017-09-10
refactor(testing): use typescript hosted tsconfig.json parser

vvakame:
I think We should use tsconfig.json parser that hosted by typescript.
tsconfig.json is not json. It can write a comment, It has a `extend` feature.

### DIFF
--- a/src/tsconfig.json
+++ b/src/tsconfig.json
@@ -1,26 +1,63 @@
 {
   "compilerOptions": {
-    "alwaysStrict": true,
-    "allowSyntheticDefaultImports": true,
-    "allowUnreachableCode": false,
-    "declaration": true,
-    "experimentalDecorators": true,
-    "forceConsistentCasingInFileNames": true,
-    "jsx": "react",
-    "jsxFactory": "h",
-    "lib": [
+    /* Basic Options */
+    "target": "es2015",                       /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', or 'ESNEXT'. */
+    "module": "es2015",                       /* Specify module code generation: 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */
+    "lib": [                                  /* Specify library files to be included in the compilation:  */
       "dom",
       "es2015"
     ],
-    "module": "es2015",
-    "moduleResolution": "node",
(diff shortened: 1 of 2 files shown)


vvakame:
I overwrite src/tsconfig.json, but options is kept everything.
I did `tsc --init` and copy values.
`npm run build && npm run test` is passed in my local environment.

src/tsconfig.json
```diff
@@ -1,26 +1,63 @@
...
+    // "removeComments": true,                /* Do not emit comments to output. */
+    // "noEmit": true,                        /* Do not emit outputs. */
+    // "importHelpers": true,                 /* Import emit helpers from 'tslib'. */
+    // "downlevelIteration": true,            /* Provide full support for iterables in 'for-of', spread, and destructuring when targeting 'ES5' or 'ES3'. */
+    // "isolatedModules": true,               /* Transpile each file as a separate module (similar to 'ts.transpileModule'). */
+
+    /* Strict Type-Checking Options */
+    // "strict": true,                        /* Enable all strict type-checking options. */
+    "noImplicitAny": true,                    /* Raise error on expressions and declarations with an implied 'any' type. */
+    // "strictNullChecks": true,              /* Enable strict null checks. */
```
vvakame: BTW, I strongly recommended to turn on this options.
`tsc --init` generate `"strict": true` option by default.

kensodemann:
@vvakame - thank you very much for this pull request.

Can you provide some information on what advantages implementing this has over what we currently have implemented? We would like to fully understand the advantages of this over what we currently have before deciding if we should implement it.

Thank-you!

vvakame:
Now, `scripts/jest.preprocessor.js` doesn't handle tsconfig.json just like a tsc command.
tsconfig.json is not json. It can write a comment, It has a `extend` feature.

This changes is not a important to this project directly.
But I think this code is important to this project.
`stencil` command is not read tsconfig.json file directly now. https://github.com/ionic-team/stencil/blob/4b84ad22ac8ed7cd9356cff1435a4e35e0c9a31d/src/compiler/transpile/compiler-options.ts#L5-L29
This behavior will confusing about TypeScript users. This code snippet can uses for other part of this project. 😉

kensodemann:
Thank you for submitting this pull request. It is greatly appreciated.

 After discussing this change internally, we have decided to close this PR without merging the change at this time.

PR #153, opened 2017-09-13
feat(testing): expose the testing utilties

kensodemann:
This exposes the `transpile`, `render`, and `register` in `@stencil/core`.

Ideally, I would like to have tests import `@stencil/core/testing`

Also, I am running into a bit of a wall with rendering when the class has already been transpiled. I can get this to render:

```ts
class IonTest {
  render() {
    return [
      h('div', 0, 'hi')
    ];
  }
}
```

but not this:

```ts
var IonTestTwo = (function() {
  function IonTestTwo() { }
  IonTestTwo.prototype.render = function() {
    return [
      h('div', 0, 'hi')
    ];
  };
  return IonTestTwo;
});
```

But I want to get this much in as a baby step before I mess it up. This seems so close...


adamdbradley:
Also add these new files to package.json `files`

PR #158, opened 2017-09-15
fix(proxy): proxy all method calls

manucorporat:
So, still using ES6 proxies, so @adamdbradley it is up to you to merge it... I have figured out that we don't even need this PR to make `ion-menu` to work... (`getInstance()`) has been replaced with `componentOnRead()`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2817,13 +2817,6 @@
             }
           }
         },
-        "string_decoder": {
-          "version": "1.0.1",
-          "bundled": true,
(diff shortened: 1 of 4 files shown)


adamdbradley:
Holding off on using JS proxy for the time being. I do think this will be the end solution, but since it can't be polyfilled we'll have to wait a little bit longer to use it mainstream when all the browsers support it. (which is looking great, it's already debatable if we can use JS proxy today)

PR #161, opened 2017-09-16
feat(testing): move the testing utilities into their own module

kensodemann:
This changes things so that the testing utilities are pulled in via import statements as such:

```ts
import { register, render } from '@stencil/core/testing';
```

And the filter ends up looking like this:

```JavaScript
const stencil = require('@stencil/core/testing');

module.exports = {
  process(src, path) {
    if (path.endsWith('.tsx')) {
      return stencil.transpile(path, './src');
    }
    return src;
  }
};
```


kensodemann:
The global `h()` issue is still kicking my ass though... :(

src/testing/util.ts
```diff
@@ -19,9 +19,10 @@ export function render(platform: PlatformApi, html: string): Promise<HostElement
   return waitForLoad(platform, node);
 }
```
kensodemann: Transpiler was throwing errors about rootDir without this being set.

src/testing/util.ts
```diff
@@ -19,9 +19,10 @@ export function render(platform: PlatformApi, html: string): Promise<HostElement
...
 
-export function transpile(tsFileName: string): string {
+export function transpile(tsFileName: string, srcDir: string): string {
   const config = mockBuildConfig();
   config._isTesting = true;
+  config.srcDir = srcDir;
```
kensodemann: Transpiler was throwing rootDir error without this set properly. Filter sends in './src' (typically)

package.json
```diff
@@ -17,7 +17,9 @@
     "dist/testing",
     "dist/util",
     "dist/index.js",
-    "dist/index.d.ts"
+    "dist/index.d.ts",
+    "testing.d.ts",
+    "testing.js"
```
adamdbradley: Is it possible to put `testing` inside of the `dist` directory?

kensodemann: TLDR; No, under `dist` doesn't work, but I think the latest commit is cleaner. Have a look.

More info...

Putting it under `dist` was actually my first attempt. Didn't work. So I went and looked at the [Typescript docs for module resolution](https://www.typescriptlang.org/docs/handbook/module-resolution.html).

Here is a synopsis:

The rules for module resolution of `import { x } from 'Foo'` are basically:

1. node_modules/Foo.(ts | tsx | d.ts)
2. node_modules/Foo/package.json w/ "types" specified
3. node_modules/Foo/index.(ts | tsx | d.ts)

Similar in node-land, but looking for 'js' files and package.json with "main" specified.

The `import { component } from '@stencil/core'` uses item number 2 for resolution (`node_modules/@stencil/core` has a `package.json` with "main" and "types" specified, both pointing to files in `dist/`).

So for resolving `@stencil/core/testing` I can do one of the following:

1. `node_modules/@stencil/core/testing.d.ts`
2. `node_modules/@stencil/core/testing/package.json` with "main" and "types" specified
3. `node_modules/@stencil/core/testing/index.d.ts` (plus .js for node land)

See my latest commit where I switched from strategy 1 to strategy 3 above. I think that is cleaner.

kensodemann: Thinking about this more, I think option number 2 may make this more maintainable. Let me have a stab at that and see how it looks before we do anything else with this PR.

testing/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "@stencil/core",
+  "version": "0.0.1",
+  "license": "MIT",
+  "main": "../dist/testing/index.js",
+  "types": "../dist/testing/index.d.ts",
```
kensodemann: Rather than having `testing.*` at the root level (original code) or `testing/index.*` that needed to potentially be maintained, this just points to the `dist/testing/index.*` files which are just built from the source as part of the build process.

PR #165, opened 2017-09-18
Fix path to util for logger

mitchellsimoens:
Either I was partly wrong about the path or something changed to the building over the weekend (I'm afraid it's likely the former). Either way, after installing `next` from npm, this path doesn't escape the `dist` dir.


adamdbradley:
No problem, and thanks for fixing this!

Issue #192, opened 2017-09-28
Use own service worker


kubii:
**Stencil version:**

```
 @stencil/core@0.0.6-2
```

**I'm submitting a ...**
[ x ] bug report
[ ] feature request
[ ] support request

**Current behavior:**
When serviceWorker.swSrc is specified in stencil.config.js file and I try to build the project then I get the following message:
```
The configuration options 'swSrc' will be ignored in the context of Workbox's generateSW mode.
```
According to the Workbox documentation the swSrc config is only valid with Workbox's injectManifest(). The goal was to use my own service worker implementation and build the manifest info into it.

**Expected behavior:**
If swSrc is defined then use injectManifest() to use the specified service worker and inject the manifest into it.

**Steps to reproduce:**
- clone stencil-starter project
- specify serviceWorker.swSrc config
- run the build task

**Related code:**

```
exports.config = {
  bundles: [
    { components: ['my-name'] }
  ],
  collections: [
    { name: '@stencil/router' }
  ],
  serviceWorker: {
    swSrc: 'pathToServiceWorkerImplementation'
  }
};
```

**Other information:**
Created a fork and changed the related code so the injectManifest() is used when swSrc is defined: [commit](https://github.com/kubii/stencil/commit/268487cf61665f7ad7c3bacfbf77ad450491c87b).
Let me know what do you think about this solution.

jgw96:
Looking into this now @kubii , thanks!

jgw96:
@kubii, would you mind making a PR for this? Im testing your fork, but it seems that its missing some of the stuff from the latest version of stencil/core. Besides that, it looks correct to me. Also, note to self, we are going to need some docs around this.

kubii:
@jgw96 
Sure, I can create the PR later today.

PR #222, opened 2017-10-07
Use own service worker

kubii:
Pull request for #192 

If swSrc config is not specified then the service worker
will be genereted otherwise injectManfiest will be executed.

Notes:
1; Now there is an inconsistency with the base paths since the base path of the `copy` functionality is the `/src` directory but the base path of the `swSrc` is the root of the project. Couldn't find any option for workbox to set the base path of the `swSrc`. Maybe we could prefix the path with `src/`. What do you think? Does it make sense?

2; When using `swSrc` then the config won't be passed to WorkboxSW's constructor. This is normal since WorkboxSW needs to be instantiated manually by the user.
All of these things are documented in the workbox docs. Quote:
> Note: This option is only valid when used with generateSW(). When using injectManifest(), you can explicitly pass the desired value in to the WorkboxSW() constructor in your swSrc file.

### DIFF
--- a/src/compiler/service-worker/generate-sw.ts
+++ b/src/compiler/service-worker/generate-sw.ts
@@ -1,4 +1,4 @@
-import { BuildConfig, BuildContext } from '../../util/interfaces';
+import { BuildConfig, BuildContext, ServiceWorkerConfig } from '../../util/interfaces';
 import { catchError } from '../util';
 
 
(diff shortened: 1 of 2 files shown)


adamdbradley:
Great, we'll take a look, thanks! @jgw96

jgw96:
This is looking pretty good to me @kubii . On your second point, i think this is fine since most people using this feature will be service worker "power users" and will probably need to be familiar with the workbox docs anyways. We can also link to that section in our [service worker docs](https://github.com/ionic-team/stencil-site/blob/master/docs-md/advanced/service-worker/index.md). 

On your first point, if im understanding correctly your saying that if a user uses this `swSrc` functionality then the service worker will not be copied over correctly with our default copy task?

kubii:
Hi @jgw96,
The service worker will be copied over correctly. Let me explain the problem through an example:
Let's assume we have `source.js` and `sw.js` in the `src` directory.
```
.
├── src
│   ├── source.js
│   └── sw.js
```

Now I want to copy over `source.js` and define `sw.js` as my custom service worker. The related config looks like this:
```
...
copy: {
    source: { src: 'source.js' }
},
serviceWorker: {
    swSrc: 'src/sw.js'
}
...
```
The `copy` is relative to the src directory but the `serviceWorker` is relative to the project root. I think that might be a bit confusing.

jgw96:
Ahh i got ya. Hmm, i think this is fine though actually, just something we will have to document. Ill test, we can get this merged and then ill write docs for it. Thanks again for the PR!

jgw96:
Thanks for the PR @kubii ,its working great!

PR #224, opened 2017-10-07
feat(interfaces): make  EventEmitter generic for better type checking

### LINKED ISSUE #210: util/interfaces: make EventEmitter generic for better type safety
**Stencil version:** 
```
 @stencil/core@0.0.6-10
```

[ ] bug report
[x] feature request

**Current behavior:**

`EventEmitter` interface has turned off proper type checking -> any

```ts
export interface EventEmitter {
  emit: (data?: any) => void;
}
```

**Expected behavior:**

Better DX would be having generic type with default values if user doesn't want to provide any payload

```ts
export interface EventEmitter<T=any> {
  emit: (data?: T) => void;
}
```

even better provide an EventEmitter class so we can directly initialise the `@Event` emitter instead of using just type annotations

I can send a PR, if yo're ok with this :)

Hotell:
Closes #210

### DIFF
--- a/src/util/interfaces.ts
+++ b/src/util/interfaces.ts
@@ -42,8 +42,8 @@ export interface EventListenerCallback {
 }
 
 
-export interface EventEmitter {
-  emit: (data?: any) => void;
+export interface EventEmitter<T= any> {
+  emit: (data?: T) => void;
 }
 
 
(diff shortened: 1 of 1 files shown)


adamdbradley:
Great thinking, thanks!

PR #233, opened 2017-10-08
docs(.github): tweak issue template to be more readable

Hotell:
Just made few tweaks so filling issues is less annoying. I filled them a few in recent days so this is my suggestions for better issue creation DX :)


jgw96:
Thanks for the PR!

PR #250, opened 2017-10-18
Ensure duplicate classes are preserved in DOM

CrshOverride:
Unfortunately, there are some CSS frameworks that rely on duplicate positional class names. While the CSS spec says duplicate names should be ignored, most browsers aren't implemented that way.

Previously, the compiler was updated to preserve duplicate class names in the resulting components.

Recently, a change was introduced into the rendering engine that removes those duplicate class names. The underlying issue is the use of `classList` when diffing new elements during `patch`. `classList.add` will only add a single instance of a given class to the list which prevents duplicates and breaks rendering.

This PR updates the diffing algorithm to build  a vanilla `classList` array using `classNames`, performs a diff using that array, and then sets the `className` property of the element using the resulting array.


jthoms1:
Not sure about the event decorator change.  Can you remove this and possibly place in another PR?

CrshOverride:
@jthoms1 Hah. Sure thing. It wasn't supposed to land here to begin with.

jthoms1:
I will merge after 0.0.6 which is coming very soon.  This will be released in the first prerelease of 0.0.7.

PR #261, opened 2017-10-24
feat(polyfill): add template element polyfill

### LINKED ISSUE #260: IE11 does not load styles due to lack of template element support
**Stencil version:**
```
@stencil/core@0.0.6-22
```

**I'm submitting a:**
- [x] bug report
- [ ] feature request
- [ ] support request => Please do not submit support requests here, use one of these channels: https://forum.ionicframework.com/ or https://stencil-worldwide.slack.com

**Current behavior:**
IE11 does to load styles due to lack of [&lt;template&gt;](https://html.spec.whatwg.org/multipage/scripting.html#the-template-element) element support:
image

**Expected behavior:**
IE11 should create a proper `<template>` element and load styles

**Steps to reproduce:**
1. Install and run https://github.com/ionic-team/stencil-app-starter
2. Open in IE11, there will be an error in the console

**Related code:**
This line https://github.com/ionic-team/stencil/blob/master/src/client/platform-client.ts#L316 can throws the error since `content` property does not exist
```tsx
 const styleElm = templateElm.content.cloneNode(true) as HTMLStyleElement;
```

**Other information:**
- Happens because IE11 does not support [&lt;template&gt;](https://html.spec.whatwg.org/multipage/scripting.html#the-template-element) element: https://caniuse.com/#feat=template
- I have a fix with the additional polyfill https://github.com/webcomponents/template, will submit a PR for review shortly

RomanMinkin:
Fixes #260

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -149,6 +149,7 @@ Web Components, specifically Custom Elements, are natively supported in Chrome a
 
 For the small minority of browsers that do not support modern browser features and APIs, Stencil will automatically polyfill them on-demand. What this means is that for browsers that already support the feature natively, they will not have to download and parse any unnecessary JavaScript. The great news is that in today's web landscape, most modern APIs are already shipping for what Stencil requires. Polyfills which are loaded on-demand include:
 
+ - [Minimal polyfill for <template>](https://github.com/webcomponents/template)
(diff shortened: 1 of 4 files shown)


adamdbradley:
this is great! Thank you for adding this and testing out IE!

PR #275, opened 2017-11-02
added Stylus support incl mocks and specs

kristianmandrup:
All tests pass:

```bash
Test Suites: 54 passed, 54 total
Tests:       795 passed, 795 total
Snapshots:   0 total
Time:        10.038s
```

Please verify and take it for a spin ;)
See [Stylus](http://stylus-lang.com/), a powerful alternative to SASS. 

Note that you can use this with PostCSS via [poststylus](https://github.com/seaneking/poststylus) and there are dozens of additional plugins... Pure NodeJS bliss ;)


mlynch:
@kristianmandrup thank you so much for the PR. 

This is a good opportunity for us to think ahead a bit. We want to keep dependencies for Stencil to an absolute minimum. As such, I don't think we want to merge this into master as-is.

What if this could be a plugin so users could optionally include it? Maybe it's time to figure out a plugin system? cc @jthoms1 @adamdbradley

kristianmandrup:
Yeah, would be super neat with a plugin system. I understand your concerns. Must be some ready made plugin systems out there to "plug in" ;)

Stylus is way more light weight (pure Node lib) than SASS, which requires native C bindings etc. Also much less verbose syntax and more extension/plugin friendly and can do mostly all the same things... such as nesting, imports, variables, functions, .... 
Unfortunately Stylus is not as well known as SASS and perhaps not as much community around it, but I'd personally take it any day over SASS.

adamdbradley:
@mlynch Yeah agreed, I think we can start allowing plugins so stencil itself doesn't need to contain all possible dependencies. Under the hood we're already using rollup, so I think we could piggyback off of their existing system, and configure it via a `plugins` array in the `stencil.config.js` file. Thoughts?

adamdbradley:
@kristianmandrup Also completely agree the Sass dependency is quite heavy, and would love to remove it from the default stencil dependencies.

mlynch:
@adamdbradley I like that, their plugin system is so simple. I think it's fair that we have our own place to configure plugins, that's how rollup and webpack do it as well.

kristianmandrup:
I'd suggest first cleaning up the internal interfaces/structures for iterating on and deciding what to do in the build process.

Instead of hardcoded "entry points", such as:

- `includedStylusFiles`
- `stylusConfig`
- `changeHasStylus`
- `stylusBuildCount`
- `includedStylusFiles`

Use a more generic, extensible structure and then turn you own default choices into "default" baked in plugins as well ;)

```js
interface BuildConfig {
  // ...
  styling: Styling[]
}

interface Styling {
    name: stylus: {
    changed?: boolean,
    buildCount?: number
    config?: Object
    includedFiles?: string[]
}
```

```js
// stencil.config.js

export {
  // ...
  plugins: {
    // reference config extensions in npm modules of plugins
    stylus: '@stenciljs/stylus-styling/plugin-config/default.config', 
    sass: '@stenciljs/sass-styling/plugin-config/holy-grail.config'    
  }
}
```

Each of the plugins could then ship with their own re-configuration/extension of the `buildConfig`.
Then the build process should have no knowledge of the specific naming of any build config entries, but just be able to traverse it in a generic way, oblivious to which plugins are installed and what config they have - loose coupling!

Your thoughts?

kristianmandrup:
As an example: 

```js
function processIncludedStyles(config: BuildConfig, ctx: BuildContext, moduleFile: ModuleFile) {
  const processStyle = config.createProcessStyle(config, ctx, moduleFile)
  config.styling.map(styleConfig => processStyle)
}
```

So much cleaner :)

kirkhoff:
I would love to see stylus support for stencil, and thanks for putting up this PR! As a workaround, I've simply linked to the generated css file and relied on npm scripts to generate the css file from stylus prior to building. 

Alternatively, if we were following the webpack model, we could include a stylus loader to inject css directly into the `styles` property when setting up the stencil Component.

adamdbradley:
Also, we haven't forgotten about this. We've got a plugin system now, and we'll be able to add stylus and less to the list. Right sass is still an internal plugin, but the building blocks are in place that we can pull it out of the core of stencil, then we can add others. We'll let you know when it's ready. thanks!

bfmatei:
Hello guys,

I implemented a demo about PostCSS integration in the current plugins system. You can find it here: https://github.com/bfmatei/stencil-boilerplate/blob/master/plugins/postcss.js

It's the first version and it's just a base thing. From my opinion you can literally remove the oob support for everything except pure CSS and make optional plugins like the above one in order to enable endless opportunities.

kristianmandrup:
Sweet :) Could you or someone else please do the same for Stylus support ;) Thanks!

adamdbradley:
We'll be creating plugins for autoprefixer, less, and stylus. Sass has already been completed and can be found here: https://github.com/ionic-team/stencil-sass

This will take affect in the `0.6.0` release. Thanks!

kristianmandrup:
Awesome! Thanks :) Please make an update to the docs on the main site as well :)
Cheers! Keep up the good work!

PR #369, opened 2017-12-19
fix(testing): fix preprocessor to only process .tsx files with stenci…

danbucholtz:
This PR fixes several issues in `@ionic/core` for testing, and greatly speeds up the time for testing. Originally, whenever importing a `.d.ts` file, it would take ~300 seconds to execute the tests. When I made the change to stop processing `.d.ts` files, it took it down to ~5.5 seconds for the tests to execute. When I stopped compiling .ts files with Stencil, it took the time down to ~1 second.

Thanks,
Dan

### DIFF
--- a/scripts/packages/testing/jest.preprocessor.js
+++ b/scripts/packages/testing/jest.preprocessor.js
@@ -2,7 +2,6 @@ var path = require('path');
 var ts = require('typescript');
 var testing = require('./index');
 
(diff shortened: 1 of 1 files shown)


adamdbradley:
Oh sweet, good catch. @jthoms1 would you be able to test the testing test tests? Test.

Issue #390, opened 2017-12-31
Implementation of supportsCustomElements, supportsEsModules are not correct


musou1500:
**Stencil version:**

```
 @stencil/core@0.1.1-0
```

**I'm submitting a:**

- [x] bug report
- [ ] feature request
- [ ] support request => Please do not submit support requests here, use one of these channels: https://forum.ionicframework.com/ or https://stencil-worldwide.slack.com

**Current behavior:**
I think implementation of `supportsCustomElements`, `supportsEsModules` are not correct.
https://github.com/ionic-team/stencil/blob/master/src/client/loader.ts#L38-L44

**Other information:**

I fixed it. Can I send PR?
https://github.com/ionic-team/stencil/compare/master...musou1500:fix-check-feature-function

musou1500:
@jgw96 thank you. I sent PR.

PR #394, opened 2018-01-05
fix: fix incorrect functions implementation

musou1500:


musou1500:
fix https://github.com/ionic-team/stencil/issues/390

adamdbradley:
Ahh! Good catch, thanks!

PR #429, opened 2018-01-17
fix #375

### LINKED ISSUE #375: html after a <svg> node fails to render
**Stencil version:**

```
 @stencil/core@0.1.0
```

**I'm submitting a:**

- [x] bug report
- [ ] feature request
- [ ] support request => Please do not submit support requests here, use one of these channels: https://forum.ionicframework.com/ or https://stencil-worldwide.slack.com

**Current behavior:**

Inside the return block for the `render` function, any html added after the inclusion of a `<svg>` node will not render. The `<svg>` itself will render, but nothing after it.

**Expected behavior:**

The `<svg>` node should not prevent subsequent html from rendering.

**Steps to reproduce:**

1. Add any valid `<svg>` node to the return block in your render function.
2. Ensure you have some sort of html that you expect to be rendered after the `<svg>` node.
3. View the result in the browser, and notice that the svg renders but nothing after it.

I've simplified my problem and included it below.

**Related code:**

```tsx

@Component({
  tag: 'svg-oops'
})
export class SvgOops {
  render() {
    return (
      <div>
        <span>foo</span>
        <svg />
        <span>bar</span>
      </div>
    );
  }
}

```

**Other information:**

tungd:
Hi, I see @adamdbradley is assigned, but this is urgent for me so I just go ahead and investigate this any way. 

The issue is the SVG context is not reset after create the first SVG element (`isSVGMode` never set to `false`). Any element created after the `<svg/>` actually is a `SVGElement` thus not rendered by browser. I've added a simple test case, not sure if it covers all the cases though.

screen shot 2018-01-18 at 1 59 01 am

### DIFF
--- a/src/core/renderer/patch.ts
+++ b/src/core/renderer/patch.ts
@@ -139,6 +139,8 @@ export function createRendererPatch(plt: PlatformApi, domApi: DomApi): RendererA
           }
         }
       }
+
+      isSvgMode = false;
     }
 
     return vnode.elm;
--- a/src/core/renderer/test/patch-svg.spec.ts
+++ b/src/core/renderer/test/patch-svg.spec.ts
@@ -38,5 +38,16 @@ describe('renderer', () => {
       expect(elm.firstChild.namespaceURI).toEqual(SVGNamespace);
       expect(elm.firstChild.firstChild.namespaceURI).toEqual(XHTMLNamespace);
     });
+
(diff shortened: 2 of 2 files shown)


adamdbradley:
🚀 AWESOME!!! Thanks for the PR and unit tests! Looks great, we'll get another release out soon, thanks!

PR #439, opened 2018-01-24
fix(build): update uglify-es to 3.3.8

nlx-sascha:
This update fixes an error in function inlining within loops


adamdbradley:
Awesome thank you!

PR #451, opened 2018-01-26
Read user tsconfig

bfmatei:
Hello,

This pull request should read the user’s tsconfig file, fallback to default settings in case the config is malformed or contains missing props and translate tsconfig paths mapping in actual paths.

Feel free to have a look and point out any mistakes :).


bfmatei:
Hello @adamdbradley ,

I did some improvements (moved away from readFileSync to async readFile - used from compilerCtx in both cases - tsconfig parse and paths resolver) in the latest commit.

I tried to understand what do you mean by:
> One thing to change is to pass `resolvePaths` the `compilerCtx`

1. Create an instance of the paths resolver in compilerCtx to make it available widely
2. Pass the compilerCtx object to the paths resolver in order to use the fs?

I just wanted to make sure as the second option was already implemented.

adamdbradley:
Sorry, yeah good question. Here's an example where we're passing in the compilerCtx: https://github.com/ionic-team/stencil/blob/2ba8d189df876294c134f1197a2c2619ea85d594/src/compiler/bundle/rollup-bundle.ts#L28

And here's how it's being used:
https://github.com/ionic-team/stencil/blob/2ba8d189df876294c134f1197a2c2619ea85d594/src/compiler/bundle/rollup-plugins/transpiled-in-memory.ts#L4

bfmatei:
Thanks for pointing out. Latest push should fix it.

bfmatei:
Hello again @adamdbradley ,

You can find the corresponding tests here: https://github.com/bfmatei/stencil/blob/read-user-tsconfig/src/compiler/bundle/rollup-plugins/test/paths-resolution.spec.ts

I created a basic suite for testing all possible cases: UNIX paths, Windows paths and mixed paths.

Thanks,
Bogdan

bfmatei:
Just a side node: I brought the branch in sync with the latest master branch after the latest merges

bfmatei:
Hello!

I brought to branch to the latest master and also i fixed two things:
- add await to the plugin initialiser (the plugin was not loaded anymore before this)
- enforce absolute path on module files

I also tested using the latest build (0.3.0-5) on my boilerplate.

adamdbradley:
merged!!! Should go out in our next release, thanks!!

warapitiya:
How can I use this? im getting following error?

```
Bundling Warning
           '@validators' is imported by
           src/components/abc/abc.js, but could not be
           resolved – treating it as an external dependency
```

MattMcFarland:
@warapitiya I too am having problems getting this to work.  It appears that the rollup-plugin @bfmatei wrote does not account for `baseUrl` - so if you set your `baseUrl` to `./` it should work.  

I basically went from this:

```
compilerOptions: {
  ...
  baseUrl: 'src',
  paths: {
     '@pages/*': ['pages/*'],
     '@data/*': ['data/*'],
     ...
  }
}
```

To this:
```
compilerOptions: {
  ...
  baseUrl: './',
  paths: {
     '@src/*': ['src/*]'
     ...
  }
}
```
I found a workaround that seems to work for me:

MattMcFarland:
Created issue #1291

Issue #162, opened 2017-09-16
Improve External Module Bundling and Load Through ES Module Imports


adamdbradley:
Build a graph of all the shared modules and rollup each bundle as far down as possible. Then use ES module imports to load the shared bundles. For browsers that do no support ES modules it'll fallback to the current process.

kristianmandrup:
So what to do until this issue is resolved? Any recommendations?

kylecordes:
I think this is already implied by the description, but I'll write it out just in case....

Hopefully the ultimate goal here is effective sharing of library code across multiple separately developed Stencil components and further across Stencil consuming code:

* Developer A publishes a Stencil component that uses RxJS
* Developer B publishes a Stencil component that uses RxJS
* Developer C uses Angular, which uses RxJS
* Developer C's application directly uses RxJS, and is is written with Angular, using the Stencil components from A and B
* When it all ends up running on the page, there is only a single copy of RxJS shared by all of the above

This should be implementable today for current browsers using ES modules; fallback to static bundling would need to happen at the highest (**application**) level to get correct results for all browsers. I don't see a way to get correct results for all browsers if the Stencil build does the fallback bundling. You would end up with (my in my example) more than one copy of RxJS on the page.

FelschR:
Any update on this?

adamdbradley:
@FelschR Yup, we've got a branch going and actively workin on it.

jgw96:
Just wanted to post an update here as this issue is getting a decent amount of traction on the stencil slack. The team is actively working on this and we are pretty close to having it working. This is probably our biggest refactoring of the internals of the Stencil compiler to date, or atleast one of the biggest, so its taking some time.

rolandjitsu:
Is this solved now?

kristianmandrup:
@jgw This is likely the most anticipated update as well. Not using Stencil until this critical issue is resolved. What is the timeline? Before end of January? Alpha, Beta? 
Thanks :)

rolandjitsu:
@kristianmandrup I might be mistaken, but it seems like the built files do not contain the imports that I have in my components. That leads me to believe that this is partially solved?

kristianmandrup:
What is the expected timeline on this? You expect to have a beta release of this critical feature/improvement end of January 2018? 

Thanks :)

adamdbradley:
[...]

PR #453, opened 2018-01-26
Esmodules

jthoms1:
This pull request is to change the stencil packaging process so that we use rollup to manage multiple entry points. Depends on `rollup@0.55.0`.


jthoms1:
Tests are failing because of `node_modules/rollup/dist/typings/Graph.d.ts`.  The types of rollup are currently not setup correctly.  All tests pass on local machines after this file is updated.

jthoms1:
This PR fixes #162.

adamdbradley:
🚀

Issue #454, opened 2018-01-29
DashCase attribute are not watched


dpellier:
**Stencil version:**

```
 @stencil/core@0.2.3
```

**I'm submitting a:**

- [x] bug report
- [ ] feature request
- [ ] support request => Please do not submit support requests here, use one of these channels: https://forum.ionicframework.com/ or https://stencil-worldwide.slack.com

**Current behavior:**
When using a Prop like:
`@Prop() maxLength: number;`
`@Watch('maxLength')` is never called
When looking in the code the `attributeChangedCallback` is never called, seems to be because the `HostElementConstructor.observedAttributes` is filled with 'maxLength' instead of 'max-length' on `defineComponent`.

**Expected behavior:**

When using a Prop like:
`@Prop() maxLength: number;`
`@Watch('maxLength')` should be called

**Steps to reproduce:**

```tsx
class SomeElement {
  @Prop() maxLength: number;

  @Watch('maxLength')
  protected onMaxLengthChange(_newValue: string): void {
    console.log('maxLength has changed');
  }
}
```

**Other information:**
Pull Request: https://github.com/ionic-team/stencil/pull/455

jgw96:
Hey! Thanks for opening an issue with us and thanks for the PR! I just tried to reproduce this locally and unfortunately could not. Mind posting a repo we could use to reproduce this?

dpellier:
Sure, here is a small dummy component that reproduce the problem: https://github.com/dpellier/stencil-issue

adamdbradley:
I think this may have been fixed in https://github.com/ionic-team/stencil/commit/73aabd247259221235791dc54c39d7469c279d11

Would you be able to test the latest and double check it's working?

dpellier:
Indeed it work now with 0.3.0.
Thanks 👍

adamdbradley:
great, thanks for testing!

liho98:
I have the exactly same problem with the Author when using Stencil JS v1.17.3

PR #455, opened 2018-01-29
fix(data-parse): use dashCase as attribName in memberMeta

dpellier:
When having a Prop like: `@Prop() maxLength: number;` the `attributeChangedCallback` is never called.
see https://github.com/ionic-team/stencil/issues/454


dpellier:
Fixed in 0.3.0

Issue #448, opened 2018-01-25
SVG not fully rendered


jrmlt:
**Stencil version:**
```
 @stencil/core@0.1.6
```

**I'm submitting a:**
- [X] bug report
- [ ] feature request
- [ ] support request

**Current behavior:**
Since version 0.1.6 and still in 0.2.3 when rendering an `svg` element, only the first path is rendered.

**Expected behavior:**
The `svg` is fully rendered

**Steps to reproduce:**
Render an `svg` element with multiple sub-elements, only the first one is rendered:
```tsx
render() {
    return (
      <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 129.8 51" fill="transparent" stroke-width="0.5" stroke-linecap="square">
        <path class="splash-bg" d="M0 .1v37.2h129.5V.1z"/>
        <path class="splash-letter" d="M10.7 26.6V10.8h3.7l7.5 9v-9h3.9v15.8h-3.4l-7.8-9.4v9.4z"/>
      </svg>
    );
  }
```

**Related code:**

Seems to be linked to commit 7ffb4ef

**Other information:**

jgw96:
Thanks for opening an issue with us. Note to team, this is only with inline SVGs and does not impact ion-icon

aabm00:
Hi!!

I am just starting with stencil and it's great - Congratulations!!!
When trying to incorporate in the render, SVGs I have found the same problem with Svg rendering. Looking at the #504 post  I understand that the problem is still in process to be solved.  Am I Ok?

There is something I'd want to clarify. Looking at the Stencil-App-Started package.json it looks that is working with the stencil-core 0.2.3 but the last version of stencil-core is 0.4.0-1 so Can I just update the version in the package of the  Stencil-App-Started to 0.4.0-1 so I can get the last fixed?

Thanks

PR #460, opened 2018-01-31
Fix/svg

tungd:
Hi, it turns out my previous fix for SVG rendering is not totally correct. It breaks when the SVG has two or more child nodes, the `isSvgMode` was reseted after the first child node. This update will only reset  the SVG context when exiting SVG element. There's some other tests broken but I'm able to run `jest -t 'render'`.


tungd:
This probably will fix #448

adamdbradley:
thanks!!

sayjava:
Was this fix part of the 0.3 release? I am still not able to render more than 1 svg child node after upgrading.

tungd:
@sayjava I believe it's not. You can check the [Change Log](https://github.com/ionic-team/stencil/blob/master/CHANGELOG.md) and commit log.

aabm00:
Hi 

I have the same problem, only the first svg child is rendered, so is it still in process to solve the problem?

Thanks

Issue #469, opened 2018-02-02
Custom relative wwwIndexHtml broken


finneganh:
**Stencil version:**

```
 @stencil/core@0.2.3
```

**I'm submitting a:**

- [x] bug report
- [ ] feature request
- [ ] support request => Please do not submit support requests here, use one of these channels: https://forum.ionicframework.com/ or https://stencil-worldwide.slack.com

**Current behavior:**
A relative `wwwIndexHtml` config value gets converted into the absolute path of `wwwDir`, ignoring `wwwIndexHtml` entirely. Additionally, since `wwwDir` is a directory, this means that no index.html file is written at all.

**Expected behavior:**
The index.html file is written to the path at `wwwIndexHtml` relative to `wwwDir`.

**Steps to reproduce:**
1. Set a `wwwIndexHtml` value in the config
1. Run `stencil build`
1. See that no `wwwIndexHtml` file is written

**Other information:**

Looks like a copy-paste error in `validate-paths.ts`. Submitting a PR to fix.

adamdbradley:
merged 🎉 thanks!

PR #470, opened 2018-02-02
Fixes absolutizing of wwwIndexHtml

finneganh:
If wwwIndexHtml is not absolute, it is changed to wwwDir, which seems
like a copy-paste oversight.

This commit changes it to use the value of wwwIndexHtml relative to
wwwDir.

Fix for https://github.com/ionic-team/stencil/issues/469

### DIFF
--- a/src/compiler/config/validate-paths.ts
+++ b/src/compiler/config/validate-paths.ts
@@ -86,7 +86,7 @@ export function validatePaths(config: Config) {
     config.wwwIndexHtml = normalizePath(path.join(config.wwwDir, DEFAULT_INDEX_HTML));
   }
   if (!path.isAbsolute(config.wwwIndexHtml)) {
-    config.wwwIndexHtml = normalizePath(path.join(config.rootDir, config.wwwDir));
+    config.wwwIndexHtml = normalizePath(path.join(config.wwwDir, config.wwwIndexHtml));
   }
 
   if (typeof config.publicPath !== 'string') {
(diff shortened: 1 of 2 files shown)


src/compiler/config/validate-paths.ts
```diff
@@ -86,7 +86,7 @@ export function validatePaths(config: Config) {
     config.wwwIndexHtml = normalizePath(path.join(config.wwwDir, DEFAULT_INDEX_HTML));
   }
   if (!path.isAbsolute(config.wwwIndexHtml)) {
-    config.wwwIndexHtml = normalizePath(path.join(config.rootDir, config.wwwDir));
+    config.wwwIndexHtml = normalizePath(path.join(config.wwwDir, config.wwwIndexHtml));
```
adamdbradley: Awesome thanks for finding this! So one thing we've been doing is that if the file is relative, it should be relative to the stencil.config.js file, so basically the project's root. So in this case, i think it'd be normalizePath(path.join(config.rootDir, config.wwwIndexHtml)); Thoughts?

finneganh: That's totally your call. Looking over the file it's inconsistent what paths are absolutized against. I see build is also against `wwwDir`. Dunno if output files should be relative to `wwwDir` or if these all need a bit of a rename.

adamdbradley: 🤔 very good point! Let's go with it, thanks!

PR #488, opened 2018-02-06
feat(onReady): onReady callback

manucorporat:
There are two use cases for `element.componentOnReady()`

```ts
// returning a promise
element.componentOnReady().then(() => {
   console.log("component is ready");
}

// passing a callback
element.componentOnReady(() => {
   console.log("component is ready");
});
```

but non of them work in vanilla JS
```html
<script>
  const el = document.queryElement('ion-menu-controller');
  el.componentOnReady().then(() => {
     el.open('right');
  });
</script>
</body>
```

The reason is that at this point the CustomElement has not been registered yet and `componentOnReady` is undefined.

I propose using a object property that references a callback function that is called when the element gets initialised, this "third" approach would work even if the custom element is not registered yet:

```js
el.componentOnReady = function() {
   console.log('component is ready');
}
```

**The problem** is that the property `componentOnReady` conflicts with the stencil's HostElement prototype that also has a componentOnReady: https://github.com/ionic-team/stencil/blob/master/src/core/instance/init-host-element.ts#L34

Because object properties have precedence over prototype properties, stencil runtime would have to "delete" the property:
```js
delete el.componentOnReady;
```
in order for `prototype.componentOnReady` to be called in future usages.

**If we don't delete the `componentOnReady` property, other code paths that expect a promise from this function will break.**

It's a well-known anti-pattern to use "delete" because it cases deoptimization in most of the JS engines.

That's why I propose to use a different event-like property: `onReady`

```js
el.onReady = function() {
   console.log('component is ready');
}
```

### DIFF
--- a/src/compiler/app/build-core-content.ts
+++ b/src/compiler/app/build-core-content.ts
@@ -272,6 +272,7 @@ const RESERVED_PROPERTIES: string[] = [
   '$onRender',
   '$',
   'componentOnReady',
+  'onReady',
 
 
   /**
--- a/src/core/instance/init-component-instance.ts
+++ b/src/core/instance/init-component-instance.ts
@@ -67,6 +67,12 @@ export function initComponentLoaded(plt: PlatformApi, elm: HostElement, hydrated
       // fire off the ref if it exists
       callNodeRefs(elm._vnode);
 
+      // fire off the user's elm.onReady function
+      if (elm.onReady) {
+        elm.onReady(elm);
+        elm.onReady = null;
+      }
+
       // fire off the user's elm.componentOnReady() callbacks that were
       // put directly on the element (well before anything was ready)
       if (elm._onReadyCallbacks) {
--- a/src/core/instance/test/init.spec.ts
+++ b/src/core/instance/test/init.spec.ts
@@ -14,11 +14,11 @@ describe('instance init', () => {
       let called1 = false;
       let called2 = false;
 
-      let p1 = elm.componentOnReady().then(() => {
+      const p1 = elm.componentOnReady().then(() => {
         called1 = true;
       });
 
-      let p2 = elm.componentOnReady().then(() => {
+      const p2 = elm.componentOnReady().then(() => {
         called2 = true;
       });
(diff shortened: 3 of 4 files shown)


adamdbradley:
Closing for now, we'll revisit later

adamdbradley:
I think the best option is to use:

```
customElements.whenDefined('my-cmp').then(...)
```

What’s happening is a script can execute before the lazy loaded core file is downloaded, which defines the components (and has polyfills if need be). So the race condition is if the script is ran befoe the downloaded core file runs.

`customElements.whenDefined()` is a standard: https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry/whenDefined

However, it’s different than stencil’s `componentOnReady` because it resolves once a tag is defined as a custom element, but it doesn’t take into consideration which exact element is ready, or if the stencil component has resolve it’s own `componenDidLoad`.

`componentOnReady` on the other hand fires after the component has rendered, and the component’s `componentDidLoad` is resolved, and all of it’s child components have finished loading too.

Next gotcha was that `customElements.whenDefined()` wouldn’t be polyfilled yet by IE11, so had to add this: https://github.com/ionic-team/stencil/commit/8d92cc13296c621622a4223cf81e25320e57bcaa

adamdbradley:
Removing the whenReady temporary polyfill in favor using componentOnReady, as of https://github.com/ionic-team/stencil/commit/57d14c7fed0df33b3cdfa78d9561abb64d49d1c9

Note: You can still use `whenReady`, just needs to be used after the fully custom elements polyfill has applied.

julesjanssen:
So, just to be clear: how should we detect if a Stencil component is ready when using them within a vanilla JS environment?

Using `customElements.whenDefined` doesn't seem to work anymore in browsers where `customElements` needs to be polyfilled. (I think it worked in earlier versions ~0.7.6 of the build?)

Using `element.componentOnReady` both as a Promise and with a callback doesn't fire in at least Firefox (I'm not sure if this is caused by the need the polyfill anything).

PR #497, opened 2018-02-07
fix(copy): wait for processCopyTaskDestDir execution

dpellier:
Otherwise the `allCopyTasks` of `copyTasks` function is always empty when resolving.


adamdbradley:
Gahh! Great catch! Thanks

Issue #498, opened 2018-02-07
componentDidLoad() is fired twice


manucorporat:
1. create a html
2. use a stencil component with children nodes

```
<ion-router>
   <ion-route />
</ion-router>
```

The lifecycle method `componentDidLoad` is called twice in `ion-router`

PR #499, opened 2018-02-07
fix(init): componentDidLoad called twice

manucorporat:
Tries to fix: https://github.com/ionic-team/stencil/issues/498

@adamdbradley I am not sure it is a good fix or that is actually solves the root of the problem. could you look at it? The fact that componentDidLoad runs twice sometimes breaks some components in ionic-core. Also having some unit tests for the lifecycle events would be great


adamdbradley:
going to look into different ways to handle this. I think this way will not work 100% of the time and a second change won't get rendered.

adamdbradley:
I take that back, this should be good to go, thanks!

PR #503, opened 2018-02-08
fix(svg): isSvgMode must be reset when an svg node is patched

yesobo:
patchVNode function does not reset isSvgMode after patching svg nodes, therefore, if the next element in the DOM tree has to be created by domApi.$createElement funcion, it is created as an SVG Element.

The bug can be tested in a [demo I created](https://github.com/yesobo/stencil-svg-bug) based on stencil-app-starter with @stencil/core 0.3.0 where it is unable to add a new div element follwing a svg node already rendered in the DOM tree.

PR #460 fixes subsequent SVG nodes created in the same patchVNode stage but it does'nt take into account when a previous SVG element was already rendered in a previous stage


yesobo:
Forgot to uncomment line with fix

PR #504, opened 2018-02-08
fix(svg): isSvgMode must be reset when an svg node is patched

yesobo:
patchVNode function does not reset isSvgMode after patching svg nodes, therefore, if the next element in the DOM tree has to be created by domApi.$createElement funcion, it is created as an SVG Element.

The bug can be tested in a [demo I created](https://github.com/yesobo/stencil-svg-bug) based on stencil-app-starter with @stencil/core 0.3.0 where it is unable to add a new div element follwing a svg node already rendered in the DOM tree.

PR #460  fixes subsequent SVG nodes created in the same patchVNode stage but it does'nt take into account when a previous SVG element was already rendered in a previous stage

### DIFF
--- a/src/core/renderer/patch.ts
+++ b/src/core/renderer/patch.ts
@@ -339,6 +339,9 @@ export function createRendererPatch(plt: PlatformApi, domApi: DomApi): RendererA
       // and also only if the text is different than before
       domApi.$setTextContent(elm, newVNode.vtext);
     }
+
+    // reset svgMode when svg node is fully patched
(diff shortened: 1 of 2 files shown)


manucorporat:
@yesobo thanks for the PR! can you add an unit test that fails without your fix?

yesobo:
The test added in https://github.com/ionic-team/stencil/pull/504/commits/9f66e68dac62079458124379e1fb19b639c66e9c fails because an added div Node is created as an SVGElement intead of the expected HTMLDivElement

adamdbradley:
awesome thanks!

Issue #567, opened 2018-02-22
add config option for custom file system


Xiphe:
**Stencil version:**

```
 @stencil/core@0.6.5
```

**I'm submitting a:**

[x] feature request

**Current behavior:**

Stencil uses a virtual file system under the hood which uses `fs-extra`, which uses default node `fs`

**Expected behavior:**

Since a big part of the stencil api is file/folder based it would be great if we could customise the file system that is being used by the virtual system of stencil.

given config:
```js
module.exports = {
  fs: myCustomFs
  /* ... */
};
```

I would expect the stencil vfs to use `myCustomFs` instead of `fs-extra`.

This would:
 - simplify integration tests
 - enable use of other data sources for "file" inputs like
   - http
   - memory
   - db

**Other information:**

I already tried this:

```js
const { Compiler } = require('@stencil/core/compiler');
const { loadConfig } = require('@stencil/core/server');
const myCustomFs = require('./myCustomFs');

const config = loadConfig(__dirname);
config.fs = myCustomFs;

const compiler = new Compiler(config);
compiler.sys.fs.disk = myCustomFs;

compiler.build();
```

But it feels really like a hack and fails when transpiling modules with typescript.

---

Is this a direction you'd like for this tool? If yes I'd be happy to provide more information and/or work on a PR.

Xiphe:
related to https://github.com/ionic-team/stencil/issues/474 https://github.com/ionic-team/stencil/issues/523

Xiphe:
Ok, I think I located the core problem: typescript does not use the virtual file system for module lookups. So the source files actually need to exist on disk.

Stencil could provide a `resolveModuleNames` on it's `tsHost` supporting lookups in the virtual file system.

I'm currently spiking if this would work somehow.

Xiphe:
Found it!

the stencil [compiler-host](https://github.com/ionic-team/stencil/blob/master/src/compiler/transpile/compiler-host.ts) does not implement a `directoryExists` method. Causing typescript to use it's own implementation for `directoryExists` which uses the real file system.

I'll provide a PR to fix this.

adamdbradley:
Thanks for digging into this! Yeah we shouldn't be too far away from what you're looking for, all of our internal tests are using an in-memory fs so something must be off or not exposed properly. If you can provide a PR that'd be really helpful, thanks.

Xiphe:
I'll have a look into the tests, maybe my custom FS is lacking something.
[...]

PR #570, opened 2018-02-24
Do not use typescript default fs for directory lookup

Xiphe:
This PR adds a `directoryExists` method to `compiler-host.ts`. Typescript will otherwise use its own implementation which uses the default node file system instead of ours.

This is required to support builds with a virtual file system and `suppressTypeScriptErrors: false`

ref https://github.com/ionic-team/stencil/issues/567

### DIFF
--- a/src/compiler/transpile/test/transpile.spec.ts
+++ b/src/compiler/transpile/test/transpile.spec.ts
@@ -1,6 +1,8 @@
 import { TestingCompiler } from '../../../testing';
 import { wroteFile } from '../../../testing/utils';
+import { normalizePath } from '../../../compiler/util';
 import * as path from 'path';
+import * as fs from 'fs';
 import * as ts from 'typescript';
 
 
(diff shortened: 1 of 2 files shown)


src/compiler/transpile/test/transpile.spec.ts
```diff
@@ -155,4 +157,67 @@ describe('transpile', () => {
...
+      return originalReadFileSync(filePath);
+    }
+    c.fs.statSync = (itemPath) => {
+      itemPath = normalizePath(itemPath);
+      if (itemPath.indexOf(nodeModulesDir) === 0 || itemPath.indexOf(distDir) === 0) {
+        return fs.statSync(itemPath);
+      }
+
+      return originalStatSync(itemPath);
+    }
```
Xiphe: Not sure about the monkey-patching here. This could also be a feature of the `testing-fs` but that would've required more refactoring.

PR #577, opened 2018-02-27
Stylus/Less Warnings

bfmatei:
Warn when Stylus and/or Less are used without corresponding plugins


adamdbradley:
merged! thanks!

PR #596, opened 2018-03-05
Image decode definition

arjunyel:
[https://twitter.com/addyosmani/status/938078402430382080](https://twitter.com/addyosmani/status/938078402430382080)

Just curious, do these definitions have to be manually maintained or are they pulling from somewhere? Thanks!


jthoms1:
Mostly we depend on typescript for these definitions. 
https://github.com/Microsoft/TSJS-lib-generator

But in the case of JSX it is all internal.

PR #608, opened 2018-03-07
fix(polyfill): added weakmap as polyfill for IE 9 and 10

illiano:
i have added weakmap as polyfill for Internet Explorer 9 and 10. Now it is possible to run Stencil Apps also in IE 9 / 10


adamdbradley:
We haven't fully tested IE9 and IE10, and I'm sure there are many other features that wouldn't work in them so we haven't listed 9 and 10 as officially supported. I guess before we attempt to do that I'd like to get more full testing in those browsers and see if it'll be possible or not.

adamdbradley:
After some more discussion we decided to not include the WeakMap polyfill. Currently we do not support IE9 and 10, and if we start to go down the path of trying to support those two then many other issues will probably arise. However, it would be interesting to have this as a separate file for those who need to support 9 and 10, but again we can't guarantee it'll be bug free without using up a lot of valuable development time.

Issue #606, opened 2018-03-06
Stencil polyfill breaks MobX on IE11


finneganh:
**Stencil version:**

```
 @stencil/core@0.6.18
```

**I'm submitting a:**

[X] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://forum.ionicframework.com/ or https://stencil-worldwide.slack.com

**Current behavior:**

If Stencil in on a page in IE11, importing MobX will cause a "Assignment to read-only properties is not allowed in strict mode" error.

**Expected behavior:**

Stencil and MobX should be able to co-exist.

**Steps to reproduce:**

1. Add the following tag after the Stencil loader: `<script src="https://cdnjs.cloudflare.com/ajax/libs/mobx/3.6.1/mobx.umd.js"></script>`
1. Load the page in IE11 with the debugger open
1. See the error when MobX tries to assign `ObservableArray.prototype.find`

**Related code:**

```tsx
// insert any relevant code here
```

**Other information:**

This may be related to the fact that the `Array.find` polyfill is read-only, based on mobxjs/mobx#1028

Adding an `Array.find` polyfill in a `<script>` block above the Stencil loader that has `writable: true` in its `defineProperty` fixes things.

It may be that the polyfills should be defined as "writable," but I don't know enough about polyfill / `defineProperty` convention to say whether that's correct.

finneganh:
FWIW the https://polyfill.io/ code seems to use `writable: true` in their polyfills: https://github.com/Financial-Times/polyfill-service/blob/master/polyfills/_ESAbstract/CreateMethodProperty/polyfill.js

adamdbradley:
Adding writable makes sense. Would someone be able to try that out? Thanks

finneganh:
I can check through the polyfills and try to get their `defineProperty` calls to be similar to what polyfill.io does.

finneganh:
OK, sent in #612

PR #612, opened 2018-03-08
Makes polyfills writable properties

finneganh:
Defining polyfills as non-writable causes "Assignment to read-only
properties is not allowed in strict mode" errors in certain
circumstances, such as using MobX and its ObservableArray subclass of
Array.

This follows the conventions from polyfill.io around the options passed
to `defineProperty`.

Fixes ionic-team/stencil#606


adamdbradley:
🎉 thanks!!!!

PR #627, opened 2018-03-16
support crossorigin for dynamic import

WClouds:
add crossorigin attribute for script tag


adamdbradley:
This looks great! We'll get `0.7.0` shipped, then merge it in so we can do more internal testing. Thanks

adamdbradley:
Merged thanks!

PR #631, opened 2018-03-17
fix(types): Render HydrateAnchor

arjunyel:
"While string index signatures are a powerful way to describe the “dictionary” pattern, they also enforce that all properties match their return type. This is because a string index declares that obj.property is also available as obj["property"]. " https://www.typescriptlang.org/docs/handbook/interfaces.html


adamdbradley:
thanks

Issue #632, opened 2018-03-18
node-http-server.spec.ts fails on Windows


tja4472:
**Stencil version:**

```
@stencil/core@0.7.1
```

**I'm submitting a:**

[x] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://forum.ionicframework.com/ or https://stencil-worldwide.slack.com

**Current behavior:**

Test fails on Windows

**Expected behavior:**

Test passes on Windows

**Steps to reproduce:**

npm test node-http-server.spec.ts

```bash
 FAIL  src\sys\node\test\node-http-server.spec.ts
  node-http-server
    × getFilePath w/ baseUrl and dir (66ms)
    × getFilePath w/ baseUrl (1ms)
    × getFilePath, defaults w/ querystring and hash (1ms)
    × getFilePath, defaults (4ms)

  ● node-http-server › getFilePath w/ baseUrl and dir

    expect(received).toBe(expected)

    Expected value to be (using Object.is):
      "/my/network/drive/some/file/system/dir/data.json"
    Received:
      "\\my\\network\\drive\\some\\file\\system\\dir\\data.json"

      20 |       } as d.OutputTargetWww
      21 |     ];
    > 22 |     validateConfig(config);
      23 |     outputTarget = config.outputTargets[0];
      24 |     const url = '/some/web/url/data.json?v=123#hello';
      25 |

      at Object.it (src/sys/node/test/node-http-server.spec.ts:22:19)

  ● node-http-server › getFilePath w/ baseUrl

    expect(received).toBe(expected)

    Expected value to be (using Object.is):
      "/www/data.json"
    Received:
      "\\www\\data.json"

      34 |         type: 'www',
      35 |         baseUrl: '/docs'
    > 36 |       } as d.OutputTargetWww
      37 |     ];
      38 |     validateConfig(config);
      39 |     outputTarget = config.outputTargets[0];

      at Object.it (src/sys/node/test/node-http-server.spec.ts:36:19)

  ● node-http-server › getFilePath, defaults w/ querystring and hash

    expect(received).toBe(expected)

    Expected value to be (using Object.is):
      "/www/data.json"
    Received:
      "\\www\\data.json"

      42 |     const p = getFilePath(outputTarget, url);
      43 |     expect(p).toBe('/www/data.json');
    > 44 |   });
      45 |
      46 |   it('getFilePath, defaults w/ querystring and hash', () => {
      47 |     config = new TestingConfig();

[...]

PR #633, opened 2018-03-18
node-http-server.spec.ts fails on Windows

tja4472:
#632


adamdbradley:
Thanks for catching this!

Issue #637, opened 2018-03-20
validate-config.spec.ts fails on Windows


tja4472:
**Stencil version:**

```
@stencil/core@0.7.2
```

**I'm submitting a:**

[x] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://forum.ionicframework.com/ or https://stencil-worldwide.slack.com

**Current behavior:**

Test fails on Windows
**Expected behavior:**

Test passes on Windows
**Steps to reproduce:**

npm test validate-config.spec.ts

```bash
 FAIL  src\compiler\config\test\validate-config.spec.ts
  validation
    √ should default watch to false
    √ should set devMode to false
    √ should set devMode to true
    √ should default devMode to false
    √ should set default generateDocs to false (1ms)
    √ should set generateDistribution to be defined
    √ should default dist false and www true
    √ should require at least one output target
    √ should set devInspector false
    √ should set devInspector true
    √ should default devInspector false when devMode is false (1ms)
    √ should default devInspector true when devMode is true
    √ should set default generateDocs to false
    √ should set generateDistribution to be defined (1ms)
    √ should default dist false and www true
    √ should require at least one output target
    √ should throw error for invalie outputTarget type (1ms)
    √ should default add www type to outputTarget
    √ should default outputTargets with www
    enableCache
      √ set enableCache true (8ms)
      √ set enableCache false (7ms)
      √ default enableCache true (1ms)
    buildAppCore
      √ set buildAppCore true (1ms)
      √ set buildAppCore false
      √ default buildAppCore true (1ms)
    es5 build
      √ set buildEs5 false (1ms)
      √ set buildEs5 true (1ms)
      √ prod mode default to both es2015 and es5
      √ dev mode default to only es2015 (1ms)
    include/exclude globs
      × should default include glob (9ms)
      √ should default exclude glob (1ms)
    hydrate css
      √ should set hydratedCssClass
      √ should default hydratedCssClass (1ms)
    hashed filenames
      √ should throw error when hashedFileNameLength too large
      √ should throw error when hashedFileNameLength too small (1ms)
      √ should set from hashedfilenamelength
      √ should set hashedFileNameLength from function (1ms)
      √ should set hashedFileNameLength
      √ should default hashedFileNameLength (1ms)
      √ should default hashFileNames to false in watch mode despite prod mode
      √ should default hashFileNames to true in prod mode
      √ should default hashFileNames to false in dev mode
[...]

PR #638, opened 2018-03-20
test(normalize): fix validate-config.spec on windows

tja4472:
#637 validate-config.spec.ts fails on Windows


adamdbradley:
Thanks for fixing these tests on windows!

PR #645, opened 2018-03-20
fix: Avoid Object.entries (breaks Node 6)

mattdsteele:
I'm on Windows and a lot of tests still fail, but a quick `npm link` against a project worked, so caveat emptor :)


adamdbradley:
thanks for catching this!

Issue #545, opened 2018-02-16
Stencil unescapes unicode characters in the content attribute incorrectly


richardruiter:
**Stencil version:**

```
 @stencil/core@0.6.0-0
```

**I'm submitting a:**

[x] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://forum.ionicframework.com/ or https://stencil-worldwide.slack.com

**Current behavior:**

there is a bug in stencil that as unescapes unicode characters in the content attribute incorrectly

**Expected behavior:**

**Steps to reproduce:**

```css
.myclass:before {
   content: "\F113";
}
```

In the component is compiled to
```js
static get style() { return ".myclass:before {content: \"\F113\";}" }
```

When inspected in the browser you see that the slash is removed
```css
.myclass:before {content: "F113";}
```

adamdbradley:
Ah good catch, we'll get it fixed up. I thought we had some unit tests around this but apparently they're no good 😬 

Thanks for filing the issue

PR #657, opened 2018-03-22
fix(styles): escape unicode chars & octal literals in css

### LINKED ISSUE #656: Styles compiler breaks bundle/css when escaping octal literals
**Stencil version:**

```
 @stencil/core@0.7.2
```

**I'm submitting a:**

[x] bug report
[ ] feature request

**Current behavior:**
White screen & generated build does not load

**Expected behavior:**
Generated build to work!

**Steps to reproduce:**
1. Create a component
2. Apply a style with a rule that sets content attribute to what would look like an octal literal to the compiler in JS.
3. Build or Serve the component
4. An Error is thrown (in browser console) and nothing renders

**Related code:**
```css
// css file
.test-class:before { content: "\2014 \00A0" }
```

```tsx
// tsx file
@Component({ tag: 'test-cmp', styleUrl: 'test-cmp.css' })
export class TestCmp {
  render() {
    return (<div class="test-class">Test</div>);
  }
}
```

**Other information:**
```
SyntaxError: Octal escape sequences are not allowed in strict mode.
```

Related to #545 and b32d9da2ac

kshaaban-:
Fixes #656

Seems to be caused by [the replace statement here](https://github.com/ionic-team/stencil/blob/b32d9da2acfa68ed38ad0c2af4f832193ee8771c/src/compiler/style/style.ts#L149) as it escapes the first escape character it encounters. Which then breaks the escapes applied by the call to `escapeCssForJs` as it looks for an octal literal.

This reverts the change introduced for #545 , and instead fixes it by expanding the RegEx for identifying octal literals in the `escapeCssForJs` function to also identify non-digit characters.

### DIFF
--- a/src/compiler/style/style.ts
+++ b/src/compiler/style/style.ts
@@ -144,13 +144,10 @@ function hasPluginInstalled(config: d.Config, filePath: string) {
 export async function setStyleText(config: d.Config, compilerCtx: d.CompilerCtx, buildCtx: d.BuildCtx, cmpMeta: d.ComponentMeta, styleMeta: d.StyleMeta, styles: string[]) {
   // join all the component's styles for this mode together into one line
 
-  // debugging css unicode escape issues read by js is probably the funnest thing ever
-  styleMeta.compiledStyleText = styles.map(style => {
-    return style.replace('\\', '\\\\');
-  }).join('\n\n');
-
   if (config.minifyCss) {
-    styleMeta.compiledStyleText = await minifyStyle(config, compilerCtx, buildCtx.diagnostics, styleMeta.compiledStyleText);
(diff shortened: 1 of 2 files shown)


kshaaban-:
Hmm, not sure if it's better to change the regex to

```tsx
return style
  .replace(/\\[\A-Z0-7]/gi, (v) => '\\' + v)
```

Global Insensitive A-Z & 0-7 vs. Non-digit chars & 0-7
🤷‍♂️

adamdbradley:
This is great! Thanks! Regarding the regex, I'm not sure. I say we go with what we got and fix if needed. Also, thanks for the unit tests!

PR #666, opened 2018-03-24
Correct link to the Stencil Worldwide Slack

calendee:
Clicking the current link for Stencil Worldwide Slack in the [CONTRIBUTING.MD file](https://github.com/ionic-team/stencil/blob/master/.github/CONTRIBUTING.md) results in a web page that says "This invite link is no longer active."  

NOTE: This only happens when you are NOT logged into the group already.  If you ARE logged into the group, the current link works.


adamdbradley:
thanks!

Issue #670, opened 2018-03-25
build-modules.spec.ts fails on Windows


tja4472:
**Stencil version:**

```
@stencil/core@0.7.7
```

**I'm submitting a:**

[x] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://forum.ionicframework.com/ or https://stencil-worldwide.slack.com

**Current behavior:**

Test fails on Windows

**Expected behavior:**

Test passes on Windows

**Steps to reproduce:**

npm test bundle-modules.spec

```bash
Windows PowerShell
Copyright (C) Microsoft Corporation. All rights reserved.

PS C:\VSCode\stencil\fork\stencil> npm test bundle-modules.spec

> @stencil/core@0.7.7 test C:\VSCode\stencil\fork\stencil
> jest "bundle-modules.spec"

  console.log src\compiler\bundle\rollup-bundle.ts:57
    { Error: Could not resolve './../../../../__tmp__in__memory__/cmp-a.js' from entry:cmp-a.cmp-b
        at error (C:\VSCode\stencil\fork\stencil\node_modules\rollup\dist\rollup.js:168:15)
        at C:\VSCode\stencil\fork\stencil\node_modules\rollup\dist\rollup.js:21536:25
        at <anonymous> code: 'UNRESOLVED_IMPORT' }

 FAIL  src\compiler\bundle\test\bundle-modules.spec.ts
  bundle-module
    build
      × should not rebuild without js changes (1030ms)
      × should build 2 bundles of 3 components (57ms)
      × should include json files (52ms)

  ● bundle-module › build › should not rebuild without js changes

    expect(received).toEqual(expected)

    Expected value to equal:
      []
    Received:
      [{"absFilePath": null, "code": "UNRESOLVED_IMPORT", "header": "build error", "language": "javascript", "level": "error", "lines": [], "messageText": "Could not resolve './../../../../__tmp__in__memory__/cmp-a.js' from entry:cmp-a.cmp-b", "relFilePath": null, "type": "build"}, {"absFilePath": null, "header": "build error", "level": "error", "lines": [], "messageText": "rollup died
        at Object.<anonymous> (C:\\VSCode\\stencil\\fork\\stencil\\src\\compiler\\bundle\\rollup-bundle.ts:61:19)
        at Generator.throw (<anonymous>)
[...]

PR #671, opened 2018-03-25
test(windows): fix build-modules.spec.ts

tja4472:
#670 build-modules.spec.ts fails on Windows


adamdbradley:
What is the `TjaTestingCompiler` used for vs. just the TestingCompiler?

tja4472:
The TjaTestingCompiler will set the config.rootDir to 'C:\' for Windows machines so that internally the tests will be using 'C:\src\' rather than '/src/.

In addition it has methods that will convert the '/src/' paths used in .spec files to 'C:\src\' for Windows machines.

At the moment any code under test which does a path.resolve() on a Windows machine will return 'C:\..' which more than likely will cause the test to fail, as internally the test is expecting '/src...';

adamdbradley:
Is there any way this logic can be put into the TestingCompiler? I'd like to not make a specific class for just windows machines to work.

tja4472:
In theory the code in TjaTestingCompiler can replace the code in TestingCompiler .

I had it as a separate class as I have no way of testing on non windows machines.

tja4472:
Tested with ubuntu.

adamdbradley:
Going to close this PR since there are conflicting issues. We'll keep looking and various approaches to fix this. Thanks

Issue #676, opened 2018-03-28
Cannot set property 'ES6Promise' of undefined


marisaoliva:
**Stencil version:**

```
 @stencil/core@<version>
```

**I'm submitting a:**

[X ] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://forum.ionicframework.com/ or https://stencil-worldwide.slack.com

**Current behavior:**

In windows lastest version chrome or the lastest chronium version. when you compile for the first time show the error

Uncaught TypeError: Cannot set property 'ES6Promise' of undefined
    at mutua-components-guide.core.pf.js:50
    at mutua-components-guide.core.pf.js:50

**Expected behavior:**
normal behavior

**Steps to reproduce:**
1. compile files
2. open windows lastest version chrome 
3. saves files
4. error
-->

**Related code:**

```tsx
// insert any relevant code here
```

**Other information:**

jgw96:
@marisaoliva I have not seen this issue on my surface book running the latest version of windows 10 and latest stable version of Chrome. Do you have a repo I could use to reproduce this?

bitflower:
Might be related https://github.com/ionic-team/stencil-postcss/issues/6

jgw96:
Thanks @bitflower . @marisaoliva are you using PostCSS ?

jagreehal:
@jgw96 @marisaoliva I get this error and I am using PostCSS

jgw96:
Hello all! Since it looks like this is an issue specifically with PostCSS I am going to be closing this issue in favor of https://github.com/ionic-team/stencil-postcss/issues/6 . Ill ping someone on the team who knows about the PostCSS plugin. Thanks for using Stencil everyone!

marisaoliva:
Hello, I'm not use postcss, I use sass, and  I also  building es5 

thanks!!

bfmatei:
PR created to fix this issue.

brandyscarney:
I am able to reproduce this in `@ionic/core` by running the following command:

```
npm run dev --no-open --es5
```

The initial build works fine, but after I make a change to a file (I specifically changed a `*.scss` file) after the rebuild I am seeing the following error in Chrome:

```
ionic.core.pf.js:49 Uncaught TypeError: Cannot set property 'ES6Promise' of undefined
```

This error does not exist in Firefox.

tyvdh:
[...]

PR #681, opened 2018-03-29
fix(loader): force window as global object in promise polyfill

bfmatei:
Using `type="module"` on the polyfills tag forces `this` to be `undefined` ([spec](https://www.ecma-international.org/ecma-262/8.0/#sec-module-environment-records-getthisbinding)).

This PR updates the promise polyfill to current latest version (4.2.4) and also enforces `this` as the global object.

Issue: https://github.com/ionic-team/stencil/issues/676


adamdbradley:
This error was a side effect of a build error. In general the promise polyfill shouldn't even be loaded with a type=module script. Thanks for debugging though! It helped pin point the root issue, thanks.

PR #712, opened 2018-04-10
Fix the <dialog> TypeScript interface

### LINKED ISSUE #711: Cannot create a `<dialog>` element with the `open` attribute
**Stencil version:**

```
@stencil/core@0.7.7 
```

**I'm submitting a:**

[x] bug report
[ ] feature request

**Current behavior:**

I'm trying to add the `open` attribute to a `<dialog>` element, but it does not work

**Expected behavior:**

This should be fine, since `<dialog>` is an HTML element and has an `open` property

**Steps to reproduce:**
Try the below component out and you'll get a TypeScript compilation error

**Related code:**

```tsx
import { Component, Prop } from '@stencil/core';

@Component({
  tag: 'os-modal',
  shadow: true
})
export class OSModal {
  @Prop() open: boolean = false;

  render() {
    return (
      <dialog open={this.open}>
        <slot name="header" />
        <slot />
      </dialog>
    );
  }
}
```

**Other information:**
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog

alexlafroscia:
I used the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement) as a reference for the interface to define.

Closes #711

### DIFF
--- a/src/declarations/jsx.ts
+++ b/src/declarations/jsx.ts
@@ -40,7 +40,7 @@ declare global {
       del: JSXElements.DelHTMLAttributes;
       details: JSXElements.DetailsHTMLAttributes;
       dfn: JSXElements.HTMLAttributes;
-      dialog: JSXElements.HTMLAttributes;
+      dialog: JSXElements.DialogHTMLAttributes;
       div: JSXElements.HTMLAttributes;
       dl: JSXElements.HTMLAttributes;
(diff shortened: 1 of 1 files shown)


adamdbradley:
Awesome thanks!

PR #725, opened 2018-04-12
fix(css-shim): Fix relative paths

### LINKED ISSUE #602: CSS variable shimming messes up relative URLs in CSS
**Stencil version:**

```
 @stencil/core@0.6.18
```

**I'm submitting a:**

[X] bug report
[ ] feature request

**Current behavior:**

The `<style>` that the CSS variable shimming applies can cause relative resources to be loaded from the wrong server.

If the CSS file is loaded from another domain and contains relative URLs, the setting of `innerHTML` will cause those URLs to be loaded relative to the document’s domain, rather than the domain that the CSS file came from.

This can cause 404s, since the resources aren’t available, which in our experience it breaks `@font-face` declarations in IE, overwriting the correct URLs with ones that 404.

**Expected behavior:**

The resources loaded by the shimmed CSS should be the same as those in the original CSS, relative to the domain that the CSS came from.

**Steps to reproduce:**

1. Use IE 11, which needs CSS variable support shimmed.
1. `<link>` to a CSS file on another domain with relative URLs (_e.g._ `url(/images/logo.png)`) that also includes CSS variables (or has `--` somewhere in it). You can use our styles at: https://patterns.boston.gov/css/public.css
1. In the web inspector, notice the 404s for assets

**Related code:**

https://github.com/ionic-team/stencil/blob/6f27283729b3f4a20acf3c718ed3a6474515d14a/src/client/css-shim/init-css-shim.ts

**Other information:**

juanmiguelbesada:
Fix #602 

I'm not an expert in regex, so maybe there is a better one or a simple way of doing this. So please review and comment.

Note that this fix just prepend the "import path" to every url() in imported css. This will generate paths like "assets/css/../fonts/testfont.tff".

### DIFF
--- a/src/client/css-shim/init-css-shim.ts
+++ b/src/client/css-shim/init-css-shim.ts
@@ -34,6 +34,9 @@ function loadLinkStyles(doc: Document, customStyle: CustomStyle, linkElm: HTMLLi
 
     if (hasCssVariables(text)) {
       const styleElm = doc.createElement('style');
+      if (hasRelativeUrls(text)) {
+        text = fixRelativeUrls(text, url);
+      }
       styleElm.innerHTML = text;
       linkElm.parentNode.insertBefore(styleElm, linkElm);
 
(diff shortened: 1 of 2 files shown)


adamdbradley:
Cool thanks!

Issue #486, opened 2018-02-06
Code splitting - chunk file names should be versioned


cjorasch:
**Stencil version:**
```
 @stencil/core@0.3.0
```

**I'm submitting a:**
- [x] bug report

**Current behavior:**
I have been working with the new code splitting features and am very impressed with the progress.  Awesome job delivering an auto code splitting approach with support for ES5!

One thing I noticed is that when you do a production build the `chunk.js` files that get generated do not use the filename hash versioning scheme that is used by the bundles.  For example, the output might include a file called `chunk1.js`.  This can cause problems in file caches if you build a new version and `chunk1.js` changes to include entirely different code.  Immutable bundle files should use immutable chunk references to avoid breaking changes.

**Expected behavior:**
Chunk files should include a content hash in the file name similar to what is used for bundles.

jthoms1:
I created an issue with rollup to see what their preferred approach would be. https://github.com/rollup/rollup/issues/1995  Most likely we are going to need to create a PR to fix.

jthoms1:
It appears that this they are working on a PR to fix this within rollup. https://github.com/rollup/rollup/issues/1995#issuecomment-372680114

jthoms1:
PR is now open and being discussed. https://github.com/rollup/rollup/pull/2068

manucorporat:
https://github.com/rollup/rollup/pull/2068#issuecomment-376060844

@adamdbradley @jthoms1 looks like it was merged!

jthoms1:
Yep this should be fixed with their next release.

PR #742, opened 2018-04-17
chore(): increment version of rollup.

jthoms1:
This PR will close issue #486.


src/compiler/bundle/test/bundle-modules.spec.ts
```diff
@@ -110,11 +110,12 @@ describe('bundle-module', () => {
 
       const r = await c.build();
       expect(r.diagnostics).toEqual([]);
+      console.log(r.filesWritten);
```
juanmiguelbesada: Is this console.log expected?

jthoms1: yep, for now. I am troubleshooting an issue. :)

PR #746, opened 2018-04-17
fix(queue): fix high priority queue

manucorporat:


jthoms1:
What is the likelyhood you write some tests for this?

PR #751, opened 2018-04-19
Prevent IE11 from StackOverflowing when using base64 values as CSS fallback

### LINKED ISSUE #749: IE11: 'Out of stack space' when using data URI as fallback CSS variable
**Stencil version:**

```
 @stencil/core@0.7.25
```

**I'm submitting a:**

[x] bug report
[ ] feature request

**Current behavior:**
Starting with 0.7.25, using a data URIs as a fallback for CSS variables causes a stack overflow in IE11. i.e.:

```
// in stylesheet
div {
  background: var(--overwritable-val, url('data:image/svg+xml, ....'));
}
```

will throw an error and not render properly.

**Steps to reproduce:**

Sample project here: https://github.com/mattdsteele/stencil-ie11-svg-vars
Clone, run `npm run dev --es5` and open in IE11.

You should see this error: `SCRIPT28: Out of stack space`

Other browsers show the intended effect (display the SVG tiger).

The offending CSS is here: https://github.com/mattdsteele/stencil-ie11-svg-vars/blob/master/src/components/my-component/my-component.css#L4-L6

**Other information:**

The use case for this is to let external users overwrite a background SVG.

Possibly related: #393 

I'll try to put together a Karma test as well, just need to learn how to do it first :)

mattdsteele:
Fixes #749 

(This is my first non-trivial PR, so nits on style/syntax are welcome. Thanks!)

### DIFF
--- a/src/client/css-shim/style-properties.ts
+++ b/src/client/css-shim/style-properties.ts
@@ -81,7 +81,6 @@ export class StyleProperties {
     if (property) {
       if (this.hasNonNestedSemicolon(property)) {
         property = this.valueForProperties(property, props);
-
       } else {
         // case (2) variable
         let fn = (prefix: any, value: any, fallback: any, suffix: any) => {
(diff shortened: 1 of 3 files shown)


adamdbradley:
awesome thanks! And thanks for adding unit tests 🚀

PR #779, opened 2018-05-05
[FEATURE] Keep gitkeep files when emptying a directory

Serabe:
Both `stencil-app-starter` and `stencil-component-starter` launch
concurrently `stencil build --dev --watch` and `stencil-dev-server`.
This was causing a problem where launching it twice was required to see
the built app.

The first approach was to add a `.gitkeep`, and though that requires some
more tweaking (see https://github.com/ionic-team/stencil-app-starter/pull/50), it is a good one. Sadly, compiler completely empties the output targets
thus removing the `.gitkeep` file.

As using this kind of file is common in the git community, this PRs checks
for the existence of such a file and recreates it after emptying the dir.

### DIFF
--- a/src/compiler/build/empty-dir.ts
+++ b/src/compiler/build/empty-dir.ts
@@ -1,4 +1,5 @@
 import * as d from '../../declarations';
+import * as p from 'path';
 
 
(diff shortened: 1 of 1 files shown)


Serabe:
karma tests seem to be failing due to it not being able to load the browsers.

PR #780, opened 2018-05-06
chore(sys): calculate relative path to 'sys-util.js'

fourcube:
Calculate the path to the `sys-util.js` file just like it's done in the node-sys.ts. I needed this change to programatically invoke the stencil compiler from within a third party package.


adamdbradley:
cool thanks!

PR #784, opened 2018-05-08
Remove console.log

juanmiguelbesada:
I think this console.log is not expected to be here. I'm correct?


jthoms1:
Sorry fixed this without seeing the PR 47f549542d4ade82f8496b3eeed059941394e0f4.  You are correct though and thanks for opening it.  Just wish I saw it before my change.

PR #787, opened 2018-05-08
refactor(core): reduce size stencil core

manucorporat:


manucorporat:
Reopening since this will be eventually merged once there are not higher priority issues in the queue :)

Issue #786, opened 2018-05-08
CORS issue in loader.


mattcarroll-cs:
**Stencil version:**
```
 @stencil/core@0.8.0
```

**I'm submitting a:**

[x] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://forum.ionicframework.com/ or https://stencil-worldwide.slack.com

**Current behavior:**
The current esm loader process fails when requests are cross domain.

**Expected behavior:**
Components should load even if scripts resource are cross domain.

**Steps to reproduce:**
Attempt to load scripts resources cross domain.

**Related code:**

src/client.loader.ts 
```tsx
  } else {
    // let's do this!
    x.src = resourcesUrl + appCore;
    x.setAttribute('type', 'module');
    x.setAttribute('crossorigin', true);
}
```

`crossorigin` needs to be set to `use-credentials`

Creating a PR for the fix.

mattcarroll-cs:
https://github.com/ionic-team/stencil/pull/788

PR #788, opened 2018-05-08
fix(loader): set crossorigin to use-credentials

mattcarroll-cs:
Updates the `crossorigin` attribute on the script tags generated from the loader.ts file to the value of 'use-credentials' instead of true. 
https://github.com/ionic-team/stencil/issues/786

### DIFF
--- a/src/client/loader.ts
+++ b/src/client/loader.ts
@@ -110,7 +110,7 @@ export function init(
     // let's do this!
     x.src = resourcesUrl + appCore;
     x.setAttribute('type', 'module');
-    x.setAttribute('crossorigin', true);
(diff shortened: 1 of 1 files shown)


mattcarroll-cs:
karma tests are failing due to: `BrowserStackTunnel: binary not present`

adamdbradley:
I think it's good to go, thanks!

tavikukko:
was this ever merged? checked the master branch and seems that it still sets 'crossorigin' = true. was it reverted for a reason?

John0x:
@adamdbradley could you please revert to this PR?
Having issues with CORS, because this was apparently reverted.

PR #792, opened 2018-05-09
test(slot): Add failing test demonstrating rendering/array bug with slots

mattdsteele:
Here's a potentially relevant Slack conversation: https://stencil-worldwide.slack.com/archives/C789G3X1R/p1525872210000427

This adds a test which fails in Firefox and Edge. The page renders like this in Chrome:
image

But renders out-of-order in ES5 browsers:
image

This seems *really* similar to the "slot-array-top" test, which oddly enough is still passing. So I'm not sure what's going on.

Let me know if you need any more details. Thanks!

### DIFF
--- a/test/karma/src/components.d.ts
+++ b/test/karma/src/components.d.ts
@@ -543,6 +543,39 @@ declare global {
 declare global {
 
   namespace StencilComponents {
+    interface SlotArrayBasic {
+
+    }
+  }
+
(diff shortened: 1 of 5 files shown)


adamdbradley:
Cool thanks for the test! The merge conflict is cuz I renamed a directory, but we'll be able to get it sorted, thanks

adamdbradley:
fixed! https://github.com/ionic-team/stencil/pull/792 thanks

PR #826, opened 2018-05-28
perf(compiler) Use Set over array

tanguyantoine:
Set has a build-in `has` methods which is more efficient that `indexOf` of array.


adamdbradley:
Cool thanks

PR #841, opened 2018-06-09
fix(cssvars): implements scoped css var shim

manucorporat:
Implements a CSS var shim with support for scoped components:

- CSS parser
- CSS Scope resolver
- Direct Acyclic Graph algorithm resolution for CSS var values
- CSS template generation and serialization
- High level API to be integrated into the render loop of stencil


rinckemd:
Should :root { vars } in a link element in head be picked up by css-shim?  If I change from :root to something else it works fine. I'm trying to add global css custom properties.  Thanks

PR #847, opened 2018-06-12
fix(compiler): Remove duplicate TranspileResults declaration

tanguyantoine:
because it is export within `build.ts` and `transpile.ts` then it raises an error warning within src/declaration/index.ts

* https://github.com/ionic-team/stencil/blob/master/src/declarations/build.ts#L136
* https://github.com/ionic-team/stencil/blob/master/src/declarations/transpile.ts#L4

imported 

* https://github.com/ionic-team/stencil/blob/master/src/declarations/index.ts#L3
* https://github.com/ionic-team/stencil/blob/master/src/declarations/index.ts#L29


tanguyantoine:
Looks to be fixed :-)

PR #867, opened 2018-06-20
test(): add tests for dynamic slot content when nesting custom elements

cary-smith:
**Stencil version:**

```
 @stencil/core@0.9.9
```

**I'm submitting a:**

[x] bug report
[ ] feature request

**Current behavior:**

So there are two slightly related issues going on that I have added tests for:

1. Dynamic slot content is not consistently rendered in the proper location
2. Dynamic slot content does not stay in the intended order

In both cases a parent element is mapping child elements into a custom element that is part of their template. On initial render everything is exactly as it should be, but if a state change updates the mapped elements, then things start to break down.

In shadow-native browsers, the order is maintained when new values are added, but the elements are added inside the shadow DOM instead of being appended to the element as they appeared initially so the component can slot them correctly. 

In the case of polyfilled browsers like Firefox, the content is always rendered in the correct location, but the order is incorrect when new values are added. 

**Expected behavior:**

1. Rendering locations should be consistent
2. Order of output should be consistent to intent

**Steps to reproduce:**

I added the `shadow-dom-slot-mapped` and `shadow-dom-slot-mapped-order` tests. The first fails in Chrome due to render location, the second fails in Chrome (due to render location) and in Firefox because it is out of order.

**Other information:**

I unfortunately do not know the code that may be causing these issues, but figured the tests would help you track it down.


adamdbradley:
This is awesome, thanks for the unit tests!

Fixed one test so far: https://github.com/ionic-team/stencil/commit/2fddcbde060f22a660b0b71e61db3db18ec8d061

cary-smith:
Seems the better approach to assisting (especially if I have no idea where its happening beyond 'somewhere in the slotting mechanism'). i will try to continue this approach for bug when I run across them.

PR #880, opened 2018-06-25
feat(compiler): Add typescript transformer customization

tanguyantoine:
It opens typescript internal behaviours and empower component building with StencilJS.

Ex: You can create a plugin which automatically scope all component tag names

sample of `stencil.config.js`
```js
const MyTransformer = require('my-custom-transformer');

exports.config = {
  namespace: "go-goats",
  outputTargets: [
    {
      type: 'dist'
    },
    {
      type: 'www',
      serviceWorker: false,
    }
  ],
  customTransformers: {
    prependBefore: [MyTransformer()]
  }
}

```


tanguyantoine:
closed in favor of #881

PR #881, opened 2018-06-25
feat(compiler): Add typescript transformer customization

tanguyantoine:
It opens typescript internal behaviours and empower component building with StencilJS.

Ex: You can create a plugin which automatically scope all component tag names

sample of `stencil.config.js`
```js
const MyTransformer = require('my-custom-transformer');

exports.config = {
  namespace: "go-goats",
  outputTargets: [
    {
      type: 'dist'
    },
    {
      type: 'www',
      serviceWorker: false,
    }
  ],
  customTransformers: {
    prependBefore: [MyTransformer()]
  }
}

```


adamdbradley:
Sorry, but before we add something like this to the compiler we'd like to discuss it further in slack. Currently we're leaning toward not exposing this as a public api, which may work great to start, but over time may cause many breaking changes and incompatible plugins, which ultimately would making building stencil apps more of a challenge and times goes on. Happy to talk over the ideas in slack and come up with a future proof way to accomplish what you're looking for. Thanks

tanguyantoine:
Well, I tried to reach out to you multiple time, created multiple PR and asked for feedback without response. 
We achieved our goal in a different way and works like a charm.

Keep going guys, you do a great job. 👍

PR #885, opened 2018-06-26
fix(init-css-shims): CSS_URL_REGEXP check relative to root paths

JoshVazq:
Fixed: CSS_URL_REGEXP was checking only for absolute url ("http://domain/path") and taking relative to root url ("/relative/to/root") as relative, replacing it for "relative/path//relative/to/root"


adamdbradley:
Awesome work! Thanks!

PR #902, opened 2018-07-05
SlotAttributes interface must include slot attribute

dfeyer:
This is useful when you want to pass a slot down to an other slot


dfeyer:
Not really sure if the broken test are related to this PR, with your help I can maybe solve the problem.

adamdbradley:
cool thanks

PR #914, opened 2018-07-09
chore(): replace uglify with terser

manucorporat:
DO NOT MERGE

Uglify-es is not longer maintained and there are bug reports affecting stencil that WILL never be fixed.

Terser is the fork of uglify where rollup, webpack and two of the main maintainers of uglify are involved.


adamdbradley:
Will do, we'll get this in. Currently working on another branch to bring in terser.

PR #922, opened 2018-07-11
adjusted es5-disabled information to reflect that stencil-component-s…

franktopel:
…tarter no longer has a "dev" script

Fixes https://github.com/ionic-team/stencil-component-starter/issues/60


adamdbradley:
Good call, thanks!

PR #932, opened 2018-07-16
Transpile test to es2015 and not es5

Serabe:
There is a problem with es6 classes and es5 not being fully compatible.
In particular, calling super in constructor is a problem if code is
compiled down to es5.

AFAIK, there is no simple way to keep compatibility in this issue
between es5 and es6 without user intervention.


manucorporat:
Since node 6 supports es6, I think this is good to merge

Issue #940, opened 2018-07-18
[Bug] Livereload doesn't work when stencil app is called by IP address


RafaelKr:
**Stencil version:**

```
 @stencil/core@0.10.7
```

**I'm submitting a:**

[x] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**
The fix of my issue #921 caused a regression where livereload isn't working when you load the page by IP address from another network device.

**Expected behavior:**
I would like to submit a PR, where a script is injected which creates the iframe on the client side using `location.origin` instead of injecting the iframe directly.

**Other information:**
I already prepared this PR locally and it's working, also the tests are running. But if I want to test it in a stencil application `npm start` hangs up while building (also when I do a `git reset --hard <commit_hash_of_0.10.7>`, so it's not related to my code changes). I already posted this problem in the `contributors` channel of the stencil slack.
If you can help me there, I'll also update [stencil/.github/CONTRIBUTING.md](stencil/.github/CONTRIBUTING.md), if it's helpful.

Update: My PR is perfectly working in a newly created stencil app. More Info in the slack channel. I submitted the PR: #941

PR #941, opened 2018-07-18
fix(dev-server): set the correct domain for dev-server iframe

RafaelKr:
This PR moves the creation of the dev server iframe to the client side by injecting a script, which uses `location.origin` + `{{DEV_SERVER_URL}}` as the src-attribute of the created iframe.

This ensures, that the dev-server is working for localhost and also over the network, when the stencil app is called over an IP address.

Related issue: #940


RafaelKr:
This PR is obsolete. I'm currently creating a new one.

PR #952, opened 2018-07-23
fix(set-accesor): boolean values

manucorporat:
- Setting boolean true / false to an attribute should render as a boolean attribute in HTML

```html
// tappable: false
<div></div>
```

```html
// tappable: true
<div tappable></div>
```

- Passing a boolean to a string prop, should cast the boolean to string and render as string

```html
// str: false
<my-component str="false"></my-component >
```

```html
// str: true
<my-component  str="true"></my-component >
```

- Passing a string to an attribute should always render the string

```html
// aria-hidden: "false"
<div aria-hidden="false"></div>
```

```html
// aria-hidden: "true"
<div aria-hidden="true"></div>
```

### DIFF
--- a/src/core/test/render.spec.ts
+++ b/src/core/test/render.spec.ts
@@ -373,8 +373,7 @@ describe('instance render', () => {
         empty: '',
         class: 'a b c my-class',
         number: '12',
-        appear: 'true',
-        'no-appear': 'false'
+        appear: '',
       });
     });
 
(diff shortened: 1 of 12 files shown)


src/core/test/render.spec.ts
```diff
@@ -373,8 +373,7 @@ describe('instance render', () => {
         empty: '',
         class: 'a b c my-class',
         number: '12',
-        appear: 'true',
-        'no-appear': 'false'
```
manucorporat: Funny fact, today the 'no-appear' attributes appears :)

Issue #950, opened 2018-07-22
🐛 Error thrown when using HTTPS protocol on dev server


CookieCookson:
**Stencil version:**
```
 @stencil/core@0.10.7
```

**I'm submitting a:**

[x] bug report
[ ] feature request

**Current behavior:**
When setting the `protocol` property to `https` on the `devServer` object in `stencil.config.js`, an error is thrown and no dev server is started.

**Expected behavior:**
The dev server should start up and use the HTTPS protocol.

**Steps to reproduce:**
1. Edit `stencil.config.js`
2. In the `devServer` object, set `protocol` to `https`
3. Run `stencil build --dev --watch --serve`

**Related code:**
```
...
devServer: {
    protocol: "https"
},
...
```

**Other information:**

Terminal log:
```bash
[08:13.6]  transpile started ...
[08:15.7]  transpile finished in 2.07 s

[ ERROR ]bject]

[08:15.7]  module map started ...
[08:15.7]  generate styles started ...
[08:16.2]  generate styles finished in 475 ms
[08:16.4]  module map finished in 750 ms
[08:16.4]  generate bundles started ...
[08:17.7]  generate bundles finished in 1.23 s
[08:17.7]  generate app files started ...
[08:20.1]  generate app files finished in 2.46 s
```

adamdbradley:
Thanks for opening the issue. If anyone is able to assist fixing this please hit us up on the stencil slack channel. Thanks

CookieCookson:
@adamdbradley You mentioned in the PR that the `devcert-san` package and its dependencies are too involved for this, would you be happy for me to submit a PR using a different package?

e.g. `webpack-dev-server` uses https://www.npmjs.com/package/selfsigned for generating SSL certificates, which has a lot less dependencies!

PR #959, opened 2018-07-26
fix(dev-server): uncomment ssl generation

CookieCookson:
Fixes https://github.com/ionic-team/stencil/issues/950

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
   },
   "dependencies": {
     "chokidar": "2.0.3",
+    "devcert-san": "0.3.3",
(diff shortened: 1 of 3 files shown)


adamdbradley:
I looked into the devcert-san and devcert packages and their dependencies is much more involved that I'd like to add to stencil at the moment. It also appears it has changed significantly since we used it for the original dev-server. I think we'll hold off for a bit and reevaluate later. Thanks

CookieCookson:
Ok - I’ll try and see if there are any other lighter weight packages we can use for SSL generation.

Issue #961, opened 2018-07-27
CLI perf: migrate to turbocolor


manucorporat:
https://github.com/jorgebucaran/turbocolor

Rollup already did: https://github.com/rollup/rollup/pull/2339

> This PR swaps chalk for turbocolor. It ought to give us a humble perf boost as turbocolor loads >20x faster and applies styles >18x faster than chalk. Terminal color detection is also built-in, so I was able to simplify the code in one place.

AND most important:

**No dependencies!**

manucorporat:
cc @jorgebucaran

adamdbradley:
You had me at "no dependencies" ❤️

PR #964, opened 2018-07-28
refactor(turbocolor): rename to tc and save horizontal space

jorgebucaran:
Howdy! Rename `turbocolor` to `tc` to save horizontal space. Just a stylistic preference. I didn't have a chance to comment in https://github.com/ionic-team/stencil/issues/961 before 6822c23, but I would have advocated for that there. If you prefer the more readable (but longer)`turbocolor`, that's understandable too.

Let me know if this looks right. 

Cheers! 👋 

cc @adamdbradley & @manucorporat


src/sys/node/bundles/sys-util.js
```diff
@@ -1,4 +1,4 @@
-exports.turbocolor = require('turbocolor');
+exports.tc = require('turbocolor');
```
manucorporat: Personally, i am fine with the PR, but this line.

In order to reduce the number of "dependencies", we bundle devDependencies in sysUtils (for faster installation), I think it makes sense to keep the full name of the package here.

Then, i am fine to use the `tc` alias.

```
const tc = this.tc = sysUtil.turbocolor;
```

manucorporat:
cc @jthoms1 CI is failing but i don't think it's related with the changes implemented in this PR

src/sys/node/bundles/sys-util.js
jorgebucaran: @manucorporat Sure. I changed this line and also the NodeLogger constructor to grab the library from `sysUtil.turbocolor`. Did I miss anything?

PR #1005, opened 2018-08-09
Update NodeResolveConfig

acoreyj:
Update `NodeResolveConfig` `customResolveOptions` to allow `string[]`. Some node-resolve options, such as `moduleDirectory`, can take a string array.


jthoms1:
Looks good to me.  Just checked against the 'resolve' module that 'rollup-plugin-node-resolve' uses.

PR #1027, opened 2018-08-21
feat(dev-server): add ssl option to DevServerConfig

adamlacombe:
A temporary solution until a lightweight SSL generation package is implemented (#959 / #950)


CookieCookson:
@adamlacombe LGTM! Have you added something to your project's scripts to generate the SSL automatically before the build script runs?

adamlacombe:
Nah, had a LetsEncrypt cert on my machine for something else so I just set the dev-server domain to match and gave it a different port.

PR #1042, opened 2018-08-27
feat(compiler): methods should return a promise

manucorporat:


mattdsteele:
What are the implications of this change? Will we able to invoke methods without waiting for `componentOnReady`, or was this intended for another purpose?

julesjanssen:
It'd be really interesting to know why this change was made.

PR #1072, opened 2018-09-07
insert style[data-styles] tag after meta[charset] to prevent parsing …

alexanderknapstein:
…errors while behaving right order of styles.

Hi @adamdbradley,

as mentioned a few days ago, I would like to please you to check this litte optimization, whereby the style[data-styles] (and so all the other following) are inserted after the meta[charset] tag, which caused rendering errors.

Thanks, Alex


adamdbradley:
awesome, thanks!

PR #1073, opened 2018-09-07
feat(puppeteer): allow specification of Chrome executable

mattdsteele:
In some environments (such as behind corporate proxies) it's difficult to download Chromium with Puppeteer.

There are a handful of environment variables to prevent Puppeteer from downloading Chromium during the install: https://pptr.dev/#?product=Puppeteer&version=v1.8.0&show=api-environment-variables

When configured this way, there needs to be a way to launch Puppeteer against the system Chrome (or another Chromium install). This adds a config option, which can get passed into `puppeteer.launch()`

### DIFF
--- a/src/declarations/testing.ts
+++ b/src/declarations/testing.ts
@@ -104,6 +104,11 @@ export interface TestingConfig {
   browserArgs?: string[];
 
   /**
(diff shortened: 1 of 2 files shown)


adamdbradley:
good thinkin, merged! Thanks

PR #1097, opened 2018-09-17
WIP test: add failing test to demo my lack of understanding re: focus API

mattdsteele:
Don't merge as is, but more this is to help me understand how to use the Puppeteer API and compare it against Stencil elements. I figure once it's working, it out I'll update the fork and others could use it as a reference.

My use case is: I have a component with a `focusOnOpen` prop. When set, a part of the template is focused. I'd like to verify this behavior.

What's happening: I've tried a few approaches to do comparisons of the Stencil E2EElement I get with `page.find` and the Puppeteer element I get with `page.evaluateHandle`, but the test hangs when I do any comparisons.

I'm probably using the API wrong, but can't quite figure out what's going on. Any help would be appreciated!

### DIFF
--- a/test/example/src/components.d.ts
+++ b/test/example/src/components.d.ts
@@ -57,6 +57,13 @@ export namespace Components {
     'onMyWindowEvent'?: (event: CustomEvent<number>) => void;
   }
 
+  interface FocusCmp {
+    'focusOnOpen': boolean;
+  }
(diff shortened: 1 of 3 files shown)


adamdbradley:
In this case, `document.activeElement` is within the browser context within puppeteer, and `document` isn't the same object in the node runtime as the one in the browser. We'll have to think of a good API to help test the focus. Thanks for opening the PR.

PR #1103, opened 2018-09-19
mock-doc: Add getComputedStyle

mattdsteele:
As noted in Slack - we've got a few instances where we need `getComputedStyle` even in our unit tests. This stubs out the API good enough to pull a few out of the e2e bucket.


adamdbradley:
thanks!

PR #1105, opened 2018-09-21
style(compiler): Adjusted the whitespace to remove tslint errors

simplyspoke:
The **indentTypes** function was also trimming the white space from the comment before the code which would get picked up by tslint. This minor adjustment sidesteps that irritant.


adamdbradley:
great thanks!

Issue #1023, opened 2018-08-20
EventEmitter Prevent Default and dynamic event names


JasonCubic:
**I'm submitting a:**

[ ] bug report
[x ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

I see you are using custom events for your events.  I am hoping you can put some improvements into those events.  There are a few features that would really help.  First is the ability to preventDefault using a function that is pre-attached to the event.

If you're using custom events something like this to attach the function:

```js
Object.defineProperties(myEvent, {
    preventDefault: {
      value: preventDefault,
      writable: true
    }
  });
```

And then this would be the function that is attached:
```js
function preventDefault() {
  Object.defineProperty(this,
    "defaultPrevented", {
      get: function () {
        return true;
      }
    }
  );
}
```

the problem being how inside of the stencil component would you get a handle to the event that was just dispatched so that you could check if the default was prevented.  Maybe you could have emit return the event it just dispatched?

```js
const myEventHandle = this.myEventEmitter.emit();
```

Another thing is if there is an ability to set the event names using props?  The reason this would even be useful is that some libraries (like bootstrap for example) have periods in the event names.  But Frameworks like vue do not bind to event names with periods in them very easily.  I am thinking about something like this:

```js
@Event({eventName: this.propStringThatCouldBeChangedAsNeeded }) showEventEmitter: EventEmitter;
```

If any of this can already be done somehow them please let me know how.  Thank you!

jthoms1:
You can actually name Events today. This is just not very well documented.  Currently there are 4 options that can be passed to the Event Decorator.

Below shows what the default options would be. These can be changed.  Also note that by default custom events are cancellable.  So you you should be able to call `event.preventDefault()` on them.

https://developer.mozilla.org/en-US/docs/Web/API/Event/cancelable

```js
@Event({
  eventName: 'showEventEmitter',
  bubbles: true,
  cancelable: true,
  composed: true
})
showEventEmitter: EventEmitter
```

Thanks!  I am leaving this issue open to remind us that we need documentation on the event decorator.

JasonCubic:
How do you detect if an event was cancelled inside your component?

PR #1162, opened 2018-10-19
feat: get event from EventEmitter.emit allow check on e.defaultPrevented

MrAntix:
Allows check on emitted event.defaultPrevented

```
const e = this.myEmitter.emit();
if (e.defaultPrevented) ... [don't] do something
```


MrAntix:
ref: https://github.com/ionic-team/stencil/issues/1023

adamdbradley:
thanks!

PR #1163, opened 2018-10-19
docs: add detail to generated event docs

### LINKED ISSUE #1161: Add Event Type to Generated Docs
**Stencil version:**

```
 @stencil/core@<0.14.0>
```

**I'm submitting a:**

[ ] bug report
[ x ] feature request

**Current behavior:**

Documentation generated for custom events lack type information for what is emitted

**Expected behavior:**

The generated readmes and json docs include type information for the emitted event details

**Steps to reproduce:**

Generate documentation

**Other information:**

It looks like the compiler can store and eventType attribute when the EventEmitter is given one, so this information is ready to grab, just needs to be extracted into the documentation. 

I'll try to add a PR for this by the weekend, if I get some time after work.

mattcosta7:
fix #1161 

This adds a detail key to the events doc-json and adds a column for Detail to the generated readmes

It does not list void, since this value is used whether the EventEmitter is passed void or no value, where no value may not necessarily be correctly void

### DIFF
--- a/src/compiler/docs/docs-util.ts
+++ b/src/compiler/docs/docs-util.ts
@@ -157,6 +157,13 @@ interface RowData {
   isHeader?: boolean;
 }
 
+export function getEventDetailType(eventType: d.AttributeTypeInfo) {
+  if (eventType && eventType.text && typeof eventType.text === 'string' && eventType.text !== 'void') {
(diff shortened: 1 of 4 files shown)


adamdbradley:
thanks!

Issue #1176, opened 2018-10-23
Build date month is one off


rjimenezda:
**Stencil version:**
```
 @stencil/core@latest
```

**I'm submitting a:**

[x] bug report
[x] feature request

**Current behavior:**

The build date output on the banner has the UTC Month, which made me very confused as to why I had a month old build.

**Expected behavior:**

Well, I would personally expect the build date would use the UTC Month + 1, but maybe this is 100% intended.

**Steps to reproduce:**

Generate a minified build with a banner.

I've submitted a PR (#1177) with the proposed change but I mean it's just adding one to the time, so feel free to discard it.

adamdbradley:
merged

PR #1177, opened 2018-10-23
Fix build month in build timestamp

rjimenezda:
Related issue: #1176

Maybe this was intended, but I was going insane checking the build time thinking it was a month old :sweat_smile:


adamdbradley:
thanks

Issue #1008, opened 2018-08-10
support for delegatesFocus


MrAntix:
**Stencil version:**

```
 @stencil/core@0.11.3<version>
```

**I'm submitting a:**

[ ] bug report
[x] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**
not supported

**Expected behavior:**

https://developers.google.com/web/fundamentals/web-components/shadowdom#focus

jthoms1:
@adamdbradley thoughts?

zoofadoofa:
This is not really an edge case, anything that is slotting in content into shareable components will need delegatesFocus
See Demo

adamdbradley:
https://github.com/ionic-team/stencil/commit/f45c919a9675827a5006702218982022b3e0ec0d

PR #1190, opened 2018-10-30
feat(component): allow delegatesFocus on the component decorator

MrAntix:
ref: https://github.com/ionic-team/stencil/issues/1008


adamdbradley:
I like this, but wonder if we should instead have the options object passed in, rather than making a new option to the component decorator.

manucorporat:
Agreed with Adam, i think it should be an object, probably allowing the `shadow` property to be both boolean or `ShadowDomOptions`:

```
 @Component({
     shadow: {
         delegateFocus: true
     }
  })
```

adamdbradley:
@MrAntix would you be able to make the changes so `shadow` could either be a boolean, or an object? If it's an object we merge it with `{ mode: 'open' }`. And if it's a boolean, the object should be `{ mode: 'open' }`.

MrAntix:
hows that look?

adamdbradley:
Looking great. One thing is that in most cases, `shadow: true` will be the case, and the options would be used less often. So I think the net result of less could we be to not always build an options object, where every component would set `{mode:'open'}`, but rather have the run do the merge instead.

src/core/host-snapshot.ts
```diff
@@ -55,7 +55,7 @@ export function initHostSnapshot(domApi: d.DomApi, cmpMeta: d.ComponentMeta, hos
       // this component is using shadow dom
       // and this browser supports shadow dom
       // add the read-only property "shadowRoot" to the host element
-      domApi.$attachShadow(hostElm, { mode: 'open' });
+      domApi.$attachShadow(hostElm, cmpMeta.shadowRootInit);
```
manucorporat:
```suggestion
      domApi.$attachShadow(hostElm, { mode: 'open', cmpMeta.shadowRootInit });
```

src/declarations/component.ts
```diff
@@ -28,13 +28,14 @@ export interface ComponentMeta {
...
   encapsulationMeta?: number;
+  shadowRootInit?: Readonly<ShadowRootInit>;
   assetsDirsMeta?: AssetsMeta[];
   componentConstructor?: ComponentConstructor;
   componentClass?: string;
   dependencies?: ComponentDependencies;
   jsdoc?: JsDoc;
   styleDocs?: StyleDoc[];
-  hmrLoad?: () => void;
+  hmrLoad?: () => void;
```
manucorporat: revert

src/compiler/transpile/datacollection/component-decorator.ts
```diff
@@ -107,3 +116,7 @@ function getHostMeta(diagnostics: d.Diagnostic[], hostData: d.HostMeta) {
 
   return hostData;
 }
+
+function isShadowDomOptions(o: any): o is d.ShadowDomOptions {
+  return typeof(o) != typeof(true);
```
manucorporat:
```suggestion
  return o && typeof(o) === 'object';
```

src/compiler/transpile/datacollection/component-decorator.ts
```diff
@@ -47,14 +46,24 @@ It will be removed in future versions. Please use the "hostData()" method instea
...
+    jsdoc: serializeSymbol(checker, symbol)    
   };
 
-  // normalizeEncapsulation
-  cmpMeta.encapsulationMeta =
-      componentOptions.shadow ? ENCAPSULATION.ShadowDom :
+  // shadow dom options
+  if(componentOptions.shadow){
+    cmpMeta.encapsulationMeta = ENCAPSULATION.ShadowDom
+    if(isShadowDomOptions(componentOptions.shadow)) {
```
manucorporat:
```suggestion
    if(isShadowDomOptions(componentOptions.shadow)) {
        cmpMeta.shadowRootInit = componentOptions.shadow;
```

manucorporat: remove the DEFAULT_SHADOW_ROOT_INIT from here, it should be provided by the runtime automatically

src/compiler/transpile/datacollection/component-decorator.ts
```diff
@@ -47,14 +46,24 @@ It will be removed in future versions. Please use the "hostData()" method instea
...
+      cmpMeta.shadowRootInit = Object.assign({}, DEFAULT_SHADOW_ROOT_INIT, {
+        delegatesFocus: componentOptions.shadow.delegatesFocus
+      });
+    }else {
+      cmpMeta.shadowRootInit = DEFAULT_SHADOW_ROOT_INIT;
```
manucorporat: not needed

src/compiler/transpile/datacollection/component-decorator.ts
```diff
@@ -47,14 +46,24 @@ It will be removed in future versions. Please use the "hostData()" method instea
     assetsDirsMeta: [],
     hostMeta: getHostMeta(diagnostics, componentOptions.host),
     dependencies: [],
-    jsdoc: serializeSymbol(checker, symbol)
+    jsdoc: serializeSymbol(checker, symbol)
```
manucorporat: revert

MrAntix:
all done, do you need the commits squashed?

manucorporat:
@MrAntix don't worry about it! we can squash them before merge

@adamdbradley, @jthoms1 the PR looks good to me!

tehdb:
please merge and publish 🙏

zoofadoofa:
This is not really an edge case, anything that is slotting in content into shareable components will need delegatesFocus
[See Demo](https://jsfiddle.net/zoofadoofa/6hvrzkf4/)

MrAntix:
Could this be released? I won't have capacity to upgrade to v1 as the breaking changes are too big and my project is going live with 0.18.x

romulocintra:
Any news ;) any help needed ?

driskull:
This would be nice to have

adamdbradley:
Closing in favor of https://github.com/ionic-team/stencil/pull/1190

MrAntix:
Thanks

Issue #1194, opened 2018-10-30
Doc generation issue on Windows env


toams69:
Hi everyone.

I am face to an issue on my env (windows 10). Run stencil build with option --docs add some carriage return before the tag  (and I guess after the generated content also)

see attached screenshot
![2018-10-30_21-26-35](https://user-images.githubusercontent.com/814788/47751590-8042d680-dc92-11e8-91ea-97430878b517.png)

I found the root cause and have a fix.
The issue is coming from this compiler function updateReadme.
I can do a PR if this issue is not already fixed :)

Regards
Thomas

PR #1195, opened 2018-10-30
fix(compiler): fix doc generation issue for windows

toams69:
This fix is related to this issue
https://github.com/ionic-team/stencil/issues/1194


src/compiler/docs/generate-readmes.ts
```diff
@@ -116,18 +116,15 @@ async function updateReadme(config: d.Config, compilerCtx: d.CompilerCtx, readme
 
   const content: string[] = [];
 
-  const existingLines = existingContent.split(/(\r?\n)/);
```
toams69: because here a Regexp is used delimiter is present inside the array existingLines

adamdbradley:
cool thanks

PR #1199, opened 2018-11-01
feat(types): add support for JSX required props

manucorporat:
```
@Prop() required!: string;
@Prop() optional?: string;
@Prop() alsoOptional = 'defaultValue';
```
and we could reflect that in the JSX types

Use cases:
- ion-tab, ion-tab-button  (both need the “tab” prop)
- ion-router and stencil-router: (path and component props)
- many others

screenshot 2018-11-01 at 13 17 52


snaptopixel:
This seems like an odd choice, since the bang in typescript is exactly the opposite meaning (this value **can** be undefined) why not just add `required` to the @Prop config?

PR #1232, opened 2018-11-19
feat(cli): add --dry-run flag

natemoo-re:
This PR adds a `--dry-run` flag to `@stencil/cli`.

The `--dry-run` flag is intended to enable `stencil build` to run without actually writing any files to an outputTarget. It is a boolean, defaulting to `false`.

**Usage**
```bash
stencil build --dry-run
```
```bash
stencil build --dev --watch --dry-run
```

**Why?**

This feature is useful for:
- New `@stencil/cli` users
- Users that wish to quickly check for diagnostic errors (particularly Stencil's custom diagnostics)
- Diagnostic tools that wish to run Stencil in the background (in order to integrate Stencil's custom diagnostics into an editor, for example)

Note that this implementation still steps through the entire build process short of writing the generated files. Everything remains in the `inMemoryFs` rather than being written to disk. *Are there any steps that can be skipped to improve build times without sacrificing diagnostic coverage?*

**Alternative Names**
- `--dry`
- `--no-emit`
- `--diagnostics-only`

**See Also**
- [npm install --dry-run](https://docs.npmjs.com/cli/install)
- [typescript getPreEmitDiagnostics](https://github.com/Microsoft/TypeScript/blob/901476fae3064435b48412a189f7120f3bdc9dc4/src/compiler/program.ts#L201)

### DIFF
--- a/src/compiler/build/build.ts
+++ b/src/compiler/build/build.ts
@@ -68,8 +68,8 @@ export async function build(config: d.Config, compilerCtx: d.CompilerCtx, buildC
     const cmpRegistry = await generateBundles(config, compilerCtx, buildCtx, entryModules, rawModules);
     if (buildCtx.hasError || !buildCtx.isActiveBuild) return buildCtx.abort();
 
-    // should we write any files?
-    // if we're in diagnostics mode, the answer is no
+    // should we write these files?
+    // if we're in dryRun mode, the answer is no
     if (config.flags.dryRun) return buildCtx.finish();
 
     // generate the app files, such as app.js, app.core.js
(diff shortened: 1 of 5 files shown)


src/compiler/build/build.ts
```diff
@@ -65,6 +68,10 @@ export async function build(config: d.Config, compilerCtx: d.CompilerCtx, buildC
     const cmpRegistry = await generateBundles(config, compilerCtx, buildCtx, entryModules, rawModules);
     if (buildCtx.hasError || !buildCtx.isActiveBuild) return buildCtx.abort();
 
+    // should we write these files?
+    // if we're in dryRun mode, the answer is no
+    if (config.flags.dryRun) return buildCtx.finish();
```
manucorporat: we should wait for `await buildCtx.validateTypesBuild();`

adamdbradley:
Thank you for the PR, but closing due to the conflicts. Please ping us in the stencil contributor chat if you'd like to work on this further and we can help out. Thanks.

PR #1241, opened 2018-11-24
fix(component-manifest): expect css from preprocessors

MrAntix:
```collection-manifest.json``` in compiled libs contain references to the linked stylesheets which may be scss files

Assuming these will be converted to css files by a preprocessor

ref: https://github.com/ionic-team/stencil-sass/issues/12 where node-sass will attempt to process the referenced file, which may not have access to imports from the original component project


adamdbradley:
Sorry for the late response, but the codebase has changed enough we won't be able to merge this PR.

Issue #1076, opened 2018-09-09
Crossorigin "True" on Scripts Prevents Sending Auth


lupiter:
**Stencil version:**

```
 @stencil/core@<version>
```

**I'm submitting a:**

[ ] bug report
[X] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**
Currently the [loader](https://github.com/ionic-team/stencil/blob/3a95a15062b7f105a0042800e27bffafd58c48e1/src/client/loader.ts) sets "crossorigin" to be "true" (interpreted as "anonymous", not sure why it isn't set explicitly to that). This means that those files can't be hosted on a server requesting auth, for example if the entire Stencil page is behind a proxy that checks for cookies, no cookies will be sent with the request. This property was [briefly](https://github.com/ionic-team/stencil/commit/351fd73) set to "use-credentials" in 0.8.1, and reverted back in 0.8.2 (with the slightly cryptic message ["fixes CORS issue in unpkg"](https://github.com/ionic-team/stencil/commit/dc482f3).

**Expected behavior:**
I would expect that for a `dist` build, "crossorigin" would be "anonymous" and for a `www` build "crossorigin" would be set to "use-credentials" or unset.

If there are cases I haven't thought of, please let me know. I'd also be happy with a config option to change the value.

[Slack Thread](https://stencil-worldwide.slack.com/archives/C79EANFL7/p1536127733000100)

PR #1259, opened 2018-12-03
fix: set script crossorigin=use-credentials for better Safari support

mattdsteele:
This PR sets the `crossorigin` attribute of the `<script type="module">` loader to `use-credentials`.

This solves an issue as described in this Slack thread: https://stencil-worldwide.slack.com/archives/C79EANFL7/p1536127733000100

Essentially, when compiled Stencil files are loaded from a server that uses session cookies to protect resources, they fail to load in Safari on macOS and iOS. This appears to happen because the current `crossorigin=true` attribute does not pass session cookies through on Safari.

Demo of the differing `crossorigin` behaviors here: https://protected-stencil.herokuapp.com/

The code for this demo is here; it's an Express app that [403s for any JS files](https://github.com/mattdsteele/protected-stencil-demo/blob/master/app.js#L14) that don't contain a cookie: https://github.com/mattdsteele/protected-stencil-demo

A few files are being loaded to demonstrate the varying cross-origin behavior:

* `no-co.js` - fails in Chrome & Safari
* `co-true.js` - fails in Safari, passes in Chrome
* `co-usecredentials` - passes in Safari and Chrome
* `mycomponent.js` - Stencil component built with current compiler - fails in Safari, passes in Chrome
* `myfixedcomponent.js` - Stencil component built with PR - passes in both Chrome & Safari

From what I can tell, the [spec](https://html.spec.whatwg.org/multipage/urls-and-fetching.html#cors-settings-attributes) only allows for "anonymous" or "use-credentials", not `true`, which is probably why the behavior is different.

### DIFF
--- a/src/client/loader.ts
+++ b/src/client/loader.ts
@@ -69,7 +69,7 @@ export function init(
     // let's do this!
     x.src = resourcesUrl + appCore;
     x.setAttribute('type', 'module');
-    x.setAttribute('crossorigin', true);
+    x.setAttribute('crossorigin', 'use-credentials');
   }
 
   x.setAttribute('data-resources-url', resourcesUrl);


mattdsteele:
It appears a change similar to this was added in 351fd731482866e222ef39ea297be34b6befc7ac but then reverted in dc482f3b28c537eef2be65dd671175fc3f07c955 a few days later. Any recollection on what was going on?

mattdsteele:
Bumping - we're running into this issue more as we're rolling out more Stencil components to iOS devices and it's preventing us from a wider rollout.

@manucorporat, do you recall why you removed `crossorigin=use-credentials` for fetch requests?

@mattcarroll-cs are you running into the same issue I'm seeing since the revert?

federicocosta:
We are also having this issue. Tried this and it worked.
Would be great if it could be merged

mattdsteele:
A workaround we've found is to re-bundle the Stencil components using Webpack (we're doing this inside our Angular apps via the [standard Framework Integration](https://stenciljs.com/docs/angular) approach). It uses Webpack's bespoke loading rather than `fetch()`, which appears to preserve session cookies.

Still an issue if you're using a script tag, but it's better than nothing!

Mouvedia:
related: #1076

adamdbradley:
This should no longer be required since Stencil v1. Now developers are creating the script tags directly and should be able to set the attributes themselves.

PR #1287, opened 2018-12-13
Change constructor creation to avoid CSP problems

gc392:
Having an strict policy on Content Security Policy (CSP) configuration triggers errors using the 'eval' like function creation


adamdbradley:
Sorry but we won't be able to use this change since `class` will break within browsers that do not support it. We'll be sure to address this in our upcoming refactor, thanks.

Issue #1218, opened 2018-11-11
Cannot listen for events emitted at onComponentDidUnload


rdesistoz:
**Stencil version:**

```
 @stencil/core@0.15.1
```

**I'm submitting a:**

[x] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**
When an event is emitted when a child component unloads, the event either does not fire or does not get picked up by any listeners.

**Expected behavior:**
It should be possible to listen for unload events when a child component unloads.

**Steps to reproduce:**
Demo: https://github.com/rdesistoz/broken-unload

**Related code:**

```tsx
@Event() dynCompDidUnload: EventEmitter<void>;
  componentDidUnload() {
    this.dynCompDidUnload.emit();
  }

  @Listen('dynCompDidUnload')
  onDynUnloaded(event: CustomEvent) {
    console.log(event, 'unloaded'); // never happens
  }
```

**Other information:**

Nightbr:
Same here `@stencil/core@0.15.2`

Nightbr:
Upgrading to `@stencil/core@0.16.0-0` but still the same bug.

rdesistoz:
Just tried this in 0.17.0 anyone else test this and still have issues?

Nightbr:
I still have the issue in 0.17.0:

I have my `soon-component`:
```
  componentDidUnload() {
    console.log('component unload', this.el);
    this.soonComponentDidUnload.emit();
  }
```
and my container `soon-component-container`:

```
  @Listen('soonComponentDidUnload')
  async onComponentUnLoad() {
    console.log('component UNLOAD from container');
    await this.loadComponents();
    this.orderlist();
    this.updateComponents();
  }
```

When I remove a `<soon-component>` from DOM:
![image](https://user-images.githubusercontent.com/4228646/51755423-efc43880-20be-11e9-96df-b0da0e675a5a.png)

Only `componentDidUnload` of the child component is triggered. I don't know why the parent is not receiving the event...

robmarti:
I just tried out the demo provided by @rdesistoz with `@stencil/core@0.17.0`. The bug seems to be still present when using the `@Listen()` decorator. However, when binding the handler directly to a component in JSX it works.

**my-component.tsx**
```
onDynUnloaded(event: CustomEvent) {
    console.log(event, 'unloaded');
  }

[...]

PR #1293, opened 2018-12-18
fix(events): remove event listeners after lifecycle hook call

robmarti:
This PR relocates the removal of a certain elements' event listeners to after its `componentDidUnload` method call.
As stated in [#1218](https://github.com/ionic-team/stencil/issues/1218), when an event is emitted within a component's `componentDidUnload` method, it can no longer be consumed by its listeners.
By moving the `plt.domApi.$removeEventListener(elm)` and `plt.hasListenersMap.delete(elm)` calls to after the `componentDidUnload` call, events fired within `componentDidUnload`  can still be consumed by all listeners.


adamdbradley:
Good call, thanks!

PR #1336, opened 2019-01-28
update domApi to handle template children automatically

MrAntix:
changed the domApi to handle template elements children as being in the 'content' property
allows the vdom patch to add nodes to the correct place in this instance

one thing, was wondering if patch should be changed instead?

might be faster as there would be less calls to $nodeIs guard function (new bit)
and feels like I am changing expectations of an $appendChild call

however, do you ever need to have childnodes directly on a template?
also it would require more changes and need to be implemented in anything using domAPI

### DIFF
--- a/src/declarations/dom.ts
+++ b/src/declarations/dom.ts
@@ -3,6 +3,7 @@
 export interface DomApi {
   $doc: HTMLDocument;
   $nodeType(node: any): number;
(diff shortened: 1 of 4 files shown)


MrAntix:
fixes https://github.com/ionic-team/stencil/issues/1169
related to https://github.com/ionic-team/stencil/pull/1334

MrAntix:
I'm going to close this. The benefit seems very small
Sorry for the noise

PR #1348, opened 2019-01-31
Add close event to Dialog

arjunyel:
I believe that I should be able to do an inline event listener for the close event. Example is the returning value from dialog section https://demo.agektmr.com/dialog/ https://html.spec.whatwg.org/multipage/interactive-elements.html#the-dialog-element


adamdbradley:
We can't merge changed JSX yet until we update how `h` is imported. Once we get that updated we'll be able to merge this.

PR #1349, opened 2019-01-31
refactor(docs): replace stencil.config.js with stencil config

ps011:
Refactored the comment in component-style file


adamdbradley:
thanks!

Issue #1144, opened 2018-10-15
--prerender for app starter freezes


big-red-dragon:
**Stencil version:**

```
 @stencil/core@0.13.2
```

**I'm submitting a:**

[x] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**

Prerendering with `--prerender` in the (default) `www` mode just freezes after `[56:11.8]  prerender, started: / ...`. No error when using `--debug`.

**Expected behavior:**

build succeeds

**Steps to reproduce:**

1. `npm init stencil`
2. select `app`
3. `npm run build`

**Related code:**

```tsx
// insert any relevant code here
```

**Other information:**

matdec01:
@destructive-dragon Had the same issue. Upgrading to the latest stable (0.14.x) seems to have fixed the problem.

CRHain88:
v0.16 fails.

```sh
[17:07.4]  prerendering started ...
[17:07.4]  prerender, started: / ...
[17:08.2]  prerender, finished: / in 807 ms
[17:08.2]  prerendering failed in 810 ms

[ WARN  ]  Bundling Warning
           Generated an empty bundle

[ ERROR ]  TypeError [ERR_INVALID_ARG_TYPE]: The "url" argument must be of type string. Received type
           undefined at Url.parse (url.js:152:11) at Object.urlParse [as parse] (url.js:146:13) at
           getWritePathFromUrl
           (/Users/crhain88/Development/stencil-app-starter/node_modules/@stencil/core/dist/compiler/index.js:8936:38)
           at
           /Users/crhain88/Development/stencil-app-starter/node_modules/@stencil/core/dist/compiler/index.js:15262:26
           at Generator.next (<anonymous>) at
           /Users/crhain88/Development/stencil-app-starter/node_modules/@stencil/core/dist/compiler/index.js:32:71
           at new Promise (<anonymous>) at __awaiter
           (/Users/crhain88/Development/stencil-app-starter/node_modules/@stencil/core/dist/compiler/index.js:28:12)
           at writePrerenderDest
           (/Users/crhain88/Development/stencil-app-starter/node_modules/@stencil/core/dist/compiler/index.js:15260:12)
           at
           /Users/crhain88/Development/stencil-app-starter/node_modules/@stencil/core/dist/compiler/index.js:15246:19

[...]

PR #1366, opened 2019-02-08
fix: export DEFAULT_MODE from host-config to use it in platform-server and to prevent undefined as a bundle id.

ChBernat:
This should fix 

> ENOENT: no such file or directory, open
           '/Users/.../www/build/app/undefined.es5.entry.js' at

of issue [https://github.com/ionic-team/stencil/issues/1144](url). It is uses the very similar logic you've already used when getting bundleIds as array in a different place of an application thus I thought it will be a good idea to stick to this logic.

One more thing - when I was debugging I've noticed that ionic's components are giving me bundleIds for ios and md; Would it be a good idea to take that into notice at some point in time?

### DIFF
--- a/src/compiler/prerender/host-config.ts
+++ b/src/compiler/prerender/host-config.ts
@@ -283,7 +283,7 @@ export function mergeUserHostConfig(userHostConfig: d.HostConfig, hostConfig: d.
 }
 
 
-const DEFAULT_MODE = 'md';
+export const DEFAULT_MODE = 'md';
 const MAX_LINK_REL_PRELOAD_COUNT = 6;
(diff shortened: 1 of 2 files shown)


ChBernat:
@adamdbradley Hey, could you take a look at those tests? I scrolled through them a few times and honestly it does not seem like a proper behaviour to me. Additionally all other Pull Requests are failing on those 3 as well which does not seem to be a  coincidence. 

Also, this fix is sort of a hot topic for me as we are using Stencil to build our first product at what.digital and we need prerendering to satisfy our SEO needs. 🙏

PR #1373, opened 2019-02-14
feat(polyfills): add polyfill for composedPath() to work with IE

ps011:
This PR will add a polyfill for Event.composedPath() function to get that working in Internet Explorer. 
Since this is [not supported by IE](https://developer.mozilla.org/en-US/docs/Web/API/Event/composedPath) and we need support for that


src/client/polyfills/composedpath.js
```diff
@@ -0,0 +1,21 @@
...
+    this.path = [];
+    while (target.parentNode !== null) {
+      this.path.push(target);
+      target = target.parentNode;
+    }
+    this.path.push(d, w);
+    return this.path;
+    }
+  }
+})(Event.prototype, document, window);
```
arjunyel:
```suggestion
})(Event.prototype, document, window);

```

Add newline at end

ps011:
@adamdbradley , anything else need to be done in this ?

simonhaenisch:
Was about to start working on this, however I just saw that this polyfill was already added in the meantime:

https://github.com/ionic-team/stencil/blob/62839d8e81c1d68bbc177ff35c4f1dd03cad8496/src/client/polyfills/dom.js#L55-L73

Thanks for the PR though!

ps011:
@simonhaenisch  was the polyfill added after 0.15.2 ?

simonhaenisch:
It was added in `v1.0.0-beta.8` (381ded9d08fec13c8c39fcc7015e358a5ee48e6d).

PR #1393, opened 2019-02-28
feat: Improve error msg with effective config attributes name

peterpeterparker:
Improve error msg:

```
[ ERROR ]  copy task, "to" property must exist if "from" property is an absolute path:
```

to

```
[ ERROR ]  copy task, "dest" property must exist if "src" property is an absolute path:
```

as the attributes' names in the Stencil config are dest/src and not to/from


manucorporat:
Fixed in https://github.com/ionic-team/stencil/commit/b574b86903900e3ff427296caae7f9ffda0fe163

Sorry i was not able to merge your PR :/

peterpeterparker:
@manucorporat no worries, I even didn't forgot about this ;)

Issue #1269, opened 2018-12-05
IE11: Polyfill Infinite Loop - es6-collections vs core-js/map


cary-smith:
**Stencil version:**

```
 @stencil/core@0.16.0
```

**I'm submitting a:**

[x] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**

The [es6-collections](https://github.com/WebReflection/es6-collections) library is run first and works great. 

When added to an Angular 6 project (which includes the core-js map/weakmap/etc polyfills), the constructor of the polyfill is modified due to an [iterable shortcoming](https://github.com/zloirock/core-js/blob/99ea4e25f1e4705725a1acd464f4872fb577926a/modules/_collection.js#L57) (not sure exactly what). Once the constructor is modified, this line
```js
if (!this || this.constructor !== b) return new b(a);
```
...from es6-collections will always identify the constructor as not 'b', so it runs into a stack overflow in IE11 while loading an angular application.

**Expected behavior:**

The Ancient One (IE11) still functions as expected: Slowly, painfully, and with lots of excess code to make it behave until we can put it on the trash heap.

**Steps to reproduce:**

I was able to reproduce it in the following:

1. run `git clone -b IE11-polyfill-fun git@github.com:cary-smith/stencil-component-starter.git`
2. run `npm i`
3. Open the `node_modules/@stencil/core/dist/client/polyfills/es5/map.js` file
4. Add a console log just under the line mentioned in the Current Behavior section of this ticket

**Related code:**

```js
(function(e) {
  function d(a, b) {
    function c(a) {
      console.log("IE still exists and makes us all sad...");
      if (!this || this.constructor !== c) return new c(a);
```
5. Run the application via `npm start`
5. Open in IE11 and watch that console log GO! Wheeeee!

**Other information:**

This blocks our ability to stay current with all the new kick-assery you folks keep implementing and may side-affect other devs at some point. We want to stay in the loop!

Thanks for the awesome software.

warapitiya:
I'm facing same issue. I think `es6-collections` is colliding with other existing polyfill.

warapitiya:
[...]

Issue #1385, opened 2019-02-21
Make Polyfills Optional


romulocintra:
**I'm submitting a:**

[ ] bug report
[x ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**

Stencil add needed Polyfills for different things into components/Apps

**Expected behaviour:**

I would like to make this process optional in case i want to have centralized Polyfill strategy or service. Considering also to have optional outputTarget for Polyfill code

romulocintra:
@adamdbradley and @manucorporat  there is a lot of noise about Polyfill IE11 in #1269  and I would like to know if there is any plan to go with that decision or there is anything else planned. 

My proposal about have it as optional, that way will be more decoupled and easy to teams to adapt if they need to change.

Proposal 1 : 

```code 
// stencil.config.ts
 outputTargets: [
   {
    type : 'polyfills',
    exclude : [ 'Element.remove' , 'Object.values' ] // Or  false to exclude all,
    include :  ['core-js.js' , '../../array-ponyfill.js']  // OR  function() 
   }
```

PR #1395, opened 2019-03-01
IE11: Polyfill Infinite Loop - es6-collections vs core-js/map #1269

roNn23:
Hey! We want to use your awesome lib, but sadly Stencil isn't working with our stack (AngularJS) on IE11. We've got the same issue as mentioned in #1269 and fixed it by replacing the deprecated es6-collection lib with the more recent es6-shim lib. Because of this, we had to change some of the polyfills and refactored duplicated code. Please look into it and merge it, if possible. Thanks a lot!


romulocintra:
Just add #1385  to make it optional !!!!

fionawhim:
We were just bit by this. If core-js runs second, it tries to wrap / extend the es6-collection polyfill classes, modifying the `constructor` value, which causes the es6-collection constructors to go into an infinite loop.

Our hacky solution was to clear out the es6-collection versions of Map and Set before running core-js. 

Something more stable / contained would be ideal; it was a scary situation that upgrading Stencil broke the app on the page that we include Stencil in.

stefanmeschke:
Any updates here? :)
Would like to see this merged? ;)

dgibson666:
Any chance this can be fixed/implemented? We've been patiently awaiting a fix for months and are stuck on .15.2 as a result. I wonder how many other people are in the same boat.

mattdsteele:
This should be resolved with Stencil One, as per @adamdbradley; it's using a different polyfill. You can try the prereleases and see the breaking changes here: https://github.com/ionic-team/stencil/blob/core-refactor/BREAKING_CHANGES.md

@manucorporat might be good to add the `fixed-with-stencil-one` label to this ticket :)

manucorporat:
Fixed https://github.com/ionic-team/stencil/commit/5afaa7eee04e09ae626a06317ca82aec0bfd87c3

PR #1404, opened 2019-03-06
Fail the build for reserved members

sslotsky:
I propose that the build should fail if reserved words are used as component props. Currently, a warning is issued, which is too easy to ignore. If this has a potential to cause a real issue, it should be an error.

An alternative would be to create an eslint rule for stencil. Or, if it's really not an issue, perhaps it should not be a warning? Discussion appreciated.


adamdbradley:
Thanks for the PR, but at this time we'll keep the warnings the same. Thanks

PR #1452, opened 2019-04-11
chore(): polish breaking changes

shreeshbhat:


BREAKING_CHANGES.md
```diff
@@ -146,7 +146,7 @@ The event name should be provided excl
```
shreeshbhat: Also, I am not sure what excl means here.

PR #1463, opened 2019-04-13
feat (Puppeteer): Added devtools option in puppeteer.launch

jeanbenitez:
NOTE: I closed my previous PR to make the change in Stencil One :)

Puppeteer launch api: https://github.com/GoogleChrome/puppeteer/blob/v1.13.0/docs/api.md#puppeteerlaunchoptions


adamdbradley:
thanks!

PR #1490, opened 2019-05-01
Update readme.md / Fixed Typo in readme

jonisar:
Fixed "runs" to "run".

The verb refers to "web components" which is plural, so the correct form is "run" and not "runs". 
See: https://www.quora.com/What-is-the-third-form-of-the-word-run

Cheers


adamdbradley:
thanks

PR #1499, opened 2019-05-03
docs(breaking) fix @Method await typo

Matthew-Smith:
In the breaking changes file there is a function definition that didn't really make sense


arjunyel:
Merged, thank you!

PR #1501, opened 2019-05-03
WIP test(polyfill): Add failing IE11 test to demonstrate #1269 bug

mattdsteele:
This includes a failing test for #1269, in `es5-polyfill`.

You can see the behavior by running `npm run karma.prod` and opening Karma's Debugger in localhost. The test throws this error:

image

Commenting the `import 'core-js/es6/map'` line causes the test to pass.

I ran into a few issues getting Karma to run locally in IE11; I'll annotate the changes I made to `karma.config.js` to get it working. I'd expect these to be reverted after implementing a fix.

PR #1504, opened 2019-05-05
IE and Edge issues - Failing Tests

jereckley:
Hey! First time trying to contribute. In short the issues that are not covered by @mattdsteele here: https://github.com/ionic-team/stencil/pull/1501 : 1) CSS variables are not working as expected in IE 2) Nothing is working in Edge. After adding a polyfill for getRootNode I can get Edge working for the most part. There are still about a dozen tests failing in edge after the poly fill is added.


test/karma/karma.config.js
```diff
@@ -46,6 +46,12 @@ const localLaunchers = {
 			'--remote-debugging-port=9333'
 		]
   },
+  IEGeneral: {
+    base: 'IE'
+  },
+  EdgeGeneral: {
+    base:'Edge'
```
jereckley: Adding test support for edge and ie

simonhaenisch:
I think in the meantime there's been lots of fixes related to IE/Edge... can you check which of the cases are still not working?

jereckley:
I checked everything out yesterday and it was like a 100 time better in our projects. Thanks! There are still some issues but I think I we need to check our bases before raising any concerns.

PR #1531, opened 2019-05-15
feat(compiler):  watch  *.html files

romulocintra:
The previous versions of stencil > 0.18.1  had a watcher for *.html inside my components folder, would like to have this feature in next versions.

Use Case :

I have more than 100+ Components, to demoing them i can use the index.html or have this feature that allows me to have 

````
my-component
       my-component-demo.html
       my-component.tsx

````

and with copy have this .html file inside my `www`, allow me to navigate localhost:3333/my-component-demo.html and have multiple demos of my components isolated.


romulocintra:
@manucorporat  should I update against "one"

PR #1540, opened 2019-05-17
Build docs in dev mode

snaptopixel:
Carries over behavior from 0.x where docs are generated in dev mode as well as prod.

Useful for developing components and making sure docs are up to date in the repo.


snaptopixel:
@manucorporat here's the change I mentioned in Slack, thanks!

manucorporat:
Hi @snaptopixel this is in fact the expected behaviour, in order to make dev build super fast by default, however i did fixed it by forcing to always emit the docs when the --docs flag is used, just like 0.18.x!

Thanks a lot for pointing out to this issue!

https://github.com/ionic-team/stencil/commit/e68f56cf6fa6cb5e2997caa452586ed48f12012a

Issue #1493, opened 2019-05-01
change jest setup to use setupFilesAfterEnv instead of setupTestFrameworkScriptFile in jest preset


Tallyb:
**Stencil version:**

```
 @stencil/core@1.0.0-alpha.14
```

**I'm submitting a:**

[ ] bug report
[ X ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

Jest preset is currently using setupTestFrameworkScriptFile here: 
https://github.com/ionic-team/stencil/blob/master/scripts/packages/testing/jest-preset.js#L26

setupFilesAfterEnv is replacing setupTestFrameworkScriptFile which is deprecated. 
It is also an array, so it can be extended with additional setups easily

manucorporat:
Seems like a legit request, can you prepare a PR with the fix? even thought, `setupTestFrameworkScriptFile` is a deprecated API, i am not sure how this is an issue.

Tallyb:
I created the PR to update jest-preset. 
However, I am not sure what is the purpose of this part and how it should be updated? 
https://github.com/ionic-team/stencil/blob/master/src/compiler/config/validate-testing.ts#L83

RafaelKr:
@manucorporat did you read the last comment?
I think this should be deleted or replaced with the logic for setupFilesAfterEnv: https://github.com/ionic-team/stencil/blob/master/src/compiler/config/validate-testing.ts#L83

PR #1545, opened 2019-05-18
Jest preset

Tallyb:
closes https://github.com/ionic-team/stencil/issues/1493
Jest preset is now using setupFilesAfterEnv which is extendable


manucorporat:
That was easy! thanks a lot for the PR! ❤️

PR #1548, opened 2019-05-18
Skip puppeteer install

### LINKED ISSUE #1529: Skip puppeteer install
**Stencil version:**

```
 @stencil/core@x.x.x All
```

**I'm submitting a:**

- [ ] bug report
- [x] feature request

**Current behavior:**

Every time I run e2e tests Stencil installs puppeteer , even if I have it installed globally or under project structure(monorepo).

**Expected behavior:**

I would like to have the alternative to setup 2 things : 

1. `SkipInstall` to not install puppeteer.

-  I can use puppeteer-core in CI env
-  I can use inside monorepo  

2. `executablePath` configure on stencil.config.ts the path to chrome I would like to execute same as : 
```js
const browser = await puppeteer.launch({executablePath: '/path/to/Chrome'});
```
**Use Cases :** 

- Running e2e tests on CI 
- Want to use a different version of chrome and not chromium 
- Customized `npm install puppeteer-core@chrome-71`

romulocintra:
@adamdbradley and @manucorporat: I'm trying to close #1529 , but I have some doubts about how to proceed :

Close #1529

1. My first approach was creating a skipInstall interface where you can ignore all lazyLoaded Dependencies - Required more code and was little more complex. (Just dropped cause required a more refactoring on ensure/task.test functions)

2. Quick win, I just added an option to install puppeteer-core in case you provide a `browserExecutablePath`, should be the easy and quickest way to have it working in most CI Env's


simonhaenisch:
I just looked at your issue and I would have suggested to go with solution 2 as well... I think you can mark your PR ready for review.

romulocintra:
> I just looked at your issue and I would have suggested to go with solution 2 as well... I think you can mark your PR ready for review, lgtm.

I need to do few more test only have time during the weekend probably will prepare a legit PR around saturday

romulocintra:
@simonhaenisch  and @manucorporat  ... i just updated the lazyLoad of the module and works like a charm ... i'am little suprised , cause reading puppeteer docs difference between core and regular version , [requires change of import 'puppeteer' vs import 'puppeteer-core'](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#puppeteer-vs-puppeteer-core), but without this change is working well.

Other option will be always use pupepeteer-core, if you someone provide `browserExecutablePath ` we use provided path if not we install chrome/chromium and have a default `browserExecutablePath `  pointing to the local install.

simonhaenisch:
@romulocintra if you clean out your `node_modules`, does it still work fine without changing the import? Because I think that you might still have `node_modules/puppeteer` locally, so the import will work.

romulocintra:
All tests were made removing node_modules completely and not having puppeteer installed as global, please give it a try and confirm. I was surprised  about it working without changing the import.

Anyway, the default case is always with puppeteer install if you don't provide the browser path.

PR #1549, opened 2019-05-18
perf(): insane optimizations

manucorporat:
TodoMVC scored 2.3KB, smaller than trying to write the same app without hand, manually


test/todo-app/src/components/todo-item/todo-item.tsx
```diff
@@ -7,16 +7,16 @@ export class TodoItem {
 
   @Prop() checked: boolean;
   @Prop() text: string;
-  @Prop() index: number;
   @Event() itemCheck: EventEmitter;
   @Event() itemRemove: EventEmitter;
 
   render() {
+    const {checked, text, itemCheck, itemRemove} = this;
```
shreeshbhat: Whoa, this is an interesting approach to reduce repetition of 'this' and size. 👍

PR #1552, opened 2019-05-20
fix(hmr): call all callbacks in build event subscribers

simonhaenisch:
This fixes an HMR issue where `buildResults` was never sent to the dev-server client because only the most recent `buildFinish` event subscriber was attached.

I think the bug was introduced in https://github.com/ionic-team/stencil/commit/fbbe23260ab95de004e02ab856532bb684c74dcc#diff-06187c875ae567c88cb345d2abd4ea2b.

BTW after applying this fix locally in my project (within `node_modules/@stencil/core/dist/compiler/index.js`), I had some really weird behavior until I cleaned my `.stencil` directory. The only thing that would end up in `components.d.ts` was the last component that I had edited.

Does Stencil auto-clear the `.stencil` folder if the version that the cache was created with is a mismatch (i. e. an older version than the one currently compiling)?


adamdbradley:
thanks!!

PR #1565, opened 2019-05-22
feat(): load images with imports

manucorporat:


src/compiler/rollup-plugins/image-plugin.ts
```diff
@@ -0,0 +1,39 @@
...
+          warn.absFilePath = id;
+        }
+        return `const img = 'data:${mime};base64,${data}'; export default img;`;
+      } catch (e) {}
+      return null;
+    }
+  };
+}
+
+const MAX_IMAGE_SIZE = 4 * 1024; // 10KB
```
simonhaenisch: This looks more like 4 KB? 😄

manucorporat: "Code never lies, comments sometimes do."

src/compiler/app-core/bundle-app-core.ts
```diff
@@ -9,7 +9,7 @@ import { stencilBuildConditionalsPlugin } from '../rollup-plugins/stencil-build-
 import { stencilClientPlugin } from '../rollup-plugins/stencil-client';
 import { loaderPlugin } from '../rollup-plugins/loader';
 import { stencilExternalRuntimePlugin } from '../rollup-plugins/stencil-external-runtime';
-
+import { imagePlugin } from '../rollup-plugins/image-plugin';
```
Zucca, line 12: is there a reason why not using the rollup plugin: https://github.com/rollup/plugins/tree/master/packages/image

PR #1568, opened 2019-05-23
fix(e2e): don't intercept request

### LINKED ISSUE #1567: [1.0.0-beta.5...1.0.0-alpha.15]  puppeteer`on` method not omitted from E2EPage but breaks if used
**Stencil version:**

```
 @stencil/core@1.0.0-beta.5
```

**I'm submitting a:**

[X] bug report
[ ] feature request

**Current behavior:**

Using `E2EPage.on` to intercept a request throws an exception in stencil `one`

**Expected behavior:**
Ability to intercept requests without an exception being thrown.

This was possible  in 0.18.x

**Steps to reproduce:**

Add an `on` method...

```ts
const page = await newE2EPage();
page.on('request', () => { request.respond(...) });
```

Will throw an exception with `Request Already Handled` because of:

https://github.com/ionic-team/stencil/blob/9c63d34c684ef8451dab4c5fb88c4f22d9907fa7/src/testing/puppeteer/puppeteer-page.ts#L172

**Other information:**

This may be just a question rather than a bug, if there's a workaround for doing this in stencil one.

evaneus:
fixes #1567

### DIFF
--- a/src/testing/puppeteer/puppeteer-page.ts
+++ b/src/testing/puppeteer/puppeteer-page.ts
@@ -157,7 +157,7 @@ async function e2eSetContent(page: pd.E2EPageInternal, html: string) {
   const appUrl = (process.env as d.E2EProcessEnv).__STENCIL_APP_URL__;
   if (typeof appUrl !== 'string') {
     await closePage(page);
-    return 'invalid e2eSetContent() loader script url';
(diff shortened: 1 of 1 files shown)


evaneus:
thought all tests had run but e2e are in fact failing.  closing.

PR #1570, opened 2019-05-25
Html snapshots

Tallyb:
Adds html serializer as a jest serializer, so expect(elm).toMatchSnapshot() will work (similar to the toEqualHtml matcher). Shadow DOM is always serialized.


adamdbradley:
Looks great! Thanks!

PR #1579, opened 2019-05-30
fix(e2e): wait for idle network

bodinaren:
Sometimes when taking screenshots some of the images has not finished loading. This commit will wait for the network to become idle - indicating that all resources has been loaded - before continuing.


manucorporat:
Thanks a lot for the PR!!

PR #1581, opened 2019-05-30
feat(mock-doc): add KeyboardEvent

kensodemann:
Adds KeyboardEvent to MockDoc so it can be used in unit tests.

### DIFF
--- a/src/mock-doc/event.ts
+++ b/src/mock-doc/event.ts
@@ -56,14 +56,14 @@ export class MockCustomEvent extends MockEvent {
 }
 
 export class MockKeyboardEvent extends MockEvent {
-  code: string = null;
-  key: string = null;
-  altKey: boolean = null;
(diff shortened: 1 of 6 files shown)


src/mock-doc/event.ts
```diff
@@ -55,6 +55,26 @@ export class MockCustomEvent extends MockEvent {
 
 }
 
+export class MockKeyboardEvent extends MockEvent {
+  code: string = null;
```
manucorporat: is "null" the default value of this properties in the spec? in a real browser

kensodemann: Looked the defaults up on MDN and switched all of those values to use the defaults that were published there.

kensodemann: So this one is now defaulting to empty string, etc.

PR #1584, opened 2019-06-01
feat(compiler): support paths in compiler options

simonhaenisch:
Someone was asking about why path aliases in the `compilerOptions` section of tsconfig didn't work in Stencil, and I ended up trying to solve it.

First, I created [rollup-plugin-typescript-paths](https://github.com/simonhaenisch/rollup-plugin-typescript-paths), which uses `nodeModuleNameResolver` from the typescript api to resolve imports using the path aliases (it returns the transpiled `.js` counterpart of the `.ts` file).

Then I realized that I had an issue with const enums: Stencil removes the `paths` field from the compiler options entirely (in `transpile-service.ts`), which causes the import statements of any const enums to remain in the file even though they have disappeared during transpilation (like interfaces).

To solve this I decided to implement the resolving of path aliases in Stencil itself, because it's relatively simple and the only reason I wrote my plugin was for use with Stencil (though it should theoretically work in other contexts).

To try it out:
- `npm link` in stencil
- `npm init stencil app foobar`
- `cd foobar`, `npm link @stencil/core`
- change files:

```js
// tsconfig.json
{
  // ...
  "compilerOptions": {
    "baseUrl": ".",
    "paths": {
      "@foobar": ["src/global/app"]
    }
  }
}
```

```ts
// src/global/app.ts

export const foobar = 'foobar';
```

```tsx
// src/components/app-home/app-home.tsx

import { foobar } from '@foobar';

render() {
  return <div>{foobar}</div>;
}
```

BTW I tried writing a test but didn't have enough time to figure out how to change the compiler options within the test.


leoafarias:
Just a bump on this.. spent a lot more that I care to admit trying to see if was an issue on my end or not.

rainzhao2000:
Bump for sure, awesome of you to implement a solution!

PR #1614, opened 2019-06-10
feat(mock-doc): add NodeList

miguelyoobic95:
Customer needs to have NodeList on MockWindow.

The customer's tests fail when they try to test a Stencil component that itself uses tippy.js. Here is where it is failing with a ReferenceError because NodeList is not found, so that is essentially what I am trying to duplicate in the test w/o actually including tippy.js.

Here is the LOC in tippy.js that results in the failure in the Stencil component unit test:

https://github.com/atomiks/tippyjs/blob/8b23e4a51452b89ca9a54731bb499eac4cc2920e/src/ponyfills.ts#L3

This PR adds NodeList to MockWindow in order to facilitate these tests.


src/mock-doc/window.ts
```diff
@@ -200,6 +201,20 @@ export class MockWindow {
     return ElementCstr;
   }
 
+  get NodeList() {
+    let NodeListCstr = nodeListCstrMap.get(this);
+    if (NodeListCstr == null) {
+      const ownerDocument = this.document;
+      NodeListCstr = class extends MockElement {
```
manucorporat: This can't be right. should not it be a NodeList mock?

miguelyoobic95: Cheers for the feedback. I am not sure I fully understand how you guys are mocking each part of the DOM API but I totally misunderstood what Ken did in his PR. I have pushed another commit where I have added a MockNodeList class as this seems to be the pattern you use.

If I understand you are mapping the class members to the available properties of the DOM API objects so then when those get called in a test code (in a mock window context) they do not throw Errors ? 

Let me know if there are any more issues and I will get to it as soon as I can.

kensodemann: Per conversation with Adam and Manu, this can be merged as-is and if we need to make it more complete later we can.

PR #1655, opened 2019-06-18
"readonly" in method return causes error

kraftwer1:
When generating docs using Typedoc, I'm getting this error:

    node_modules/@stencil/core/dist/declarations/system.d.ts(71) ';' expected.
    node_modules/@stencil/core/dist/declarations/system.d.ts(71) An element access expression should take an argument.
    node_modules/@stencil/core/dist/declarations/system.d.ts(73) Declaration or statement expected.


adamdbradley:
Thank you for the PR, but closing due to the conflicts. Please ping us in the stencil contributor chat if you'd like to work on this further and we can help out. Thanks.

PR #1676, opened 2019-06-21
 fix(declarations): rollup plugin interface

simonhaenisch:
Fixes the `Plugin` interface.

The old one was quite out of sync with the actual Rollup types, which made it a bit inconvenient to write custom plugins for Stencil.

Basically what I'm doing is extending rollup's `Plugin` interface, by adding the additional `pluginContext` parameter (required by certain stencil rollup plugins) to the signature of each of the three hooks.

The hook descriptions were taken from https://rollupjs.org/guide/en#hooks.

### DIFF
--- a/src/compiler/plugin/plugin.ts
+++ b/src/compiler/plugin/plugin.ts
@@ -10,7 +10,7 @@ export async function runPluginResolveId(pluginCtx: PluginCtx, importee: string)
 
     if (typeof plugin.resolveId === 'function') {
       try {
-        const results = plugin.resolveId(importee, null, pluginCtx);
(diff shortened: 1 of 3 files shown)


simonhaenisch:
Closing, as this would create a dependency on `rollup`.

Other options would be 1) somehow copy `rollup.d.ts` into Stencil's `dist`, or 2) manually write (duplicate) the interfaces. 1) is not that feasible and 2) is cumbersome and hard to keep in sync with the actual rollup types.

Option 3) is for the user to add `rollup` into their dev dependencies so they can access the `Plugin` type from there.

PR #1685, opened 2019-06-26
fix(types): Expose ComponentInstance type

### LINKED ISSUE #1684: Type definition output error
**Stencil version:**

```
 @stencil/core@<version>
```
1.1.2

**I'm submitting a:**

[x] bug report
[ ] feature request

**Current behavior:**

After performing a build component types from `dist/types/foo/foo.d.ts` have an import like this:
```
import { ComponentInterface } from '../../stencil.core';
```
however when navigating to `stencil.core` there is no definition for `ComponentInterface` which causes build issues in consuming applications.

**Expected behavior:**

The outputted types should match.

**Steps to reproduce:**

**Related code:**

```tsx
```

**Other information:**

Current master has this [code](https://github.com/ionic-team/stencil/blob/144e966f6aeae0e8ea001616e92d1331a16ceadd/src/testing/core.ts#L10) which hides `ComponentInstance` as `ComponentInterface` then when the components are output `ComponentInterface` is not defined in `stencil.core` while `ComponentInstance` is defined.

Brocco:
fixes #1684

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -6,6 +6,7 @@ export {
   ComponentDidUpdate,
   ComponentWillLoad,
   ComponentWillUpdate,
+  ComponentInstance,
   ComponentInstance as ComponentInterface,
   StencilConfig as Config,
   EventEmitter,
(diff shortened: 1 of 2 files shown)


manucorporat:
Thanks for the PR! but `ComponentInstance` is not part of the public API
Fixed here: https://github.com/ionic-team/stencil/commit/e5bae1b8b73d87a6d0ee050c7b9696c81cc9d702

PR #1694, opened 2019-06-28
feat(testing): add MockSVGElement

kensodemann:
This change just adds some of the basic APIs and not the whole
set of them. In addition:

- the methods just return some kind of sane default value
- rather than having several sub-classes, everthing is squashed
  into MockSVGElement for simplicity

### DIFF
--- a/src/mock-doc/document.ts
+++ b/src/mock-doc/document.ts
@@ -172,7 +172,7 @@ export class MockDocument extends MockHTMLElement {
   }
 
   createElementNS(namespaceURI: string, tagName: string) {
-    const elmNs = new MockElement(this, tagName);
+    const elmNs = createElement(this, tagName);
     elmNs.namespaceURI = namespaceURI;
     return elmNs;
   }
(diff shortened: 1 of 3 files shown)


src/mock-doc/document.ts
```diff
@@ -172,7 +172,7 @@ export class MockDocument extends MockHTMLElement {
   }
 
   createElementNS(namespaceURI: string, tagName: string) {
-    const elmNs = new MockElement(this, tagName);
+    const elmNs = createElement(this, tagName);
```
manucorporat: if I am not wrong createElementNS() always creates MockElement (or svg) but never HTMLElement

manucorporat: Screenshot 2019-06-28 at 22 30 04

kensodemann: The way it was before, it always created a MockElement...

Do we need to abstract out some of the logic from `createElement()` so it can create a MockSVGElement when required?

Issue #1712, opened 2019-07-04
[Old Firefox] Syntax error in strict mode


fabiolibre:
**Stencil version:**
```
 @stencil/core@1.1.3
```

**I'm submitting a:**
[ x] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**
Trying to load the **nomodule** file on an old browser will throw a warning in the console that prevents the entire file from being executed:
> SyntaxError: in strict mode code, functions may be declared only at top level or immediately within another function

This is because the **CustomEvent** polyfill defines a function inside a conditional block, which is not suppose to happen in strict mode.

Removing **'use strict;'** from the beginning of the file fixes the issue.

**Expected behavior:**
Old browsers should be able to load the polyfills.

**Steps to reproduce:**
Open a simple page using on Firefox 33.
The stenciljs generated file won't be loaded and no custom element is defined in the page.

**Related code:**
This snippet is taken from the **dom.js** file and is the part FF complains about.
```javascript
// Polyfill CustomEvent
if (typeof window.CustomEvent !== 'function') {
  function CustomEvent(event, params) {
    params = params || { bubbles: false, cancelable: false, detail: undefined };
    var evt = document.createEvent( 'CustomEvent' );
    evt.initCustomEvent( event, params.bubbles, params.cancelable, params.detail );
    return evt;
  }

  CustomEvent.prototype = window.Event.prototype;

  window.CustomEvent = CustomEvent;
}
```
To solve the issue, just defining CustomEvent diffently, like:
```javascript
// Polyfill CustomEvent
if (typeof window.CustomEvent !== 'function') {
  CustomEvent = function(event, params) { .. }
  .
  .
  .
}
```
[This commit](https://github.com/ionic-team/stencil/commit/06986ab6b593377488737f9692b751859bf6509f) added **'use strict';** inside **getSystemLoader()**. which caused the problem.
```typescript
async function getSystemLoader(config: d.Config, corePath: string, includePolyfills: boolean) {
  const polyfills = includePolyfills ? await getAppBrowserCorePolyfills(config) : '';
  return `
'use strict';
${polyfills}
```
**Other information:**
Since the focus is to support newer browsers, a quick fix might be to remove **'use strict;'** from **getSystemLoader()**.
Alternative is to follow the use strict and modify **dom.js** like explained in the **Related code:** section.

[...]

PR #1713, opened 2019-07-04
fix(dom.js): dom.js compliant with 'use strict;'

fabiolibre:
fix for https://github.com/ionic-team/stencil/issues/1712


CSchulz:
Seems to be already fixed with bf27172c51e5a9ce7497b656daa1070de2b29f87 and released with 1.1.5

manucorporat:
@CSchulz you right! this issue should be already be fixed, thanks a lot for the PR though!

Issue #1706, opened 2019-07-03
Mock canvas element when prerendering [getContext is not a function]


renrizzolo:
**Stencil version:**
```
@stencil/core@1.1.2
```

**I'm submitting a:**
[ ] bug report
[X] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**
When using the `--prerender` flag, the prerender will fail when using `getContext()` on a canvas element. canvas is not mocked in hydrate/prerender/mock-doc.
```
prerendering failed in 1.03 s
[ ERROR ]  Hydrate Error
           [...]/dist/hydrate/app.js:6409 var ctx =
           canvas.getContext('2d'); ^ TypeError: canvas.getContext is not a function 
```
**Expected behavior:**
Prerender doesn't fail when calling `getContext()` on a canvas element.

**Steps to reproduce:**
```tsx
import { Component, h } from '@stencil/core';

@Component({
  tag: 'my-first-component',
})
export class MyComponent {
  componentDidLoad() {
    const canvas = document.createElement('canvas');
    canvas.width = 1;
    canvas.height = 1;
    const ctx = canvas.getContext('2d');
    ctx.fillStyle = 'rgba(0,0,0,0)';
    ctx.fillRect(0, 0, 1, 1);
  }
  render() {
    return (
      <p>
        My name is {this.name}
      </p>
    );
  }
}
```

**Related code:**

```tsx

```

**Other information:**
I'm not actually trying to use canvas - I would wrap the above example with `Build.isBrowser` if that were the case.
I'm importing the library [lottie-web](https://github.com/airbnb/lottie-web), which has some functions that use `getContext()`.

One solution would be to add this simple mock to the switch in the `createElement` function in dist/hydrate/index.js (not sure what the true source of that function is).

```tsx
class MockCanvasElement extends MockHTMLElement {
    constructor(ownerDocument) {
        super(ownerDocument, 'canvas');
    }
    getContext() {
        return {
            fillRect: function () { },
            clearRect: function () { },
            getImageData: function (x, y, w, h) {
                return {
                    data: new Array(w * h * 4)
                };
            },
            putImageData: function () { },
            createImageData: function () { return [] },
            setTransform: function () { },
            drawImage: function () { },
            save: function () { },
            fillText: function () { },
            restore: function () { },
            beginPath: function () { },
            moveTo: function () { },
[...]

PR #1716, opened 2019-07-08
feat(mock-doc): add canvas

renrizzolo:
As per: #1706, adds a canvas element mock to `mock-doc`, allowing pre-rendering with canvas api usage like `getContext()`.


manucorporat:
Thanks a lot for the PR!!! Love it

PR #1732, opened 2019-07-16
fix(angular): property assignment happens after append

manucorporat:


src/compiler/app-core/build-conditionals.ts
```diff
@@ -117,6 +117,7 @@ export function updateBuildConditionals(config: d.Config, b: d.Build) {
   b.member = (b.member || b.updatable || b.mode || b.lifecycle);
   b.taskQueue = (b.updatable || b.mode || b.lifecycle);
   b.constructableCSS = !b.hotModuleReplacement || !!config._isTesting;
+  b.initializeNextTick = true; // config.outputTargets.some(isOutputTargetAngular);
```
CSchulz: Is there some documentation about the other output targets like angular? I didn't find any documentation.

PR #1737, opened 2019-07-17
fix(vdom): use classList instead of className

manucorporat:
Svg.className is not a string


cary-smith:
@manucorporat - Since [IE11 does not fully support](https://caniuse.com/#search=classList) `classlist.add()/remove()` (mainly using multiple arguments), is this being polyfilled somewhere?

PR #1753, opened 2019-07-22
Browser compile

adamdbradley:
`stencil.compile(code)`


src/compiler/app-core/format-component-runtime-meta.ts
```diff
@@ -2,15 +2,15 @@ import * as d from '../../declarations';
 import { CMP_FLAGS, LISTENER_FLAGS, MEMBER_FLAGS } from '@utils';
 
 
-export function formatLazyBundleRuntimeMeta(bundleId: any, cmps: d.ComponentCompilerMeta[]): d.LazyBundleRuntimeData {
+export const formatLazyBundleRuntimeMeta = (bundleId: any, cmps: d.ComponentCompilerMeta[]): d.LazyBundleRuntimeData => {
```
manucorporat: this code does not run in the browser, right? why transform to arrow now?

src/compiler/browser/browser-scope-css.ts
```diff
@@ -0,0 +1,23 @@
+import * as d from '../../declarations';
+import { catchError } from '@utils';
+import { getScopeId } from '../style/scope-css';
+import { scopeCss as shadowCss } from '../../utils/shadow-css';
+
+
+export const scopeCss = async (code: string, tagName: string, opts: d.ScopeCssOptions = {}) => {
```
manucorporat: a lot of functions with the same name, can get tricky to maintain in the future

src/compiler/transformers/core-runtime-apis.ts
```diff
@@ -0,0 +1,38 @@
...
+  Host: `Host as ${HOST}`,
+  HTMLElement: HTML_ELEMENT,
+  proxyCustomElement: `proxyCustomElement as ${PROXY_CUSTOM_ELEMENT}`,
+  registerHost: `registerHost as ${REGISTER_HOST}`,
+  registerInstance: `registerInstance as ${REGISTER_INSTANCE}`,
+};
+
+
+export const addCoreRuntimeApi = (moduleFile: d.Module, coreRuntimeApi: string) => {
+  if (!moduleFile.coreRuntimeApis.includes(coreRuntimeApi)) {
```
manucorporat: using Set?

manucorporat: using Set?

src/runtime/initialize-component.ts
```diff
@@ -78,15 +78,19 @@ export const initializeComponent = async (elm: d.HostElement, hostRef: d.HostRef
...
+
+      const shadowCssPolyfill = (BUILD.shadowDom && cmpMeta.$flags$ & CMP_FLAGS.needsShadowDomShim);
+      const runtimeScopedCss = (BUILD.runtimeScopeCss && cmpMeta.$flags$ & CMP_FLAGS.scopedCssEncapsulation);
+
+      if (!BUILD.hydrateServerSide && (shadowCssPolyfill || runtimeScopedCss)) {
         style = await import('../utils/shadow-css').then(m => m.scopeCss(style, scopeId, false));
       }
       registerStyle(scopeId, style, !!(cmpMeta.$flags$ & CMP_FLAGS.shadowDomEncapsulation));
-      Cstr.isStyleRegistered = true;
+      Cstr.$isStyleRegistered$ = true;
```
manucorporat: love it

src/runtime/vdom/vdom-render.ts
```diff
@@ -532,8 +532,8 @@ const relocateSlotContent = (
 
               // add to our list of nodes to relocate
               relocateNodes.push({
-                slotRefNode: childNode,
-                nodeToRelocate: node
+                $slotRefNode$: childNode,
```
manucorporat: good caught!

src/compiler/app-core/format-component-runtime-meta.ts
adamdbradley: Not everything is going to run in the browser, but in updating the ones that are I probably updated some others too. And I guess for consistency

src/compiler/browser/browser-scope-css.ts
adamdbradley: Yeah the shadow-css.ts file and fn needs to be renamed.

src/compiler/transformers/core-runtime-apis.ts
adamdbradley: moduleFile needs to be serializable so we can cache it.

PR #1762, opened 2019-07-24
fix(runtime): Add regex to clear multiple spaces and line breaks in classList.

### LINKED ISSUE #1757: IE11 - Component does not load when class attribute is used in JSX template
**Stencil version:**
```
 @stencil/core@1.1.8
```

**I'm submitting a:**
[x] bug report
[ ] feature request

**Current behavior:**
Web Component does not load in IE11 with `@stencil/core` versions 1.1.8 and above when `class` attribute is used in JSX template.

**Expected behavior:**
Component should load in IE11. 

**Steps to reproduce:**
Create a Stencil starter project using the CLI.
```
npm init stencil
Pick a starter > component
Project name > any-name
```

in `my-component.tsx`, add class attribute in the div tag:
```tsx
import {Component, Prop, h, Element} from '@stencil/core';
import {format} from '../../utils/utils';

@Component({
  tag: 'my-component',
  styleUrl: 'my-component.css',
  shadow: false
})
export class MyComponent {
  /**
   * The first name
   */
  @Prop() first: string;
  
  /**
   * The middle name
   */
  @Prop() middle: string;
  
  /**
   * The last name
   */
  @Prop() last: string;
  
  @Element() el: HTMLElement;
  
  private getText(): string {
    return format(this.first, this.middle, this.last);
  }
  
  render() {
    return <div class="message">Hello, World! I'm {this.getText()}</div>;
  }
}
```

Add `--es5` in package.json's `start` script:
```
"start": "stencil build --dev --watch --serve --es5"
```

then run the app in IE11.

Console shows the following error:
image
**Related code:**

AlexSerrano22:
I figured out that when you are using template strings and it has line breaks the last fix does not cover this case properly, I did change the approach to use a regex to clear the extra spaces and line breaks from classList value.

it should fix #1757

### DIFF
--- a/src/runtime/vdom/set-accessor.ts
+++ b/src/runtime/vdom/set-accessor.ts
@@ -120,7 +120,5 @@ export const setAccessor = (elm: HTMLElement, memberName: string, oldValue: any,
   }
 };
 
-const parseClassList = (value: string | undefined | null): string[] => {
-  value = value ? value.trim().replace(/\s+/g, ' ') : value;
-  return (!value) ? [] : value.split(' ');
-};
+const parseClassList = (value: string | undefined | null): string[] =>
+  (!value) ? [] : value.trim().split(/\s+/);
--- a/src/runtime/vdom/set-accessor.ts
+++ b/src/runtime/vdom/set-accessor.ts
@@ -120,5 +120,7 @@ export const setAccessor = (elm: HTMLElement, memberName: string, oldValue: any,
   }
 };
 
-const parseClassList = (value: string | undefined | null): string[] =>
-  (!value) ? [] : value.split(' ').filter(c => c);
+const parseClassList = (value: string | undefined | null): string[] => {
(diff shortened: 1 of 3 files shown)


src/runtime/vdom/set-accessor.ts
```diff
@@ -120,5 +120,7 @@ export const setAccessor = (elm: HTMLElement, memberName: string, oldValue: any,
   }
 };
 
-const parseClassList = (value: string | undefined | null): string[] =>
-  (!value) ? [] : value.split(' ').filter(c => c);
+const parseClassList = (value: string | undefined | null): string[] => {
+  value = value ? value.trim().replace(/\s+/g, ' ') : value;
+  return (!value) ? [] : value.split(' ');
```
manucorporat: it could be:
```
const parseClassList = (value: string | undefined | null): string[] => {
    (!value) ? [] : value.trim().split(/\s+/)
```

right?

Issue #1765, opened 2019-07-26
Can't import multiple components on same page


misterjojo:
**Stencil version:**

```
 @stencil/core@1.2.0
```

**I'm submitting a:**

[x] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**
I try to import two stencil components into one single page like this with esm import. 
On chrome 75, it works on first access, if i refresh the page, only one component is loaded. i've got these errors in console :

````
component-b/dist/component-b/p-wfisufbz.entry.js:1 Failed to load resource: the server responded with a status of 404 (Not Found)
p-5c72aa3d.js:1 TypeError: Failed to fetch dynamically imported module: http://localhost:8080/component-b/dist/component-b/p-wfisufbz.entry.js
d @ p-5c72aa3d.js:1
p-5c72aa3d.js:1 TypeError: a is not a constructor
    at p-5c72aa3d.js:1
d @ p-5c72aa3d.js:1
p-5c72aa3d.js:1 Uncaught (in promise) TypeError: Cannot read property 'k' of undefined
    at p-5c72aa3d.js:1
````
It appears that component-a entries are loaded from component-b base path.
I've only seen this issue on Chrome and it appears only if you **don't have DevTools open**.

**Expected behavior:**
Components should load correctly.

**Steps to reproduce:**
You can reproduce the problem on this simple repo : 
https://github.com/misterjojo/stencil-import-issue

manucorporat:
Opened an issue in V8: https://bugs.chromium.org/p/v8/issues/detail?id=9558
I guess we will need to find a workaround... in the meantime

PR #1766, opened 2019-07-26
fix(import): use arrow fn instead of Function

misterjojo:
This PR fix issue #1765 .
It seems that sometimes (After a refresh and if DevTools is close), Chrome was using the wrong function when importing. 
This is weird and i can't explain why so if anyone has an explanation, i'll take it.

Another working solution was to use another variable name across components instead of 'w', i found it less elegant.

### DIFF
--- a/src/client/import-shims.ts
+++ b/src/client/import-shims.ts
@@ -44,7 +44,7 @@ export const patchBrowser = async (): Promise<d.CustomElementsDefineOptions> => 
 export const patchDynamicImport = (base: string) => {
   const importFunctionName = getDynamicImportFunction(NAMESPACE);
   try {
-    (win as any)[importFunctionName] = new Function('w', 'return import(w);');
+    (win as any)[importFunctionName] = (w: string) => import(w);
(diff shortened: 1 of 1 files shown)


manucorporat:
wow! this is insane. seems like a chrome bug... i will look into it

misterjojo:
I just realized my fix won't work on ie11 as it doesn't support arrows fonctions.

As i mentioned it works if we replace the variable name 'w'. We could use NAMESPACE to be sure it is unique. I'll try to update my pr with that fix.

manucorporat:
Created another PR: https://github.com/ionic-team/stencil/pull/1768
The main problem is not IE11, we transpile to es5, the problem is Edge, when it sees import() it will crash at parse time! we have to hide it to browsers that does not support dynamic import()

Once Edge dies (and gets replaced by the chromium based one) we will be able to remove all this code!

misterjojo:
Thanks for the clarification, i'm closing this PR in favor of #1765 .

PR #1779, opened 2019-08-01
fix(compiler): change props union type parsing

### LINKED ISSUE #1777: Escaped pipes in docs-readme markdown table are causing issues
**Stencil version:**

```
 @stencil/core@1.1.3
```

**I'm submitting a:**

[x] bug report
[ ] feature request

**Current behavior:**

When generating readme.md with the internal docs compiler, union types are output with single escaped special character `\|`:
https://github.com/ionic-team/stencil/blob/30aa0c792aa80545e0aaad02a830016ef0307d36/src/compiler/docs/readme/docs-util.ts#L35

This is causing issue for markdown parsers, storybook.js included as seen in this issue: https://github.com/storybookjs/storybook/issues/6869

**Expected behavior:**

Currently both `markdown-to-jsx` and `simple-markdown` libraries do not parse these escaped pipe characters correctly (nor does markdown preview in webstorm), resulting in broken tables. I except the fact that string literals when used in JavaScript regex ignores [single escaped special characters](https://www.regular-expressions.info/characters.html), as opposed to double escaped `\\|`, to be the main reason why this breaks markdown table parsing.

To make it easier to parse, and to avoid workarounds or new exception case supported in popular markdown parsers such as `simple-markdown`, in a non-standard way (as JS regex can't parse single escaped special characters), I would suggest to parse union types to `or` instead of `\|`.

**Steps to reproduce:**

Some simple code to illustrate, with sample markdown table generated from stencil.js, can be found here in test repo: https://github.com/robaxelsen/markdown-to-jsx-table-pipe-test

**Related code:**

See repo above.

**Other information:**

I am working on a PR for this.

robaxelsen:
Closes #1777.

### DIFF
--- a/src/compiler/docs/readme/docs-util.ts
+++ b/src/compiler/docs/readme/docs-util.ts
@@ -34,7 +34,7 @@ export class MarkdownTable {
 
 function escapeMarkdownTableColumn(text: string) {
   text = text.replace(/\r?\n/g, ' ');
-  text = text.replace(/\|/g, '\\|');
+  text = text.replace(/\|/g, 'or');
   return text;
 }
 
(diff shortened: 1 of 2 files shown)


adamdbradley:
thanks!

PR #1788, opened 2019-08-05
System.js build: Fix leaking variables, improve resource resolution logic

### LINKED ISSUE #1456: Component lazy loader detects wrong base URL
**Stencil version:**

```
 @stencil/core@0.18.1
```

**I'm submitting a:**
[x ] bug report
[ ] feature request

**Current behavior:**
Including components built by stencil dynamically can cause the lazy loader to find the wrong base URL for loading the rest of the component.

**Expected behavior:**
I should be able to dynamically add web components to a page.

**Steps to reproduce:**

Try to include two separate Stencil components on a page by appending script tags.
You will see in the browser network tab that after the initial request for component1.js and component2.js, the next requests for component bundles will only be to example2.com (the last script on the page):

```tsx
setTimeout(() => {
        [
          "https://example1.com/dist/component1.js",
          "https://example2.com/dist/component2.js"
        ].forEach(src => {
          const script = document.createElement('script');
          script.src = src;
          document.body.appendChild(script);
        })
      })
```

**Other information:**
In conversations in the Stencil Slack, someone else noted that the loader (in the initial component.js file) is missing some arguments that are present in development but lost after a build. I think it's the namespace, but I haven't seen any documentation about how that *should* work. 
For reference, my component when in development has the self-calling function with these arguments:

`(window, document, "namespace","namespace",0,"namespace.core.js", ...`

And when built with npm run build:

`(window,document,0,0,0,0,0,0, ...`

### LINKED ISSUE #1786: System.js build is leaking variables to window object
**Stencil version:**

```
 @stencil/core@1.2.3
```

**I'm submitting a:**

[x] bug report
[ ] feature request

**Current behavior:**
The system.js build output in `dist/<namespace>/<namespace>.js` contains the inlined contents of the css polyfill. All variables defined in this polyfill will be set on the global `window` object.

**Expected behavior:**
The inlined polyfill should be executed inside a self-executing anonymous function to avoid polluting the global `window` namepsace.

**Steps to reproduce:**

- Run a stencil build `stencil build`
- Inspect the end of the `dist/<namespace>/<namespace>.js` file, you will find the compiled CSS polyfill code unwrapped inside that file - operating in the global `window` context.
- Include that system.js file in your project and check `window.CustomStyle`, or `window.fixRelativeUrls`, or `window.clean` or `window.parse`, they all exist.

### LINKED ISSUE #1787: System.js build can resolve wrong resources url when included with defer attribute
**Stencil version:**

```
 @stencil/core@1.2.3
```

**I'm submitting a:**

[x] bug report
[ ] feature request

**Current behavior:**
Loading stencil's system.js loader with a `defer` flag can resolve the wrong resource path. For example, this will not work:

```html
<script defer src="path/to/stencil/system-js-build.js"></script>
<script src="path/to/some/other/lib"></script>
```

In the case above, the logic to resolve the resource url of the chunks will fail and point to `path/to/some/other/lib`.

https://github.com/ionic-team/stencil/blob/0b4d8142f19d508f5256558dddc9259349f12e92/src/compiler/component-lazy/generate-system.ts#L50-L59

The resolution logic selects all script tags, and then iterates backwards until it finds a script with a `src` or a `data-resources-url` attribute. At the time of executing the deferred script, the other script tag has already made it to the DOM and will be selected instead.

**Expected behavior:**
The system.js resource url should be resolved correctly, even if the entry file is loaded with a `defer` attribute.

**Suggested fixes:**
I guess an IE11 compatible approach would be to give that script tag a unique id or attribute for selection. E.g. `<script data-stencil-namespace="myproject" src="/path/to/stencil/system-js.build.js"></script>`.

Another workaround could be to modify the selector logic to first find a script with the `data-resources-url` attribute, and only if not found continue with the src resolution logic. This way, specifying the `data-resources-url` attribute will always lead to the correct path.

mofux:
- Wraps the polyfills and loading logic inside a self-executing function to avoid leaking variables to the window object. Fixes #1786 
- First looks for a script with `data-stencil-namespace` or `data-resources-url` to resolve the resources url, instead of always using the last script in the DOM. Fixes #1787, fixes #1456

### DIFF
--- a/src/compiler/component-lazy/generate-system.ts
+++ b/src/compiler/component-lazy/generate-system.ts
@@ -71,6 +71,9 @@ if (win.__stencil_cssshim) {
 } else {
   start();
 }
+
+// Note: using .call(window) here because the self-executing function needs
+// to be scoped to the window object for the ES6Promise polyfill to work
 }).call(window);
 `;
 }
--- a/src/compiler/component-lazy/generate-system.ts
+++ b/src/compiler/component-lazy/generate-system.ts
@@ -49,7 +49,7 @@ async function getSystemLoader(config: d.Config, corePath: string, includePolyfi
 ${polyfills}
 
 var doc = document;
-var scriptElm = doc.querySelector('script[data-resources-url]');
+var scriptElm = doc.querySelector('script[data-stencil-namespace="${config.fsNamespace}"]') || doc.querySelector('script[data-resources-url]');
 if (!scriptElm) {
   var allScripts = doc.querySelectorAll('script');
   for (var x = allScripts.length - 1; x >= 0; x--) {
--- a/src/compiler/component-lazy/generate-system.ts
+++ b/src/compiler/component-lazy/generate-system.ts
@@ -45,17 +45,21 @@ async function getSystemLoader(config: d.Config, corePath: string, includePolyfi
   const polyfills = includePolyfills ? await getAppBrowserCorePolyfills(config) : '';
   return `
 'use strict';
+(function () {
 ${polyfills}
 
 var doc = document;
-var allScripts = doc.querySelectorAll('script');
-var scriptElm;
-for (var x = allScripts.length - 1; x >= 0; x--) {
-  scriptElm = allScripts[x];
-  if (scriptElm.src || scriptElm.hasAttribute('data-resources-url')) {
-    break;
+var scriptElm = doc.querySelector('script[data-resources-url]');
+if (!scriptElm) {
+  var allScripts = doc.querySelectorAll('script');
+  for (var x = allScripts.length - 1; x >= 0; x--) {
+    scriptElm = allScripts[x];
+    if (scriptElm.src) {
+      break;
+    }
   }
 }
+
 var resourcesUrl = scriptElm ? scriptElm.getAttribute('data-resources-url') || scriptElm.src : '';
 var start = function() {
   var url = new URL('${corePath}', resourcesUrl);
@@ -67,5 +71,6 @@ if (win.__stencil_cssshim) {
 } else {
   start();
 }
+}).call(window);
 `;
 }
--- a/src/compiler/component-lazy/generate-system.ts
+++ b/src/compiler/component-lazy/generate-system.ts
@@ -49,12 +49,12 @@ async function getSystemLoader(config: d.Config, corePath: string, includePolyfi
 ${polyfills}
 
 var doc = document;
-var scriptElm = doc.querySelector('script[data-stencil-namespace="${config.fsNamespace}"]') || doc.querySelector('script[data-resources-url]');
+var scriptElm = doc.querySelector('script[data-stencil-namespace="${config.fsNamespace}"]');
 if (!scriptElm) {
   var allScripts = doc.querySelectorAll('script');
   for (var x = allScripts.length - 1; x >= 0; x--) {
     scriptElm = allScripts[x];
-    if (scriptElm.src) {
+    if (scriptElm.src || scriptElm.hasAttribute('data-resources-url')) {
       break;
     }
   }


src/compiler/component-lazy/generate-system.ts
```diff
@@ -67,5 +71,6 @@ if (win.__stencil_cssshim) {
 } else {
   start();
 }
+}).call(window);
```
manucorporat: do we need call() here?
can't it be `})()`?

src/compiler/component-lazy/generate-system.ts
```diff
@@ -45,17 +45,21 @@ async function getSystemLoader(config: d.Config, corePath: string, includePolyfi
...
 ${polyfills}
 
 var doc = document;
-var allScripts = doc.querySelectorAll('script');
-var scriptElm;
-for (var x = allScripts.length - 1; x >= 0; x--) {
-  scriptElm = allScripts[x];
-  if (scriptElm.src || scriptElm.hasAttribute('data-resources-url')) {
-    break;
+var scriptElm = doc.querySelector('script[data-resources-url]');
```
manucorporat: I would rather not modify this code, since it's very easy to break.c

src/compiler/component-lazy/generate-system.ts
mofux: That's what I've tried initially, but it turned out that the function scope was not bound to `window` when using `})()`. This did break the embedded ES6Promise polyfill unfortunately. Using `.call(window)` therefore explicitly sets the function scope to the `window` object, making the ES6Promise polyfill work again.

mofux:
I've updated the script resolution logic to first look for a script with the `data-stencil-namespace` property set to the `namespace` property configured in the `stencil.config.ts`. This is the most reliable way to tell the loader which script it should use to resolve the `resourcesUrl` from.

manucorporat:
@mofux I think it's called `[data-namespace]`, not `[data-stencil-namespace]`, no?
would you mind creating a different PR with the fix for the defer though?

Btw, would you mind adding the explanation of call(window) as comment?

Awesome job figuring all this out!

mofux:
@manucorporat Sure, I'll split out the defer fix to a different PR. 

The `[data-stencil-namespace]` attribute was introduced by myself and shouldn't be referred to anywhere outside the system.js loader itself... Do you want me to change it to `[data-namespace]` instead? I think `data-stencil-namespace` is a better name because it is scoped to stencil.

mofux:
Making two PRs out of this at the same time seems... complicated 🤔 The changes of one PR would conflict with the other. I could make one PR (defer) based on the other (var leak), but then the second PR would be identical to this one 😅

I guess it's best to remove the var leak stuff from this PR, then get this one merged, and then add another PR for the var leak based on the updated master branch?

I'd like this PR to become the one that fixes the defer stuff, because the comments and referenced issues mostly talk about that topic.

manucorporat:
So far, I can't merge the current PR since it would break existing applications. Also what's the point of adding defer? just curious

mofux:
> So far, I can't merge the current PR since it would break existing applications

The changes should be fully backwards-compatible now. There is no need to specify the `data-stencil-namespace` attribute with the script tag, but if you do, chances that the correct script is picked are much higher 😅 If not specified it will simply continue with the old lookup logic instead.

> Also what's the point of adding defer? just curious

You can use the `defer` or `async` attributes to [delay execution of the script](https://peter.sh/experiments/asynchronous-and-deferred-javascript-execution-explained/). IMO this is very useful when combining it with prerendered components because it can speed up the time to the first meaningful draw (the JS parsing can run after the page has already painted).

manucorporat:
@mofux I know... but modern browser (all but IE11) will use type="module" which will not even run all this code. I guess why try to optimize IE11. 

ok, now you fixed it: https://github.com/ionic-team/stencil/pull/1788/commits/0bc27ec200ffef3b8371a6c8a90efa3f6deaf4b3

:)

mofux:
> but modern browser (all but IE11) will use type="module" which will not even run all this code. I guess why try to optimize IE11.

Well, we had to opt out of the type="module" script because of two reasons:
- It doesn't contain all the polyfills. We had a problem with Firefox ESR 60 that has no native support for `customElements`, so the module script failed on it.
- The CMS of my customer doesn't support the ES6 modules (it does some magic around the scripts for it's live editing features)

IMO it is a wrong assumption to assume that the `nomodule` script is only used in IE11. It works perfectly fine in all other browsers as well, and there may be many reasons to prefer the `nomodule` script over the `module` script.

mofux:
@manucorporat I just noticed commit https://github.com/ionic-team/stencil/commit/3a44defb1bfd9fbb8f92389309640fd08c2eded0. Will this solve the leaking variables problems, or is this addressing a different problem?

CRHain88:
> The changes should be fully backwards-compatible now. There is no need to specify the `data-stencil-namespace` attribute with the script tag, but if you do, chances that the correct script is picked are much higher 😅 If not specified it will simply continue with the old lookup logic instead.

Is there a way to check to see if the namespace is a part of the script URL? I'm seeing an issue where this code is intermittently changing the URL of a third-party script that exists on the page outside of the Stencil Web Component and changing the filename to `corePath`.

For example, it's changing `https://foo.com/analytics.js` to `https://foo.com//p-80f7b62d.system.js`, breaking both the web component and the third-party script.

I've opened issue #1827 with more details.

PR #1814, opened 2019-08-16
feat(cli): add component generator

simonhaenisch:
```sh
stencil g
stencil generate
stencil generate path/to/my-comp
```

kapture-2

- fails when not run in a Stencil project root, i. e. not in the folder that contains `stencil.config.ts`
- fails when the component name doesn't contain any dashes
- allows to prefix the component name with a path (see gif)
- respects `srcDir` from `stencil.config.ts`
- generates all files into subfolders of `<srcDir>/components`
- doesn't allow overwriting existing files (`wx` flag)
- `shadow: true` is set
- `styleUrl` is only set if a stylesheet has been created

Things I'm not sure about:

- Should I have used `graceful-fs` over `fs` (don't see the benefit)?
- Was it correct to add `inquirer` to the list of packages to copy the license for (prod build)?
- Those CI tests also fail on master, so most likely unrelated? The tests passed locally, except for the karma timeouts.

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,6 +66,7 @@
     "@types/fs-extra": "^8.0.0",
     "@types/glob": "^7.1.1",
     "@types/graceful-fs": "^4.1.2",
+    "@types/inquirer": "6.5.0",
     "@types/is-glob": "^4.0.0",
     "@types/jest": "^24.0.13",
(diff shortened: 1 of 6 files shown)


adamdbradley:
🎉 thanks!!

PR #1820, opened 2019-08-20
improve(cli): replace inquirer with prompts package

simonhaenisch:
Greatly reduced the `dist/cli/index.js` file size (from 1.5 MB).

image


adamdbradley:
🎉

PR #1821, opened 2019-08-20
Old FF can't parse const/let

fabiolibre:
Firefox 33 is not able to load system.js because it uses const and let, which are not supported.


adamdbradley:
Merged here: https://github.com/ionic-team/stencil/commit/d5434e3db8d174d589b27c26fdbd8a60aa3ee221

Thanks!

Issue #1829, opened 2019-08-24
Add MouseEvent for tests


justinwaite:
**Stencil version:**

```
 @stencil/core@1.3.1
```

**I'm submitting a:**

[x] feature request

**Current behavior:**
Currently MouseEvent is not accessible during jest tests.

**Expected behavior:**
The user should be able to mock MouseEvent events for their unit tests.

**Steps to reproduce:**
Try using MouseEvent in a unit test for a component and you will receive a MouseEvent is not defined error.

**Other information:**

See #1581 when the KeyboardEvent was added.

I plan on creating a PR for this to mirror the work done for KeyboardEvent.

PR #1830, opened 2019-08-24
feat(mock-doc): add MouseEvent

justinwaite:
Adds MouseEvent to MockDoc so it can be used in unit tests.

Modeled after #1581 

Resolves issue #1829


manucorporat:
Thanks a lot for the PR! merged!

PR #1832, opened 2019-08-24
fix(readme generator): fix path sep on windows

### LINKED ISSUE #1708: relative path in docs incorrect on when build on windows
**Stencil version:**

```
 @stencil/core@1.1.3
```

**I'm submitting a:**

[x] bug report
[ ] feature request

**Current behavior:**

Links to other components use system path separator instead of URL path separator.

Take [this readme](https://github.com/mikaelkaron/stencil-xstate-router/blob/master/packages/stencil-xstate-router/src/components/xstate-router-navigo/readme.md) as an example. Because it was built on windows it uses `\` as a path separator instead of the correct `/`.

> I'm making an assumption here that the separator is reversed because I use a windows machine. I have not looked at the code or verified with other generated `readme.md` files from a linux system.
>
> In some way I guess it's my own fault for using windows as a dev system (it's not the first time I report path related bugs in a project) but I'm sure I'm not the only one.

**Expected behavior:**

Paths in `.md` files can safely be assumed to be intended to be consumed on the internet, so the path separator should _always_ be `/`.

**Steps to reproduce:**

Build a multi-component stencil project on a windows system and check the generated `readme.md` files.

**Related code:**

```tsx
```

**Other information:**

rtpHarry:
fixes #1708 

I was just trying to make some Ionic docs changes and ended up down a rabbit hole. I'm submitting this one line fix in case that's all that needs to be done, but I have only tested it on my own Windows 10 computer and it resolved the issue.

### DIFF
--- a/src/compiler/docs/readme/markdown-dependencies.ts
+++ b/src/compiler/docs/readme/markdown-dependencies.ts
@@ -50,7 +50,7 @@ export function depsToMarkdown(config: d.Config, cmp: d.JsonDocsComponent, cmps:
 function getCmpLink(config: d.Config, from: d.JsonDocsComponent, to: string, cmps: d.JsonDocsComponent[]) {
   const destCmp = cmps.find(c => c.tag === to);
   if (destCmp) {
-    const cmpRelPath = config.sys.path.relative(from.dirPath, destCmp.dirPath);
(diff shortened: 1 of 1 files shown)


manucorporat:
Thanks for the PR, but we already have a function to do this: https://github.com/ionic-team/stencil/commit/22d9c9306830c9345a8b9e5d43e1ed7be93d30d3

Still, thanks a lot for your time!!

PR #1838, opened 2019-08-28
feat(testing): add template rendering

manucorporat:


romulocintra:
Saving our day!!!!! Almost replicated Templatizer behaviour from Polymer 1 to have something like this ... handy for composition example creating data-grid with cmp's great @manucorporat

PR #1863, opened 2019-09-12
correct minify comments

mexx:
Currently the documentation comment are swapped for the `minifyJs` and `minifyCss`.


arjunyel:
Thank you for the PR!

Issue #1891, opened 2019-09-25
InnerHTML property of svg-style returns undefined on IE11.


sjanczewski:
**Stencil version:**
```
 @stencil/core@1.5.1
```

**I'm submitting a:**
[x] bug report
[  ] feature request
[  ] support request

**Current behavior:**
InnerHTML property of `style` tag inside of `svg` tag returns undefined on IE11. So `clean()` function throws the error.

**Expected behavior:**
No thrown errors.

**Steps to reproduce:**
Create `svg` tag with `style` tag inside.

**Related code:**

```tsx
// src\client\polyfills\css-shim\scope.ts
export function addGlobalStyle(globalScopes: CSSScope[], styleEl: HTMLStyleElement) {
  const css = parseCSS(styleEl.innerHTML); // <<< innerHTML is not available on IE11
  css.styleEl = styleEl;
  globalScopes.push(css);
}
```

```tsx
// src\client\polyfills\css-shim\css-parser.ts
function clean(cssText: string) {
  return cssText.replace(RX.comments, '').replace(RX.port, ''); // <<< cssText is undefined on IE11
}
```

**Other information:**
Instead of using `innerHTML`, would be better to use `textContent` property. The `textContent` works properly on all browsers and is more performant (does not parse HTML, and is faster). `style` tag contains just text without HTML. So using `textContent` is more appropriate.

This is known issue. See below please:
https://stackoverflow.com/questions/28129956/get-innerhtml-of-svg-tag-result-in-undefined-in-ie

manucorporat:
Fixed! thanks a lot for the PR!

PR #1892, opened 2019-09-25
fix(css-shim): Replaced innerHTML with textContent

sjanczewski:
Issue: #1891 

Instead of using `innerHTML`, would be better to use `textContent` property. 
The `textContent` works properly on all browsers and is more performant (does not parse HTML, and is faster). `style` tag does not contains HTML. So seems to be using `textContent` is more appropriate.

### DIFF
--- a/src/client/polyfills/css-shim/scope.ts
+++ b/src/client/polyfills/css-shim/scope.ts
@@ -17,7 +17,7 @@ export function parseCSS(original: string): CSSScope {
 }
 
 export function addGlobalStyle(globalScopes: CSSScope[], styleEl: HTMLStyleElement) {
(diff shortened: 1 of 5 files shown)


src/client/polyfills/css-shim/scope.ts
```diff
@@ -17,7 +17,7 @@ export function parseCSS(original: string): CSSScope {
 }
 
 export function addGlobalStyle(globalScopes: CSSScope[], styleEl: HTMLStyleElement) {
-  const css = parseCSS(styleEl.innerHTML);
+  const css = parseCSS(styleEl.textContent || '');
```
sjanczewski: To ensure that `parseCSS()` always receives String value as we don't have 100% control on `textContent` property.

PR #1893, opened 2019-09-26
improved resource-url extraction from script tag

nopantsmonkey:
Current code only get last script tag from the document and use it as source for stencil component which in my case is not a valid location for stencil component. I have updated the code with the previously added data attribute 'data-resource-url' and i am querying the script tag which contains the data-resource-url to get valid path for stencil component lazy loading.

The current code enforces developers to only add the script at the end of the document and it does not work a lot of time when the scripts are being injected by CI or other scripts and it keep on breaking and generating invalid src for stencil component script.


nopantsmonkey:
can anyone have a look at AppVeyor CI build? some other PRs also seems to be failing due to this.

Issue #1897, opened 2019-09-26
Element.getRootNode not guarding against being called with undefined nodes


lotz:
**Stencil version:**

```
 @stencil/core@1.5.2
```

**I'm submitting a:**

[x] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**

Now that `Element.isConnected` has been polyfilled, the `Element.getRootNode` polyfill is being called many more times. Sometimes, for some reason, it is called with nodes that are undefined, causing errors that prevent other scripts from initializing on the page. This can break the entire page loading.

**Expected behavior:**

The `Element.getRootNode` polyfill should guard against undefined nodes to avoid these errors.

**Steps to reproduce:**

**Related code:**

```tsx
/*!
Element.getRootNode()
*/
(function(c){
    function d(a){
        a=b(a);
        return a&&11===a.nodeType?d(a.host):a  // <-- added `a&&` here
    }
    function b(a){
        return a&&a.parentNode?b(a.parentNode):a  // <-- added `a&&` here
    }
    "function"!==typeof c.getRootNode&&(c.getRootNode=function(a){
        return a&&a.composed?d(this):b(this)
    })
})(Element.prototype);
```

Adding the `a&&` guard to the `d(a)` and `b(a)` functions will fix this issue. I will submit a PR with this change.

**Other information:**

PR #1898, opened 2019-09-26
fix(polyfill): guard against undefined nodes in getRootNode

lotz:
Fixes https://github.com/ionic-team/stencil/issues/1897.


manucorporat:
Thanks a lot for the PR!
https://github.com/ionic-team/stencil/pull/1898

PR #1950, opened 2019-10-14
fix(vdom): add dev check for unexpected vnode

manucorporat:
Screenshot 2019-10-14 at 21 21 05
Screenshot 2019-10-14 at 21 20 58

### DIFF
--- a/src/runtime/vdom/h.ts
+++ b/src/runtime/vdom/h.ts
@@ -10,6 +10,7 @@
 import * as d from '../../declarations';
 import { BUILD } from '@build-conditionals';
 import { isComplexType } from '@utils';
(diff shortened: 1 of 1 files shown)


src/runtime/vdom/h.ts
```diff
@@ -30,6 +31,12 @@ export const h = (nodeName: any, vnodeData: any, ...children: d.ChildType[]): d.
       } else if (child != null && typeof child !== 'boolean') {
         if (simple = typeof nodeName !== 'function' && !isComplexType(child)) {
           child = String(child);
+
+        } else if (BUILD.isDev && child.$flags$ === undefined) {
```
mikaelkaron: is it possible `child` could have been a `string`?

PR #1961, opened 2019-10-17
enhancement: Add better TS definition support to style attribute

kevinclarkadstech:
- Add csstype library
- Change style prop from an index signature object to use csstype

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,6 +88,7 @@
     "core-js-builder": "^3.2.0",
     "css-what": "3.2.0",
     "cssnano": "4.1.10",
+    "csstype": "2.6.7",
     "execa": "^2.0.3",
     "exit": "^0.1.2",
     "fast-deep-equal": "^2.0.1",
--- a/src/declarations/jsx.ts
+++ b/src/declarations/jsx.ts
@@ -1,4 +1,4 @@
-
+import * as CSS from 'csstype';
 declare namespace LocalJSX {
   export interface Element {}
   export interface IntrinsicElements {}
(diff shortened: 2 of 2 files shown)


adamdbradley:
This will have to be updated because the `csstype` package wouldn't be installed by the user since it'd be a devDependency. Additionally, we'd prefer it to not be another dependency, so we'll have to talk how to proceed on this.

kevinclarkadstech:
> This will have to be updated because the `csstype` package wouldn't be installed by the user since it'd be a devDependency. Additionally, we'd prefer it to not be another dependency, so we'll have to talk how to proceed on this.

But when installing stencil as a dev dependency for doing a project, i.e. ionic pwa toolkit, won't that install dependencies stencil needs?

I understand about the dependency though..well the DOM attributes are fairly stable...maybe just copy over the content of the definitions from ```csstype```?

kevinclarkadstech:
?

simonhaenisch:
@kevinclarkadstech the goal is to keep the dependencies as few as possible, and at the moment `typescript` is the only dependency that is installed by Stencil. Any other dependencies are bundled into Stencil, so the user doesn't have to worry about these. And only the needed parts of each dependency are bundled which keeps Stencil lean and quick to install.

manucorporat:
Style could also accept css variables:
```
style={{
 '--variable': something
}}
```

I don't think this types would be flexible enough

manucorporat:
Thank you very much for this PR, but It's very unlikely to merge this PR, it would add a new dependency and it would not be compatible with css variables out of the box.

kevinclarkadstech:
That makes sense.  However, I still consider this to be an open issue @manucorporat @adamdbradley .  Is this something you will work on in the future?  The HTML style attribute is very stable, you could just copy and paste from csstype lib into your d.ts file if you don't want another dependency.  The reason it's less of an issue for me is because I know css very well and know about the different camel casing vs kebab casing after using React + JSX, but I can imagine a developer newer to JSX and Stencil would find lack of TypeScript support pretty frustrating.

PR #1991, opened 2019-11-06
chore(compiler): Update to typescript 3.7

Nicolaidavies:
Update to the latest version of Typescript.

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "vermoji": "node scripts/vermoji.js"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


adamdbradley:
Thanks

robinsummerhill:
Typescript 3.7.2 has a breaking change around type information output for getters that is causing a lot of headaches (https://github.com/microsoft/TypeScript/issues/33939). It also means that component libraries build with Stencil v1.8.0+ are incompatible with Angular 8.

PR #1995, opened 2019-11-07
fix(polyfills): update baseURI for base w/out href

### LINKED ISSUE #1994: baseURI polyfill fails if base tag has no href
**Stencil version:**
1.7.5 (latest)

**I'm submitting a:**
[X] bug report - MAJOR
[ ] feature request

**Current behavior:**
In IE11, If the document has a base tag without a href attribute, the baseURI in dom.js returns undefined.  This breaks hydration, and NO COMPONENTS RENDER.

For instance, this occurs if the base tag is like this:
```
<base target="_parent">
```
Storybook, for instance, has this base tag on every storybook story page.

**Expected behavior:**
baseURI should use the document's URL if no href is provided by &lt;base&gt;.

**Steps to reproduce:**
Git repo to reproduce in Stencil + Storybook configuration:
https://github.com/bbellmyers/stencil-base-example

**Related code:**
See repo

**Other information:**
Causes error in the IE11 console like this:
Stencil-1 7 4-versus-IE11-Storybook

bbellmyers:
Fixes #1994 

If document base tag only has a target, and does not have an href, baseURI polyfill currently returns undefined.  Update the polyfill to check first if base.href has a value, otherwise returns document.URL as normal.

### DIFF
--- a/src/client/polyfills/dom.js
+++ b/src/client/polyfills/dom.js
@@ -44,7 +44,7 @@ if (typeof document.baseURI !== 'string') {
     configurable: true,
     get: function () {
       var base = document.querySelector('base');
-      if (base) {
+      if (base && base.href) {
         return base.href;
       }
       return document.URL;


bbellmyers:
There appear to be other polyfills in the codebase, for instance, in app.  Should this also be fixed?

bbellmyers:
@manucorporat @adamdbradley - thoughts?

simonhaenisch:
@bbellmyers what other polyfills are you referring to?

bbellmyers:
@simonhaenisch I think it’s just these two files. I had initially only fixed one, but added a second commit to fix the other one too.

Issue #2012, opened 2019-11-15
Support ShadowDom Polyfills (not just Scoped Fallback)


umdjm:
**Stencil version:**
```
 @stencil/core@1.8.0
```

**I'm submitting a:**
[X] bug report
[X] feature request
[ ] support request 

**Current behavior:**
Using a stencil component with slots doesn't work correctly in React on IE11.   

The [fallback to scoped](https://stenciljs.com/docs/component#component-options) approach doesn't work for React because the light-dom elements are moved and React gets confused.

To demonstrate, I created https://umdjm.github.io/StencilFallbackDemo/, which is adapted from the [React in One Minute Demo](https://reactjs.org/docs/add-react-to-a-website.html#add-react-in-one-minute).

On IE11, when you click "Like" the Stencil component fails to render with the template:
![](https://raw.githubusercontent.com/umdjm/StencilFallbackDemo/master/www/assets/img/IE11_React_Screenshot.png)

On Chrome, when you click "Like" the Stencil component shows the slot as expected:
![](https://github.com/umdjm/StencilFallbackDemo/blob/master/www/assets/img/Chrome_React_Screenshot.png?raw=true)

FWIW, this repro worked fine when we created the same Web Component with [LitElement](https://lit-element.polymer-project.org/).

I think the right fix for this is to support using Shadow Dom polyfills instead of the [fallback to scoped](https://stenciljs.com/docs/component#component-options) approach.  

We tried to load the ShadowDom polyfills ourselves, but ran into issues with CSS selectors.  The scoped CSS rewrites don't happen when the polyfills exist.  Stencil should provide us with a way to enable ShadowDom polyfills correctly so that slots and CSS work well.

**Steps to reproduce:**
1) Create a [stencil component](https://github.com/umdjm/StencilFallbackDemo/blob/master/src/components/my-component/my-component.tsx) that has a slot
2) Fix the fact that IE11 doesn't call render when children change by adding a mutation observer - see https://github.com/ionic-team/stencil/issues/2004#issuecomment-552691394
3) Use that component in a [React App](https://github.com/umdjm/StencilFallbackDemo/blob/master/www/index.html) 

**Related code:**

Stencil Component:
```tsx
  render() {
    return <div>
      HEADER

      <br/>

      SLOT (
        <slot />
      )

      <br/>
      FOOTER
    </div>;
  }
```

React App:
```tsx
          <my-component>
            {this.state.liked
              ? <p>Button was clicked!</p>
[...]

PR #2031, opened 2019-11-22
feat(): add forceShadowDom config option - this will disable scoped fallback in IE

schadenn:
Hi all,

this might not be the cleanest solution as I guess there's a lot of code being shipped for the scoped fallback even though you disable using it, but it's a quick fix and works together with  @webcomponents/webcomponentsjs.

Related to #2012 and various other IE issues that exist using the scoped fallback.

### DIFF
--- a/src/client/client-window.ts
+++ b/src/client/client-window.ts
@@ -17,7 +17,11 @@ export const plt: d.PlatformRuntime = {
   rel: (el, eventName, listener, opts) => el.removeEventListener(eventName, listener, opts),
 };
 
-export const supportsShadowDom = (BUILD.shadowDom) ? /*@__PURE__*/(() => !!doc.documentElement.attachShadow)() : false;
+export const supportsShadowDom =
+  BUILD.forceShadowDom ||
+  (BUILD.shadowDom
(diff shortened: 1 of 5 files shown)


viljamis:
Yes, please! Do you need any help so that we could get this merged in?

schadenn:
Actually I'm not sure if this PR has a real value as you could just "dirty hack" Stencil into ShadowDom mode by doing this before your component bundle loads:
```
if (!document.documentElement.attachShadow) {
  document.documentElement.attachShadow = true;
} 
```

That way the Stencil ShadowDom check will pass and you'll be able to try the webcomponent polyfills for `shadow: true` components.

To make this an actual feature of Stencil it would be cleaner to actually remove the "scoped fallback code" (I'm not sure what exactly this includes) from the bundle if `forceShadowDom: true`.

rwaskiewicz:
I apologize that someone on the team never responded to this PR. Since its original creation, Stencil has dropped support for IE (as of Stencil v4). As a result, we no longer support browsers that do not support shadow DOM. I'm going to close this out, but wanted to thank you for putting for the effort here and again apologize for the lack of response here. If you have any questions, please feel free to reach out.

PR #2033, opened 2019-11-22
chore(): refactor angular output target

mhartington:
👋 

This PR refactors the compiler's angular output target to use a Directive instead of proxy functions.

### Why

With recent updates to Angular's build tool (the buildOptimizer specifically) `proxyInput` and `proxyMethods` would be dropped from the final output since they are not part of a component. This would break tests in ionic's angular test setup.

### Changes proposed

Instead of using proxy functions, we use a decorator to decorate the component with the logic needed.

```ts
export function ProxyCmp(opts: { inputs?: any; outputs?: any; methods?: any }) {
  return (cls: any) => {
    if (opts.inputs) {
      proxyInputs(cls, opts.inputs);
    }
    if (opts.methods) {
      proxyMethods(cls, opts.methods);
    }
    if (opts.outputs) {
      proxyOutputs(cls, cls.el, opts.outputs);
    }
    return cls;
  };
}
```

During a build, the output would be 

```ts
export declare interface IonContent extends Components.IonContent {}
@ProxyCmp({inputs: ['color', 'forceOverscroll', 'fullscreen', 'scrollEvents', 'scrollX', 'scrollY'], outputs: ['ionScrollStart', 'ionScroll', 'ionScrollEnd'], 'methods': ['getScrollElement', 'scrollToTop', 'scrollToBottom', 'scrollByPoint', 'scrollToPoint']})
@Component({ selector: 'ion-content', changeDetection: ChangeDetectionStrategy.OnPush, template: '<ng-content></ng-content>', inputs: ['color', 'forceOverscroll', 'fullscreen', 'scrollEvents', 'scrollX', 'scrollY'] })
export class IonContent {
  ionScrollStart!: EventEmitter<CustomEvent>;
  ionScroll!: EventEmitter<CustomEvent>;
  ionScrollEnd!: EventEmitter<CustomEvent>;
  protected el: HTMLElement;
  constructor(c: ChangeDetectorRef, r: ElementRef, protected z: NgZone) {
    c.detach();
    this.el = r.nativeElement;
  }
}
```

This currently just reuses the existing logic in place, but can be optimized in the future when part of the DS plugins.

A similar PR was made against the DS plugins here 
https://github.com/ionic-team/stencil-ds-plugins/pull/12

### DIFF
--- a/src/compiler/output-targets/output-angular.ts
+++ b/src/compiler/output-targets/output-angular.ts
@@ -12,7 +12,7 @@ export async function outputAngular(config: d.Config, compilerCtx: d.CompilerCtx
   await Promise.all(
     angularOutputTargets.map(outputTarget => (
       angularDirectiveProxyOutput(config, compilerCtx, buildCtx, outputTarget))
-  ));
+    ));
 
   timespan.finish(`generate angular proxies finished`);
 }
@@ -65,6 +65,21 @@ function getProxies(components: d.ComponentCompilerMeta[]) {
     .map(getProxy)
     .join('\n');
 }
+function getProxyCmp(inputs: string[], outputs: string[], methods: string[]): string {
+
+  const hasInputs = inputs.length > 0;
+  const hasOutputs = outputs.length > 0;
+  const hasMethods = methods.length > 0;
+  const proxMeta: string[] = [];
+
+  if (!hasInputs && !hasMethods && !hasOutputs) { return''; }
+
+  if (hasInputs) proxMeta.push(`inputs: ['${inputs.join(`', '`)}']`);
+  if (hasOutputs) proxMeta.push(`outputs: ['${outputs.join(`', '`)}']`);
+  if (hasMethods) proxMeta.push(`'methods': ['${methods.join(`', '`)}']`);
+
+  return `@ProxyCmp({${proxMeta.join(', ')}})`;
+}
 
 function getProxy(cmpMeta: d.ComponentCompilerMeta) {
   // Collect component meta
@@ -72,10 +87,6 @@ function getProxy(cmpMeta: d.ComponentCompilerMeta) {
   const outputs = getOutputs(cmpMeta);
   const methods = getMethods(cmpMeta);
 
-  // Process meta
-  const hasInputs = inputs.length > 0;
-  const hasOutputs = outputs.length > 0;
-  const hasMethods = methods.length > 0;
 
   // Generate Angular @Directive
   const directiveOpts = [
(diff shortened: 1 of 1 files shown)


src/compiler/output-targets/output-angular.ts
```diff
@@ -102,19 +114,9 @@ export class ${tagNameAsPascal} {`];
   lines.push(`  constructor(c: ChangeDetectorRef, r: ElementRef, protected z: NgZone) {
     c.detach();
     this.el = r.nativeElement;`);
-  if (hasOutputs) {
```
manucorporat: I think outputs still need to be here, in the constructor, since they apply to the instance, adding own properties, not to the prototype.

src/compiler/output-targets/output-angular.ts
```diff
@@ -175,27 +177,47 @@ export const proxyInputs = (Cmp: any, inputs: string[]) => {
...
   });
 };
 
 export const proxyOutputs = (instance: any, el: any, events: string[]) => {
-  events.forEach(eventName => instance[eventName] = fromEvent(el, eventName));
+  events.forEach(eventName => (instance[eventName] = fromEvent(el, eventName)));
 };
+
+// tslint:disable-next-line: only-arrow-functions
+export function ProxyCmp(opts: { inputs?: any; outputs?: any; methods?: any }) {
```
manucorporat: Since ProxyCmp is a private API, we could use arguments instead of a single opts argument, this way it will minify a little better!
```suggestion
export function ProxyCmp(inputs: any, methods?: any) {
```

also, "output" should still be in the constructor (since it applies to own properties)

mhartington: I'm not opposed, but I think the object reads better for debugging capabilities.

hypery2k:
so this would make ionic-team/stencil-ds-plugins#12 obsolete?

src/compiler/output-targets/output-angular.ts
```diff
@@ -102,19 +113,12 @@ export class ${tagNameAsPascal} {`];
   lines.push(`  constructor(c: ChangeDetectorRef, r: ElementRef, protected z: NgZone) {
     c.detach();
     this.el = r.nativeElement;`);
-  if (hasOutputs) {
+  if(hasOutputs){
```
manucorporat:
```suggestion
  if (hasOutputs) {
```

src/compiler/output-targets/output-angular.ts
```diff
@@ -175,27 +179,44 @@ export const proxyInputs = (Cmp: any, inputs: string[]) => {
...
+        this.el[methodName].apply(this.el, args)
+      );
     };
   });
 };
 
 export const proxyOutputs = (instance: any, el: any, events: string[]) => {
-  events.forEach(eventName => instance[eventName] = fromEvent(el, eventName));
-};
+   events.forEach(eventName => instance[eventName] = fromEvent(el, eventName));
```
manucorporat: revert this two lines (spaces)

PR #2047, opened 2019-11-27
feat(test): Add two event detail jest matchers

### LINKED ISSUE #2046: Add api to check event detail that isn't the last event
Note, I do plan on opening a PR for this.

**Stencil version:**
```
 @stencil/core@1.8.1
```

**I'm submitting a:**
[ ] bug report
[x] feature request

**Current behavior:**
Right now, the `toHaveReceivedEventDetail` jest matcher only checks the [`lastEvent`](https://github.com/ionic-team/stencil/blob/master/src/testing/matchers/events.ts#L64). Jest provides [`toHaveBeenNthCalledWith`](https://jestjs.io/docs/en/expect#tohavebeennthcalledwithnthcall-arg1-arg2-) for the same purpose and stencil could provide the same for the events.

Currently, to check an event at some index, you'd have to do:

```js
const someEvent = await page.spyOnEvent('someEvent');

expect(someEvent.events[2].detail).toEqual({....});
```

**Expected behavior:**

It'd be nice to have:

```js
const someEvent = await page.spyOnEvent('someEvent');

expect(someEvent).toHaveNthReceivedEventDetail(2, {...});
```

mitchellsimoens:
Resolves #2046 

`toHaveFirstReceivedEventDetail` is like `toHaveReceivedEventDetail` except
it checks the `firstEvent` instead of always the `lastEvent`.

`toHaveNthReceviedEventDetail` is like the others except you specify an
index so that you can check other events that aren't the first and last
events in an easy way. This is like jest's [`toHaveBeenNthCalledWith`](https://jestjs.io/docs/en/expect#tohavebeennthcalledwithnthcall-arg1-arg2-)
matcher.

### DIFF
--- a/src/declarations/testing.ts
+++ b/src/declarations/testing.ts
@@ -83,6 +83,18 @@ declare global {
       toHaveReceivedEventDetail(eventDetail: any): void;
 
       /**
+       * When given an EventSpy, checks the first event has
+       * received the correct custom event `detail` data.
+       */
+      toHaveFirstReceivedEventDetail(eventDetail: any): void;
+
+      /**
+       * When given an EventSpy, checks the event at an index
+       * has received the correct custom event `detail` data.
+       */
+      toHaveNthReceivedEventDetail(index: number, eventDetail: any): void;
+
+      /**
        * Used to evaluate the results of `compareScreenshot()`, such as
        * `expect(compare).toMatchScreenshot()`. The `allowableMismatchedRatio`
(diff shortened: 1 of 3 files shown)


adamdbradley:
Thanks! Merged it into the new compiler: https://github.com/ionic-team/stencil/commit/ebc518dd258be44b9c5e74c440c5134086ab55f9

Also, would you be able to add some unit tests making sure it's doing what it should? Thanks

Issue #871, opened 2018-06-20
newly inserted, pre-hydrated DOM node gets double-wrapped content.


Xiphe:
**Stencil version:**

```
 @stencil/core@0.9.7
```

**I'm submitting a:**

[x] bug report
[x] feature request

Not sure whats the intended behaviour here :)

**Current behavior:**

Given I have a running stencil app
Using a `my-button` component rendering to

```html
<my-button class="hydrated"><button>My content</button></my-button>
```

When I deep-clone this node and append it to the DOM
Then it gets this stucture

```html
<my-button class="hydrated"><button><button>My content</button></button></my-button>
```

**Expected behavior:**

I would expect the component to pick up existing child nodes similar to the way ssr hydration works.

**Steps to reproduce:**

1. Start a stencil app
2. select a stencil component in the dev tools
3. run `$0.parentElement.appendChild($0.cloneNode(true))`

**Related code:**

`¯\_(ツ)_/¯`

**Other information:**

The whole thing seems quite complicated to me because stencil needs to know if a node has been rendered by the component itself or is `slot`-content.

I find it interesting, that a similar effect of this also happens when the cloned element and it's children contain `data-ssrc` data.

this is probably also related to https://github.com/ionic-team/stencil/issues/781

Once you finished triaging this issue I'd happily help finding a solution.

Stumbled upon this while trying to build a bridge between stencil/web-component and react: https://github.com/patternsonio/web-component-to-react

DennisJamin:
This seems to be very related to an issue that we are currently facing. I've created the following project to reproduce: https://github.com/DennisJamin/stenciljs-component-rerendering-issue

jthoms1:
@Xiphe Can I assume this is not using shadow dom?

PR #2051, opened 2019-11-29
fix(runtime): Do not double render when attaching cloned elements with Shadow DOM turned off #

### LINKED ISSUE #1948: Components without shadowdom are rendering twice when dynamically added to the DOM
**Stencil version:**
```
@stencil/core@1.7.0
```

**I'm submitting a:**

[ x] bug report
[ ] feature request

**Current behavior:**
When using a Stencil webcomponent without Shadowdom the inner contents of the webcomponent gets rendered twice when the component is added dynamically to the DOM. We've seen the behaviour occurring with AngularJS.

**Expected behavior:**
The component should only render once and cleanup previous rendered content.

**Steps to reproduce:**
- clone https://github.com/denniswiemer/stenciljs-render-issues
- `yarn` & `yarn start`
- browse to www directory
- view components which are added dynamically (timed)

patricksevat:
- In the edge case that we don't use Shadow DOM and have an cloned hydrated component that is (re-)added to the DOM,
make sure that all internal markup is removed, because Stencil will see this component as a new component and will re-add its internal markup as slot content.
This causes perceived double rendering (although technically it is correct) and this commit fixes that by keeping state by adding an 's-isc' attribute. An attribute is needed because the cloning strips off any non-native props such as comp['s-sn'].
- A real world use case where this happens is with AngularJS ng-if directive, fixes #1948

### DIFF
--- a/src/runtime/initialize-component.ts
+++ b/src/runtime/initialize-component.ts
@@ -123,4 +123,25 @@ export const initializeComponent = async (elm: d.HostElement, hostRef: d.HostRef
   } else {
     schedule();
   }
+
+  // For CLONED components that are re-added to the DOM (like ngIf in AngularJS)
+  // AND do not use shadow-dom, remove all children that are not slot content
+  if (elm.classList.contains('hydrated') && BUILD.slotRelocation && !BUILD.hydrateServerSide && !BUILD.hydrateClientSide) {
+    removeNonSlotNodes(elm);
+  }
+};
+
+const removeNonSlotNodes = (elm: Element) => {
+  Array.from(elm.children)
+  .forEach((node) => {
+    const numChildren = node.children && node.children.length;
+    // ['s-isc'] stands for is-slot-content and is only added to elements that belong to a slot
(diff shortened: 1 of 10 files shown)


patricksevat:
As this is my first PR for Stencil, I do not know all the ins and outs. If there is a better way to tackle this issue, let me know!

DennisJamin:
This will actually also fix the following issue: https://github.com/ionic-team/stencil/issues/871

DennisJamin:
This will actually also fix the following issue: https://github.com/ionic-team/stencil/issues/871 

CC @Xiphe

manucorporat:
Hi @patricksevat!
Thanks a lot for the PR! cloning a non shadow-dom component is indeed very tricky.

How about fixing this problem at the clone level, not sure i like the idea of creating a bunch or elements, then removing them (once attached already to the dom).

What if element.clone() never actually cloned the children?

patricksevat:
Closed, @manucorporat and myself are co-operating to get get this done in #2052

PR #2077, opened 2019-12-13
Add config flag for preventing script inlining

davidleger95:
We have some projects where we rely on script tags being in a certain place in the DOM. The stencil compiler moves these around when it inlines scripts. This flag would allow us to prevent inlining and keep script tags where we expect them to be in the HTML file.

### DIFF
--- a/src/compiler/html/inline-esm-import.ts
+++ b/src/compiler/html/inline-esm-import.ts
@@ -25,7 +25,7 @@ export async function optimizeEsmImport(config: d.Config, compilerCtx: d.Compile
 
   // If the script is too big, instead of inlining, we hash the file and change
   // the <script> to the new location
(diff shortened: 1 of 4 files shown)


src/compiler/html/inline-esm-import.ts
```diff
@@ -25,7 +25,7 @@ export async function optimizeEsmImport(config: d.Config, compilerCtx: d.Compile
 
   // If the script is too big, instead of inlining, we hash the file and change
   // the <script> to the new location
-  if (content.length > MAX_JS_INLINE_SIZE) {
+  if (config.allowInlineScripts && content.length > MAX_JS_INLINE_SIZE) {
```
davidleger95: Another possible solution: allow a config value for `MAX_JS_INLINE_SIZE` so can be set it to `0` to prevent inlining

PR #2083, opened 2019-12-17
fix(hydrate): don’t run initializeDocumentHydrate on unsupported browsers

### LINKED ISSUE #2068: Bug: Client-side hydration moves content out of <slot/> when Hydrate App or --prerender flag is used
**Stencil version:**
```
 @stencil/core@1.8.1
```
also tested in
```
 @stencil/core@1.9.0-1
```

**I'm submitting a:**
[x] bug report
[ ] feature request
[ ] support request

**Current behavior:**
Using Hydrate App you’re now able to generate static HTML for your components. Thank you for the amazing work on this! There’re some issues still though which make it currently impossible to use this feature (at least for our team): 

If you try to hydrate the generated HTML on the client-side, any component that has `scoped: true` instead of `shadow: true` enabled will break. The content is basically moved outside of the original slot, thus breaking for example accessibility of the markup and CSS related as well. Same thing happens in the latest Edge and IE11 for `shadow: true` components as well (because of the current scoped fallback I guess).

**Expected behavior:**
I would expect the DOM structure to NOT change so drastically during client-side hydration that it breaks CSS styles and accessibility.

**Steps to reproduce:**
1. Create a simple Stencil.js component with `scoped: true` enabled:

```js
import { Component, h } from "@stencil/core"

@Component({
  tag: "my-test",
  scoped: true
})
export class MyTest {
  render() {
    return (
      <label>
        <slot></slot>
      </label>
    )
  }
}
```

2. Create a basic HTML page that imports your component library and the simple custom element:

```html
<!doctype html>
<html>
  <head>
    <meta charset="utf-8">
    <script type="module" src="path-to-your-lib.esm.js"></script>
    <script nomodule="" src="path-to-your-lib.js"></script>
  </head>
  <body>
    <my-test>This is supposed to be inside a label</my-test>
  </body>
</html>
```

3. Build the app including `dist-hydrate-script` and render this page into string using the generated hydrate script, e.g. (simplified):

```
const results = await hydrate.renderToString(content)
return results.html
```

4. Examine the generated markup or use inspector JavaScript disabled. It’s 100% correct and accessible. If you now enable JavaScript in the browser again to do client-side hydration Stencil.js will move all content out of the slots for components that use `scoped: true`. Additionally all `<slot/>`s will also break in Edge and IE11 browsers.

**Other information:**

This is what the markup looks like for the above example JavaScript disabled. As you probably notice, the text is correctly inside a label:

Screen Shot 2019-12-08 at 11 35 48

Now if I enable JavaScript, this happens. The text content is moved outside of the label, thus breaking styles and accessibility:

Screen Shot 2019-12-08 at 11 38 26

Visually the results look like this:

Screen Shot 2019-12-12 at 20 20 02

vs.

Screen Shot 2019-12-12 at 20 20 09

viljamis:
This pull fixes #2068 partly by making the client-side hydration of prerendered content fully work in IE11 and latest Edge versions. 

With this small change `<slot/>` contents aren’t anymore breaking in IE11 and Edge when client-side hydration happens for prerendered content. This does not alter the functionality for Chrome and other modern browsers that support `attachShadow`.

### DIFF
--- a/src/runtime/client-hydrate.ts
+++ b/src/runtime/client-hydrate.ts
@@ -235,6 +235,11 @@ const clientHydrate = (
 
 
 export const initializeDocumentHydrate = (node: d.RenderNode, orgLocNodes: Map<string, any>) => {
+  // Make sure the below code isn’t executed in IE11 or Edge
+  // as it’d break prerendered content’s client-side hydration
+  if (!document.documentElement.attachShadow) {
+    return
+  }
   if (node.nodeType === NODE_TYPE.ElementNode) {
     let i = 0;
     for (; i < node.childNodes.length; i++) {


viljamis:
@manucorporat Maybe there’s a way to know in client-hydrate.ts whether the content has been prerendered already or not to fix the Chrome issue as well?

adamdbradley:
Awesome work! I had to make a few updates, and added some more tests and pushed the change here: https://github.com/ionic-team/stencil/commit/a6e573cad99488010bb847d88127e7b28fd18eb7 thanks for all the help!!

Issue #2106, opened 2020-01-06
unhandledRejection: diag.next is not iterable


kerosin:
**Stencil version:**

```
 @stencil/core@1.8.3
```

**I'm submitting a:**

[x] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**
after switching to ```pnpm``` on a private repo I get a error for the stencil build, probably related to missing imports.

```
stencil build
[51:29.8]  @stencil/core v1.8.3 
[51:29.9]  build, fls, prod mode, started ...
[51:29.9]  transpile started ...
[51:34.1]  transpile finished in 4.19 s
[51:34.1]  type checking started ...
[51:34.2]  copy started ...
[51:34.2]  generate styles started ...
[51:34.2]  bundling components started ...
[51:34.2]  copy finished (0 files) in 51 ms
[51:35.0]  generate styles finished in 820 ms

[ ERROR ]  unhandledRejection: diag.next is not iterable

[51:54.5]  bundling components finished in 20.35 s

[ ERROR ]  diag.next is not iterable
```
"**diag.next is not iterable**"

**Expected behavior:**
- print proper (error) message

**Steps to reproduce:**
- 

**Related code:**

https://github.com/ionic-team/stencil/blob/9983da0dce82b9f5aa41216c26c8c0d6ee3f79a2/src/utils/logger/logger-typescript.ts#L176

**Other information:**

adamdbradley:
PR Merged, thanks!

PR #2107, opened 2020-01-06
fix(logger): checking iterability of diag.next (instead of truthiness)

kerosin:
logger-typescript:
- checking iterability instead of truthiness

fixes error on https://github.com/ionic-team/stencil/issues/2106


adamdbradley:
Thanks!

Issue #2109, opened 2020-01-06
Template renderers remove hydrated class


abdulmhamid:
**Stencil version:**

```
 @stencil/core@1.8.1
```

**I'm submitting a:**

[x] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**
When using conditional template logic on a Stencil element in an Ember application (and presumably with other template renderers), the `hydrated` class is removed, causing the element to lose its visibility.

**Expected behavior:**
Hydration state not lost on class update. 

**Steps to reproduce:**
1. In an Ember application, use a Stencil element in an `hbs` template
2. Bind the `class` property of the element to an ember property or use a conditional class
3. Update the bound property. The `hydrated` class is removed at this point.

**Related code:**

```hbs
<stencil-element type="foo" class="bar {{if baz "biz" }}"></stencil-element>
```

**Other information:**
 I would recommend surfacing hydration through a `data-` attribute instead of using something as often mutated as a class.

abdulmhamid:
I developed my proposed solution locally but I cannot push to submit a pull request.

adamdbradley:
Thank you for opening the issue. However, at this time we've decided to not implement this feature. Instead I'd recommend always keeping the hydrated class on the element, or modifying the css of what hydrated does for your applications.

abdulmhamid:
The issue here is that Ember's template rendering engine fully rewrites the class property when bound to something that changes in the Ember code.  We are not actively removing the `hydrated` class but are dependent on the visibility setting handled in Stencil nonetheless.  With this in its current state, our only recourse is to force the hydrated class on the elements at all times, overriding the rules set by the Stencil bootstrap and we will have elements visible before they are ready.

Do you know of any alternatives to this? If not, please reconsider the suggestion. We could also have the class and data attribute coexist.  The key here is that we need a way to know whether or not the elements are ready to be visible before manually implementing the `hydrated` class.

d0whc3r:
maybe a parameter in config could be useful to indicate that you want to use attribute instead of a css class

paulcpederson:
> maybe a parameter in config could be useful to indicate that you want to use attribute instead of a css class

[...]

PR #2110, opened 2020-01-06
fix(runtime): apply hydrated marker with data- attribute instead of c…

abdulmhamid:
Relates to https://github.com/ionic-team/stencil/issues/2109.

### DIFF
--- a/src/cli/task-generate.ts
+++ b/src/cli/task-generate.ts
@@ -169,7 +169,7 @@ describe('${name}', () => {
     await page.setContent('<${name}></${name}>');
 
     const element = await page.find('${name}');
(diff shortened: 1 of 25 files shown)


adamdbradley:
Sorry this would be too much of a breaking change.

abdulmhamid:
Can you clarify how this is a breaking change?  The class seems as though it was meant to be private but it can be currently overwritten.  I updated and ran all of the tests and they passed.

Is the concern that other users are relying on the class in some other way?  If so, can the solutions coexist so we have both the class representation and a more private one (be that the data attribute or something else)?  Using class here is fragile and template renderers typically set properties fully and ignore changes from outside sources.

PR #2113, opened 2020-01-07
Every JSX element should have a "key"

kraftwer1:
Fixes:

<svg key={...} /> // Property 'key' does not exist on type 'SVGAttributes<SVGElement>'.ts(2322)


adamdbradley:
Good catch, thanks!

PR #2128, opened 2020-01-12
fix(cli): Update tests in task generator

splitinfinities:
For generated components, this gets the default up to 100% by default.


adamdbradley:
🎉

PR #2145, opened 2020-01-17
feat: Disable the dependencies part of the markdown docs.

eNepper:
I have added a dependencies: boolean property to OutputTargetDocsReadme, this allows the developer to disable the dependencies part of the markdown readme.md file.

If the property is undefined the dependencies will still be rendered.

### DIFF
--- a/src/compiler/docs/readme/output-docs.ts
+++ b/src/compiler/docs/readme/output-docs.ts
@@ -33,7 +33,7 @@ export const generateReadme = async (config: d.Config, compilerCtx: d.CompilerCt
 
 export const generateMarkdown = (config: d.Config, userContent: string, cmp: d.JsonDocsComponent, cmps: d.JsonDocsComponent[], readmeOutput: d.OutputTargetDocsReadme) => {
   //If the readmeOutput.dependencies is true or undefined the dependencies will be generated.
-  const dependencies = readmeOutput.dependencies === true || readmeOutput.dependencies === undefined ? depsToMarkdown(config, cmp, cmps) : [];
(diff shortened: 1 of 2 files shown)


src/compiler/docs/readme/output-docs.ts
```diff
@@ -31,7 +31,9 @@ export const generateReadme = async (config: d.Config, compilerCtx: d.CompilerCt
   }));
 };
 
-export const generateMarkdown = (config: d.Config, userContent: string, cmp: d.JsonDocsComponent, cmps: d.JsonDocsComponent[], footer: string) => {
+export const generateMarkdown = (config: d.Config, userContent: string, cmp: d.JsonDocsComponent, cmps: d.JsonDocsComponent[], readmeOutput: d.OutputTargetDocsReadme) => {
+  //If the readmeOutput.dependencies is true or undefined the dependencies will be generated.
+  const dependencies = readmeOutput.dependencies === true || readmeOutput.dependencies === undefined ? depsToMarkdown(config, cmp, cmps) : [];
```
manucorporat:
```suggestion
  const dependencies = readmeOutput.dependencies !== false ? depsToMarkdown(config, cmp, cmps) : [];
```

eNepper: Will this not break backwards compatibility? If readmeOutput.dependencies is undefined then undefined !== false will evaluate to true

eNepper: After looking at it again I see that I got it backwards and that your suggestion is correct.

marksdk:
Disclosure: I'm on the same team as @eNepper.

@adamdbradley Any chance you can review this? We'd love to see this change merged. Thanks!

adamdbradley:
thanks!

PR #2147, opened 2020-01-17
feat(cli): add ComponentInterface in component generator template

jeanbenitez:


simonhaenisch:
Is this recommended? I've not been doing it anywhere... couldn't be bothered for the extra import for almost no benefit, but I guess it's helpful to avoid typos with life cycle methods?

Ideally we could add that type information through the decorator, but that's a long open issue: https://github.com/microsoft/TypeScript/issues/4881.

jeanbenitez:
That is exactly why I'm encouraged to add it. Some colleagues are starting to use Stencil, and the interface provides them with that help, in addition to a small documentation on each of those methods. It is a simple help to improve the DX of newbies. They are using generate cli a lot, thanks for your contribution!

To add via decorators looks like a great solution, I hope in a future we can do it.

Thanks for answering!

simonhaenisch:
Just wondering whether we could automatically add something like this to `components.d.ts` for each component:

```ts
import { ComponentInterface } from '@stencil/core';
import { MyComponent } from './components/my-component';

declare module './components/my-component' {
  export interface MyComponent extends ComponentInterface {}
}
```

The interface declaration would get merged with the class declaration so that the class declaration extends `ComponentInterface` automatically, i. e. without needing to add `implements ComponentInterface` to each component class (got the idea from https://github.com/microsoft/TypeScript/issues/4881#issuecomment-558922745). Thoughts @manucorporat?

PR #2160, opened 2020-01-24
fix(runtime): apply hydrated marker as attribute alongside class

abdulmhamid:
Another pass at addressing https://github.com/ionic-team/stencil/issues/2109 to not break those dependent on the `hydrated` class. We could also hide the attribute bind behind a config and add either the class or the attr; that seems like unnecessary config API bloat at this time, though.

### DIFF
--- a/src/cli/task-generate.ts
+++ b/src/cli/task-generate.ts
@@ -170,6 +170,7 @@ describe('${name}', () => {
 
     const element = await page.find('${name}');
     expect(element).toHaveClass('hydrated');
+    expect(element).toHaveAttribute('data-hydrated');
   });
 });
 `;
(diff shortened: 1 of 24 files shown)


simonhaenisch:
I believe this has been solved by https://github.com/ionic-team/stencil/commit/62e2168e1f49d3778b269fae10c8b77e30dc1c49 now.

See

https://github.com/ionic-team/stencil/blob/51dd9f50022eef9f29156893974d2a935065999b/src/declarations/stencil-public-compiler.ts#L180-L189

and

https://github.com/ionic-team/stencil/blob/51dd9f50022eef9f29156893974d2a935065999b/src/declarations/stencil-public-compiler.ts#L299-L322

PR #2182, opened 2020-02-04
Feature: Add in extraSrcDirs

nfavero:
**Feature**: Add in the ability to specify extraSrc Dirs
**Reason**: This allows for external src dependencies that can be brought into the src.  This is useful for developers using aliased paths and MonoRepos (like NX)

### DIFF
--- a/src/compiler/config/test/validate-config.spec.ts
+++ b/src/compiler/config/test/validate-config.spec.ts
@@ -102,6 +102,23 @@ describe('validation', () => {
       ]);
     });
 
+    it('should add extraSrcDirs globs', () => {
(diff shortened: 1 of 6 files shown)


christian-bromann:
Thanks for the contribution @nfavero 

Unfortunately no one from the Stencil team was able to give you a response which I apologize for. We started an effort to revive all PRs and issues that have been forgotten. The team looked into this and has decided no to move forward with this. We believe that this may not be required today anymore and also unsure about the incentive to add the option. If you are still interested in this feature we would love to connect and work on understanding the reasons why this may useful for your setup. Let us know your thoughts!

Thanks again for the contribution and support for Stencil!

PR #2204, opened 2020-02-18
Add ionic-app test

simonhaenisch:
**Proposed changes:**

- Adds an `ionic-app` test to the list of tests that are run with `npm run test.analysis`. The test was created with `npm init stencil ionic-pwa ionic-app`.
- Adds a `www` output target to `hello-vdom` because it was missing and the analysis script couldn't find it.
- Adds missing fields to `hello-vdom/package.json` to stop some warnings about those fields.

**Current bugs:**

 - ~In `src/global/app.css`, the ionic imports don't work because of the tilde that's supposed to resolve to `node_modules`. It throws errors like Unable to read css import: `@import "~@ionic/core/css/core.css";` (a temporary work-around is to replace `~` with `../../node_modules/`).~ Fixed by cc7e15494e267cdef96f5a066f18a6f75fc9bdb2.
- ~In prod builds, the app doesn't load because `app-root_5.entry.js:1 [ion-router] the path does not match any route`.~ Fixed by e62f2126a3a9f30d9475f707db1b466084d0c863.
- ~In prod builds, the initial render of the `ion-split-pane` doesn't work as expected (this has been happening since at least `1.9.0-12` with our app).~ Fixed by d7793c33a64a92b9c02b176e86b7a4a1e7b94d84.
- In prod builds, within an `ion-grid.ion-no-padding`, the `ion-row`'s and `ion-col`'s padding is not working as expected (there is no padding at all between neither cols nor rows). Haven't made a repro yet but will once the router bug is solved.
- Mysterious `DOMException: Failed to execute 'insertBefore' on 'Node': The new child element contains the parent`.

All the bugs that only occur in prod builds are not caused by Ionic IMO (the only Ionic 5 bug i've had also happens in dev builds). It's not currently possible to create e2e tests for prod builds, so I can't build repros that are runnable by the test runner (adding `--prod` to the test command doesn't work).


simonhaenisch:
When I undo the line https://github.com/ionic-team/stencil/commit/cc79e39d3f874bcb1db733a4f9b15b0f0c5ccb42#diff-68526b921ad8015b94d5496a0cace164L139-R133 (`fix(css): fix node_module css imports`), which seems to be the only real change of the commit (apart from changing functions into arrow functions), then the `~` gets resolved properly again.

@adamdbradley remember why you removed `{ packageJson: true }` from the `config.sys.resolveModule` call?

simonhaenisch:
> * In prod builds, the app doesn't load because `app-root_5.entry.js:1 [ion-router] the path does not match any route`.

If I set `minifyJs: false` in the stencil config, that bug doesn't happen in prod.

```js
async writeNavStateRoot(path, direction) {
    if (!path) {
        console.error('[ion-router] URL is not part of the routing set');
        return false;
    }
    // lookup redirect rule
    const redirects = readRedirects(this.el);
    const redirect = routeRedirect(path, redirects);
    let redirectFrom = null;
    if (redirect) {
        this.setPath(redirect.to, direction);
        redirectFrom = redirect.from;
        path = redirect.to;
    }
    // lookup route chain
    const routes = readRoutes(this.el);
    const chain = routerPathToChain(path, routes);
    if (!chain) {
        console.error('[ion-router] the path does not match any route');
        return false;
    }
    // write DOM give
    return this.safeWriteNavState(document.body, chain, direction, path, redirectFrom);
}
async safeWriteNavState(node, chain, direction, path, redirectFrom, index = 0) {
    const unlock = await this.lock();
    let changed = false;
    try {
        changed = await this.writeNavState(node, chain, direction, path, redirectFrom, index);
    }
    catch (e) {
        console.error(e);
    }
    unlock();
    return changed;
}
```

gets minified to

```js
async writeNavStateRoot(t, i) {
    if (!t)
        return console.error("[ion-router] URL is not part of the routing set"),
        0;
    const o = Tt(this.el)
      , e = jt(t, o);
    let n = null;
    e && (this.setPath(e.to, i),
    n = e.from,
    t = e.to);
    const s = ((t,i)=>{
        let o = null
          , e = 0;
        for (const n of i) {
            const i = Mt(t, n);
            if (null !== i) {
                const t = Et(i);
                t > e && (e = t,
                o = i)
            }
        }
        return o
    }
    )(t, Dt(this.el));
    return s ? this.safeWriteNavState(document.body, s, i, t, n) : (console.error("[ion-router] the path does not match any route"),
    0)
}
async safeWriteNavState(t, i, o, e, n, s=0) {
    const r = await this.lock();
    let a = 0;
    try {
        a = await this.writeNavState(t,i,o,e,n,s)
    } catch (l) {
        console.error(l)
    }
    return r(),
    a
}
```

and the error happens here:

```js
return s ? this.safeWriteNavState(document.body, s, i, t, n) : (console.error("[ion-router] the path does not match any route"),
0)
```

Full files:
- [app-root.entry.js.txt](https://github.com/ionic-team/stencil/files/4219754/app-root.entry.js.txt)
- [app-root.entry.min.js.txt](https://github.com/ionic-team/stencil/files/4219757/app-root.entry.min.js.txt)

simonhaenisch:
Removing

```ts
opts.compress.booleans_as_integers = true
```

from the terser options fixes the issues with js minification.

PR #2208, opened 2020-02-21
fix(vdom): do not reset $slotRefNode$ when relocating nodes

### LINKED ISSUE #2159: Bug: The ordering of slots in nested components is incorrect in Edge.
**Stencil version:**

```
 @stencil/core@1.8.5
```

**I'm submitting a:**

[x] bug report
[ ] feature request

**Current behavior:**
The ordering of slots in nested components is incorrect in Edge.

Example: https://driskull.github.io/stencil-component-starter/

**Expected behavior:**
The component should display the following:

```
Should be first.
Should be second.
Should be third.
```

However, the following is displayed in edge browser:

```
Should be first.
Should be third.
Should be second. 
```

**Steps to reproduce:**

**Related code:**

https://github.com/ionic-team/stencil-component-starter/compare/master...driskull:master

**Other information:**

simonhaenisch:
The bug was introduced by e7b9b3807ad0250f097f6aff11c897b6192b4096 which fixed #2159.

The components are **`slot-cmp`** which just has a slot:

```tsx
<Host>
  <slot />
</Host>
```

and **`slot-parent-cmp`** which is a parent of `slot-cmp` and passes another slot into its slot:

```tsx
<Host>
  <slot-cmp>
    <slot />
  </slot-cmp>
</Host>
```

The issue occurs when wrapping the `slot-parent-cmp` in another `slot-cmp`:

```tsx
<slot-cmp>
  <slot-parent-cmp />
</slot-cmp>
```

and then causing an update (e. g. calling `forceUpdate) of the outer `slot-cmp`.

The error message is

```
DOMException: Failed to execute 'insertBefore' on 'Node': The new child element contains the parent.
```

The reason is that `vdom-render.ts` tries to re-locate nodes and somehow it tries to move the `slot-parent-cmp` in the `slot-cmp` it contains rather than the one it's contained in.

### DIFF
--- a/src/runtime/vdom/vdom-render.ts
+++ b/src/runtime/vdom/vdom-render.ts
@@ -544,7 +544,7 @@ const relocateSlotContent = (elm: d.RenderNode) => {
               relocateNodes.forEach(relocateNode => {
                 if (isNodeLocatedInSlot(relocateNode.$nodeToRelocate$, node['s-sn'])) {
                   relocateNodeData = relocateNodes.find(r => r.$nodeToRelocate$ === node);
-                  if (relocateNodeData) {
+                  if (relocateNodeData && !relocateNode.$slotRefNode$) {
                     relocateNode.$slotRefNode$ = relocateNodeData.$slotRefNode$;
                   }
                 }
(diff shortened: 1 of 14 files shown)


simonhaenisch:
Nice, this works now! I'll remove the `slot-cmp` from the `index.html` again because the `slot-cmp-container` e2e test contains the reproduction.

PR #2209, opened 2020-02-21
Don't show warning for inputs without `value` attribute

### LINKED ISSUE #2158: The "value" prop of <input> should be set after (...) but is not declared
**Stencil version:**

```
@stencil/core@1.8.3
```

**I'm submitting a:**
[x] bug report
[ ] feature request

**Current behavior:**
When rendering a `<input />` element without value, a warning is logged, mentioning the importance of order. Although there are no attributes set to order them so.

**Expected behavior:**
To *NOT* show a warning, when there is *NO* value.

**Steps to reproduce:**
Make a component that just renders an input field, without defining a value.

**Related code:**

```tsx
    render() {
        return (
            <input
                ref={ref => (this.inputElement = ref)}
                type="text"
                id="filterText"
                placeholder="Full name search..."
            />
        );
    }
```

**Other information:**
The problem is caused because the index of the `value` attribute is checked against the index of the `type` attribute. Because there is no value, it's index is `-1`, therefor a misleading warning will be logged.

tricki:
Fixes #2158 

Hide the following warning for inputs without a `value` attribute:

`The "value" prop of <input> should be set after "min", "max", "type" and "step"`

### DIFF
--- a/src/runtime/vdom/h.ts
+++ b/src/runtime/vdom/h.ts
@@ -155,6 +155,9 @@ const validateInputProperties = (vnodeData: any) => {
   const maxIndex = props.indexOf('max');
   const stepIndex = props.indexOf('min');
   const value = props.indexOf('value');
+  if (value === -1) {
+    return;
+  }
   if (value < typeIndex || value < minIndex || value < maxIndex || value < stepIndex) {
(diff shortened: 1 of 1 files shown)


adamdbradley:
Good catch, thanks!

PR #2318, opened 2020-03-26
feat(runtime): disable async queue

manucorporat:


src/compiler/output-targets/dist-hydrate-script/generate-hydrate-app.ts
```diff
@@ -119,6 +119,7 @@ const getHydrateBuildConditionals = (config: d.Config, cmps: d.ComponentCompiler
   updateBuildConditionals(config, build);
   build.lifecycleDOMEvents = false;
   build.devTools = false;
+  build.asyncQueue = false;
```
manucorporat: makes no effect

Issue #2307, opened 2020-03-24
Inconsistent rendering of a slot's fallback content


kensodemann:
**Stencil version:**

```
 @stencil/core@1.11.2
```

**I'm submitting a:**

[x] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**
The fallback content of a slot is rendered differently in evergreen browsers (works as expected ✅), old Edge ❌and IE11 ❌(rendering is different between these two). See screenshots

![default-slot-content__chrome-ff-safari](https://user-images.githubusercontent.com/1526846/77452549-10bc1280-6dc4-11ea-85f5-d4efeac3e7bd.png)
![default-slot-content__ie11](https://user-images.githubusercontent.com/1526846/77452562-13b70300-6dc4-11ea-9721-f9f2e1025b0b.png)
![default-slot-content__old-edge](https://user-images.githubusercontent.com/1526846/77452576-174a8a00-6dc4-11ea-8906-999ff5b03482.png)

**Expected behavior:**
Same as evergreen browsers in pre-chromium Edge and IE11.

**Steps to reproduce:**
Sample app: https://github.com/jcfranco/stencil-default-slot-content-old-edge-and-ie11

Components displays slotted data with a fallback value of `label`: https://github.com/jcfranco/stencil-default-slot-content-old-edge-and-ie11/blob/master/src/components/my-component/my-component.tsx#L10-L12

Tests with fallback label, with and without slotted content: https://github.com/jcfranco/stencil-default-slot-content-old-edge-and-ie11/blob/master/src/index.html#L14-L15

Note: sample uses 1.8.8, but I upgraded to 1.11.2 locally on a Windows machine, and the issue still exists.

kevin-coyle:
I've tested this on 1.12.1 and I'm still getting this issue as well, should it be reopened or should I open a new issue?

PsySolix:
Is there any progress on this? On all browsers this offers an issue... A framework for webcomponents should make sure that default/fallback slots work as intended (even with shadowDOM)..


component-a.tsx
`
render(){
....
	<div class="wrapper"><slot><input type="text" id={inputId} ... /> </slot></div>
....
}
`
Usage
`<xx-input/>` -> should render with the default input -> WORKS
`<xx-input><p>test</p></xx-input>` -> should render with ONLY the `<div><p>test</p></div>`

[...]

PR #2336, opened 2020-04-06
fix(vdom): Hide fallback slot when content is entered

kevin-coyle-unipro:
Since version 1.8.7 slotted content shows both the fallback content and the inserted content. When shadow is unavailable and it is using scoped slots.

Originally the tests passed because there didn't seem to be a test case on this scenario. In this PR we have added an additional test case to check the attribute of the slot and removed the conditional that prevented the hidden attribute from being added.

This seems to relate to issue https://github.com/ionic-team/stencil/issues/2307 which although was closed does not seem to be fixed.

### DIFF
--- a/src/runtime/vdom/test/scoped-slot.spec.tsx
+++ b/src/runtime/vdom/test/scoped-slot.spec.tsx
@@ -755,4 +755,26 @@ describe('scoped slot', () => {
     // expect(root.firstElementChild.firstElementChild.children[1].firstElementChild.children[1].nodeName).toBe('GOAT');
     // expect(root.firstElementChild.firstElementChild.children[1].firstElementChild.children[1].textContent).toBe('hey goat!');
   });
+
+  it('should hide the slot\'s fallback content when slot content passed in', async() => {
+    @Component({ tag: 'fallback-test', scoped: true })
+    class FallbackSlotTest {
+      render() {
+        return (
+          <div>
+            <slot>
+              <p>Fallback Content</p>
+            </slot>
+          </div>
+        );
+      }
+    }
+    const { root } = await newSpecPage({
+      components: [FallbackSlotTest],
+      html: `<fallback-test><span>Content</span></fallback-test>`,
+    });
+
+    expect(root.firstElementChild.children[1].nodeName).toBe('SLOT-FB');
+    expect(root.firstElementChild.children[1]).toHaveAttribute('hidden');
+  });
 });
--- a/src/runtime/vdom/vdom-render.ts
+++ b/src/runtime/vdom/vdom-render.ts
@@ -418,24 +418,22 @@ const updateFallbackSlotVisibility = (elm: d.RenderNode) => {
         childNode.hidden = false;
 
         for (j = 0; j < ilen; j++) {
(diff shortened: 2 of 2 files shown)


kevin-coyle-unipro:
Hey I'd like to see if it's possible that we can get this in. Unfortunately without some kinda of fix for this both the fallback slot and the intended slots are visible in Legacy Edge and IE11.

viljamis:
Hmm we’re seeing this issue in all browsers; latest Chrome, Safari and Firefox too. default slot content is always shown, no matter what.

src/runtime/vdom/test/scoped-slot.spec.tsx
```diff
@@ -755,4 +755,26 @@ describe('scoped slot', () => {
     // expect(root.firstElementChild.firstElementChild.children[1].firstElementChild.children[1].nodeName).toBe('GOAT');
     // expect(root.firstElementChild.firstElementChild.children[1].firstElementChild.children[1].textContent).toBe('hey goat!');
   });
+
+  it('should hide the slot\'s fallback content when slot content passed in', async() => {
+    @Component({ tag: 'fallback-test', scoped: true })
```
simonhaenisch, line 760: So this happens when setting `scoped: true`? I think it would also happen when adding `shadow: false`. Could you also add a test for that?

src/runtime/vdom/vdom-render.ts
```diff
@@ -418,24 +418,22 @@ const updateFallbackSlotVisibility = (elm: d.RenderNode) => {
...
-              }
-            } else {
-              // this is a default fallback slot node
-              // any element or text node (with content)
-              // should hide the default fallback slot node
-              if (nodeType === NODE_TYPE.ElementNode || (nodeType === NODE_TYPE.TextNode && childNodes[j].textContent.trim() !== '')) {
-                childNode.hidden = true;
-                break;
-              }
+          // this sibling node is from a different component
```
simonhaenisch, line 421: This comment was regarding the `if (childNodes[j]['s-hn'] !== childNode['s-hn'])` statement, so should also be deleted. Wondering whether this condition was important for some other edge cases though (which might also not have a test yet).

Maybe sth like

```ts
if (childNodes[j]['s-hn'] !== childNode['s-hn'] || slotNameAttr !== '') {
  // ...
}
```

would be better?

viljamis:
@kevin-coyle-unipro Do you think you’ll have time to work further with this or would you want some help here? Happy to help if needed.

kevin-coyle-unipro:
> @kevin-coyle-unipro Do you think you’ll have time to work further with this or would you want some help here? Happy to help if needed.

Hey! I'd love some help if its on offer :)

viljamis:
@kevin-coyle-unipro Would you be willing to share access to your fork, or should I create my own?

viljamis:
@simonhaenisch @kevin-coyle-unipro We ended up creating a new pull for additional fixes as that was easier. Please see https://github.com/ionic-team/stencil/pull/2650

claviska:
Thanks for the PR. We're going to focus on getting it merged via #2650 so please subscribe to that issue for updates.

Issue #2356, opened 2020-04-09
onSlotchange missing from slot element


runarberg:
**Stencil version:**

```
 @stencil/core@1.12.2
```

**I'm submitting a:**

[x] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**

When trying to react to [slot change](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSlotElement/slotchange_event) the type checking fails with:

```
[ ERROR ]  TypeScript: ./src/components/my-component/my-component.tsx:21:15
           Type '{ onSlotchange: (event: Event) => void; }' is not assignable to type 'SlotAttributes'.Property
           'onSlotchange' does not exist on type 'SlotAttributes'.

     L20:  <div style={{backgroundColor: this.color}}>
     L21:    <slot onSlotchange={this.handleSlotChange}></slot>
     L22:  </div>
```

**Expected behavior:**

We should be able to attach a slot change listener to the slot element.

**Steps to reproduce:**

See related code and try to build.

**Related code:**

```tsx
import { Component, Element, State, h } from '@stencil/core';

@Component({
  tag: 'my-component',
  styleUrl: 'my-component.css',
  shadow: true
})
export class MyComponent {
  @Element() el: HTMLElement;

  @State() color: string;

  handleSlotChange(event: Event) {
    console.log(event);
    this.color = this.el.textContent;
  }

  render() {
    return (
      <div style={{'background-color': this.color}}>
        <slot onSlotchange={(event) => this.handleSlotChange(event)}></slot>
      </div>
    );
  }
}
```

**Other information:**

I haven’t tested it but there is probably a workaround by manually attaching the listener during `componentWillRender` but I feel that is unnecessarily unergonomic.

PR #2357, opened 2020-04-09
fix(runtime): add onSlotchange to slot element attribute

runarberg:
Fixes: #2356


manucorporat:
Awesome! Thanks a lot!

Issue #2358, opened 2020-04-09
xlink:href is a deprecated SVG attribute


runarberg:
**Stencil version:**

```
 @stencil/core@1.12.3
```

**I'm submitting a:**

[x] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**

Trying to set the `href` attribute in an SVG element results in this error:

```
[ ERROR ]  TypeScript: ./src/components/my-component/my-component.tsx:17:14
           Type '{ href: string; }' is not assignable to type 'SVGAttributes<SVGElement>'.Property 'href' does not
           exist on type 'SVGAttributes<SVGElement>'.

     L16:  <svg>
     L17:    <use href={`${this.sprite}#${this.icon}`} />
     L18:  </svg>
```

I can use `xlink:href` off course—so this is a pretty minor/pedantic issue—but has been [deprecated since SVG 2](https://www.w3.org/TR/SVG2/linking.html#XLinkRefAttrs), and I would rather use the attribute according to the latest standard.

**Expected behavior:**

The build should pass with an `href` attribute on my SVG elements, removing the need to use the deprecated `xlink:href` attribute.

**Steps to reproduce:**

See code below.

**Related code:**

```tsx
import { Component, Prop, getAssetPath, h } from '@stencil/core';

@Component({
  tag: 'my-component',
  styleUrl: 'my-component.css',
  assetsDirs: ['assets'],
  shadow: true
})
export class MyComponent {
  @Prop() icon: string;

  private sprite: string = getAssetPath('./assets/sprite.svg');

  render() {
    return (
      <svg>
        <use href={`${this.sprite}#${this.icon}`} />
      </svg>
    );
  }
}
```

**Other information:**

PR #2359, opened 2020-04-09
fix(runtime): add href SVG attribute

runarberg:
Fixes: #2358


manucorporat:
Thanks a lot for the PR!

PR #2389, opened 2020-04-24
fix(): remove `ValueChange` XUL event

jpzwarte:
`ValueChange` is not a DOM event, but a XUL event: https://developer.mozilla.org/en-US/docs/Archive/Mozilla/XUL/Events/ValueChange.

So this event can be removed so web components can define their own `valueChange` event emitters.


jpzwarte:
@adamdbradley @manucorporat review please?

manucorporat:
Thanks a lot for the PR!

Issue #2349, opened 2020-04-08
1.12.3 "Composite projects may not disable declaration emit."


bitflower:
**Stencil version:**

```
 @stencil/core@1.12.3
```

**I'm submitting a:**

[x bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**
In dev mode the latest version `1.12.3` throws this error again in a `lerna` context although was fixed some time ago. It's back :-) All `tsconfig`files a re set to `declaration: true`.

**Expected behavior:**
Respects tsconfig with `declaration: true` and doesn't throw error.

**Steps to reproduce:**
Clone and run that repo: https://github.com/bitflower/stencil-lerna

![image](https://user-images.githubusercontent.com/6666636/78803993-ee2f1980-79bf-11ea-8965-0a7b8f048578.png)
![image](https://user-images.githubusercontent.com/6666636/78804025-f5eebe00-79bf-11ea-84f0-69f3b4fb3eb0.png)

bitflower:
Still happening with `1.12.6`

bitflower:
Further analysis shows that there is a difference between prod and dev build on this line:
![image](https://user-images.githubusercontent.com/6666636/80565039-c8f65f80-89ef-11ea-971a-1bfacc53d042.png)

On prod build this returns:
```
[]
```

whereas in dev mode it returns:

```
[ { file: undefined,
    start: undefined,
    length: undefined,
    messageText: 'Composite projects may not disable declaration emit.',
    category: 1,
    code: 6304,
    reportsUnnecessary: undefined } ]
```

bitflower:
The loaded config is exactly the same so it must be changed somehow during transpile:
![stencil-ts-config](https://user-images.githubusercontent.com/6666636/80640146-4c04cd80-8a63-11ea-94bd-374bee645201.gif)

bitflower:
I got it working with the help of @simonhaenisch by changing this line:
![image](https://user-images.githubusercontent.com/6666636/80646991-0699cd80-8a6e-11ea-8499-6856a5a0fd84.png)

Result:
![image](https://user-images.githubusercontent.com/6666636/80647035-15808000-8a6e-11ea-88f0-141227eb679f.png)

bitflower:
Conversation from Slack for reference:

[...]

PR #2402, opened 2020-04-29
fix(dev-mode) always add dist-types

bitflower:
With the help of @simonhaenisch I could find the reason why https://github.com/ionic-team/stencil/issues/2349 is happening. 

It is because here https://github.com/ionic-team/stencil/blob/4c70fdd2c418a1419c8c30bd585d4a081100b28b/src/compiler/config/outputs/validate-dist.ts#L47

the `buildDist` is `false` and the output target `DIST_TYPES` is not added in dev-mode. It is however needed later on.

This causes this condition to return false:
https://github.com/ionic-team/stencil/blob/63595bc540db9aa85295f5a8dc8738beb2a37611/src/compiler/transpile/ts-config.ts#L8

and `declaration` becomes `false` - which is not allowed in combination with `composite: true`.

I first tried to tackle it via this line by removing the `devMode` condition but it seems too related to other (global) decisions you've made maybe:
https://github.com/ionic-team/stencil/blob/253894d3ea154c0471e1345ad79fea3e708121d5/src/compiler/config/validate-config.ts#L52

So my solution is to always add the output target `DIST_TYPES` no matter the mode (dev or prod).

This is a huge blocker for us - would love to find a solution!

### DIFF
--- a/src/compiler/config/outputs/validate-dist.ts
+++ b/src/compiler/config/outputs/validate-dist.ts
@@ -44,6 +44,13 @@ export const validateDist = (config: d.Config, userOutputs: d.OutputTarget[]) =>
       file: join(lazyDir, `${config.fsNamespace}.css`),
     });
 
+    outputs.push({
+      type: DIST_TYPES,
+      dir: distOutputTarget.dir,
+      typesDir: distOutputTarget.typesDir,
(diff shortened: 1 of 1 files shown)


manucorporat:
Thanks a lot!

PR #2404, opened 2020-05-01
Add missing 'at' attribute to LinkHTMLAttributes

gregorym:
`as` Attribute: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link#Attributes


manucorporat:
Thanks a lot for the PR!

Issue #2397, opened 2020-04-27
Building for production doesn't support IE11


FedericoCarboni:
**Stencil version:**

```
 @stencil/core@^1.12.2
```

**I'm submitting a:**

[x] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**

Building a stencil app for production won't support IE11 but dev builds using the `--es5` flag do.

**Expected behavior:**

Production builds should support IE11.

**Steps to reproduce:**

Initialize a new Stencil app, run `stencil build` folder, serve the `www` directory. Check with IE11.
Try with dev mode run `stencil build --dev --es5 --serve`. Then check with IE11.

stanislavprokopov:
I was able to trace the issue to this part if the build code:
```
var start = function() {
    var url = new URL('./p-e55ace99.system.js', resourcesUrl);
    System.import(url.href);
  };
```
the generated index.html has this part
```<script type="module" data-resources-url="/build/"```
so when this start script is executed it tries to execute this code
```new URL('./p-e55ace99.system.js', '/build/');```
And if you run it in the IE11 dev console you will get an error ```Invalid scheme```
To fix this i changed data-resources-url to be a full url.

**Comment by pju-** [comment]
Same issue here, I can not get the production build to run in IE 11.
It will fail silently (no `console` output).

**Comment by manucorporat** [comment]
can you provide a repo case? maybe just index.html?

**Comment by pju-** [comment]
Sure, here you go:
https://github.com/pju-/stencil-issue-2397

If you run
```bash
npm init stencil (and choose the app-package or the component-package)
cd your_directory
npm install
npm run build
```
that should give you the setup :)

I have included the build files, index.html and all.
Check [here](https://github.com/pju-/stencil-issue-2397/blob/318e3cde99f37331dc1ce3d60be195dddb4623f8/www/build/app.js#L128) to see the offending line that is missing the needed absolute base URL.

[This](https://github.com/ionic-team/stencil/pull/2410) should fix it, I'm just not sure if `generate-system.ts` is a generated file.

RaviDesai:
I'm having a slightly different problem with production builds in IE.  Whenever I fire up a production build, and point IE at it, I get errors in the Console: Object doesn't support property or method 'assign' in `~dev-server` in a line that looks like `Te=Object.assign({},De, t.devServerConfig, n);`  

[...]

PR #2410, opened 2020-05-05
fix(polyfill): add proper url base to import shim

hv-pul:
URL needs absolute address in second part of constructor

This should fix https://github.com/ionic-team/stencil/issues/2397

Hope I found the right place this time ;)

### DIFF
--- a/src/compiler/output-targets/dist-lazy/generate-system.ts
+++ b/src/compiler/output-targets/dist-lazy/generate-system.ts
@@ -61,7 +61,7 @@ const getSystemLoader = async (config: d.Config, compilerCtx: d.CompilerCtx, cor
 
     var resourcesUrl = scriptElm ? scriptElm.getAttribute('data-resources-url') || scriptElm.src : '';
     var start = function() {
-      var url = new URL('${corePath}', resourcesUrl);
+      var url = new URL('${corePath}', window.location.href + resourcesUrl);
       System.import(url.href);
     };
 


src/compiler/output-targets/dist-lazy/generate-system.ts
```diff
@@ -61,7 +61,7 @@ const getSystemLoader = async (config: d.Config, compilerCtx: d.CompilerCtx, cor
 
     var resourcesUrl = scriptElm ? scriptElm.getAttribute('data-resources-url') || scriptElm.src : '';
     var start = function() {
-      var url = new URL('${corePath}', resourcesUrl);
+      var url = new URL('${corePath}', window.location.href + resourcesUrl);
```
mikeysteele, line 64: window.location.origin would be better because if for some reason there was a hash fragment in the URL it would break this.
`
var url = new URL('${corePath}', new URL(resourcesUrl, window.location.origin)) would be more reliable IMO
`

hv-pul: Good point! 

A hash fragment would actually be dropped by `new URL`, because it takes the second parameter as the base path. But for handling subpages in an SPA it should definitely point to the origin.

So it should be 
`var url = new URL('${corePath}', new URL(resourcesUrl, window.location.origin))`
for sure. I''ll update

hv-pul:
Can someone from the team have a look at this and potentially merge it? It would be fantastic to have a release that works with IE 11 again :)

flobacher:
@adamdbradley 
please take a look and merge. without this fix, the non module codepath totally breaks, so IE11 is broken. thank you very much!

bigfish:
I can confirm this fixes the issue. I was getting white screen of death on newly initialized component library. I made the fix in node_modules/@stencil/core/compiler/stencil.js  that is in this PR, rebuilt, and the demo component showed up. Please merge!

flobacher:
@bigfish 
I did the same in a www project. IE11 silently failed bootstrapping, with the fix, everything works fine!

PR #2420, opened 2020-05-06
fix(testing): MockResponse404 is not ok

luchsamapparat:
`MockResponse404` extends from `MockResponse` which sets `ok` to `true` by default. However, `ok` should only be `true`, if the status code is within the 2xx range.

Alternatively, we could change the `ok` property of `MockResponse` to a getter:

```
export class MockResponse {
  get ok() {
    return this.status >= 200 && this.status <= 299
  }
}
```

However, by doing that `ok` would be readonly and that is probably a breaking change. 🤔

### DIFF
--- a/src/testing/mock-fetch.ts
+++ b/src/testing/mock-fetch.ts
@@ -134,6 +134,7 @@ export const mockFetch = {
 };
 
 class MockResponse404 extends MockResponse {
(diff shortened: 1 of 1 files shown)


manucorporat:
Thanks for the PR!

Issue #2398, opened 2020-04-28
DetailsHTMLAttributes interface missing onToggle property


dantothefuture:
**Stencil version:**

```
 @stencil/core@1.12.6
```

**I'm submitting a:**

[x] bug report
[ ] feature request
[ ] support request

**Current behavior:**

TSX fails to complile with the following error when a `toggle` event listener is added on the `<details>` element:

> Property 'onToggle' does not exist on type 'DetailsHTMLAttributes<HTMLElement>'. ts(2322)

This issue affects types only. If you do `<details {...{ onToggle: console.log }} />`, it works just fine, but I'm guessing this isn't the way we're supposed to add event listeners in our code :)

**Expected behavior:**

The `toggle` event should be [supported](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details#Events) in `<details>` with the `onToggle` property.

**Steps to reproduce:**

1. Create a basic StencilJS project with a minimal component;
2. Try to render the following: `<details onToggle={console.log}>Content</details>`.

**Related code:**

https://github.com/ionic-team/stencil/blob/5adcdd3d601525232247a9471ae8e459ce2fbc07/src/declarations/stencil-public-runtime.ts#L778-L780

**Other information:**

Simply adding `onToggle?: (event: Event) => void;` to the interface linked above should fix it, unless I'm missing something.

manucorporat:
Can you provide a PR?

dantothefuture:
@manucorporat sure, here you go: https://github.com/ionic-team/stencil/pull/2421

PR #2421, opened 2020-05-06
Add onToggle property to the DetailsHTMLAttributes interface

dantothefuture:
This PR adds types for the `toggle` [event](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details#Events) in `<details>` with the `onToggle` property.

Closes: #2398


manucorporat:
Thanks a lot for the PR!!

Issue #2303, opened 2020-03-24
`stencil generate` output changed format with the new compiler


alexandertrefz:
Ever since the legacy compiler got removed, in [this commit](https://github.com/ionic-team/stencil/commit/63595bc540db9aa85295f5a8dc8738beb2a37611), `stencil generate` outputs a considerably different format of folders & files.
My primary concern lies in the changed naming conventions of the folder and files (and the name of the class) that get generated - perviously they had their prefix and dash stripped, now they don't anymore. That takes most of the utility of `stencil generate` away for any existing codebase, unless it fully adapts the new scheme. This change wasn't mentioned in any changelog, so I wonder wether this was accidental?

**I'm submitting a:**
[x] bug report
[ ] feature request

**Current behavior:**
`stencil generate` creates folders and files and the class with their dash-prefix

**Expected behavior:**
`stencil generate` should strip the dash-prefix like in previous versions.

**Steps to reproduce:**
Run `stencil generate`.

manucorporat:
Duplicate of https://github.com/ionic-team/stencil/issues/2347#issuecomment-624779936

Issue #2396, opened 2020-04-27
Stencil generate: generated Spec Test .spec.tsx file with import component path


milottit:
**Stencil version:**
```
 @stencil/core@@1.12.6
```

**I'm submitting a:**
[x] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**
`stencil generate test-comp` with Spec Test enabled as follow
```
? Which additional files do you want to generate? ›  
Instructions:
    ↑/↓: Highlight option
    ←/→/[space]: Toggle selection
    a: Toggle all
    enter/return: Complete answer
◉   Stylesheet (.css)
◉   Spec Test  (.spec.tsx)
◉   E2E Test (.e2e.ts)
```
will generate in `src/components/test-component/test` directory the `test-comp.spec.tsx` file with the following import statement:
```ts
import { TestComponent } from './mt-image';
```
**Expected behavior:**
It should be:
```ts
import { TestComponent } from '../mt-image'; // NOTE THE .. instead of .
```

**Culprit:**

In the file `src/cli/tasks/task-generate.ts`, the method `getSpecTestBoilerplate` has the following line:
```ts
import { ${toPascalCase(tagName)} } from './${tagName}';
```

It should be:
```ts
import { ${toPascalCase(tagName)} } from '../${tagName}';
```

alexandertrefz:
For the love of god, dear stencil team. *Stop breaking generate*! This is a massive issue for existing code bases with loads of components that all need to be refactored to the new way of generate every time you change it, and you keep changing (and breaking!!!) it without ANY community communication! At least do the changes to generate in one fell swoop so we don't have to continually refactor all existing components!

milottit:
@alexandertrefz is right. It would be nice a feature like Angular Schematics.

tricki:
The import was fixed in #2486

rwaskiewicz:
Hey all, I'm closing this issue, as it was fixed in #2486, as @tricki pointed out.

Issue #2348, opened 2020-04-08
Stencil CLI component generation config


nathanb21:
**Stencil version:**
```
@stencil/core@1.12.1
```

**I'm submitting a:**
[ ] bug report
[x] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**
`npx stencil generate app-component` provides toggles for a `.css`, `.spec.tsx`, `.e2e.ts` options.

As we are using @stencil/sass we manually rename the `.css` file to `.scss`.

I also use `scoped` instead of `shadow` so also change these both in the `.tsx` file.

**Expected behavior:**
At project generation, and within a config, the ability to specify the type of CSS files that are generated and other configurable options that could be used (custom templates?).

JasonGrant:
I would love to not have to change css to scss every time I create a new component

manucorporat:
We want to keep it simple, we encourage developers to use .css when possible, not plans for the moment to add other format options in the time being. 
That would involve a lot new moving parts, new issues, more things to support out of our control.

Issue #2347, opened 2020-04-08
Component generation regression - files/folders include 'app-' prefix


nathanb21:
**Stencil version:**

```
@stencil/core@1.12.1
```

**I'm submitting a:**

[x] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**
`npx stencil generate app-component` creates files/folders with `app-` prefix

**Expected behavior:**
Files/folders created without `app-` prefix

**Steps to reproduce:**
Run `npx stencil generate app-component`

**Related code:**

**Other information:**
In version, maybe, 1.5 it used to work the current way where `app-` was prefixed to files and folders when generating components from the CLI. This was removed in later versions. We were most recently using v1.8 and generating components without the `app-` prefix.

Without `app-` is preferred as it makes it easier to read component files and folders without having to ignore that prefix in file/folder structure.

We've updated to 1.12.1 and now `app-` is being prefixed to files/folders.

JasonGrant:
I was surprised this had changed and don't know if it is a suggested pattern change or a mistake.

manucorporat:
This is expected behaviour, we will update thee style guide. it's not a breaking changee though!

nathanb21:
@manucorporat thanks for the reply!

What's the motivation behind the change?

Ionic Core components aren't referenced like this (https://github.com/ionic-team/ionic/tree/master/core/src/components). I assume the generator wasn't used to make these but if all these files and folders had 'ion-' prefixed it'd back the list harder to read through.

I also don't believe other CLIs, such as Angular CLI, add the prefix when generating components.

manucorporat:
The prefix is a key in web components, while it's not for angular, or other frameworks, we found cases where people had different prefixes under the same collection, and sometimes that makes sense, for different teams, or really big apps.

In addition it was a lot of extra complexity in the generator code to "remove" what looks like a prefix, when sometimes it was not even a prefix. Going forward we will recomment to have the name of the component, ie the tagname which as single source of truth!

nathanb21:
@manucorporat 

Just to confirm; `npx stencil generate app-component` used to generate the following folder structure:

```
component
    component.tsx
    component.css
    component.e2e.ts
    component.spec.ts
```
The tag inside `component.tsx` was still `app-component`

The new structure is:
```
app-component
    app-component.tsx
    app-component.css
[...]

PR #2427, opened 2020-05-08
feat(cli): Add ability to specify component generation config parameters

nathanb21:
A user can define a `prefix `and `styleFormat` parameter in `stencil.config.ts` to provide more flexibility when working with the `stencil generate` CLI command.

eg.
```
import { Config } from '@stencil/core';

...
  componentGeneratorConfig: {
    prefix: 'ion',
    styleFormat: 'scss'
  }
}
```

The toggle prompts remain the same, referencing `scss` instead of `css` in the above example.

This object is optional, with a default `app` prefix and `css` format being applied.

Generated files and folders are created without the prefix within the name. Eg. `ion-button` component would create file/folder structure `button`, `button.tsx` etc.

The generation command would change to:

`npx stencil generate button` instead of `npx stencil generate ion-button`

Resolves:
https://github.com/ionic-team/stencil/issues/2303
https://github.com/ionic-team/stencil/issues/2347
https://github.com/ionic-team/stencil/issues/2348
https://github.com/ionic-team/stencil/issues/2396

### DIFF
--- a/src/cli/tasks/task-generate.ts
+++ b/src/cli/tasks/task-generate.ts
@@ -29,14 +29,17 @@ export async function taskGenerate(config: d.Config) {
     config.flags.unknownArgs.find(arg => !arg.startsWith('-')) || ((await prompt({ name: 'tagName', type: 'text', message: 'Component tag name (dash-case):' })).tagName as string);
 
   const { dir, base: componentName } = parse(input);
+  const prefix = config.componentGeneratorConfig?.prefix;
+  const componentTag = prefix ? `${prefix}-${componentName}` : `app-${componentName}`;
+  const styleExt = config.componentGeneratorConfig?.styleFormat || 'css';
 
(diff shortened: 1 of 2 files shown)


nathanb21:
I could remove the default application of `app-` if we didn't want to force a default on users, then the functionality would remain much the same as current. Let me know your thoughts :)

prodkt:
Are there any plans to move forward on this PR?

noopliez:
This would be a great help for our projects, too. Writing the tag names/prefixes manually comes always with the possibility of human error...

christian-bromann:
I am going ahead and close this one as I don't see how this fits in current context of the Stencil project. All referenced issues were closed already. If someone could provide me with more context I am happy to re-open and reconsider merging. Thank you @nathanb21 for your contribution and apologies that this patch was never picked up.

Issue #2418, opened 2020-05-06
[Bug] Stencil breaks on AMD builds in IE 11


warapitiya:
**Stencil version:**

```
 @stencil/core@1.12.5
```

**I'm submitting a:**

[x] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**
Stencil does not load Promise polyfill on IE correctly for AMD builds. This is because @stencil/core use `es6-promise` as promise polyfill and it is not patching promise auto for AMD.

**Expected behavior:**
Stencil should render polyfills for AMD build in IE 11.

**Steps to reproduce:**
[stencil-require-js](https://github.com/warapitiya/stencil-require-js)

1. Clone the repository https://github.com/warapitiya/stencil-require-js.
2. Install dependencies: yarn install
3. Build project: yarn build
4. Serve the files: npx serve
5. Navigate to URL on IE 11: http://localhost:5000/example
6. Open console to see the error.

**Related code:**

[stencil polyfill](https://github.com/ionic-team/stencil/blob/master/src/client/polyfills/promise.js)
[es6-promise-PR](https://github.com/stefanpenner/es6-promise/pull/263)

**Other information:**

manucorporat:
Would you be open to create a PR? does this mean we need to update the es6-promise polyfill?

warapitiya:
Hmm, the Issue here is that the polyfills like `es6-promise` and `fetch` are UMD supported modules. Which means we need to change the polyfills entirely.

I was hoping that we can use `core-js/es/promise` and `unfetch` instead.

Any comments?

warapitiya:
OK. Manage to get fix this by changing below:

- Remove `es6-promise` polyfill and replace it with `core-js/es/promise`.
- Use `fetch` [polyfill's IIFE version](https://cdnjs.cloudflare.com/ajax/libs/fetch/3.0.0/fetch.js).

viljamis:
Our team ran into this same issue as well just today. @manucorporat Did you get a chance to look at the pull yet?

PR #2443, opened 2020-05-15
fix(polyfill): use core-js promise and iife fetch polyfill

warapitiya:
Current polyfills are UMD supported. When appending `Stencil.js` to require.js web apps. IE breaks with an error `Promise is undefined`. This is due to [es6-promise's](https://github.com/ionic-team/stencil/blob/571ffc3f8714b9423370ee775c1f67ede767d8f4/src/client/polyfills/promise.js#L9) UMD supported check.

```
typeof define === 'function' && define.amd ? define(factory) :

```

When I do a quick scan I found out that currently fetch polyfill also does a UMD check.

This PR will replace those 2 polyfills with below:

- Remove es6-promise polyfill and replace it with core-js/es/promise.
- Use fetch polyfill's [IIFE version](https://cdnjs.cloudflare.com/ajax/libs/fetch/3.0.0/fetch.js).

Issue related: #2418


adamdbradley:
Thanks for the PR!

PR #2453, opened 2020-05-18
Verify The Prop Not Based on the Name but based on the type of functi…

nqhy:
…on itself


manucorporat:
This is not correct, the name is important, you could pass a function to a prop, without it being an event.

nqhy:
You correct i am too rush to make this change

Issue #2223, opened 2020-02-26
Moving from Puppeteer to Playwright


dmartinjs:
**I'm submitting a:**

[ ] bug report
[ x ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**
Test E2E with Puppeteer (only chromium engine)

**Expected behavior:**
Test with all Browser engines.

Did you plan to move from Puppeteer to Playwright ?

https://github.com/microsoft/playwright

janrembold:
I honestly think the whole testing integration (currently done with an outdated version of puppeteer) should be removed from `@stencil/core` package. Don't get me wrong, the helpers build around web components are a nice thing but having it hardcoded inside Stencil just feels wrong to me.

Why is there a need to integrate an opinionated testing library anyways? Wouldn't it be much smoother to just install and configure it as it is actually intended? In the best case one could create a help library that offers special utilities for Stencil.

klh:
Yeah i second removing the integration, but if tou keep it use playwright instead

danyball:
Hi @rwaskiewicz - the issue got many upvotes. Whats your strategy to go on with this idea? Has "testing" a priority at your Backlog?

rwaskiewicz:
Hey @danyball - this is something we're actively considering. While I can't promise a timeline, we're looking into ways to alleviate the painpoints of:

- Stencil keeping up with the current version of Puppeteer
- Stencil requiring the usage of Puppeteer
- Various libraries that need to be installed/how we communicate that to folks for new projects

gustavo-mungarro:
Hi @rwaskiewicz 👋 - Is there an update regarding this topic?

rwaskiewicz:
@gustavo-mungarro Not as this time, no. The team is prioritizing other work such as element internals, jest 29 support, and slot fixes at this time.

gethari:
Hi @rwaskiewicz  👋 - Is there an update regarding this topic?

rwaskiewicz:
> Is there an update regarding this topic?

The team is currently working on something to improve Puppeteer support that we hope to announce in the next month or so

rwaskiewicz:
👋 Hey folks!

Yesterday, we released a library to help with using Playwright + Stencil :tada:  

[...]

PR #2465, opened 2020-05-23
feat: integrating firefox e2e testing

romulocintra:
At this moment the support for puppeteer firefox integration still experimental but looks promising so far. I made that test can run on firefox. It opens the browser is very buggy!!

### DIFF
--- a/src/declarations/stencil-public-compiler.ts
+++ b/src/declarations/stencil-public-compiler.ts
@@ -1507,6 +1507,12 @@ export interface TestingConfig extends JestConfig {
    * Amount of time in milliseconds to wait before a screenshot is taken.
    */
   waitBeforeScreenshot?: number;
+
+
+  /**
+   * Browser to be launched when running e2e tests.
+   */
+  product?: 'chrome' | 'firefox';
 }
 
 export interface EmulateConfig {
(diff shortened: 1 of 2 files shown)


manucorporat:
Indeed! it's very interesting

cmddavid:
This would be amazing, I'm using the default testing setup for Stencil.js (Jest / Puppeteer) but am facing some feedback that we need to do cross browser testing. Since Puppeteer via Stencil.js only supports Chrome at the moment that is a problem that could result in me having to switch to another testing solution. Hope you guys see the importance of this. It's also a bit strange the Puppeteer api is limited by the Stencil.js implementation while documentation for it is difficult to find. But that's another discussion I guess.

romulocintra:
> This would be amazing, I'm using the default testing setup for Stencil.js (Jest / Puppeteer) but am facing some feedback that we need to do cross browser testing. Since Puppeteer via Stencil.js only supports Chrome at the moment that is a problem that could result in me having to switch to another testing solution. Hope you guys see the importance of this. It's also a bit strange the Puppeteer api is limited by the Stencil.js implementation while documentation for it is difficult to find. But that's another discussion I guess.

I have similar needs for cross-browser testing. This was an experiment that actually worked but Puppeteer firefox still with experimental.

 IMHO I think one of the biggest Stencil benefits is cause works as an all-in-one tool. But I believe that future projects like [playwright](https://github.com/microsoft/playwright)  that supports 3 Major engines/browsers, is also something to have in mind, probably Puppeteer will follow the same path.

For e2e test isn't hard to have your own stack, in the end, all is running on the browser

dmartinjs:
same need and I opened #2223 for [Playwright](https://playwright.dev/) intergation.

PR #2476, opened 2020-05-27
fix(spellcheck): put an any on it

adamdbradley:


ingomc:
Does this fix the spellcheck true issue, if your setting spellcheck false?
TSX
```
<input spellcheck="false"  />
```
HTML
```
<input spellcheck="true"  />
```

alexjc:
`spellcheck="false"` seems to be corrected to `true` in the current version 2.0.0.

PR #2485, opened 2020-06-03
fix(types): draggable is enumerated not boolean

### LINKED ISSUE #2484: `draggable={false}` doesn't render correctly
**Stencil version:**

```
 @stencil/core@1.14
```

**I'm submitting a:**

[x] bug report
[ ] feature request

**Current behavior:**

Image tags have `draggable` enabled by default and there's no way to turn it off currently, because using `draggable={false}` in JSX doesn't render the attribute correctly but simply omits it:

```html
<img src="http://place-hold.it/200" alt="Placeholder" draggable={false} />
```

=> `Placeholder`

**Expected behavior:**

=> `Placeholder`

**Other information:**

From https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/draggable:

`draggable` can have the following values:

* `true`: the element can be dragged.
* `false`: the element cannot be dragged.

This attribute is **enumerated and not Boolean**. A value of `true` or `false` is mandatory, and shorthand like `` is forbidden. The correct usage is ``.

**Suggested Solution:**

Fix the types so that it's `draggable?: 'true' | 'false'` instead, because as the docs state it's an enumerated attribute and not a Boolean attribute.

---

BTW I tried suggesting `draggable={"false" as any}` as a workaround, but that ends up being interpreted as truthy I guess, because it renders as `draggable="true"` 😅

---

Duplicate of #1441 which has been closed because the description wasn't very clear.

simonhaenisch:
Closes #2484.

BTW I wasn't entirely sure where to put that test... happy to move it somewhere else!

### DIFF
--- a/src/declarations/stencil-public-runtime.ts
+++ b/src/declarations/stencil-public-runtime.ts
@@ -1202,7 +1202,7 @@ export namespace JSXBase {
     contextMenu?: string;
     contextmenu?: string;
     dir?: string;
-    draggable?: boolean;
+    draggable?: 'true' | 'false';
     hidden?: boolean;
     id?: string;
     lang?: string;
--- a/src/runtime/test/jsx.spec.tsx
+++ b/src/runtime/test/jsx.spec.tsx
@@ -138,4 +138,16 @@ describe('jsx', () => {
       expect(root.textContent).toBe('on-ALLCAPS');
     });
   });
+
+  describe('attributes', () => {
+    describe('draggable', () => {
+      for (const value of [undefined, 'true', 'false'] as const) {
+        it(String(value), async () => {
+          const { root } = await newSpecPage({ components: [], template: () => <img draggable={value} /> });
+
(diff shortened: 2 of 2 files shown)


simonhaenisch:
**Before:**

```
    attributes
      draggable
        ✓ undefined (5 ms)
        ✕ true (8 ms)
        ✕ false (4 ms)

● jsx › attributes › draggable › true

    expect(received).toBe(expected) // Object.is equality

    Expected: "<img draggable=\"true\">"
    Received: "<img draggable=\"\">"

  ● jsx › attributes › draggable › false

    expect(received).toBe(expected) // Object.is equality

    Expected: "<img draggable=\"false\">"
    Received: "<img>"
```

**After:**

```
   attributes
      draggable
        ✓ undefined (6 ms)
        ✓ true (3 ms)
        ✓ false (3 ms)
```

simonhaenisch:
@manucorporat am I correct to assume that you have already fixed this in 0d72aeea51cb754615fed74020b035b1853ad740?

PR #2486, opened 2020-06-03
fix(cli): correct import in generated spec tests

tricki:
Since generated tests are now placed in a `tests` folder (baff1dc), the import path had to be adjusted.


adamdbradley:
thanks for fixing that

PR #2493, opened 2020-06-06
Fix CSS compilation for multiple styleURLs (Close #2432)

### LINKED ISSUE #2432: Imported CSS is not compiled/transformed
**Stencil version:**

```
 @stencil/core@v1.13.0
 @stencil/core@v1.13.1-0
```

**I'm submitting a:**

[x] bug report
[ ] feature request

**Current behavior:**

When importing CSS files, Stencil compiler does not always process each file, creating bundles with with raw/unprocessed CSS.
Given a component using `importUrls` for two separate CSS with identical contents, the `dist` generated stylesheets contain different content.
Sometimes with the contents of the file being imported, and sometimes just the raw import text (e.g. `@import "~tachyons";`)

I also saw hints of race conditions, as sometimes file `A` would be processed, and sometimes file `B`.

_Screenshot shows results of a single build, and results of two input files (`app-root.css` and `app-root.alt.css`)_

**Expected behavior:**

All files being imported by Stencil components should be processed using any plugins, and content inlined in the generated files.
In case there are issues importing or processing files, the compiler should throw an error and fail the build.

**Steps to reproduce:**

A repository for reproducing the issue can be found here: https://github.com/alex-ketch/stenciljs-postcss-import-issue

**Other information:**

Possibly related issues:

- https://github.com/ionic-team/stencil/issues/2292
- https://github.com/ionic-team/stencil/issues/2300
- https://github.com/ionic-team/stencil-postcss/issues/25

alex-ketch:
This PR fixes compilation of multiple CSS files as documented in #2432
They underlying issue was with the stateful nature of the RegEx `exec` command when using the `g` flag. Moving it into the closure of the function makes sure it's reset for each file, fixing the problem.

I have a [test comparing the contents](https://github.com/alex-ketch/stenciljs-postcss-import-issue/blob/master/src/components/app-root/app-root.spec.ts) of the compiled files in the reproduction fork here, but I haven't found a good way to test for it in this repo.

Exporting and testing the `updateCssImports` function works until it tries to read the file contents at [`compilerCtx.fs.readFile`](https://github.com/ionic-team/stencil/blob/master/src/compiler/style/css-imports.ts#L104), where the `mockCompilerCtx` seems to return `undefined` for the file contents.

Could you advise on how you'd like to see this PR proceed please @adamdbradley or @manucorporat?
Thanks!

### DIFF
--- a/src/compiler/style/css-imports.ts
+++ b/src/compiler/style/css-imports.ts
@@ -126,6 +126,7 @@ export const getCssImports = async (config: d.Config, compilerCtx: d.CompilerCtx
   const importeeExt = filePath.split('.').pop().toLowerCase();
 
   let r: RegExpExecArray;
+  const IMPORT_RE = /(@import)\s+(url\()?\s?(.*?)\s?\)?([^;]*);?/gi;
   while ((r = IMPORT_RE.exec(styleText))) {
     const cssImportData: d.CssImportData = {
       srcImport: r[0],
@@ -167,8 +168,6 @@ export const getCssImports = async (config: d.Config, compilerCtx: d.CompilerCtx
 
   return imports;
 };
-
-const IMPORT_RE = /(@import)\s+(url\()?\s?(.*?)\s?\)?([^;]*);?/gi;
 
 export const isCssNodeModule = (url: string) => url.startsWith('~');
 


src/compiler/style/css-imports.ts
```diff
@@ -123,9 +123,10 @@ export const getCssImports = async (config: d.Config, compilerCtx: d.CompilerCtx
   styleText = stripCssComments(styleText);
 
   const dir = dirname(filePath);
-  const importeeExt = filePath.split('.').pop().toLowerCase();
+  const importedExt = filePath.split('.').pop().toLowerCase();
 
   let r: RegExpExecArray;
+  const IMPORT_RE = /(@import)\s+(url\()?\s?(.*?)\s?\)?([^;]*);?/gi;
```
simonhaenisch, line 129: Moving this regex here doesn't change the code? If there was a problem with accessing `IMPORT_RE` , it would be sth like an "`IMPORT_RE` is not defined" runtime error, right? Can you explain how this solves your issue?

src/compiler/style/css-imports.ts
```diff
@@ -123,9 +123,10 @@ export const getCssImports = async (config: d.Config, compilerCtx: d.CompilerCtx
   styleText = stripCssComments(styleText);
 
   const dir = dirname(filePath);
-  const importeeExt = filePath.split('.').pop().toLowerCase();
+  const importedExt = filePath.split('.').pop().toLowerCase();
```
simonhaenisch, line 126: "Importee" is the file being imported, vs. "importer", the file that imports the importee. So the current variable name is correct.

alex-ketch: Thanks for clarifying, I've removed the commit changing the variable name

src/compiler/style/css-imports.ts
alex-ketch: It's not variable scoping issue, but rather that the [`RegExp.prototype.exec()` is stateful](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec) when using the `/g` flag.
This causes subsequent iteration loops to have stale state and doesn't match import statements in the second file.
Moving the regex inside the function closure means that the state is fresh for each file, and successfully resolves all the imports.

simonhaenisch: Ah yup makes sense, thanks for explaining! TIL ;)

I'm just a contributor, not a maintainer, but I'll approve this so they'll at least know someone else had a look already 🤓

adamdbradley:
awesome, thanks for the fix!

Issue #2366, opened 2020-04-13
prerender: rehydrated components (pages) added twice to the DOM


peterpeterparker:
**Stencil version:**
"stencil/core": "^1.12.3"

**I'm submitting a:**
[X] bug report

**Current behavior:**
When I upgrade my apps to Stencil v1.12.3 and run a **prerender** build (`npm run build --prerender`), the components (pages) are added twice to the DOM!

I can confirm that if I run `npm run start` these are correctly added only once.

Not sure if the problem is linked to Stencil or the Ionic router, which I'm using.

**Expected behavior:**
Component added only once to the DOM.

**Steps to reproduce:**

I provide a sample repo.

```
git clone https://github.com/peterpeterparker/twice-component
npm run build
```

and then serve`www` with a local web server.
**Screenshot:**

<img width="1340" alt="Capture d’écran 2020-04-13 à 09 54 57" src="https://user-images.githubusercontent.com/16886711/79103822-969cff00-7d6d-11ea-9952-644f56dddd33.png">

Paul-Hebert:
@peterpeterparker , were you able to resolve this in your project? Any info on a workaround? Thanks!

peterpeterparker:
@Paul-Hebert I wasn't. Still unsure if something not correctly configured or "just" an issue in case of re-hydratation when the Ionic router is used.

I actually reported it today and they (Ionic team) might have a look at it.

Paul-Hebert:
Thanks @peterpeterparker ! I'm also in the awkward spot of not knowing whether I'm doing something wrong or if there's a bug. Would you mind updating me if you learn more? I'll post back here if I can figure it out. Thanks!

peterpeterparker:
@Paul-Hebert sure

Paul-Hebert:
By the way, here's a simplified example  that reproduces this  issue for me:

## My simplified components

```tsx
import { Component, h } from "@stencil/core";

@Component({
  tag: "c4-button",
  shadow: true,
})
export class C4Button {
  render() {
    return (
      <button>
        <slot />
      </button>
    );
  }
}
```

```tsx
import { Component, h} from "@stencil/core";

@Component({
  tag: "c4-page",
})
export class C4Page {
  render() {
    return (
      <div>
        <slot />
      </div>
    );
  }
}
```

## My simplified document I'm hydrating

```html
<!DOCTYPE html>
<html>

<head>
  <script src="/stencil/stencil-components.js"></script>
</head>

<body>
[...]

PR #2508, opened 2020-06-12
fix(#2366): rehydrated components added twice

peterpeterparker:
### Introduction

This is my suggestion to solve #2366. I am honestly not sure it is the proper solution and would appreciate any hints.

I am also not against that PR being rejected as long as the problem, I communicated in April 2020,  finds  a solution.

Indeed I am not able to upgrade our applications and therefore are kind of blocked  with Stencil v1.8. I would be really happy to upgrade. 

Moreover, I noticed that  prerendering solved would enhance by more  than 10% our load time, therefore, more really more  than in need  and looking forward  to a solution.

### Current behavior:

With prerendering, components are not rehydrated and are added twice to the DOM

### Expected behavior:

Components  are rehydrated with a single instance to the DOM

### Root cause of issue

I mostly tested the issue with the `ion-router`, more precisely noticed the  issue with the `ion-nav` and a "page" (`app-home`) being added twice two the DOM.

I did a comparison with the Stencil router, with which I did not face the issue, and noticed that the root case was the missing component reference (`c-id`).

Concretely, currently these elements would be prerendered as following:

```
<ion-nav s-id=14>
     <!--s.14.0.0.0.-->
     <app-home s-id=15>
```

But if they would be added as following, then the issue would be solved:

```
<ion-nav s-id=14>
     <!--s.14.0.0.0.-->
     <app-home s-id=15 c-id=14.0.0.1>
```

Adding `c-id=14.0.0.1` was the solution here.

A spent some time debugging `vdom-annotation.ts` and figured out that this "missing annotation" was a result of the VDOM representation.

Indeed, if the `ion-nav` is referenced as `parent` of the `app-home`, the contrary was not correct, the `app-home` was not referenced as `child` of the `ion-nav` but its `slot`.

ion-nav element:

```
ion-app
    ion-nav
        slot
```

app-home element:

```
ion-nav
    app-home
         ion-content
```

### Suggestion of solution

As both elements are kind of not fully linked, my idea is to add a post processing to the insertion of the annotation which checks if the node as received a `c-id` and if not, as a a parent which contains a `comment` for the slot. If so, it calculate the `c-id` and adds the attribute

### Conclusion

As I said in my introduction, my solution might not be correct, it of course works out according my tests but you might find some reason why not. In such case, I am not against rejecting this PR, a cool, but would really, really appreciate if we can move forward with prerendering while using the Ionic Router.

### DIFF
--- a/src/runtime/test/hydrate-shadow-parent.spec.tsx
+++ b/src/runtime/test/hydrate-shadow-parent.spec.tsx
@@ -452,4 +452,78 @@ describe('hydrate, shadow parent', () => {
       </ion-tab-button>
     `);
   });
+
+  it('nested cmp-b, parent slot', async () => {
+    @Component({ tag: 'cmp-a', shadow: true })
+    class CmpA {
+      render() {
+        return <slot></slot>;
+      }
+    }
+    @Component({ tag: 'cmp-b' })
+    class CmpB {
+      render() {
+        return (
+          <Host>
+            <slot></slot>
+          </Host>
+        );
+      }
+    }
+    // @ts-ignore
+    const serverHydrated = await newSpecPage({
+      components: [CmpA, CmpB],
+      html: `<cmp-a><cmp-b>cmp-a-light-dom</cmp-b></cmp-a>`,
+      hydrateServerSide: true,
+    });
+
+    expect(serverHydrated.root).toEqualHtml(`
+      <cmp-a class="hydrated" s-id="1">
+        <!--r.1-->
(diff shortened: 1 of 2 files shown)


manucorporat:
Thanks a lot for the PR! it was merge in this commit: https://github.com/ionic-team/stencil/commit/e1527738bcbac87d9cfbdcf492a681a72f56d699

peterpeterparker:
Awesome! Thx :)

Issue #2425, opened 2020-05-07
Unable to point `testEnvironment` to a node module in `stencil.config.ts`


toureh:
**Stencil version:**

```
 @stencil/core@1.12.2
```

**I'm submitting a:**

[x] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**
Tests do not run when pointing `testEnvironment` to a node module in `stencil.config.ts`. Jest looks for the module in `/path/to/my/project/stencil.config.ts/<module name>`

**Expected behavior:**

Tests should run when pointing `testEnvironment` to a node module in `stencil.config.ts` as described in [Configuring Jest Guide](https://jestjs.io/docs/en/configuration.html#testenvironment-string).

**Steps to reproduce:**

1. Set `testEnvironment` to "jsdom" in `stencil.config.ts`.
2. Run tests with `npm test`
3. Observe error:
```bash
[ ERROR ]  runJest: ● Validation Error: Test environment
           C:/path/to/my/project/stencil.config.ts/jsdom
           cannot be found. Make sure the testEnvironment configuration option
           points to an existing node module. Configuration Documentation:
           https://jestjs.io/docs/configuration.html

```

**Related code:**
I set `testEnvironment` to "jsdom" in `stencil.config.ts` like this:

```javascript
  testing: {
    testEnvironment: "jsdom"
  }
```

toureh:
I've opend a PR for this: https://github.com/ionic-team/stencil/pull/2534

PR #2534, opened 2020-06-25
fix(testing): fix testEnvironment validation

toureh:
Hoping to resolve https://github.com/ionic-team/stencil/issues/2425.

After this change I am able set `testEnvironment` as a node module in `stencil.config.ts`  like "jsdom".


adamdbradley:
Thanks for opening the issue and PR. We addressed the issue here: https://github.com/ionic-team/stencil/commit/645f3a0ba98ec322b1f2f21feb68caa813ab30e5

thanks

PR #2552, opened 2020-07-05
Support multiple globalStyle

### LINKED ISSUE #2550: Support multiple globalStyle
**Stencil version:**
```
 @stencil/core@1.15.0
```

**I'm submitting a:**
[ ] bug report
[x] feature request

**Current behavior:**
We can pass a single `globalStyle` to Stencil.

**Expected behavior:**
I have a collection of web components with shadow dom disabled. I use SASS.
I want to output one css file per theme.

reda-alaoui:
Fix #2550

### DIFF
--- a/src/compiler/build/compiler-ctx.ts
+++ b/src/compiler/build/compiler-ctx.ts
@@ -39,7 +39,7 @@ export class CompilerContext implements d.CompilerCtx {
   rollupCacheNative: any = null;
   rootTsFiles: string[] = [];
   tsService: d.TsService = null;
-  cachedGlobalStyle: string;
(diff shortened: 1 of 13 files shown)


adamdbradley:
Sorry, but originally this is how we have the global styles and eventually decided it was better for the compiler to just have one global css file entry for the index file to include. This way prerendering knows which file to optimize and the filename it's going to use. Instead we recommending having one global css file, and from within that file you can add multiple css `import`s. The css imports will be grouped together, rather than separate http requests. Thank you for the PR, but decided to close this for now.

reda-alaoui:
Having a single css entry prevents the existence of one global style per theme.

Anyway, I ended up using styleMode in each web-component, so I don't need this feature anymore.

PR #2549, opened 2020-07-03
fix(compiler): fix transitive module dependencies in unit tests (#2178)

chrisdarroch:
This should resolve a few `Cannot use import statement outside a module` errors when spec files are run. 

There's a slightly deeper problem here re: who decides what is a module.  Both typescript and jest can be configured to accept different extensions as module types, but neither configuration is honoured by the stencil compiler when spec files are run.  Perhaps a better fix would be to inspect the `opts` in this helper function to determine which extensions to add here.

Adding `js` to the list of files to compile may introduce potential slowness or problems, as more files will be run through the typescript compiler now.

Issue #2178, opened 2020-02-03
SyntaxError: Cannot use import statement outside a module


wschaef:
**Stencil version:**
```
 @stencil/core@1.8.6
```

**I'm submitting a:**

[x] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**
I'm using a lib named Elix (https://component.kitchen/elix) and it works for npm start. 
But I get an ERROR on npm-test:
```
 SyntaxError: Cannot use import statement outside a module

      2 | import { Model, Data, Element, Relation } from '../../global/model/index';
      3 | import { MatchResults, RouterHistory } from '@stencil/router';
    > 4 | import 'elix/define/Tabs'
        | ^
      5 | 
      6 | 
      7 | @Component({

sformer.js:537:17)
      at ScriptTransformer.transform (xxxxxx/node_modules/@jest/transform/build/ScriptTransformer.js:579:25)
      at Object.<anonymous> (src/components/app-home/app-home.tsx:4:1)
```

**Expected behavior:**
npm test doesn't fail

**Steps to reproduce:**
Create Stencil a default Project with "npm init stencil" and select app
**Related code:**

`npm install elix`

add to any of stencil component. E.g. app-home
`import 'elix/define/Tabs'`

run `npm test`

```tsx
// insert any relevant code here
```

**Other information:**
I've searched throught stenciljs and jest documentation and tried to solve it by adding 
```
  testing: {
    transformIgnorePatterns: ["/node_modules/", "/node_modules/elix/.*"]
  }
```
to stencil.config.ts

chrisdarroch:
I've been running in to this one, too.

Jest doesn't work with ES modules (see https://github.com/facebook/jest/issues/9292, https://github.com/facebook/jest/issues/9622, and https://github.com/facebook/jest/pull/8751), so at the moment, it will bail on any code it encounters with an `import` statement inside it.

From what I can gather, Stencil is compiling the Typescript component file, then passing that code to Jest. What seems to be going wrong is that dependencies of the component file are *not* transpiled, so any `import` statements in those will cause the test to bail.

In theory, if the spec tests were to use the full CJS output for a given component, it would work. In practice, I'm not sure what needs to change. It doesn't seem like something fixable just by changing the jest parameters... but maybe I'm missing something.

Shirinrpatel:
[...]

PR #2558, opened 2020-07-07
Fix transitive js modules in tests (#2178)

chrisdarroch:
This is a follow-up from my previous PR #2549, which didn't quite fix the problem 🤦 

This time, I've spent a bit more time understanding the internal repo structure, and have written a test that exercises the previously-faulty behaviour. It should fail if you alter the `shouldTranspileModule` function to exclude `js` files.

I leave it to you to decide whether to run the test in CI :)


adamdbradley:
Awesome fix and love all the tests, thanks!

PR #2597, opened 2020-07-28
Fix syntax error in generated types

### LINKED ISSUE #2596: Syntax error in generated types file when using `dist-custom-elements-bundle` and `dist` together
**Stencil version:**

```
 @stencil/core@1.17.2<version>
```

**I'm submitting a:**

[x] bug report

**Current behavior:**

When using `dist` and `dist-custom-elements-bundle` together, the file `index.d.ts` is generated in the `custom-elements` directory which is great! The generated file has an unterminated comment at the very end, which isn't so great :(

It looks like the comment above/for `setAssetPath()` is the culprit

**Expected behavior:**

Types file is free of syntax errors

**Steps to reproduce:**

Use the output targets mentioned and you'll see

**Other information:**

snaptopixel:
Fixes #2596


snaptopixel:
@adamdbradley @manucorporat quick one, hope this can make into the next release :)

adamdbradley:
gahh, good catch, thanks

PR #2613, opened 2020-08-05
feat(testing): add collectCoverageFrom jest parameter

### LINKED ISSUE #2557: Testing: Not possible to set collectCoverageFrom (and possibly other options)
**Stencil version:**

```
 @stencil/core@1.16.0
```

**I'm submitting a:**

[x] bug report
[ ] feature request

**Current behavior:**

Setting `testing.collectCoverageFrom` is not passed to Jest.

**Expected behavior:**

It should pass it through

**Steps to reproduce:**

Add `collectCoverageFrom` with a value (i.e. with a single file, or a set with negation) and set `collectCoverage` to true. The coverage generated will not match the expectation passed in.

**Related code:**
https://github.com/ionic-team/stencil/blob/master/src/testing/jest/jest-config.ts#L39
https://github.com/ionic-team/stencil/blob/master/src/testing/jest/jest-config.ts#L45-L49

**Other information:**

https://github.com/ionic-team/stencil/blob/master/src/testing/jest/jest-config.ts#L39

There is no `default` defined for this option (see [collectCoverageFrom docs](https://jestjs.io/docs/en/configuration#collectcoveragefrom-array) and, being `undefined` means that it is not looped over in the function here:
https://github.com/ionic-team/stencil/blob/master/src/testing/jest/jest-config.ts#L45-L49

It's also worth noting that `collectCoverageFrom` is potentially not the only option affected by this as anything set to `undefined` in the defaults will not be set (unless set in the subsequent block of `if` statements)

I'll raise a PR for this if I get a chance

**Workaround**
Assume that all files loaded by Jest will be covered and use `coveragePathIgnorePatterns` to negate anything you don't want.

If a file is not covered in your output, it means you have not loaded it in any of your tests.

Note: Puppeteer doesn't contribute to coverage so don't expect `e2e` tests to add to coverage 😉

irega:
Hi!

This PR allows us to collect the code coverage from other not tested code files, adding the *collectCoverageFrom* jest parameter.

Resolves #2557

### DIFF
--- a/src/testing/jest/jest-config.ts
+++ b/src/testing/jest/jest-config.ts
@@ -55,6 +55,9 @@ export function buildJestConfig(config: d.Config) {
   if (isString(stencilConfigTesting.collectCoverage)) {
     jestConfig.collectCoverage = stencilConfigTesting.collectCoverage;
   }
+  if (Array.isArray(stencilConfigTesting.collectCoverageFrom)) {
+    jestConfig.collectCoverageFrom = stencilConfigTesting.collectCoverageFrom;
+  }
   if (isString(stencilConfigTesting.coverageDirectory)) {
     jestConfig.coverageDirectory = stencilConfigTesting.coverageDirectory;
   }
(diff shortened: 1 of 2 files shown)


adamdbradley:
thanks!

Issue #2491, opened 2020-06-06
Stencil 1.14.0 doesn't render anything with taskQueue: 'immediate'


michael42:
**Stencil version:**

```
 @stencil/core@1.14.0
```

**I'm submitting a:**

[X] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**
I wanted to try out `taskQueue: 'immediate'`, but stencil 1.14.0 doesn't render anything. I slimmed down my root component to:
```javascript
export class RootPage {
    constructor() {
        console.log("RootPage");
    }
    connectedCallback() {
        console.log("RootPage connectedCallback");
    }
    componentWillLoad() {
        console.log("RootPage componentWillLoad");
    }
    render() {
        console.log("RootPage render");
        return <p>Test!</p>;
    }
}
```
But it only ever reaches the `connectedCallback`:
```
RootPage
RootPage connectedCallback
```

**Expected behavior:**
Stencil should support `taskQueue: 'immediate'` and render something.

**Steps to reproduce:**
1. `npm init stencil` => ionic pwa
2. Update to @stencil/core 1.14.0
3. Set `taskQueue: 'immediate'` in `stencil.config.js`
4. Start the dev server and open the page in the browser

**Related code:**
I debugged the initialization and found the current `scheduleUpdate` implementation very suspicious:
```tsx
const scheduleUpdate = (hostRef, isInitialLoad) => {
    if (BUILD.taskQueue && BUILD.updatable) {
        hostRef.$flags$ |= 16 /* isQueuedForUpdate */;
    }
    if (BUILD.asyncLoading && hostRef.$flags$ & 4 /* isWaitingForChildren */) {
        hostRef.$flags$ |= 512 /* needsRerender */;
        return;
    }
    attachToAncestor(hostRef, hostRef.$ancestorComponent$);
    // there is no ancestorc omponent or the ancestor component
    // has already fired off its lifecycle update then
    // fire off the initial update
    const dispatch = () => dispatchHooks(hostRef, isInitialLoad);
    return BUILD.taskQueue ? writeTask(dispatch) : dispatch; // <- shouldn't this be `dispatch()`?
};
```
After patching the line locally, the component renders fine.

maxjustus:
This seems to have started with 1.13.0 and is still broken in 1.17.3 Task queue immediate just renders nothing when enabled

PR #2630, opened 2020-08-11
Fix taskQueue: "immediate" rendering nothing

maxjustus:
With `taskQueue: "async"` and `"congestedAsync"` `scheduleUpdate` calls `writeTask`, passing it the `dispatch` callback. `writeTask` calls the passed in dispatch callback in `process.nextTick`

Before this change, when the taskQueue: was set to `immediate`, the dispatch callback was never called. Resulting in no components rendering in the DOM.

Fixes https://github.com/ionic-team/stencil/issues/2491


adamdbradley:
thanks for the fix!

PR #2637, opened 2020-08-14
docs(changelog): example to restore 1.x defaults

bodinaren:
Add a section to make it easy for devs to restore their Stencil 1 defaults. I'm not really sure about `taskQueue` though. The goal is to make it easy for people to do the upgrade easy without breaking, but changing the `taskQueue` isn't a breaking change and so I didn't want to suggest it here. Feel free to overrule me on that though :)


adamdbradley:
thanks!

PR #2652, opened 2020-08-27
fix: add ref attribute to SlotAttributes

claviska:
Quick PR to enable `ref` on slots:

```html
<slot ref={el => this.mySlot = el} />
```


claviska:
Moving this to a draft PR since it seems to fail in components that _don't_ use a shadow DOM.

bitflower:
Sounds interesting....

manucorporat:
Closing, since <slot> can be virtual when not using shadow-dom mode, thanks for the PR Cory!

PR #2689, opened 2020-10-02
fix(mock-doc): set hostname when location is updated

### LINKED ISSUE #2688: window.location.hostname not populated when window.location.href set.
**Stencil version:**
```
 @stencil/core@2.1.0-2
```

**I'm submitting a:**

- [x] bug report
- [ ] feature request

**Current behavior:**
In a jest spec unit test test that the `window.location.hostname` is set when the `href` is set:
```ts
import { newSpecPage } from '@stencil/core/testing';

describe('location.hostname', () => {
  it('has a hostname', async () => {
    const page = await newSpecPage({
      url:'http://test.example.com/',
      components: [],
      html: `<html></html>`,
    });
    expect(window.location.hostname).toEqual('test.example.com');
  });
});
```
output:
```
 FAIL  src/components/location-hostname/test/location-hostname.spec.tsx
  ● location-hostname › has a hostname

    expect(received).toEqual(expected) // deep equality

    Expected: "test.example.com"
    Received: ""

       8 |       html: `<html></html>`,
       9 |     });
    > 10 |     expect(window.location.hostname).toEqual('test.example.com');
         |                                      ^
      11 |   });

      at Object.<anonymous> (src/components/location-hostname/test/location-hostname.spec.tsx:10:38)
```

As we can see the hostname is not set.

**Expected behavior:**
The hostname should be populated.

**Steps to reproduce:**
See current behaviour.

**Related code:**
https://github.com/ionic-team/stencil/blob/master/src/mock-doc/location.ts#L22

**Other information:**
I have a PR ready to submit to fix this, but the contributing guidelines said to create an issue first.

cam-narzt:
The hostname was not updated when a new href is set on the window location, it looks like this was an oversight, this commit makes it so the hostname is populated like everything else.

chore(test): add test for MockLocation behaviour

Closes #2688

### DIFF
--- a/src/mock-doc/location.ts
+++ b/src/mock-doc/location.ts
@@ -20,6 +20,7 @@ export class MockLocation implements Location {
     this._href = url.href;
     this.protocol = url.protocol;
     this.host = url.host;
+    this.hostname = url.hostname;
     this.port = url.port;
     this.pathname = url.pathname;
     this.search = url.search;
--- a/src/mock-doc/test/location.spec.ts
+++ b/src/mock-doc/test/location.spec.ts
@@ -1,0 +1,95 @@
+import { MockWindow } from '../window';
+
+describe('location.href', () => {
+  let win: MockWindow;
+  beforeEach(() => {
+    win = new MockWindow(`
+      <html>
+        <head>
+        </head>
+      </head>
+      `);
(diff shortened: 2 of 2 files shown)


adamdbradley:
great fix, thanks!

PR #2694, opened 2020-10-06
fix(runtime): do regular clone of normal slotting

justusromijn:
The extra option `cloneNode` has a drawback for angularJS. When an ng-if is true at its initial render, the originally slotted content is not visible. This is not a problem for ng-if blocks that start out with a falsy value and are appearing later.

By adding a check to not patch up any non-stencil nodes, this problem seems resolved.

I would like to get some feedback from the maintainers whether this has a chance of making the codebase, it is quite a blocker for our team to upgrade to 1.8.x and higher.

I could not find any tests related to the dom-extra's, please point me in the correct direction if this needs proper coverage.

Here's a [demo repository](https://github.com/justusromijn/stencil-clonenode-repro) with a stencil generated app (1.8.x), loading a very basic angularJS binding context to demonstrate the problem. In the readme are the steps to reproduce the different problems, and a way to apply the patch of this PR to see the solution.


justusromijn:
I'm in the process of creating a demonstration app to show the problem with the cloneNode patch, once finished I will link it from the PR.

justusromijn:
I've included a test scenario as requested. Its not perfect with the inclusion of the angular library, I could not find a way to prevent double loading when running karma tests (but if I do not load it from the karma config, it seems to also not load from the index file itself). Maybe the Ionic Team can help me out there?

PR #2696, opened 2020-10-07
feat(docs-custom): add config argument

jgroth:
I would like to have access to the logger when using the `docs-custom` output target, at the moment only the `docs` generated by Stencil is given to the `generator` function.

Sending only the logger to `generator` seems a bit weird and since the logger is available on `config` I added that as an argument instead. There might be other stuff on `config` that also could be of use to anyone using the `docs-custom` output target.

If there is another way to do this, please let med know! :slightly_smiling_face: 
I'm also not sure if there is anything else I need to update, all the tests pass at least! :slightly_smiling_face:


adamdbradley:
looks good, thanks!

PR #2736, opened 2020-11-06
Update link to puppeteer's key names

bjankord:
Recently I was working on setting up a test and wasn't sure if the key I needed pass was named 'Enter' or 'Return'. As I dug into the E2EElement interface and was reading about the press method, I found a link to the Keys that puppeteer supports. I've updated the comment to use the new link as the old one was 404ing.


bjankord:
cc/ @manucorporat @adamdbradley

claviska:
Thanks!

PR #2738, opened 2020-11-09
Allow class component to be referred to in JSX by the class object

truongsinh:
Allow class component to be referred to in JSX by the class object, instead of tag name. E.g:

```
return <HelloWord />
```

Slack discussion: https://stencil-worldwide.slack.com/archives/C79EANFL7/p1604929720060900


scabana:
Got a feeling that the json doc will be missing stuff for fields: 
```typescript
dependents: string[];
dependencies: string[];
dependencyGraph: JsonDocsDependencyGraph;
```

christian-bromann:
I looked into this PR and try to apply it on the current code base. While it actually allows to do what it claims I don't see an advantage of it right now as the user would loose all type safety using the class component instead of the tag name. There would be more to add for this to happen. Hence we will close it but make sure to add a note to our roadmap to see whether we want/can make this possible.

PR #2437, opened 2020-05-13
Create 2020-05-13.md

mattdsteele:

PR #2745, opened 2020-11-19
fix(events): map onFocusIn/Out to correct events

### LINKED ISSUE #2628: issues with focusin event in JSX
**Stencil version:**

```
 @stencil/core@1.17.2
```

**I'm submitting a:**

[x] bug report
[ ] feature request

**Current behavior:**

If i render something like this:

```jsx
<div onFocusIn={e => console.log("focusin", e)}>
  <button>hello world!</button>
</div>
```

I _will never_ see "focusin" logged in the console when the button receives focus. 

It's possible the same issue exists for `focusout` event, but I have not tested this.

**Expected behavior:**

If i render something like this:

```jsx
<div onFocusIn={e => console.log("focusin", e)}>
  <button>hello world!</button>
</div>
```

I _will_ see "focusin" logged in the console when the button receives focus.

**Steps to reproduce:**

The above code should be sufficient

**Other information:**

After inspecting what listeners are registered on the div, I see stencil has added a listener for `focusIn` instead of `focusin`. Event names are case-sensitive so this does not work. Here's a screenshot of the registered listeners in the devtools:

image

Since this is an issue with how stencil maps `onFocusIn` in JSX to the DOM event `focusin`, it is possible to workaround this issue by using `onFocusin` in JSX, and telling typescript to ignore the wrong name:

```jsx
{/* @ts-ignore */}
<div onFocusin={e => console.log("focusin", e)}>
  <button>hello world!</button>
</div>
```

But ideally this shouldn't be necessary!

claviska:
Fixes #2628.

Support for `onFocusIn` and `onFocusOut` were added in #2437. These events are supported by [all browsers](https://caniuse.com/focusin-focusout-events) and are widely used, but [they're not defined as properties in the spec](https://github.com/whatwg/html/issues/3514) yet. As such, they don't exist as properties on the `window` object.

The end result is the attached listeners being `focusIn` and `focusOut` instead of `focusin` and `focusout`. ~This PR adds an array of event transformations to fall back on for edge case events such as these.~

**Update:** we've decided to change the types to `onFocusin` and `onFocusout` so we don't modify the runtime. Implemented in 513c571.


claviska:
I took a look at what Preact does and it turns out [they don't support either event](https://github.com/preactjs/preact/issues/1611#issuecomment-490027816) and [neither does React](https://github.com/facebook/react/issues/6410). Notably, the reasoning is:

> The supported part is actually driven by the fact that the event exists on the Global Event Interface and they do not. Just try it in your console 'onfocusin' in window.

However, in both Preact and React the `onFocus` and `onBlur` methods bubble up, whereas natively (and in Stencil) they do not. It's worth mentioning that, currently, you can make this work with `@ts-ignore` and lowercasing:

```tsx
render() {
  return (
    // @ts-ignore
    <div onFocusin={() => console.log('focus')} onFocusout={() => console.log('blur')}>
      <div contenteditable>Edit me</div>
    </div>
  );
}
```

But this doesn't seem like a good solution. 😕

adamdbradley:
I think the best solution would be to change the JSX types from `onFocusIn` to `onFocusin`. It'd force the correct case, and we wouldn't have to add to the runtime for these special cases.

claviska:
I’ll update it. How do we feel about both variations working once those properties make it to the window object though? 🤔

WickyNilliams:
I'd be happy with the `onFocusin`/`onFocusout` approach

PR #2749, opened 2020-11-24
fix(client): ios9 strict mode does not allow function declarations in a lexically nested statement

huanz:


huanz:
@adamdbradley

huanz:
@adamdbradley

Issue #2728, opened 2020-11-04
Module ?worker has no exported member 'workerPath'


realleocaster:
**Stencil version:**

```
 @stencil/core@2.0.3
```

**I'm submitting a:**

[x] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**
workerPath is not exported when trying to import from "myworker.ts?worker"

**Expected behavior:**
workerPath is defined and one can instantiate new web workers using the path

**Steps to reproduce:**
```
import { workerPath } from "myworker.worker.ts?worker";

let worker = new Worker(workerPath);
```

[ ERROR ]  TypeScript: ./src/components/....tsx:3:10
           Module '"*?worker"' has no exported member 'workerPath'.

      L3:  import { workerPath } from './myworker.worker.ts?worker';

PR #2754, opened 2020-11-27
Update *?worker declaration

George-Payne:
`*?worker` declaration is outdated / doesnt match current behaviour.

Updated declaration based on `import * as worker from './worker.worker.ts?worker;`

image

and https://stenciljs.com/docs/web-workers#advanced-cases


realleocaster:
Fixes #2728

adamdbradley:
thanks!

PR #2773, opened 2020-12-17
fix(client): Test for presence of replace-Method of CSSStyleSheet

MBlanquett:
Test for presence of replace-Method of CSSStyleSheet as we rely on this method. Not every browser that implements CSSStyleSheet does also implement CSSStyleSheet.replace()


MBlanquett:
[](url) #2773

claviska (approved):

PR #2779, opened 2020-12-23
Immutable props fix (fixes #2433)

### LINKED ISSUE #2433: Props are immutable by default but still able to manipulate without mentioning mutable: true
Been stuck where documentation refers props are immutable by nature but explicitly we can set them mutable by assigning mutable:true, but in my case i can still change the prop value without setting mutable property, 

```tsx
import { Component, ComponentInterface, h, Prop} from '@stencil/core';

@Component({
  tag: 'app-button',
  styleUrl: 'app-button.css'
  shadow: true,
})
export class AppButton implements ComponentInterface {
@Prop() message : string = "Hello Everyone!!"
componentDidLoad(){
  
  this.message ="kkkk"
  console.log(this.message)
}
render() {
    return (
      <p>{this.message}</p>
    );
    }

}
```

Instead of showing "Hello Everyone!!" as output it shows kkkk

claviska:
This fixes [a longstanding bug](https://github.com/ionic-team/stencil/issues/2433) where the warning for immutable prop changes doesn't get shown. It also adds tests to ensure the warning is emitted at the correct times.

Side note: this will probably confuse people once it gets merged because a lot of projects are likely changing props without the `{ mutable: true }` flag. I've already noticed a few in Stencil Router. 😆

### DIFF
--- a/src/runtime/proxy-component.ts
+++ b/src/runtime/proxy-component.ts
@@ -2,7 +2,7 @@ import type * as d from '../declarations';
 import { BUILD } from '@app-data';
 import { consoleDevWarn, getHostRef, plt } from '@platform';
 import { getValue, setValue } from './set-value';
-import { MEMBER_FLAGS } from '../utils/constants';
+import { HOST_FLAGS, MEMBER_FLAGS } from '../utils/constants';
 import { PROXY_FLAGS } from './runtime-constants';
 
 export const proxyComponent = (Cstr: d.ComponentConstructor, cmpMeta: d.ComponentRuntimeMeta, flags: number) => {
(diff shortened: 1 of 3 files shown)


NRaf:
@claviska Just updated to Stencil 2.4.0 and noticed this warning starting to come through my unit tests. I'm not actually mutating any props through my component logic, however, I do mutate props in my unit tests. Is there any way to disable the warning during tests? Or is there a recommended alternative?

allinne:
My unit tests  pass even in 2.0.1 with an error:
`@Prop() "state" on <my-component> is immutable but was modified from within the component.` 
The test looks like:
```
const myComponent = new MyComponent();
myComponent.state = 'error';
```
To pass the tests I needed make all props mutable but it is not the right way. Please, clarify this moment.

PR #2788, opened 2021-01-05
fix(mock-doc): make MockAttributeMap iterable

simonhaenisch:
TL;DR Proposed changes:

- implement `[Symbol.iterator]` in `MockAttributeMap` to make it iterable
- implement `get [Symbol.toStringTag]` in `MockAttributeMap`
- don't pass symbols into `isNaN` in attribute proxy
- prettier formatted a couple things

---

When writing E2E tests I ran into different `PrettyFormatPluginError` errors, when Jest would try to pretty-format an `E2EElement` while iterating over its attributes:

```
PrettyFormatPluginError: Cannot read property 'name' of undefinedTypeError: Cannot read property 'name' of undefined
```

or

```
PrettyFormatPluginError: _val$hasAttribute.call is not a functionTypeError: _val$hasAttribute.call is not a function
```

```
at node_modules/jest-matcher-utils/node_modules/pretty-format/build/plugins/DOMElement.js
```

---

This error might only happen in my case because I'm using `jest@26` and I think Stencil still defaults to `jest@25`, but doesn't complain if a higher major version is used cause it's likely compatible.

Anyway, after debugging this for half a day, I figured out that `pretty-format` fails to print the attributes because `MockAttributeMap` is not really iterable but it tries to iterate it (passed into `Array.from`, see [facebook/jest/pretty-format/src/plugins/DOMElement.ts#L87](https://github.com/facebook/jest/blob/7430a7824421c122cd07035d800d22e1007408fa/packages/pretty-format/src/plugins/DOMElement.ts#L87) and [facebook/jest/pretty-format/src/plugins/DOMElement.ts#L87](https://github.com/facebook/jest/blob/7430a7824421c122cd07035d800d22e1007408fa/packages/pretty-format/src/plugins/DOMElement.ts#L92)), and ends up being an Array of `undefined`s (not sure why).

The solution is to add an implementation of `[Symbol.iterator]`. Another symbol that pretty-format tries to access is `Symbol.toStringTag` so I've also implemented that.

---

Also, the attribute proxy uses `isNaN` which breaks when called on a symbol:

```
PrettyFormatPluginError: Cannot convert a Symbol value to a numberTypeError: Cannot convert a Symbol value to a number
        at isNaN (<anonymous>)

      at Object.get (node_modules/@stencil/core/mock-doc/index.cjs:21:10)
          at Proxy.toString (<anonymous>)
          at Array.map (<anonymous>)
```

thus I've added a check for that.

### DIFF
--- a/src/mock-doc/test/attribute.spec.ts
+++ b/src/mock-doc/test/attribute.spec.ts
@@ -1,12 +1,20 @@
+import { MockAttr, MockAttributeMap } from '../attribute';
 import { MockDocument } from '../document';
 import { MockElement, MockHTMLElement } from '../node';
 import { XLINK_NS } from '../../runtime/runtime-constants';
-import { MockImageElement } from '../element';
 
 describe('attributes', () => {
   let doc: MockDocument;
   beforeEach(() => {
     doc = new MockDocument();
+  });
+
+  it('attribute map is iterable', () => {
+    const map = new MockAttributeMap();
+    const attr = new MockAttr('attr', 'value');
+    map.setNamedItem(attr);
+
+    expect(Array.from(map)[0]).toBe(attr);
   });
 
   it('should get attributes by index', () => {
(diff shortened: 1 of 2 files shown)


src/mock-doc/test/attribute.spec.ts
```diff
@@ -1,14 +1,22 @@
...
   beforeEach(() => {
     doc = new MockDocument();
   });
 
+  it('attribute map is iterable', () => {
+    const map = new MockAttributeMap();
+    const attr = new MockAttr('attr', 'value');
+    map.setNamedItem(attr);
+
+    expect(Array.from(map)[0]).toBe(attr);
```
simonhaenisch, line 17: I've confirmed that this test works by removing the `[Symbol.iterator]` implementation. `Array.from(map)` would then yield `[undefined]` instead of `[MockAttr]`.

Issue #2837, opened 2021-03-01
Export of readme markdown generators as part of Stencil Core Compiler API


legaev-stas:
**Stencil version:**

```
 @stencil/core@2.4.0
```

**I'm submitting a:**

[ ] bug report
[x ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**
Stencil generates a great API documentation. But if the default generator does not satisfy business needs, the consumers could generate [custom documentation](https://stenciljs.com/docs/docs-custom) out of JsonDocs sources. But in this case the consumer has to create it from the scratches, while in some business cases the consumer might want to customize only some parts of default document.

Example: 
I need to modify default document generation process in the following way:
- to export all CSS Custom Props from all of the recursive dependencies. So that, parental element lists all nested customizable CSS props 
- to add additional Interfaces section

Both of my changes do not modify existing markdown generators. These features only require data modification and creation of new Interface markdown generator + document template modification. I don't have any option how to re-use default markdown generators except copy their sources and their dependencies.

**Expected behavior:**
Stencil exports markdown generators. They possibly could be exported as part of [Stencil Core Compiler API](https://stenciljs.com/docs/compiler-api). It could be exported as single `docReadmeUtils` object that has references to all of the generators.

With this feature consumers who want to implement partial redesign of default document generation process could reuse default markdown generators.

jordan-boyer:
This feature would really be great. 

For example I really want to add example of my stencil composent using framework (vue, react).

I like the default formatter but i don't really want to write all the docs for the major framework, and I know it's possible with the custom doc but I don't want to copy paste your code for generating some part of the docs that I like

beckysolomon:
This feature would be so helpful. I want to output usage and more robust documentation for methods, but I don't want to start from scratch. Would be nice to base it off the current default behavior and then customize from there.

rwaskiewicz:
Hey folks 👋 

I apologize that it took someone on the team so long to reply to this issue. 

I discussed this with the team last night, and have agreed we'd rather pursue a different way of going about this. At this time, we feel that the various `*ToMarkdown` functions are best kept internal to Stencil. 

However, we do wonder if there is a better compromise somewhere between creating a custom generator and copying the existing `*ToMarkdown` functions. Would adding an _override_ of existing markdown generators solve the use case here? With a theoretical override functionality, projects could (in theory):
[...]

PR #2842, opened 2021-03-05
feat(compiler): apiDocsGenerationUtils are exported as part of compiler module

legaev-stas:
This is proposed solution for [Issue 2837](https://github.com/ionic-team/stencil/issues/2837)


rwaskiewicz:
Hey @legaev-stas 👋 

I apologize that it took so long for someone on the team to respond to this pull request. While the team and I appreciate the effort you went through for this work, we don't feel that this change is a good fit for Stencil at this time. Right now, we'd prefer to keep these APIs internal to Stencil. However, we do believe that there is a possibility for reaching some middle ground between writing a custom Docs generator, and having to accept all of Stencil's defaults. I've added a comment to the original issue [here](https://github.com/ionic-team/stencil/issues/2837#issuecomment-1490227737). Let us know what you think, and we can move forward from there. Thanks!

PR #2846, opened 2021-03-09
feat: expose `Fragment` on `h`

simonhaenisch:
This way it's possible to set `jsxFragmentFactory` to `h.Fragment` in tsconfig, so that `<></>` syntax can be used without having to import `Fragment` explicitly (just importing `h` is sufficient).


simonhaenisch:
The broken test is from eec7651fa723658c2c8d853dc44b6316709b2317 (see https://github.com/ionic-team/stencil/runs/1796291494). All other tests pass locally.

manucorporat:
This is kind of cool!, i like it. My only concern is so performance regression because adding a property to an arrow function is disables some V8 optimization.

Wondering how we could check that

simonhaenisch:
Interesting, I didn't know that. Is this different when using a normal `function h` instead of assigning an anonymous arrow function to `h`? Do you have a link that explains this optimization? Is there a way to measure the impact of this?

BTW In the meantime what I did was to patch `h` at runtime with a globalScript.

```ts
import { h, Fragment } from '@stencil/core;

export default () => {
  if (!('Fragment' in h)) {
    h.Fragment = Fragment;
  }
}
```

(breaks spec tests with `newSpecPage` though)

simonhaenisch:
Hm, idea: 

```ts
const h = () => { /* ... */ }

export default h;

export const Fragment = (_, children) => children;
```

```ts
import h from './h'

export { h }
```

simonhaenisch:
Never mind, mixing default and named exports doesn't seem to work with Rollup (see [repl](https://rollupjs.org/repl/?version=2.41.5&shareable=JTdCJTIybW9kdWxlcyUyMiUzQSU1QiU3QiUyMm5hbWUlMjIlM0ElMjJtYWluLmpzJTIyJTJDJTIyY29kZSUyMiUzQSUyMmltcG9ydCUyMGglMjBmcm9tJTIwJy4lMkZoLmpzJyUzQiU1Q24lNUNuaCgpJTNCJTVDbiU1Q25jb25zb2xlLmxvZyhoLkZyYWdtZW50KSUzQiUyMiUyQyUyMmlzRW50cnklMjIlM0F0cnVlJTdEJTJDJTdCJTIybmFtZSUyMiUzQSUyMmguanMlMjIlMkMlMjJjb2RlJTIyJTNBJTIyY29uc3QlMjBoJTIwJTNEJTIwKCklMjAlM0QlM0UlMjBjb25zb2xlLmxvZygnSSUyMGFtJTIwaC4nKSUzQiU1Q24lNUNuZXhwb3J0JTIwZGVmYXVsdCUyMGglM0IlNUNuJTVDbmV4cG9ydCUyMGNvbnN0JTIwRnJhZ21lbnQlMjAlM0QlMjAoKSUyMCUzRCUzRSUyMGNvbnNvbGUubG9nKCdJJTIwYW0lMjBGcmFnbWVudC4nKSUzQiUyMiUyQyUyMmlzRW50cnklMjIlM0FmYWxzZSU3RCU1RCUyQyUyMm9wdGlvbnMlMjIlM0ElN0IlMjJmb3JtYXQlMjIlM0ElMjJlcyUyMiUyQyUyMm5hbWUlMjIlM0ElMjJteUJ1bmRsZSUyMiUyQyUyMmFtZCUyMiUzQSU3QiUyMmlkJTIyJTNBJTIyJTIyJTdEJTJDJTIyZ2xvYmFscyUyMiUzQSU3QiU3RCU3RCUyQyUyMmV4YW1wbGUlMjIlM0FudWxsJTdE)).

tanner-reits:
Hi @simonhaenisch I closed this PR as we were purging stale branches from the repo. If this is still functionality you'd like to see implemented, please fork the repo and open a new PR, or open an issue!

Issue #2848, opened 2021-03-10
An OffscreenCanvas could not be cloned because it was not transferred.


arsenpellumbi:
**Stencil version:**

```
 @stencil/core@latest
```

**I'm submitting a:**

[x] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**

Trying to transfer the OffscreenCanvas from main thread to a web-worker using StencilJS web-worker proxy. This prevents from using web-gl in web-workers.

Error:
`cv-canvas-3469fcdb.js:65 Uncaught (in promise) DOMException: Failed to execute 'postMessage' on 'Worker': An OffscreenCanvas could not be cloned because it was not transferred.`

**Expected behavior:**

All transfarable objects should be transferred as well, not only ArrayBuffers.
(https://developer.mozilla.org/en-US/docs/Web/API/Transferable)

**Steps to reproduce:**

Create a web-worker containing a simple method that accepts an OffscreenCanvas object as parameter.
Create a component that contains a canvas in its DOM. In componentDidLoad() event, call the web-worker method sending the OffscreenCanvas from canvas. It will give an error.

**Related code:**
Component: canvas.tsx
```tsx
import { Component, h, State } from '@stencil/core';
import { init } from './web-gl.worker.ts';

@Component({
  tag: 'cv-canvas',
  styleUrl: 'cv-canvas.scss',
  shadow: true,
})
export class CvCanvas {
  canvas: HTMLCanvasElement;

  componentDidLoad() {
    const offscreenCanvas = this.canvas.transferControlToOffscreen();
    init(offscreenCanvas);
  }

  render() {
    return (
      <canvas ref={el => this.canvas = el} style={{"width": `${window.innerWidth}px`, "height": `${window.innerHeight - 120}px`, "top": "120px", "position": "fixed"}}></canvas>
    );
  }

}
```

Web-Worker: web-gl.worker.ts
```tsx
let canvas: OffscreenCanvas;

export const init = async (offscrenCanvas: OffscreenCanvas) => {
  canvas = offscrenCanvas;
  return canvas;
};
```

**Other information:**

Error happens in worker-plugin.ts. 
```tsx
    const postMessage = (w) => (
      w.postMessage(
        [workerMsgId, pendingId, exportedMethod, args],
        args.filter(a => a instanceof ArrayBuffer)
      )
    );
```

[...]

PR #2849, opened 2021-03-10
fix(worker-plugin): transfer OffscreenCanvas

arsenpellumbi:
Issue: https://github.com/ionic-team/stencil/issues/2848


manucorporat:
could we make it a small utility function? like `const isTransferable`?

arsenpellumbi:
> could we make it a small utility function? like `const isTransferable`?

The problem is that the check is used in two separated contexts, one is in the "getWorkerIntro" which is attached to the worker, and the other is in the "WORKER_HELPERS" which is attached to the file that instantiates the worker.
One simple way is to create a string containing the function and attach it to both "getWorkerIntro" and "WORKER_HELPERS", but I don't see any benefit doing that.

manucorporat:
ah you right! merging i am actually thinking of using getTransferables in both contexts

PR #2863, opened 2021-03-22
Add peer dependencies for test related dependencies

jharrilim:
Running `stencil test --spec --e2e` after initializing a project yields:

```txt
[13:37.0]  @stencil/core
[13:37.3]  v2.4.0 📷

[ ERROR ]  Please install missing dev dependencies with either npm or yarn.
           npm install --save-dev @types/jest@26.0.20 jest@26.6.3 jest-cli@26.6.3 @types/puppeteer@5.4.2
           puppeteer@5.5.0
```

At a glance, it seems like this project aims to have 0 dependencies so I did not add it there. Adding it to peer dependencies acknowledges that this should be done earlier on, and is an expected part of the process.

### DIFF
--- a/package.json
+++ b/package.json
@@ -118,6 +118,13 @@
     "webpack": "^4.46.0",
     "ws": "7.4.2"
   },
+  "peerDependencies": {
+    "@types/jest": "^26.6.3",
+    "@types/puppeteer": "^5.4.2",
+    "jest": "^26.6.3",
+    "jest-cli": "^26.6.3",
+    "puppeteer": "5.5.0"
+  },
   "engines": {
     "node": ">=12.10.0",
     "npm": ">=6.0.0"


rwaskiewicz:
Hey @jharrilim, thanks for the PR! To make sure I understand the problem you're trying to solve correctly, can you confirm that you see this issue after creating a new Stencil project using the Stencil CLI?

jharrilim:
> Hey @jharrilim, thanks for the PR! To make sure I understand the problem you're trying to solve correctly, can you confirm that you see this issue after creating a new Stencil project using the Stencil CLI?

Indeed, I do see the issue after creating a new Stencil project using the Stencil CLI:

image

jharrilim:
Actually this may not necessarily be the best place to add it.

I suppose it would be better suited as either:
1. A dependency of `@stencil/core` (might be more appropriate since `stencil` configures jest and offers the `stencil test` command)
2. A dev dependency of [stencil-component-starter](https://github.com/ionic-team/stencil-component-starter) (might be more appropriate if it is preferred to have jest and it's swathe of dependencies upfront)

I personally like the first option, as I would have to download Jest anyway, and since it is managed by stencil, it is probably best if stencil manages this version as well, even though it would require adding a rather large dependency.

rwaskiewicz:
All great points!

To start I really like this idea:

> A dev dependency of stencil-component-starter (might be more appropriate if it is preferred to have jest and it's swathe of dependencies upfront)

This helps with the developer experience quite a bit, since one would expect dependencies to be in a 'boilerplate' project that the `create-stencil` CLI creates (but run into the exact issue you described above where they need to install a bunch of testing dependencies).

At the moment, the `create-stencil` CLI uses [one](https://github.com/ionic-team/create-stencil/blob/280f3d1644fb1c8791d7473328696a2738dc9cb0/src/starters.ts#L12) [of](https://github.com/ionic-team/create-stencil/blob/280f3d1644fb1c8791d7473328696a2738dc9cb0/src/starters.ts#L18) [three](https://github.com/ionic-team/create-stencil/blob/280f3d1644fb1c8791d7473328696a2738dc9cb0/src/starters.ts#L24) repositories for spinning up a new project:
- [ionic-team/ionic-pwa-toolkit](https://github.com/ionic-team/ionic-pwa-toolkit) for PWAs
- [ionic-team/stencil-app-starter](https://github.com/ionic-team/stencil-app-starter) for applications
- [ionic-team/stencil-component-starter](https://github.com/ionic-team/stencil-component-starter) for component(s) (which you've pointed out already 🙂)

if you're interested, we'd be grateful for PRs in each of there repos, as we think they'd greatly improve the testing dev experience.  Making these dev dependencies sounds like a good idea to me, I think that's a good call out on your part. If  you're not able to/interested, that's totally cool - if you could let us know that'd be great too.

As far as updating `@stencil/core` goes:

> A dependency of @stencil/core (might be more appropriate since stencil configures jest and offers the stencil test command)

Let me talk to the team about this further - we've discussed this recently, and I think need to come to consensus on the trade-offs here. I 100% agree with your point there and think it's totally valid.  What I would like to dive in more into is some of the consequences of that decision (if they exist).

jharrilim:
> if you're interested, we'd be grateful for PRs in each of there repos, as we think they'd greatly improve the testing dev experience. Making these dev dependencies sounds like a good idea to me, I think that's a good call out on your part. If you're not able to/interested, that's totally cool - if you could let us know that'd be great too.

Done! Here are the PRs:
https://github.com/ionic-team/ionic-pwa-toolkit/pull/116
https://github.com/ionic-team/stencil-app-starter/pull/76
https://github.com/ionic-team/stencil-component-starter/pull/98

rwaskiewicz:
Thank you so much 👏 👏 👏 

I had a few comments on each - if you could take a look I'd appreciate it!

jharrilim:
@rwaskiewicz Thanks for the suggestions, I have applied all of them!

rwaskiewicz:
Thanks @jharrilim ! I've approved/merged your PRs and am going to close this one since those three supersede this one.  Thanks again!

Issue #2886, opened 2021-04-16
Markdown file changes don't trigger build in watch mode


janrembold:
**Stencil version:**

```
 @stencil/core@latest
```

**I'm submitting a:**

[x] bug report
[x] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**

- Markdown file changes do not trigger the build task in watch mode
- This leads to not updating `docs.json` created by the `docs-json` output target
- While using Storybook this forces the developer to restart the whole build process to see the documentation output in a Storybook story

**Expected behavior:**

Markdown files trigger a build in watch mode. IMHO this should be either default while using the `docs-json` output target or at least configurable.

**Steps to reproduce:**

- Simply set up a new project
- Add `docs-readme` output target
- Let stencil create a readme.md inside a component folder
- Start `build --watch` or `build --watch --docs` (doesn't matter)
- Change readme.md
- See that nothing happens

**Related code:**
https://github.com/ionic-team/stencil/blob/master/src/compiler/build/watch-build.ts#L161

```tsx
  const srcFiles = await compilerCtx.fs.readdir(config.srcDir, {
    recursive: true,
    excludeDirNames: ['.cache', '.git', '.github', '.stencil', '.vscode', 'node_modules'],
    excludeExtensions: [
      '.md',
      '.markdown',
      '.txt',
      '.spec.ts',
      '.spec.tsx',
      '.e2e.ts',
      '.e2e.tsx',
      '.gitignore',
      '.editorconfig',
    ],
  });
```

**Other information:**

I would like to provide a PR for this issue. Could someone from the core team give me a hint what kind of fix/change would be of interest for you? 

I'm thinking about different possible solutions:
1. Add new configurations for `watch.excludeDirNames` and `watch.excludeExtensions` 
2. Exclude `.md` and `.markdown` only if output targets `docs-readme` or `docs-json` are not set

What do you think?

janrembold:
I added a simple non-breaking PR for only a single additional config option `watchExcludeExtensions`. 
I would be very happy to receive feedback on this.

benmb1984:
It would be great if this simple solution will be implemented. I have a Stencill app that is consuming md files but now I have to start a new build for each change in my md files. @janrembold, why is your pr not merged?

Are this issues also related to this missing config option? 

- #2560 
- #2780 
- #1822

PR #2887, opened 2021-04-16
feat(build): configure watch exclude extensions

janrembold:
Make excluded extensions for watch task configurable.

See: https://github.com/ionic-team/stencil/issues/2886


janrembold:
tl;dr: This is a non-breaking optional configuration for the excluded extensions used by the watch task

PR #2899, opened 2021-05-03
feat(compiler): @Prop getters / setters

### LINKED ISSUE #230: Not possible to expose a getter with @Method
**Resources:**
Before submitting an issue, please consult our [docs](https://stenciljs.com/).

**Stencil version:** (run `npm list @stencil/core` from a terminal/cmd prompt and paste output below):

```
$ npm list @stencil/core
my-name@0.0.1 /Users/samuelmortenson/repos/stencil-getter
└── @stencil/core@0.0.6-10
```

[x] bug report
[ ] feature request

**Current behavior:**
If you add a `@Method` to a component and use the `get` keyword in front of the function name, the component will compile incorrectly leading to a `Uncaught ReferenceError: Method is not defined` error when loaded in a web browser.

**Expected behavior:**
I would expect that Stencil elements could expose getter/setter methods using `@Method`. These accessors are super useful, especially when trying to emulate the behavior of something like the `<input>` element's "value" attribute. It can get set as an initial property, and the attribute never changes in the DOM, but when `input.value` is called the current value is returned. Would be great to do something similar in Stencil.

**Steps to reproduce:**

1. `git clone git@github.com:mortenson/stencil-getter.git`
1. `cd stencil-getter && npm install && npm start`

See `Uncaught ReferenceError: Method is not defined` in your web browser.

**Related code:**

Here's the syntax I would expect to expose a getter/setter:

```
  @Method()
  get value() {
    return Math.random();
  }
```

(ref https://github.com/mortenson/stencil-getter/blob/master/src/components/my-name/my-name.tsx#L12)

**Other information:**

Here's the TypeScript documentation on Accessors (get/set): https://www.typescriptlang.org/docs/handbook/classes.html#accessors

### LINKED ISSUE #1359: @Prop decorator does not work on a getter/setter
**Stencil version:**

```
 @stencil/core@0.17.2
```

**I'm submitting a:**

[ ] bug report
[X] feature request

**Current behavior:**

When adding a \@Prop decorator to a getter/setter property, the code compiles, and the property is added to the generated interface, but accessing the prop in the client js does not execute the getter, and setting the value does not execute the setter.

To give you an example of where this would be useful, I'm working on a set of mapping components and have a property for the zoom level. The zoom property allows the user to set a zoom programatically and also exposes the current zoom that may be a result of end user mouse interaction with the map. Using the scroll wheel will change the zoom level too.

Currently every time the map is zoomed by mouse interaction, I have to sync the stencil zoom prop with the zoom from my internal mapping api. This causes complications because the `@Watch` decorator also fires when the value is modified by the component internally. See StackOverflow link below. If getter/setter was supported, the stencil prop could act as a more direct passthrough, and I could fetch the zoom level from the internal api only when it's requested.

**Expected behavior:**

**Steps to reproduce:**

**Related code:**
```tsx
@Prop()
get zoom(): number {
return 10;
}
set zoom(newZoom: number) {
console.log('setting the zoom')
}
```

**Other information:**

Follow up to discussion here: https://github.com/ionic-team/stencil/issues/230
Opening this as a new ticket because it should be a \@Prop decorator and not a \@Watch

Also see this StackOverflow post that describes problems you run into when you have to make the prop mutable and sync it by modifying the value. https://stackoverflow.com/questions/52229575/how-to-mutate-a-property-within-a-stenciljs-component-without-triggering-watch

### LINKED ISSUE #2879: Will this project support setters for @Props?
Duplicate of https://github.com/ionic-team/stencil/issues/1359, but that one is locked and left open without an answer.

I would like to at least know if the team is even contemplating this possibility. I would like to be able to do this because I want an easy way for a user of my component to set a value without creating a new prop, assiging it to a lower-level object instead. This would be ideal to make code cleaner, together with all the needs voiced in the original issue.

Sorry for duplicating the issue, I didn't find a better way to do this.

johnjenkins:
Allowing the @Prop decorator to use `get` / `set` provides many quality of life improvements for developers and can improve the api of their components (see linked issues for many valid, uncontested use cases). 
Validation & Guards. Transforms. Read-only properties. All things that may be achievable at the moment but can be extremely cumbersome to implement.

With this PR:
`
// This prop will be readonly (and will add '(readonly)' to the docs description)
@Prop()
get myReadonlyProp() {
  return 'a string';
}`

`// getter & setter with validation
private _agent = 'James Bond'; 
@Prop()
get agent() {
  return this._agent;
}
set agent(agent: string) {
  if (
       typeof agent === 'string' && 
       (agent === 'xxx' || agent === 'James Pond Robocod')
  ) this._agent = agent;
}`

This PR utilises most of the features from the current @Prop decorator. 
Decorator attributes: `attribute` and `reflect` work as expected. 
Prop setters work as expected with @Watch decorated methods (the Watch method will only be called after the setter has actually changed the property value).

fixes #1359 
fixes #2879 
fixes #230

### DIFF
--- a/src/compiler/docs/generate-doc-data.ts
+++ b/src/compiler/docs/generate-doc-data.ts
@@ -116,6 +116,9 @@ const getRealProperties = (properties: d.ComponentCompilerProperty[]): d.JsonDoc
 
       optional: member.optional,
       required: member.required,
+
+      getter: member.getter,
+      setter: member.setter,
     }));
 };
 
@@ -134,6 +137,9 @@ const getVirtualProperties = (virtualProps: d.ComponentCompilerVirtualProperty[]
 
     optional: true,
     required: false,
+
+    getter: false,
+    setter: false,
   }));
 };
 
--- a/src/compiler/docs/readme/markdown-props.ts
+++ b/src/compiler/docs/readme/markdown-props.ts
@@ -34,5 +34,7 @@ const getAttributeField = (prop: d.JsonDocsProp) => {
 };
 
 const getDocsField = (prop: d.JsonDocsProp) => {
-  return `${prop.deprecation !== undefined ? `<span style="color:red">**[DEPRECATED]**</span> ${prop.deprecation}<br/><br/>` : ''}${prop.docs}`;
+  let docField = `${prop.deprecation !== undefined ? `<span style="color:red">**[DEPRECATED]**</span> ${prop.deprecation}<br/><br/>` : ''}${prop.docs}`;
+  if (prop.getter && !prop.setter) docField+= ` (readonly)`;
+  return docField;
 };
--- a/src/compiler/transformers/decorators-to-static/convert-decorators.ts
+++ b/src/compiler/transformers/decorators-to-static/convert-decorators.ts
@@ -79,6 +79,8 @@ const removeStencilDecorators = (classMembers: ts.ClassElement[]) => {
         return ts.updateProperty(m, newDecorators, m.modifiers, m.name, m.questionToken, m.type, m.initializer);
       } else if (ts.isGetAccessor(m)) {
         return ts.updateGetAccessor(m, newDecorators, m.modifiers, m.name, m.parameters, m.type, m.body);
+      } else if (ts.isSetAccessor(m)) {
+        console.log('A get accessor should be decorated before a set accessor');
       } else {
         console.log('unknown class node');
       }
--- a/src/compiler/transformers/decorators-to-static/convert-decorators.ts
+++ b/src/compiler/transformers/decorators-to-static/convert-decorators.ts
@@ -77,6 +77,8 @@ const removeStencilDecorators = (classMembers: ts.ClassElement[]) => {
         return ts.updateMethod(m, newDecorators, m.modifiers, m.asteriskToken, m.name, m.questionToken, m.typeParameters, m.parameters, m.type, m.body);
       } else if (ts.isPropertyDeclaration(m)) {
         return ts.updateProperty(m, newDecorators, m.modifiers, m.name, m.questionToken, m.type, m.initializer);
+      } else if (ts.isGetAccessor(m)) {
+        return ts.updateGetAccessor(m, newDecorators, m.modifiers, m.name, m.parameters, m.type, m.body);
       } else {
         console.log('unknown class node');
       }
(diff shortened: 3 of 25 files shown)


arjunasuresh3:
Any plans for merging this feature?

theo-staizen:
Thank you @johnjenkins for your hard work, you legend! Your PRs are all highly useful features!

rmegaide:
> hi @rmegaide < you can use my fork if you like (at your own risk), I try to keep it up-to-date with upstream. I use it personally, and, as far as I can attest, it works fine :). get only props are very useful to me.
> 
> swap out `"@stencil/core": "npm:@johnjenkins/stencil-community@^2.11.1"`
> 
> in your package.json and npm i ... should work although anecdotally that doesn't always work for windows users

This worked great, thanks a lot.
Will it be merged any time soon ?

vectorjohn:
Just for what it's worth, this PR would be greatly appreciated and used immediately if I could. One of my main problems with Stencil is the fact that Props are not really type safe because realistically they are all `unknown`. Using setters or getters would allow to validate and possibly coerce input values. For example, a Prop that I expect to contain a number is going to come in a `"5"` because it's an attribute. Using a setter, I could convert it to a number. `Watch` almost works for this but not really. To get a similar effect I would have to make a duplicate private property for every Prop, and remember to only read from that one. Which just adds bloat and noise to components.

Chili6666:
Will this feature ever merged to the master? My fingers are crossed since 12 month. Using get/set properties will make our life much more easier.

ysulyma:
This would be really useful to me. I try to imitate existing DOM APIs as much as possible, but there are several properties which behave differently in HTML vs the DOM. For example, the `form` attribute on a `<button>` is a string (the ID of a form), while `HTMLButtonElement["form"]` is an `HTMLFormElement`. To replicate this functionality, I want a setter for the `form` prop that accepts either a string or a form element.

tfrijsewijk:
@johnjenkins Hi John 👋 Great work! Let's hope the Stencil Team can have a look this year 🚂

christian-bromann:
@johnjenkins Hey John 👋 thanks so much for raising this PR. I would love to revive this, would it be possible to update the branch?

ajgagnon:
One of the big problems this will solve. Right now in order pass arrays or objects to components, we need to do this:

```
<script>
   document.querySelector('todo-list-item').itemLabels = ['non-urgent', 'weekend-only'];
</script>
```

With a getter/setter, we could pass a JSON string. Then in the setter, we could `JSON.parse()` if it's a string.

What's needed to revive this? It's the top requested PR on Stencil right now.

PR #2900, opened 2021-05-05
fix(rollup): Add `preserveSymLinks: true` to rollup config (#2840)

### LINKED ISSUE #2840: Rollup can't find CSS sibling files for components when running in Bazel sandbox in Stencil >= 1.15.0
**Stencil version:**

```
 @stencil/core >=1.15.0 (tested on newest releases as well 2.5.2)
```

**I'm submitting a:**

- [x] bug report
- [ ] feature request

**Current behavior:**

This change only happens when upgrading from version 1.14.0 to version 1.15.0 and is affected at every version >= 1.15.0.
When running in a [Bazel](https://github.com/ionic-team/stencil/compare/v1.14.0...v1.15.0) sandbox, rollup is unable to resolve imports for the corresponding CSS file for a stencil component.

Say I have a component as the one pasted below, the build will work fine running directly via stencil, but then when executing through Bazel, Rollup says it can't resolve the import:
```
[ ERROR ]  Rollup: Unresolved Import
           Could not resolve './my-comp?tag=my-comp' from
           ./src/components/my-comp/my-comp.tsx at error
```
```tsx

@Component({
  tag: "my-comp",
  styleUrl: "my-comp.css",
})
export class MyComp {
  render() {
    return (
      <Host>
        <slot></slot>
      </Host>
    );
  }
}
```

**Expected behavior:**
Rollup should find the files if they are located in the correct place. 

**Steps to reproduce:**

I have taken the time to [make an example repo](https://github.com/loudmouth/bazel-stencil) exemplifying the issue and have left run instructions there. The repo illustrates that everything works with 1.14.0 but not with 1.15.0 and above

https://github.com/loudmouth/bazel-stencil

Note that simply omitting component CSS all together, but continuing to include `globalStyle`, the build succeeds with all expected outputs, leading me to believe this is specific only to the way CSS files set via `styleUrls` are resolved 

**Related code:**

**Other information:**

I was reviewing the [changes from 1.14.0 to 1.15.0](https://github.com/ionic-team/stencil/compare/v1.14.0...v1.15.0) and was curious if possibly this issue could be due to a change in the way [CSS file paths are normalized](https://github.com/ionic-team/stencil/compare/v1.14.0...v1.15.0#diff-a90a836949139b0a395b6ecd00c565f300cdee1fbee4efe26f62bab55b1e6524R143).

(If that hyperlink doesn't work, it's in the diff from 1.14.0 to 1.15.0 at src/utils/normalize-path.ts, line 143)

johnjenkins:
Adding `preserveSymlinks: true` to rollup config, fixes relative imports from symlinked files.
I can't see any negative impact of this, all tests pass and works in local projects. 
If it is an issue, can make it configurable. 

fixes #2840

### DIFF
--- a/src/compiler/bundle/bundle-output.ts
+++ b/src/compiler/bundle/bundle-output.ts
@@ -120,6 +120,7 @@ export const getRollupOptions = (
     treeshake: getTreeshakeOption(config, bundleOpts),
     inlineDynamicImports: bundleOpts.inlineDynamicImports,
     preserveEntrySignatures: bundleOpts.preserveEntrySignatures ?? 'strict',
+    preserveSymlinks: true,
 
     onwarn: createOnWarnFn(buildCtx.diagnostics),
 


loudmouth:
I've tested this branch, and can confirm that it gets bazel builds working again!

loudmouth:
One note that I find strange:  adding the follownig to the stencil.config does not solve the issue, but building stencil from source with this change does 🤷 

```js
  nodeResolve: {
    customResolveOptions: {
      preserveSymlinks: true,
    },
  },
```

bitflower:
Cool, could this solve the Stencil <> Stencil hot reload scenario in `lerna` environments? I will try!

tcarrio:
This would definitely help us out too, we're using Bazel and symlinks are absolutely required so rollup needs to be able to work with them. This could at the least be behind a configuration option.

PR #2923, opened 2021-06-01
refactor(canvas-mock): add todataurl in canvas mock

kherP:


src/mock-doc/element.ts
```diff
@@ -372,6 +372,7 @@ export class MockCanvasElement extends MockHTMLElement {
           data: new Array(w * h * 4),
         };
       },
+      toDataURL() {},
```
rwaskiewicz, line 375: I think this should be hoisted to be a method on `MockCanvasMethod`, rather than the object returned by `getContext` - it appears that `toDataURL` is an instance method on a Canvas element according to https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/toDataURL (please let me know if my understanding is incorrect though!).

rwaskiewicz (changes requested):
I apologize that it took so long for a member of the team to take a look at this! If you're willing/able, I had a small ask to move the declaration for `toDataURL`. If not, I completely understand. Thanks again for you contribution, and again I apologize for the lack of response here

christian-bromann:
This was implemented in #5773 and released as part of Stencil [v4.18.3](https://github.com/ionic-team/stencil/tree/v4.18.3)!

Issue #2232, opened 2020-03-02
Nested slots: Nodes are removed/added from DOM instead of updated


simonvizzini:
**Stencil version:**

```
 @stencil/core@1.8.11
```

**I'm submitting a:**

[x] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**
We have a few "design components" which provide nothing more but a default slot and some styling. Nesting more than one such component causes issues when the consuming component has to update due to State/Prop changes. The child nodes inside the wrapper component will get removed from the DOM and then re-added, instead of getting updated. This causes issues especially when trying to use CSS transitions: instead of updating the classList of an element, the whole element gets removed and added, and thus the transition will not get triggered.

This issue does not happen when native shadow DOM is enabled for the components, but only when using the slot polyfill. Unfortunately we still have to support browser that do not support native shadow DOM

**Expected behavior:**
DOM nodes inside nested slots should get updated instead of getting removed completely from the DOM and re-added.

**Steps to reproduce:**

A full example that reproduces the issues can be found here:
https://github.com/simonvizzini/stenciljs-slot-bug-example

It uses a DOM MutationObserver to visualize the issue. The first example is using only a single wrapper component and it works as expected, no DOM mutations are logged and the background transition works fine. The second example uses two nested wrapper components and will log DOM mutations when increasing the counter. Inspecting the logged mutations will reveal that stencil is removing and then adding DOM nodes, thus breaking the CSS transition.

**Related code:**

```tsx
// First wrapper component
@Component({
  tag: "my-wrapper-one",
  styleUrl: "wrapper-one.css"
})
export class WrapperOne {
  render() {
    return (
      <Host>
        <slot></slot>
      </Host>
    );
  }
}

// Second wrapper component, which uses 'my-wrapper-one' internally
@Component({
  tag: "my-wrapper-two",
  styleUrl: "wrapper-two.css"
})
export class WrapperTwo {
  render() {
    return (
      <Host>
        <my-wrapper-one>
          <slot></slot>
        </my-wrapper-one>
      </Host>
    );
  }
}

// Using 'my-wrapper-one' is fine, but with 'my-wrapper-two' the transition doesn't work anymore
@Component({
  tag: "my-component",
  styleUrl: "my-component.css"
})
export class MyComponent {
  @State() count: number = 0;
  private increase = () => {
[...]

PR #2938, opened 2021-06-25
fix(shadowDomShim): slot behaviour

### LINKED ISSUE #1968: Parent with Child Slot rerender bug
**Stencil version:**

```
 @stencil/core@1.7.4
 ---
 update: 15.11.2019
 ---
 @stencil/core@1.8.0
---
 update: 15.01.2020
 ---
 @stencil/core@1.8.5
```

**I'm submitting a:**

[x] bug report
[ ] feature request

**Current behavior:**
Here is a demo: https://dutscher.github.io/stencilslots/

Here is the code which doesnt work from <my-parent /> with comments
```tsx
render() {
    const classes = [];

    if (this.disabled) {
      classes.push(`${this.componentName}--disabled`);
    }

    return (
      <Host class={classes.join(' ')}>
        Demonstrate slot rerender bug
        {/* this dont rerender slot */}
        <div>
          {/* remove condition: works */}
          {/* wrap with <div />: works */}
          {this.disabled && !!this.tag && (
            <div class="tag" innerHTML={this.tag}/>
          )}
          {/* remove div: works */}
          <div>
            before slot&gt;
            <slot/>
            &lt;after slot
          </div>
        </div>
      </Host>
    );
  }
```
The `<slot />` renders first very well, when a rerender in the parent happens, the `<slot />` going away. There is a ugly workarround, you can put a `<div />`. But logical you dont need this `<div/>`.
https://github.com/dutscher/stencilslots/blob/master/src/components/parent-workaround/parent.tsx

I got a scenario there i need more than one `<div />`.

I guess its a vdom diff problem.

**Expected behavior:**
TSX works without extra empty `<div />` and the `<slot />` is rendered after parent is rerendered.

**Steps to reproduce:**
https://dutscher.github.io/stencilslots/

**Related code:**
https://github.com/dutscher/stencilslots/blob/master/src/components/parent-broken/parent.tsx
https://github.com/dutscher/stencilslots/blob/master/src/components/parent-workaround/parent.tsx

Cheers and thanks for the attention :)

### LINKED ISSUE #1993: Different results in the SSR Vs CSR and problem with Vue SSR.
**Stencil version:**

```
 @stencil/core@1.7.4<version>
```

**I'm submitting a:**

[x] bug report
[ ] feature request

**Current behavior:**
The stencil are creating different results in the RSS VS Client, this is a problem for the other libs/frameworks which have a vdom, for example is very bad to try use any lib or component built with the Stencil in a Nuxt Project Universal (SSR). The Vue'll throwing few errors when hydrating on CSR.

I'm using a plugin in my [Nuxt Project](https://github.com/Gomah/nuxt-stencil), for use my Stencil components. Because the SSR and CSR have different outputs, I thing which is a problem with the StencilJS hydrate.

**Expected behavior:**
The code in SSR and CRS are the same, and Vue works fine.

**Steps to reproduce:**
1- clone my example `git clone https://github.com/patrickalima98/stencil-nuxt.git`
2- `npm i`
3- `npm run dev`
4- Open your browser in `http://localhost:3000`
5- Open the console browser and see the errors.

**Other information:**

On the assertNodeMatch:

SSR:
```html
<bm-button color="is-primary" class="hydrated" s-id="1">
  <button class="button is-primary" c-id="1.0.0.0">
    <!--s.1.1.1.0.--><!--t.0.1--> Click Me! 
  </button>
</bm-button>
```

CSR, before vue hydration:

```html
<bm-button color="is-primary" class="hydrated" s-id="1">
  Click Me!
</bm-button>
```

After vue & stencil hydration:

```html
<bm-button color="is-primary" class="hydrated">
<!---->
  <button class="button is-primary hydrated">
    Click Me!
  </button>
</bm-button>
```

image

### LINKED ISSUE #1997: Virtual DOM bug: <slot /> array on second render
**Stencil version:**

```
 @stencil/core@1.7.5
```

**I'm submitting a:**

[x] bug report
[ ] feature request

**Current behavior:**

When rendering an array into a `<slot />`, during the initial render, array items are added at the correct location in the DOM.

On a second render, a new array element is rendered as a child of the parent's parent.

```html
<my-select class="hydrated">
  <!---->
  <select>
    <option value="foo">foo</option>
  </select>
  <option value="bar">bar</option>
</my-select>
```

**Expected behavior:**

On the second render, the new array element is a DOM sibling to other array elements.

```html
<my-select class="hydrated">
  <!---->
  <select>
    <option value="foo">foo</option>
    <option value="bar">bar</option>
  </select>
</my-select>
```

**Steps to reproduce:**

Repro link: https://github.com/petermikitsh/stencil-vdom-bug

**Related code:**

```tsx
import { Component, State, h } from "@stencil/core";

@Component({
  tag: "my-component",
  styleUrl: "my-component.css",
  shadow: true
})
export class MyComponent {
  connectedCallback() {
    window.setTimeout(() => {
      this.options = ["foo", "bar"];
    }, 4000);
  }

  @State() options = ["foo"];

  render() {
    return (
      <div>
        <div>
          <my-select>
            {this.options.map(opt => (
              <option key={opt} value={opt}>
                {opt}
              </option>
            ))}
          </my-select>
        </div>
      </div>
    );
  }
}
```

```tsx
import { Component, h } from "@stencil/core";

@Component({
  tag: "my-select",
  shadow: false
})
export class MySelect {
  render() {
    return (
      <select>
        <slot />
      </select>
    );
  }
}
```

### LINKED ISSUE #2004: Slot Changes Re-Rendered in Wrong Location on IE11/Edge
**Stencil version:**
```
 @stencil/core@1.7.5
```

**I'm submitting a:**
[ X] bug report
[ ] feature request
[ ] support request

**Current behavior:**
Adding child element to a slot on IE11 or Edge displays below template.

**Expected behavior:**
Element displays adjacent to original slot content.

**Steps to reproduce:**
Open https://umdjm.github.io/StencilFallbackDemo/ on IE11
Code https://github.com/umdjm/StencilFallbackDemo

**Related code:**

```tsx
  render() {
    return <div>
      HEADER
      <br/>
      SLOT GOES HERE (
        <slot />
      )
      <br/>
      FOOTER
    </div>;
  }
```
```Usage
    <my-component>
      <p> Initial Slot Data </p>
    </my-component>
```
**Other information:**
## IE Screenshot
IE Screenshot
## Chrome Screenshot
Chrome Screenshot

### LINKED ISSUE #2036: Virtual DOM bug: SSR don't put the content of the slots in correct place in CSR
**Stencil version:**

```
 @stencil/core@1.8.1
```

**I'm submitting a:**

[ x ] bug report
[ ] feature request

**Current behavior:**

When I use any stencil components without Shadow Dom, and insert content for slots (default and named) the hydrate not put the content of slots in the correct place in SSR, the code from the SSR is OK, but the error start in the client side.

The code generated before Stencil Hydrate, here the render of the component is OK.

```html
<my-component class="hydrated sc-my-component-h sc-my-component-s" s-id="1">
  <!--r.1-->
  <!--o.0.1-->
  <!--o.0.2-->
  <div c-id="1.0.0.0" class="sc-my-component sc-my-component-s">
    <!--s.1.1.1.0.-->
    <!--t.0.1-->
    Default slot
    <div c-id="1.2.1.1" class="end sc-my-component sc-my-component-s">
      <!--s.1.3.2.0.end-->
      <span c-id="0.2" slot="end">
        End slot
      </span>
    </div>
  </div>
</my-component>
```

image

The code after Stencil Hydrate, here the texts isn't in the correct place.

```html
<my-component class="hydrated sc-my-component-h sc-my-component-s">
  <!--r.1-->
  <!---->
    Default slot
    
  <!----><span slot="end">
        End slot
      </span>
  <div class="sc-my-component sc-my-component-s">
    <!--s.1.1.1.0.-->
    <div class="end sc-my-component sc-my-component-s">
      <!--s.1.3.2.0.end-->
      
    </div>
  </div>
</my-component>
```

image

**Expected behavior:**

The default slot to display the "Default slot" text  and the named slot to display the "End slot" span.

```html
<my-component>
  <div>
    Default slot
    <div>
      <span>
        End slot
      </span>
    </div>
  </div>
</my-component>
```

**Steps to reproduce:**

```bash
$ git clone https://github.com/patrickalima98/stencil-slot-bug.git
$ cd stencil-slot-bug
$ npm i
$ node ./server.js
# Open your browser in http://localhost:3333
```

**Related code:**
```tsx

@Component({
  tag: 'my-component',
  styleUrl: 'my-component.css',
  shadow: false,
})
export class MyComponent {
  render() {
    return (
      <div>
        <slot/>
        <div class='end'>
          <slot name='end'></slot>
        </div>
      </div>);
  }
}
```

**Other information:**

### LINKED ISSUE #2257: Conditional slots are not rendered correctly in IE11/Edge
**Stencil version:**

```
 @stencil/core@1.9.0
```

**I'm submitting a:**

[x] bug report
[ ] feature request

**Current behavior:**
When a slot within a component is dynamically added or removed, stencil (es5 builds only) does not render correctly.

**Expected behavior:**
If a slot within a component is removed dynamically, the corresponding content should also be removed. If the slot is added again, the content should also reappear.

**Steps to reproduce:**
___
Issue A (toggle slot directly):

***Component***
```tsx
@Component({
  tag: 'dynamic-slot',
  styles: `
    div {
      display: block;
      width: 200px;
      height: 100px;
      background-color: lightcoral;
    }
  `,
  shadow: true
})
export class DynamicSlot implements ComponentInterface {
  @State() toggle: boolean;

  @Listen('click')
  clickHandler() {
    this.toggle = !this.toggle;
  }

  render() {
    return (
      <div>
        <p>{this.toggle ? 'Toggle true' : 'Toggle false'}</p>
        {this.toggle ? <slot /> : null}
      </div>
    );
  }
}
```
***HTML***
```html
<dynamic-slot>
  <a>Should be rendered if toggle is true</a>
</dynamic-slot>
```
***Before click***
The text should be hidden because toggle is false.
IssueA-before-click

***After first click***
Correct rendering after the first click.
IssueA-after-first-click

***After second click***
The text should be removed because toggle is false.
IssueA-after-second-click

***After nth click***
The text does not toggle. Always same result.
IssueA-after-second-click

___
Issue B (toggle slot with div wrapper):

***Component***
```tsx
import {
  Component,
  ComponentInterface,
  h,
  State,
  Listen,
} from '@stencil/core';

@Component({
  tag: 'dynamic-slot',
  styles: `
    div {
      display: block;
      width: 200px;
      height: 100px;
      background-color: lightcoral;
    }
  `,
  shadow: true
})
export class DynamicSlot implements ComponentInterface {
  @State() toggle: boolean;

  @Listen('click')
  clickHandler() {
    this.toggle = !this.toggle;
  }

  render() {
    return (
      <div>
        <p>{this.toggle ? 'Toggle true' : 'Toggle false'}</p>
        {this.toggle ? <div><slot /></div> : null}
      </div>
    );
  }
}
```
***HTML***
```html
<dynamic-slot>
  <a>Should be rendered if toggle is true</a>
</dynamic-slot>
```
***Before click***
The text should be hidden because toggle is false.
IssueB-before-click

***After first click***
Correct rendering after the first click.
IssueB-after-first-click

***After second click***
Correct rendering after the second click (text was removed).
IssueB-after-second-click

***After nth click***
The text should be displayed because toggle is true again.
IssueB-after-third-click

---

**Related code:**

```tsx
```

**Other information:**
In the pre-release @stencil/core 1.9.0-14 it still worked.

### LINKED ISSUE #2259: React fails to manage children in Stencil <slot />
**Stencil version:**
@stencil/core@1.8.1 

**I'm submitting a:**

[X] bug report
[ ] feature request

**Current behavior:**
Given: one or more nested Stencil container tags with Shadow DOM off, which insert DOM levels between the Host and the `<slot />`.
* component-a: Inserts a div between host and `<slot />`.
* component-b: Renders its `<slot />` inside component-a
When React manages the children of the Stencil container tag, it fails to add and remove them correctly.  This happens on every render **after** the first one.
1) Added items are rendered outside the intended slot.
2) Removing items throws a React error and the app fails to render.

**Expected behavior:**
Adding and removing items from the children array should all render in the expected slot.

**Steps to reproduce:**
See this github repo for an example.
React app that uses Stencil components:  https://github.com/bbellmyers/stencil-slots
Stencil component library: https://github.com/bbellmyers/stencil-slot-order

**Related code:**

Component a:
```tsx
<div class="component-a-inserted">
    <slot />
</div>
```

Component b:
```tsx
<component-a>
    <slot />
</component-a>
```

React render:
```jsx
<component-b>
  {children.map((child, index) => (
    <div className="example"
      key={`item-${index}`}>child {child}</div>
  ))}
</component-b>
```
If the children array is longer or shorter, it fails to render correctly on the second render.  re-rendering with any array of the same length works.

**Other information:**

### LINKED ISSUE #2366: prerender: rehydrated components (pages) added twice to the DOM

### LINKED ISSUE #2641: Elements rendered outside of slot when shadow not enabled
**Stencil version:**
```
 @stencil/core@1.17.3
```

**I'm submitting a:**
[ x ] bug report
[ ] feature request

**Current behavior:**
I am seeing the behavior described in #897. The only difference is I do not have shadow enabled. If I enable shadow I see the correct behavior.

**Expected behavior:**
Children added to a slot after initial render should show up inside the designated slot.

**Steps to reproduce:**

**Related code:**

**Other information:**

### LINKED ISSUE #2801: Slotted text node in scoped component is not relocated in the DOM properly when rendered with Maquette
**Stencil version:**

```
 @stencil/core@2.4.0-3
```

**I'm submitting a:**

[x] bug report
[ ] feature request

**Current behavior:**

The second `calcite-label` element rendered using this code reproduces this issue, which is not placing the slotted text node in the proper place in the DOM:

```
<!DOCTYPE html>
<html dir="ltr" lang="en">
  <head>
    <meta charset="utf-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1.0, minimum-scale=1.0, maximum-scale=5.0" />

    <script type="module" src="/build/stencil-starter-project-name.esm.js"></script>
    <script nomodule src="/build/stencil-starter-project-name.js"></script>
  </head>
  <body>
    <script type="module">
      import { h, createProjector } from "https://unpkg.com/maquette@3.3.7/dist/index.js";
      document.addEventListener('DOMContentLoaded', function () {
        const domNode = document.body;
        const projector = createProjector();
        const test = {
          renderMaquette: function () {
            return h("div", [
              h("calcite-label", ["I am a text node and won't get properly relocated"]),
              h("calcite-label", [h("span", ["I am a span and will get properly relocated"])]),
            ])
          }
        };
        projector.append(domNode, test.renderMaquette);
        projector.scheduleRender();
      });
    </script>
  </body>
</html>
```
image

**Expected behavior:**

Sending in plain text (without any wrapping element) to a scoped Stencil component properly places the content in the default `<slot />` position.

**Steps to reproduce:**

1. `git clone https://github.com/eriklharper/stencil-maquette-bug`
2. `npm i && npm start`
3. Open browser developer tools and inspect both rendered `calcite-label` elements.  Note that the first one is not placing the text content in the proper place while the second one is.

**Other information:**

Repro environment: https://github.com/eriklharper/stencil-maquette-bug

### LINKED ISSUE #2877: Slot CRUD does not work properly when shadow is false
**Stencil version:**

```
 @stencil/core@2.5.2
```

**I'm submitting a:**

[x] bug report
[ ] feature request

**Current behavior:**

>  Slot increase delete add update when shadow is false does not work properly
1. When there is no slot inside Web Components, the user passes the slot to be present, but if shadow
  Is not displayed when true
  ```ts
  @Component({
    tag: 'pui-example',
    shadow: false,
  })
  export class Example implements ComponentInterface {
  render() {return (<Host></Host>)}
  }
  ```
  The wrong presentation flattened DOM:
  ```html
  <pui-example class="hydrated"><!----><div><span>outside default slot</span></div></pui-example>
  ```
  Expected behavior flattened DOM
  ```html
  <pui-example class="hydrated"><!----><div hidden=""><span>outside default slot</span></div></pui-example>
  ```
2. Initialize the hidden slot, pass it to the slot node when used, and display the slot inside the component with parameters that can correctly process the slot to its actual location. However, if the internal slot of the component is hidden here, the external slot node disappears and the external slot node should be moved back to its original position for hiding  

3. When you outsource an incoming slot and receive it internally, delete the outsourced slot. The hidden property of the internal slot-fb is not removed

**When shadow is true, the browser will handle it just fine**

**Expected behavior:**

Slot increase, delete and update when shadow is false will work normally

**Steps to reproduce:**

**Related code:**

  ```tsx
  import { Component, ComponentInterface, Prop, Host } from '@stencil/core';
  @Component({
    tag: 'pui-example',
    shadow: false,
  })
  export class Example implements ComponentInterface {
    /**
     * Whether or not to show slot
    */
    @Prop() enabled = false;
    render(): VNode {
      return (
        <Host>
          {this.enabled && (
            <div data-txt="slot wrapper">
              <slot>
                <span>fallback default slot</span>
              </slot>
            </div>
          )}
        </Host>
      );
    }
  }
  ```
  DOM nodes to use and display：
  ```html
  // Use the code(Vue) showSlot = true
  <pui-example>
    <div v-if="showSlot" data-txt="outside default slot">
      <span>outside default slot</span>
    </div>
  </pui-example>
  // flattened DOM: ====》Incoming slot should be hidden
  <pui-example class="hydrated">
    <!---->
    <div data-txt="outside default slot">
      <span>outside default slot</span>
    </div>
  </pui-example>
  ```
  When the following statement is executed：
  ```js
  document.getElementsByTagName('pui-example')[0].enabled = true
  ```
  This is normal as shown below
  ```html
  <pui-example class="hydrated"><!----><div data-txt="slot wrapper"><slot-fb hidden=""><span>fallback default slot</span></slot-fb><div data-txt="outside default slot"><span>outside default slot</span></div></div></pui-example>
  ```
   When the following statement is executed：
  ```js
  document.getElementsByTagName('pui-example')[0].enabled = false
  ```
  Incoming slot nodes are deleted： ====》**It should be moved back to its original position and then hidden**
  ```html
  <pui-example class="hydrated"><!----></pui-example>
  ```

**Other information:**

### LINKED ISSUE #2878: Problem with hide fallback slot
**Stencil version:**

```
 @stencil/core@2.5.2
```

**I'm submitting a:**

[x] bug report
[ ] feature request

**Current behavior:**

Fallback slot messes between parent and child components

**Expected behavior:**

Correctly handle fallbak slot issues for parent and child components

**Steps to reproduce:**

```tsx
// my-component

@Component({
  tag: 'my-component',
  styleUrl: 'my-component.css',
  shadow: false,
})
export class MyComponent {

  render() {
    return <div><slot><span>my-component fallback</span></slot></div>;
  }
}
// my-button

@Component({
  tag: 'my-button',
  styleUrl: 'my-button.css',
  shadow: false,
})
export class MyButton {

  render() {
    return <div>
      <my-component>
        <div data-txt="Insert into the default slot of my-component">
          <slot data-tip="lun-button The default slot of the lun-button"><span>LunButton fallback</span></slot>
        </div>
      </my-component>
    </div>;
  }
}
```
```html
<!DOCTYPE html>
<html dir="ltr" lang="en">
  <head>
    <meta charset="utf-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1.0, minimum-scale=1.0, maximum-scale=5.0" />

    <script type="module" src="/build/lun-button.esm.js"></script>
    <script nomodule src="/build/lun-button.js"></script>
  </head>
  <body>
    <!-- <my-component first="Stencil" last="'Don't call me a framework' JS"></my-component> -->
    <my-button></my-button>
  </body>
</html>
```
<div style="color:red;">The result of the actual rendering，The page is blank<div>  
```html
<my-button class="hydrated">
    <!---->
    <div>
      <my-component class="hydrated">
        <!---->
        <div>
          <slot-fb hidden=""><span>my-component fallback</span></slot-fb>
          <div data-txt="Insert into the default slot of my-component">
            <slot-fb data-tip="lun-button The default slot of the lun-button" hidden=""><span>LunButton fallback</span>
            </slot-fb>
          </div>
        </div>
      </my-component>
    </div>
  </my-button>
```  
The actual rendering hides the parent component's slot fallback as well  

Snipaste_2021-04-09_11-11-47

**Other information:**

1. I think there should be a boundary between the elements of the parent component and the elements of the child component, and should not be mixed
2. In fact, the problem of slot passthrough cannot be handled correctly
```jsx
<div><slot name"abc" slot="abc"></slot></div>
```

### LINKED ISSUE #2905: Vue.js hydrate
**Stencil version:**
```
 @stencil/core@2.5.2
```

**I'm submitting a:**
[x] bug report
[ ] feature request
[ ] support request

**Current behavior:**

**Expected behavior:**

**Steps to reproduce:**

1. Create a Stencil component with a `<slot />`.
2. Setup a Nuxt.js or Vue.js project with SSR.
3. Use the Stencil component in a Vue.js component and put some content in the slot.
4. SSR the Vue.js component.

**Related code:**

Stencil component
```tsx
import { Component, Host, h } from '@stencil/core';

@Component({
  tag: 'bs-slot-component',
  shadow: false,
  scoped: true,
})
export class SlotComponent {
  render() {
    return (
      <Host>
        <slot />
      </Host>
    );
  }
}
```

Vue.js component
```vue
<template>
  <bs-slot-component>
    <div>Content in slot</div>
  </bs-slot-component>
</template>
```

Nuxt.js module to hydrate Stencil components
```js
import { renderToString } from 'path/to/hydrate';

export default function () {
  this.nuxt.hook('render:route', async (_, page) => {
    const render = await renderToString(page.html);
    page.html = render.html;
  })
}
```

**Other information:**

image

### LINKED ISSUE #2937: Inconsistent behavior: slot-fb breaks styling of default slot content in component with `shadow: false`
**Stencil version:**
```
 @stencil/core@2.6.0
```

**I'm submitting a:**
[x] bug report
[ ] feature request

**Current behavior:**
Slot default content is rendered inside the component `slot-fb`, which may cause styling issues. For example, this breaks layout styling when the parent element defines a flex layout, since it will not affect the children of `slot-fb`. With `shadow: true` this problem doesn't occur.
This could also pose an issue on legacy browser that don't support the shadow DOM.

**Expected behavior:**
The component should look and behave the same with `shadow: true` and `shadow: false`. `slot-fb` shouldn't have an effect on the style/layout of the component.

**Steps to reproduce:**
If you use the component as follows, with `shadow: true` the layout is rendered correctly, as soon as `shadow: false, scoped: true` is set, the component is rendered incorrectly:
```HTML
  <my-component></my-component>
```

This works correctly in both cases, because the passed slot content is added as direct children instead of being wrapped in `slot-fb`:
```HTML
  <my-component>
    <div class="item">Passed #1</div>
    <div class="item">Passed #2</div>
    <div class="item">Passed #3</div>
  </my-component>
```

**Related code:**

```tsx
import { Host, Component, h } from '@stencil/core';

@Component({
  tag: 'my-component',
  styles: `
    :host {
      display: flex;
      gap: 0.5rem;
    }

    .item, ::slotted(.item) {
      border: 2px solid #ccc;
      padding: 0.25rem;
    }
  `,
  shadow: false,
  scoped: true,
})
export class MyComponent {
  render() {
    return (
      <Host>
        <slot>
          <div class="item">Default #1</div>
          <div class="item">Default #2</div>
          <div class="item">Default #3</div>
        </slot>
      </Host>
    );
  }
}
```

**Other information:**
Perhaps the slot default content should be rendered outside the slot component in the same way as passed slot content is handled, or this problem can be circumvented with a different component structure.

### LINKED ISSUE #2997: Nested slots mis-ordered when not using Shadow DOM
**Stencil version:**

```
 @stencil/core@2.0.1
```

**I'm submitting a:**

[x] bug report
[ ] feature request

**Current behavior:**

When a parent component's template contains a child component which it passes its slot content into, the slot content can end up mis-ordered compared to the child component's template. I believe the child component's props being based on the parent component's state may be related to the issue.

For example, in my use case "State: true" should come before "Hello" but it ends up backwards. (code examples below)

```
Hello

State: true
```

If the components re-render, or I turn on Shadow DOM, that fixes the issue (turning on Shadow DOM will not work for my real-world use case)
  
**Expected behavior:**

I would expect the order of the text above to be swapped:

```
State: true

Hello
```

**Steps to reproduce:**

Here is a demo and reproduction repo:

- [Demo](https://mystifying-borg-a6eba2.netlify.app/index.html)
- [Minimum reproducible test case repo](https://github.com/cloudfour/stencil-nested-slot-bug-repro)

**Related code:**

Here are two components I used for a minimum reproducible test case:

Parent component:

```tsx
import { Component,  h, Host } from '@stencil/core';

@Component({
  tag: 'my-component',
})
export class MyComponent {
  render() {
    return <Host>
      <div>
        <nested-component state={true}>
          <slot />
        </nested-component>
      </div>
    </Host>;
  }
}
```

Nested Component

```tsx
import { Component, Host, Prop, h } from '@stencil/core';

@Component({
  tag: 'nested-component',
})
export class NestedComponent {
  @Prop() state: boolean;

  render() {
    return (
      <Host>
        <div>State: {this.state.toString()}</div>
        <slot />
      </Host>
    );
  }
}
```

Here's how I'm using this component:

```html
<my-component>
  <p>Hello</p>
</my-component>
```

**Other information:**

#2232 sounds related, but not quite the same. That issue deals with issues on re-renders. My issue is with the initial render.

### LINKED ISSUE #3278: bug: Failed to execute 'removeChild' on 'Node'
### Prerequisites

### Stencil Version

2.14.1

### Current Behavior

I created a simple web component without parameters but with a slot inside using stencil. This component is bundled using the @stencil/react-output-target package (0.3.1).

```js
@Component({
  tag: 'tds-grid',
})
export class Grid {
render() {
  return (<div>
    <slot />
  </div>
);
```
When the children of the component change dynamically, we got the  Failed to execute 'removeChild' on 'Node' error.

This issue has been listed in the ionic-framework projet and is reproductible with stencil.
https://github.com/ionic-team/ionic-framework/issues/18782

### Expected Behavior

=> should not generate a fatal error and content must be rerender without crashing.

### Steps to Reproduce

1. Create a web component as describe in the current behaviour section
2. Generate this component for react using the @stencil/react-output-target package
3. Create a react application, import the  generated react component and try to update the content of this component dynamically as it is described in the ionic framework issue https://github.com/ionic-team/ionic-framework/issues/18782

### Code Reproduction URL

this code is easily reproductible => https://github.com/ionic-team/ionic-framework/issues/18782

### Additional Information

_No response_

### LINKED ISSUE #3413: bug: client side re-hydration of slotted elements leads to the element being rendered twice
### Prerequisites

### Stencil Version

2.16.1

### Current Behavior

I am trying to setup an express server, which recieves an html string from a proxied server, prerenders the components on it and returns the result to the client.
The problem is, that some components don't seem to get re-hydrated correctly on the client. More specifically it seems to not recognize components, which are slotted inside other components.

Having the following code structure:
```
<app-section>
      <app-banner>...</app-banner>
</app-section>
```
Leads to the `app-banner` appearing twice on the page. Both of them look broken in another way:
broken
 
Removing the `app-section` component around the `app-banner` leads to the banner being re-hydrated correctly:
ok

The section isn't doing anything fancy. It just renders its slotted content with a wrapping element.

### Expected Behavior

Slotted components should only appear once on the re-hydrated page.

### Steps to Reproduce

- Create a new Component which renders its slotted content with a wrapping `Host` Element.
- Create another component which will be nested (slotted) inside the first component.
- Render yet another component inside the `render()` function of the second component.
- Prerender the Components on the server using the `renderToString` function.
- Look at the source-code of the re-hydrated page to see, that the component appears twice on the page.

### Code Reproduction URL

https://github.com/eWert-Online/stencil-ssr-hydrate-reproduction

### Additional Information

- It has to be an hydration issue, because if you don't pre-render it on the server, everything works fine. 
- If you disable JavaScript in the browser it also looks fine. The Problem is with the re-hydration of the rendered markup on the client, not with the pre-rendering on the server.
- The issue does not go away if you wrap the app-section `<slot />` in another element (e.g. `<div><slot /></div>`)

### LINKED ISSUE #3593: bug: component content duplicated (with hydrate app on a SSG)
### Prerequisites

### Stencil Version

2.13.0

### Current Behavior

Hi, greetings.
I am using stencil with **[hydrate app output target](https://stenciljs.com/docs/hydrate-app)**, since I need to use my library of components with a static site generator (Eleventy). On stenciljs everything works fine, but on eleventy, I am experiencing a buggy behavior with a specific component: **gwds-button**. This component prints a simple a tag along with some css classes: 

image

On eleventy, I am using another component (**gwds-loot-at-me**), which has as slotted content a gwds-button component (along with other content such as a paragraph). Here is where the strangeness begins: two link elements gets printed inside gwds-button, where just one is expected to be display.

image

Now, if I place a gwds-button **outside** of another component, the button behaives as expected, tag duplication does not ocurrs. The duplication only seems to happen when gwds-button is nested inside another gwds-component.

### Expected Behavior

gwds-component to display only one a tag.

### Steps to Reproduce

1. clone the repo
2. npm install
3. npm start
4. open the site
5. see the first big component with green background. you will see two pink buttons, and two white buttons, and you should see one pink and one white only.
6. To check the markup, go to src/en/index.html

### Code Reproduction URL

https://github.com/bsastregx/gwds-eleventy

### Additional Information

May be this helps, not sure: https://github.com/ionic-team/stencil/issues/2611 | https://github.com/ionic-team/stencil/issues/3413

### LINKED ISSUE #3913: bug: slot element loses its parent reference and disappears when its parent is rendered conditionally
### Prerequisites

### Stencil Version

2.13.0

### Current Behavior

When I try to render the default slot within the conditionally rendered parent element in the **no-shadow** component and the parent element is changed, the slot relocation algorithm loses the parent element reference and the slot element is not relocated (re-rendered) into the new parent element.

### Expected Behavior

Slot element should be relocated to the new parent element.

### System Info

_No response_

### Steps to Reproduce

In the reproduction project:
 1. you will see the typography element is rendered correctly and the slot will be displayed within the "span" element in the "app-typography" component DOM structure
 2. when the user clicks on the "Toggle Element type" button, the handler function finds the typography element and changes its "element" attribute (which controls the rendered element tag of the slot's parent)
 3. then you will see the "span" element turns to the "p" element within the "app-typography" component DOM structure and the slot disappears (not relocated)

### Code Reproduction URL

https://github.com/yigityuce/stencil-conditional-rendering-issue/commits/master

### Additional Information

The issue might be here in the renderer (the old and new node tags checking)
https://github.com/ionic-team/stencil/blob/main/src/runtime/vdom/vdom-render.ts#L153

The goal is to follow the accessibility rules and the sementic web. I want to render the hX tags (h1, h2, ..., h6) correctly regarding the typography as Material UI does. (https://mui.com/material-ui/react-typography/#changing-the-semantic-element)

And the inspiration comes from the ionic-button component: https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/button/button.tsx#L302

### LINKED ISSUE #3977: bug: Component markup disappears when overwriting default slot using text
### Prerequisites

### Stencil Version

3.0.0-beta.1

### Current Behavior

When the following component is defined:

```ts

import { h, Component, ComponentInterface, Element, Host } from '@stencil/core';

@Component({
  tag: 'custom-element-button',
  styleUrl: 'button.scss',
  shadow: false,
  scoped: false,
})
export class Button implements ComponentInterface {
  @Element() host;

  render(): HTMLElement {
    return (
      <Host class="test-button">
        <button>
          <div class="content-wrapper">
            <span class="prefix">
              <slot name="prefix" />
            </span>
            <slot />
            <slot name="suffix" />
          </div>
        </button>
      </Host>
    );
  }
}
```

When modifying default slot content using javascript:

```html
<custom-element-button>Button</custom-element-button>
```
```js
document.querySelector('custom-element-button').innerText = "newButtonText"
```

All the markup inside disappears and only the text remains.

### Expected Behavior

Same as with shadow DOM

### System Info

```shell
N/A

not relevant, happening in multiple OS's, browsers, and node versions
```

### Steps to Reproduce

Download the linked repository, install, start the project, and type into the input element

### Code Reproduction URL

https://github.com/oscargm/stencil-slot-problem

### Additional Information

In the reporduction repository you'll find a demo using components with:

- shadowDOM (works fine)
- scopedCSS without shadowDOM (issue reproducible)
- neither scopedCSS or shadowDOM (issue reproducible)

If this is the intended behaviour I'd like to know how to avoid this issue.

I found out that adding a span like htis:

```html
<scoped-css-button class="button with-children-element">
  <span>Button</span>
</scoped-css-button>
```

and changing the text of the span everything works fine, or even with the prefix/suffix named slots changing the textContent or innerText it works fine.

### LINKED ISSUE #4284: bug: Slot content went missing within dynamic component
### Prerequisites

### Stencil Version

3.2.1

### Current Behavior

I have a Stencil web component with an `<a>` tag element that needs to be changed to `<p>` tag to prevent the user from clicking it.
The dynamic component is working, however I noticed that the content that goes to the <slot /> went missing.

I have a simple component code:
```tsx
import { Component, State, h } from '@stencil/core';

@Component({
  tag: 'my-component',
  styleUrl: 'my-component.css',
  shadow: true,
})
export class MyComponent {
  @State() customTag: string = 'a';

  componentDidLoad() {
    setTimeout(() => {
      this.customTag = 'p';
      alert("I'm changed!");
    }, 3000);
  }

  render() {
    return (
      <this.customTag href="#">
        <slot />
      </this.customTag>
    );
  }
}
```

In my `index.html`:
```html
<body>
  <my-component>This is a sample link</my-component>
</body>
```

It shows the slot content initially, then I tried to change the element dynamically using timeout after 3 seconds.
Then the slot content went missing as show in GIF below:

error-1

### Expected Behavior

We have the expectation that the slot content still persist after the component tag is changed after 3s:

expected

### System Info

```shell
Node version: v18.16.0
Computer: MacOS (Ventura 13.2.1)
Browser: Tried on both Firefox (12.0.1 (64-bit)) and Chrome (112.0.5615.49)
```

### Steps to Reproduce

To replicate it, I did it with Stencil Component Starter.
Please refer to the link below for reproduction.

### Code Reproduction URL

https://github.com/AaronKow/stencil-slot-missing-issue

### Additional Information

_No response_

### LINKED ISSUE #4523: bug: A slot gets the attribute hidden when it shouldn't
### Prerequisites

### Stencil Version

4.0.0 (also seen on 3.4.0)

### Current Behavior

StencilJS is adding the attribute hidden to a slot that it shouldn't. I wrote a repro accessible at the url below. There's quite a bit going on, but if I tried to simplify it further, the problem went away, so this is as minimal as I could make it. Basically I have three components:
```
text-input
  input-base
    input-group
```

image

For some reason, StencilJS adds the `hidden` attribute to the "message" slot in the `input-base` component.
I noticed that if I use `input-base` directly, it works as expected. It also works as expected if I add a wrapper div in `input-base`.

### Expected Behavior

The slot should not be hidden.

### System Info

_No response_

### Steps to Reproduce

Clone the repo, `yarn && yarn start`.

### Code Reproduction URL

https://github.com/asik/stencil-repro

### Additional Information

_No response_

### LINKED ISSUE #4525: bug: Missing slotted content when slot is forwarded from a non-shadow component
### Prerequisites

### Stencil Version

4.0.0 (also seen in 3.4.0)

### Current Behavior

I'm not sure exactly what the problem is here, but basically I have a component with this structure:

```jsx
@Component({
  tag: 'non-shadow-host',
  shadow: false,
})
export class NonShadowHost {
  render() {
    return (
      <div>
        <drop-down>
          <slot name="main-content" slot="main-content" />
          <div slot="dropdown-content-element">
            <drop-down-content>
              <slot name="dropdown-content" />
            </drop-down-content>
          </div>
        </drop-down>
      </div>
    );
  }
}
```

(the name drop-down is meaningless, I stripped out all code related to that). If I set `shadow: true`, this works, if I remove it, the slotted content in "dropdown-content" doesn't render. In the following screenshot, the selected node should get slotted but doesn't, and doesn't show up in the rendered HTML.

image

I made a copy of this component, changed the name for shadow-host and set shadow to true, and now it renders as expected.

image

image

I'm not sure if I'm trying to do something unsupported here, but I didn't see any warnings in the documentation about interop between shadow and non-shadow components.

### Expected Behavior

I would expect both components to end up rendering the same thing on screen.

### System Info

_No response_

### Steps to Reproduce

Clone the linked repo
`yarn && yarn start`
Check out example 2 labeled Missing slotted content bug in index.html.

### Code Reproduction URL

https://github.com/asik/stencil-repro

### Additional Information

_No response_

johnjenkins:
This PR looks to fix a number of issues relating to slots and how they are managed when a component is **not** `shadow: true` OR if the browser in question is using the `shadowDomShim`. 
Hence forth, I'll refer to these as 'non-shadow' components.

At present in order for non-shadow components to have similar behaviour to shadow-components (when it comes to the dom api) - stencil provides 2/3 opt-in `extras`: appendChildSlotFix, cloneNodeFix and slotChildNodesFix (https://stenciljs.com/docs/config-extras). 

'appendChildSlotFix' - when you write `myComponent.appendChild(newElement);`, the new element will arrive in an appropriate slot - not just be appended to the bottom of your non-shadow component.

'slotChildNodesFix' - when you write `myComponent.childNodes` | `myComponent.children` only the slotted nodes / elements - not all nodes in the non-shadow component dom - are returned.

I feel quite strongly, that this ^ makes little sense. 
Until very recently I 
1) assumed that `shadow: true` On an old browser would automatically try to mimic shadow components. 
2) assumed that `shadow: false` / `scoped: true` would not behave so differently regarding native APIs. 

I'm pretty sure a number of bugs reported are the result of simply not having these `extras` applied.

Additionally, at present, applying these extras will also (totally unnecessarily) apply them to `shadow: true` components on modern browsers! 

**This PR**

This PR applies both aforementioned patches automatically on all non-shadow components and also adds new patches for nearly all DOM methods & accessors relating to the addition of new nodes: 
`innerHTML`, `innerText`, `textContent`, `appendChild()`, `append()`, `prepend()`, `insertAdjacentText()`, `insertAdjacentElement()`, `insertAdjacentHTML()`, `replaceChildren()` 

With these, I believe non-shadow components feel more similar to shadow components. 
I also believe many of the kinks related to non-shadow components within frameworks (vue / react et al) will go away - I'll do more testing to confirm this. The vdom renderers of these libraries will add nodes via these methods when all is said and done. 

All 'native' methods / accessors are accessible by simply prepending '__' to the relevant thing: i.e. `myComponent.__innerHTML` or `myComponent.__appendChild(newThing)`; think of it as the equivalent of `myShadowComponent.shadowRoot.innerHTML`.
I did think about patching a mock `myComponent.shadowRoot` but it felt over-engineered and silly.

`cloneNodeFix` is still not automatically applied as I thought this was a more niche requirement(?). However now it will only apply to non-shadow components.

I've added a deprecated flag to both the 'appendChildSlotFix' and 'slotChildNodesFix' extras within stencil.config. Continuing to include them will not break anything.

Lastly, I've made a small fix to stencil's vdom-renderer. 
Here's a silly example:

```
  SlotWrapper = () => {
    if (this.aBoolean) {
      return <span><slot name="bool" /></span>
    } else return <div><slot name="bool" /></div>
  }
  render() {
    return 
      <div>
        <this.SlotWrapper />
      </div>;
  }
  ```

At present, within a non-shadow component whenever `this.aBoolean` changes, the`<slot/>` is deemed to have been removed and so all it's currently slotted children are also removed.
 
This PR also fixes that.

Fixes #2004
Fixes #2641
Fixes #1968
Fixes #1997
Fixes #2801 
Fixes #2257 
Fixes #3913
Fixes #3977
Fixes #4284
Fixes #4523
Fixes #4525
Potentially fixes #2259 
Potentially fixes #2036
Heavily referenced #2012

I’ve done quite a bit of testing in IE11, Edge 18 and all modern browsers. All current tests within stencil (spec / end-to-end via jest and karma) continue to pass.

### New commit!

Commit https://github.com/ionic-team/stencil/pull/2938/commits/f9a00632919d39886c3c950feb2c942f74ff9c8f sees an overhaul of non-shadow component fallback content.

(I did think about putting this in a separate PR, but the issues and requirements were very closely linked (and difficult to untangle) so it made more sense to add it here).

**What is the current behavior?**
In non-shadow components, slot fallback content behaviour is re-created by wrapping all fallback nodes in an `<slot-fb>` element. 
Wrapping nodes can create issues regarding css presentation (e.g. flex children)

**What is the new behavior?**
In non-shadow components,  slot fallback content nodes are no longer wrapped. 
Instead, in the same way that non-shadow components store slot meta in an empty text node, this same text node now stores data regarding fallback content.

**Does this introduce a breaking change?**
Potentially - if anyone was using the `slot-fb` for reference or styling, that will now not work. 

**Testing**
All karma & Jest tests have been updated to reflect the behaviour / lack of wrapping element.
I have also manually tested changes in an external, linked project.
Also, there was a lack of tests within Stencil for nested slots (e.g. `<slot name="content"> <slot></slot> <slot name="after"></slot> </slot>`). I've added new spec tests to cover this.

**Additionally**
I have also now patched all, `node.remove()` and `parentNode.removeChild()` on non-shadow slotted nodes - when there is slot fallback content - to assess whether fallback content should show or hide.
I have also tested and amended my changes against open issues related to slot fallback content so have marked them as fixes here.

Fixes #2937
Fixes #2878 
Fixes #2997 
Fixes #3278
Potentially fixes #2877

### New commits!

New commits address client side hydration (after using something like `renderToString()`).

Fixes #3413
Fixes #3593 
Fixes #2905
Fixes #2366
Fixes #2036
Fixes #1993

Whilst investigating, a number of these issues were related to `<slot />`s and how they are managed.
Additionally my changes around slot fallback content weren’t working with server / client hydration.
Finally, I found when working with frameworks (like nuxt) they would attempt to resolve their vdom against the real DOM after SSR by stepping through it with accessors like ‘nextSibling’. This fails for non-shadow elements so nuxt duplicates the nodes it thinks it cannot find.
To mitigate this I’ve patched these accessors just for non-shadow, client hydrated, slotted nodes.

A number of these bugs are demoed - https://stackblitz.com/edit/stencil-start-yoemx6
And here's the same demos with this PR applied - https://stackblitz.com/edit/stencil-start-brthw2

### CODE REVIEW

**Comment by eriklharper** [comment]
@johnjenkins I tried installing your temp package on my repro setup for #2801 but after I `npm i` the `npm start` command instantly fails.  Do I need to purge the npm cache or remove node_modules and reinstall everything that way?

image

**Comment by eriklharper** [comment]
Mac Catalina 10.15.7

**Comment by eriklharper** [comment]
Tried `rm -rf node_modules && npm cache clean --force && npm i` but it doesn't create even a `node_modules` folder.

**Comment by eriklharper** [comment]
I'm still seeing the issue with your fix applied.  Here's my repro case if you want to try this out yourself, its just a `stencil-component-starter` repo so you just install your package of stencil core and run `npm start`.  Here's what the bug looks like:

image

**Comment by eriklharper** [comment]
Thanks for looking into this @johnjenkins !  This has long been a thorny issue and we're currently trying to work around it by ditching scoped components altogether, but if you can come up with a fix that will be splendid!

**Comment by eriklharper** [comment]
💥  This is great @johnjenkins !  It's working for me in my repro setup! I see that the comment node appears just like the other use case that uses a wrapping element around the text.

image

**Comment by bitflower** [comment]
@eriklharper SO this is the working version?

`"@stencil/core": "npm:@johnjenkins/stencil-community@^2.4.1"`

**Comment by PaulRaschke** [comment]
Hi @johnjenkins. Many thanks for your work! It does indeed fix [the example component I posted in my issue](https://github.com/ionic-team/stencil/issues/2937). I also think your additional added features are really valuable. I'm currently using your distribution with `"@stencil/core": "npm:@johnjenkins/stencil-community"`.

But unfortunately I discovered 2 new issues introduced by your changes:

- Following up on the example component in my issue, the following problem arises once a display style is set for the items (other than `display: none`, which the browser defaults to for elements with the attribute `hidden`): the items that are used as default slot content are rendered even if slot content is passed to the component ([same behavior as reported in this issue](https://github.com/ionic-team/stencil/issues/2760#issuecomment-739004091)). The cause seems to be that the elements are no longer wrapped inside a hidden `slot-fb` element. As far as I know, the only way I can solve this is to globally force elements with the attribute `hidden` to be visually hidden via css.

```diff
import { Host, Component, h } from '@stencil/core';

@Component({
  tag: 'my-component',
  styles: `
    :host {
      display: flex;
      gap: 0.5rem;
    }

    .item, ::slotted(.item) {
+     display: block;
      border: 2px solid #ccc;
      padding: 0.25rem;
    }
  `,
  shadow: false,
  scoped: true,
})
export class MyComponent {
  render() {
    return (
      <Host>
        <slot>
          <div class="item">Default #1</div>
          <div class="item">Default #2</div>
          <div class="item">Default #3</div>
        </slot>
      </Host>
    );
  }
}
```

- Component structures with nested slots are broken. This component is rendered incorrectly and throws errors with your package:

```tsx
import { Component, Host, h } from "@stencil/core";

@Component({
  tag: "my-component",
  styles: `
    :host {
      display: block;
    }
  `,
  scoped: true,
})
export class MyComponent {
  render() {
    return (
      <Host>
        <slot name="content">
          <slot name="before">DEFAULT BEFORE</slot>
          <slot> DEFAULT CONTENT </slot>
          <slot name="after">DEFAULT AFTER</slot>
        </slot>
      </Host>
    );
  }
}
```

This component yields the following results:

```HTML
<my-component> CONTENT </my-component>
```

→ Only `CONTENT` is rendered instead of `DEFAULT BEFORE CONTENT DEFAULT AFTER` and the following error is thrown `TypeError: parentElm.classList is undefined`

```HTML
<my-component>
  <span slot="before">BEFORE</span>
  CONTENT
  <span slot="after">AFTER</span>
</my-component>
```

→ Correctly renders as `BEFORE CONTENT AFTER`, but throws the same error

```HTML
<my-component></my-component>
```

→ doesn't render at all instead of rendering as `DEFAULT BEFORE DEFAULT CONTENT DEFAULT AFTER` and throws the same error

Changing `scoped: true` to `scoped: false` results in the following behavior:

```HTML
<my-component> CONTENT </my-component>
```

→ Correctly renders as `CONTENT`, but throws `DOMException: Node.appendChild: Cannot add children to a Text`

```HTML
<my-component></my-component>
```

→ Doesn't render at all and throws the same error

None of these issues occur with `@stencil/core@2.6.0`. With `shadow: true`, everything works as expected, both with your distribution and with the official distribution.

Again, many thanks for your work. It would be great if you could take a look at these problems.

(Tested with Firefox 89.0.2 and Chrome 91.0.4472.124)
(ERROR messages as reported by Firefox)

eriklharper:
> @eriklharper SO this is the working version?
> 
> `"@stencil/core": "npm:@johnjenkins/stencil-community@^2.4.1"`

Yes indeedly doodly

PaulRaschke:
@johnjenkins

> I feel like you have more options with the items not being wrapped though

Yes, there are a few different options to solve this. Nevertheless, I think to solve this optimally, something has to be changed at compiler/runtime level, because this is probably the only way to achieve consistency with `shadow: true` and `shadow: false` and thus between browsers without the stencil user having to take action himself. I'd like to help, but I don't think I understand the stencil core and your changes sufficiently well yet. But I will certainly take a deeper look at it!
After a quick look, are there any objections to adding something in the form of:

```ts
if (hide === true) n.style.display = "none";
```

to [renderSlotFallbackContent](https://github.com/johnjenkins/stencil/blob/e5640649886e76810c3c523af2d2e542038f6624/src/runtime/vdom/render-slot-fallback.ts#L11)?

I'm not sure what the reason is for inserting the fallback slot content into the DOM in the first place?

Thanks for working on a fix for the nested slot bug!

PaulRaschke:
@johnjenkins
Happy to help :D

Thanks for the insights regarding the vDOM/DOM rendering.

Everything seems to be working flawlessly with the new version 👍 
Thanks, nice work!

vmcbaptista:
I was also facing some issues with slots when shadow: false with latest @stencil/core. Specifically, fallback content being always hidden when no content is passed into the slot.
Tried today the latest version of johnjenkins/stencil-community and it works perfectly.
Great job @johnjenkins!

@ltm I see that you have been reviewing the latest merged pull requests, would be possible to indicate any ETA for when this could be reviewed, merged and released officially on @stencil/core package?

vmcbaptista:
@johnjenkins I identified an issue with the types generated when using `johnjenkins/stencil-community`.

I have some `@Prop` with type provided by an external package.
The generated types is generating with imports like this
`import {type} from "relative_path_to_node_modules/name_of_package_containing_type/file_defining_type.d"

With @stencil/core it imports without any reference to node_modules so like this
`import {type} from "name_of_package_containing_type"

I tried several versions of your package, and all seem to have this issue.

vmcbaptista:
@johnjenkins 
I've quickly done a minimum repo https://github.com/vmcbaptista/stencil-types-issue/

This commit shows the difference that happens to the auto generated typings after updating to your version
https://github.com/vmcbaptista/stencil-types-issue/commit/beed97cefd1bea412fb7ae67ce92c117c2d85cbc

Although, in my real-life scenario the dependency is other than uuid, the same thing happens, so I guess you can rely on this 😄

vmcbaptista:
@johnjenkins it's working fine. thanks 😄

PaulRaschke:
Hi @johnjenkins, I've been using your `stencil-community` build quite extensively over the last few weeks and almost everything seems to be working fine!

Unfortunately, I've discovered another bug. The following component throws a `TypeError: fbNodes[fbSlot['s-sn']] is undefined` error when the `text` prop is updated. The error doesn't seem to completely break the component, as the `text` prop continues to be updated correctly. The same behavior is observed when either a mutable prop is updated from inside the component or a mutable or non-mutable prop is updated from outside. The bug isn't directly caused by the prop update, but instead occurs during the component update or rendering, as it also happens when updating the component with `forceUpdate(componentRef)`.

```tsx
import { Component, Host, h, Prop } from '@stencil/core';

@Component({
  tag: 'my-component',
})
export class MyComponent {
  @Prop({ mutable: true }) text: string = '';

  componentDidLoad() {
    this.text = 'Text';
  }

  render() {
    return (
      <Host>
        <slot name="content">
          <slot name="text">{this.text}</slot>
        </slot>
      </Host>
    );
  }
}
```

This bug doesn't occur with `@stencil/core@2.6.0`.

(Tested with Firefox 90.0.2 and Chrome 92.0.4515.131 with `@stencil/core@npm:@johnjenkins/stencil-community@^2.5.4`)
(ERROR messages as reported by Firefox)

PaulRaschke:
@johnjenkins - `@stencil/core@npm:@johnjenkins/stencil-community@2.5.5` fixes the issue :tada:
Strange that you couldn't reproduce the bug with `v2.5.4`. I cleaned any cache and did a clean install, but the issue persisted. If you're still interested in getting to the root of the problem, and the problem doesn't occur when you set the version to `2.5.4`, I can set up a repro repo.
BTW, I use Windows, though I don't think there could be a correlation with the problem.

I hope this PR will be reviewed and merged soon :)

Paul-Hebert:
This is great @johnjenkins ! This appears to fix the issue I reported in #2997! 

It looks like I'm still running into some issues related to #2232 but those are likely outside the scope of this PR. 

Nice work! Thanks!

(I have not tested extensively in my main repo yet.)

PaulRaschke:
Hi @johnjenkins, I think I've discovered a bug in your latest stencil-community release (`@stencil/core@npm:@johnjenkins/stencil-community@2.5.8`), which under certain circumstances also occurs with `@stencil/core@npm:@johnjenkins/stencil-community@2.5.5`:

With these two components:
```ts
import { Component, Host, h } from '@stencil/core';

@Component({
  tag: 'my-component',
})
export class MyComponent {
  render() {
    return (
      <Host>
        <my-nested-component>
          <slot name="before"></slot>
          <slot />
          <slot name="after"></slot>
        </my-nested-component>
      </Host>
    );
  }
}
```
```ts
import { Component, Host, h } from '@stencil/core';

@Component({
  tag: 'my-nested-component',
})
export class MyNestedComponent {
  render() {
    return (
      <Host>
        {/*
          With the slot wrapped in a div, the contents of the "my-component" slot
          are rendered in the correct order with v2.5.5
        */}
        <div>
          <slot />
        </div>
      </Host>
    );
  }
}
```
using them like this:
```html
<my-component>
  <div slot="before">BEFORE CONTENT</div>
  <div>CONTENT</div>
  <div slot="after">AFTER CONTENT</div>
</my-component>
```
the slots are rendered in the wrong order. As written in my comment, this issue can be "fixed" by wrapping the default slot of the `my-nested-component` in an HTML element when using `@stencil/core@npm:@johnjenkins/stencil-community@2.5.5`. This "fix" doesn't work with `@stencil/core@npm:@johnjenkins/stencil-community@2.5.8`.

(Tested with Firefox 92.0 and Chrome 93.0.4577.82)

PaulRaschke:
@johnjenkins - `@stencil/core@npm:@johnjenkins/stencil-community@2.5.9` seems to fix the issue :)
Just one issue with the latest version:
```bash
[ ERROR ]  TypeScript: ./node_modules/@stencil/core/internal/stencil-private.d.ts:3:51
           Cannot find module 'rollup' or its corresponding type declarations.

      L2:  import type { ComponentInterface, ListenOptions, ListenTargetOptions, VNode, VNodeData } from './stencil-public-runtime';
      L3:  import type { SourceMap as RollupSourceMap } from 'rollup';
      L4:  export interface SourceMap {

[36:56.0]  build failed, watching for changes... in 2.43 s
```
Everything works as expected when `rollup` is manually added as a dependency.

Thanks again for your awesome PR!

vmcbaptista:
@rwaskiewicz any estimation about when all those fixes would be merged? in my company we are currently relying on the packages of @johnjenkins because of these

joberthel:
@johnjenkins 

Using your PR together with the hydrate feature (SSR), I have the problem, that elements with slots are added multiple times to the DOM. If I find the time, I will create a small example.

danielleroux:
Any updates or timeline when the PR gets merged / released?

rwaskiewicz:
@danielleroux While I can't give an exact timeline, fixing up slot behavior is 100% something we're going to be working on this year.

phhbr:
Is there anything one can do to get this one merged?

jeco123:
Hi guys,

This fix is really important for us :-). When do you think it could be merged ?

Regards,

Jérôme

rwaskiewicz:
Hey folks,

The team will be working on slot-related fixes early this year, following the release of Stencil v3. I don't have a definitive timeline at the moment, but this area is on our radar as something to tackle early this year.

CollinHerber:
What is the status of this? I see the bug I've been waiting on for a long time get fixed two weeks ago in this PR.

rwaskiewicz:
@CollinHerber 
The status of this PR [is the same](https://github.com/ionic-team/stencil/pull/2938#issuecomment-1378794718):

> The team will be working on slot-related fixes early this year, following the release of Stencil v3. I don't have a definitive timeline at the moment, but this area is on our radar as something to tackle early this year.

JessicaSachs:
I cut a release from this for Alpha testing `npm install @stencil/core@3.3.0-dev.1685496483.a311f36 -S` - we are focused on shipping v4 of Stencil right now, but there are one or two folks that are willing to help us test out this branch to work around issues they're currently having.

Post-v4 we'll continue forward with a plan for how to handle this PR but any manual testing completed would be very helpful.

mbulfair:
I can confirm this work fixes our use case.

Observations:

1. When using nextjs some of the `<styles>` are missing, specifically around scoped components, this is due to how nextjs only inserting what it knows. @johnjenkins provided a custom server example that places the styles at the top of the `body` which works.
2. Typescript error in `stencil.config.ts`

image

choshin314:
Hi just a heads up, I brought in `@stencil/core@3.3.0-dev.1685496483.a311f36` for testing in our Nuxt 3 app.  Unfortunately for **non-shadow DOM** components w/ slots I am still seeing the same behavior as before, i.e. the server response appears correct but the custom element drops its children (and replaces w/ the ’slotted’ items only) after the vue app takes over.  On the plus side, **shadow DOM-enabled** slotted components do seem to work as expected with the if wrapping the slot items in a single root element (on the Nuxt consumer side), but I believe that also works in v3.4 so maybe not relevant to this. 

I should also note that in that Nuxt test app, we're using the plugin per the instructions for the vue output target, as well as the Stencil hydrate `renderToString` in a Nuxt server plugin.  

```
/*
/server/plugins/components-ssr.js
*/

import { defineNitroPlugin } from 'nitropack/dist/runtime/plugin'
import { renderToString } from  'path/to/hydrate'

export default defineNitroPlugin((nitroApp) => {
  nitroApp.hooks.hook('render:response', async (response) => {
    const render = await renderToString(response.body);
    response.body = render.html
  })
});


/* 
/plugins/components.js
*/

import { ComponentLibrary } from 'our/vue/package’

export default defineNuxtPlugin(nuxtApp => {
  nuxtApp.vueApp.use(ComponentLibrary)
})

/* 
Plugin source from our/vue/package
*/

import { applyPolyfills, defineCustomElements } from ‘path/to/loader';
export const ComponentLibrary = {
    async install() {
        applyPolyfills().then(() => {
            defineCustomElements();
        });
    },
};
```
Anyway, hopefully this is helpful information.  Thanks for all the work you're putting into this 🙏 🙌

choshin314:
> Hi just a heads up, I brought in `@stencil/core@3.3.0-dev.1685496483.a311f36` for testing in our Nuxt 3 app. Unfortunately for **non-shadow DOM** components w/ slots I am still seeing the same behavior as before, i.e. the ***

This was a false alarm on my part – as I was working on getting a minimum reproduction together I realized that the slotted component that I was testing had invalid html structure.  After reworking that, was able to confirm this works as expected.  Thanks again, and sorry for the distraction!

pedroresende:
I've tested this and it seems to work, except for the Hydration errors that started to appear on Next.js

pedroresende:
The only way, I've found to overcome the hydration problem was by adding a useEffect

```tsx
  const [loaded, setLoaded] = useState<boolean>(false)

  useEffect(() => {
    if (!loaded) {
      setLoaded(true)
    }
  }, [loaded])
```

pedroresende:
@JessicaSachs Since the release of version 4 was already done, could you guys take this PR into account 🙏🏻

phhbr:
Is there already an ETA for this bugfix?

antonselukh:
> Is there already an ETA for this bugfix?

i also want to know.

pedroresende:
> > Is there already an ETA for this bugfix?
> 
> i also want to know.

+1

eWert-Online:
Hello 👋 

I tested the released alpha version with our app.
The original issue we has (duplicate elements after hydration) is fixed. There are however two new issues introduced with this PR:

1. We are getting a `TypeError: e.split is not a function` when rendering some elements. The problem originates in the `setAccessor` function when calling `parseClassList`. You are expecting a string to be given to `parseClassList`, however sometimes there are [SVGAnimatedString](https://developer.mozilla.org/en-US/docs/Web/API/SVGAnimatedString)s passed to it, which leads to the error being thrown.
2. There seems to be some kind of race-condition when accessing the shadow-dom inside `componentDidLoad`.
    We are having the following code inside one of our components: 
    ```
   componentDidLoad() {
     const els = this.el.shadowRoot.querySelector("slot").assignedElements({ flatten: true });
     ...
   }
    ```
    Sometimes the `querySelector("slot")` is returning an element and sometimes it returns undefined.
    This race-condition seems to only happen with a large amount of stencil components.

I have updated my reproduction repository with this behavior: https://github.com/eWert-Online/stencil-ssr-hydrate-reproduction/tree/alpha-test 

You can also see the resulting behavior in this video:

[reproduction-video.webm](https://github.com/ionic-team/stencil/assets/7459510/89ebb443-4544-4ec0-9ec4-abe7cd0a3102)

mayerraphael:
> Hello 👋
> 
> I tested the released alpha version with our app. The original issue we had (duplicate elements after hydration) is fixed. There are however two new issues introduced with this PR:
> 
> 1. We are getting a `TypeError: e.split is not a function` when rendering some elements. The problem originates in the `setAccessor` function when calling `parseClassList`. You are expecting a string to be given to `parseClassList`, however sometimes there are [SVGAnimatedString](https://developer.mozilla.org/en-US/docs/Web/API/SVGAnimatedString)s passed to it, which leads to the error being thrown.

We are using johnjenkins/stencil-community@2.23.0 and can confirm that the e.split error with SVGs exists with this PR.
Created a bug issue some time ago: https://github.com/ionic-team/stencil/issues/4278

We worked around it by using innerHTML to set the svg.

mayerraphael:
@johnjenkins I can only reproduce the error with this PR. It seems to be related. I updated #4278 and linked a repo for reproduction.

rwaskiewicz:
Hey folks 👋 

With today's release of Stencil v4.1.0, we introduced a new flag called [`experimentalSlotFixes`](https://stenciljs.com/docs/config-extras#experimentalslotfixes). Currently, the only flag enables Stencil's pre-existing slot-related fix configuration flags. Going forward, the Stencil team is going to be working with @johnjenkins  to incrementally add the fixes in this PR into Stencil under `experimentalSlotFixes` in subsequent releases 🚀

rwaskiewicz:
The following bug(s) have fixes included in[ today's v4.2.0 release:](https://github.com/ionic-team/stencil/releases/tag/v4.2.0)
- https://github.com/ionic-team/stencil/issues/2641

rwaskiewicz:
@pedroresende 

https://github.com/ionic-team/stencil/pull/2938#issuecomment-1735648023

Can you please open a new issue with a minimal reproduction case for the team to take a look at?

pedroresende:
> @pedroresende
> 
> [#2938 (comment)](https://github.com/ionic-team/stencil/pull/2938#issuecomment-1735648023)
> 
> Can you please open a new issue with a minimal reproduction case for the team to take a look at?

Sure, thanks for the fast feedback @rwaskiewicz

pedroresende:
> > @pedroresende
> > [#2938 (comment)](https://github.com/ionic-team/stencil/pull/2938#issuecomment-1735648023)
> > Can you please open a new issue with a minimal reproduction case for the team to take a look at?
> 
> Sure, thanks for the fast feedback @rwaskiewicz

@rwaskiewicz After investigating more, I've concluded that it's actually not a stencil bug but actually it only occurs when using the React Wrapper, if I attempt to use without the wrapper everything works as expected. Therefore I've opened the ticket on stencil-ds-output-targets https://github.com/ionic-team/stencil-ds-output-targets/issues/393

tanner-reits:
Hi all!

You may have seen we've been chipping away at a list of slot-related issues (pulled from this PR) over the last several months. These changes have been placed behind the (newly introduced) [`experimentalSlotFixes`](https://stenciljs.com/docs/config-extras#experimentalslotfixes) and [`experimentalScopedSlotChanges`](https://stenciljs.com/docs/config-extras#experimentalscopedslotchanges) extras config flags. A list of issues these flags address can be found in their respective documentation.

With a majority of this work wrapped up, we're encouraging you all to enable and test out these changes so we can verify their correctness in a larger set of use cases. The plan is that these flags will be enabled by default in Stencil v5.

We do, however, want to note that these flags do **not** address SSR related issues that were mentioned in this PR. This includes #3413 and #2905. SSR related issues will be handled separately.

If anyone still experiences any of the issues claimed to be fixed by these config flags (or encounters any new issues), please open a **new** issue with a reproduction. Opening new issues will help us stay organized and prioritize fixes!

Lastly, we want to give a **HUGE** shoutout to @johnjenkins for his work on this PR! It was truly helpful as a starting point and reference while working through these issues. Thank you!

PR #2948, opened 2021-07-01
fix(slot): fix textContent when using slot polyfill

rwaskiewicz:
Creating a draft PR as we start sprint planning.


src/runtime/bootstrap-lazy.ts
```diff
@@ -137,6 +137,8 @@ export const bootstrapLazy = (lazyBundles: d.LazyBundlesRuntimeData, options: d.
         patchSlotAppendChild(HostElement.prototype);
       }
 
+      patchTextContent(HostElement.prototype);
```
rwaskiewicz, line 140: Per Nick: Consider adding a build flag to do some risk mitigation here/evaluate with the customer

rwaskiewicz:
Usurped by https://github.com/ionic-team/stencil/pull/3047

PR #2950, opened 2021-07-02
chore(format): add prettier

rwaskiewicz:
Changes:
- upgrade prettier to v2.3.2
  - lock version to prevent breaking changes in minor versions
- add prettier.dry-run package.json script
- add pipeline action to evaluate format status
- add prettierignore file for faster runs

Testing:
- I manually ran both prettier commands that I added to `package.json` to try to verify
  - We were formatting all the files we said we would
  - We weren't formatting any files in the `prettierignore` file

Considerations:
- I skipped most shims that seemed to be pulled in externally that seemed pre-minified
- This only formats TS(X) and JS(X) files, no markdown, html or css
- ~There are some `screenshot` directory items that I don't really understand why we have them checked in, they're formattted for now but that's subject to change~ I removed these from the formatter

Notes:
I'm going to fix up the git commits and rebase & merge these so that folks can adopt these changes more easily in their forks. Folks wishing to do so should
0. Make sure they have no outstanding changes that aren't committed
1. `git cherry-pick 4f9ba0f567da97868cfa417db61734f65d2938b7`
2. `npm run prettier && git commit -am "format codebase"`
3. continue working as they were


package.json
```diff
@@ -104,7 +105,7 @@
     "path-browserify": "^1.0.1",
     "pixelmatch": "4.0.2",
     "postcss": "^8.2.8",
-    "prettier": "^2.2.1",
+    "prettier": "2.3.2",
```
ltm, line 108: I know the docs state that "it’s important to have a locked down version of Prettier", but isn't that the point of `package-lock.json`?

ltm (commented):
Should we add a pre-commit hook via husky + lint-staged as well?
```
{
  "scripts": {
    "prepare": "husky install"
  },
  "devDependencies": {
    "husky": "^7.0.1",
    "lint-staged": "^11.1.1",
  },
  "lint-staged": {
    "*.{ts,tsx}": "prettier --write"
  }
}
```

package.json
rwaskiewicz: Prettier is interesting in that it's possible to have breaking changes in minor versions (like TypeScript and ts-jest), so I've made it a habit to lock it down to a specific version. Maybe a better alternative is:
```suggestion
    "prettier": "~2.3.2",
```
Thoughts?

rwaskiewicz:
@ltm I'm a bit opposed to a pre-commit hook here, the codebase is rather large and takes a bit of time to format:
```
$ time npm run prettier
....
npm run prettier  20.05s user 0.96s system 171% cpu 12.246 total
```
then running the command again (so there's no actual formatting occurring, just traversing the filesystem):
```
$ time npm run prettier
....
npm run prettier  18.95s user 0.78s system 182% cpu 10.817 total
```
unfortunately, there's no caching to be done with Prettier, so we need to spend a lot of time looking at each file every time. I tried to be as aggressive as possible in the `prettierignore` file here, but those numbers are _with_ that file sadly :-(

I think this is something @splitinfinities has brought up in the past (precommit hooks). Maybe this is something we can discuss as a team as a follow on story?

src/compiler/types/generate-app-types.ts
```diff
@@ -104,7 +115,12 @@ const generateComponentTypesFile = (config: d.Config, buildCtx: d.BuildCtx, inte
   c.push(`declare module "@stencil/core" {`);
   c.push(`        export namespace JSX {`);
   c.push(`                interface IntrinsicElements {`);
-  c.push(...modules.map(m => `                        "${m.tagName}": LocalJSX.${m.tagNameAsPascal} & JSXBase.HTMLAttributes<${m.htmlElementName}>;`));
+  c.push(
+    ...modules.map(
+      (m) =>
+        `                        "${m.tagName}": LocalJSX.${m.tagNameAsPascal} & JSXBase.HTMLAttributes<${m.htmlElementName}>;`
```
rwaskiewicz, line 121: Oh my oh my

rwaskiewicz: (this is fine, just caught me off guard)

test/ionic-app/src/components/app-home/app-home.tsx
```diff
@@ -19,13 +18,14 @@ export class AppHome {
 
         <ion-content class="ion-padding">
           <p>
-            Welcome to the PWA Toolkit. You can use this starter to build entire
```
rwaskiewicz, line 22: This one kind of surprises me, but maybe it's because I've been looking at these files for a while now 😆

test/karma/test-app/attribute-html/cmp-root.tsx
```diff
@@ -1,19 +1,33 @@
...
   @Prop() anyAttr?: any;
   @Prop() nuAttr?: number;
 
   render() {
     return [
-      <p>strAttr: <strong id="str-attr">{this.strAttr} {typeof this.strAttr}</strong></p>,
-      <p>anyAttr: <strong id="any-attr">{this.anyAttr} {typeof this.anyAttr}</strong></p>,
-      <p>nuAttr: <strong id="nu-attr">{this.nuAttr} {typeof this.nuAttr}</strong></p>
+      <p>
+        strAttr:{' '}
```
rwaskiewicz, line 14: Again, interesting 🤔

package.json
ltm: I don't have a strong opinion and I'm fine with either - my thinking was just that `package-lock.json` will lock it down to a specific version.

ltm:
@rwaskiewicz The benefit of using `lint-staged` is that it'll only run the hooks on the files staged for a commit, e.g.
```
$ time git commit -m '' src/index.ts
✔ Preparing...
✔ Running tasks...
✔ Applying modifications...
✔ Cleaning up...
Aborting commit due to empty commit message.

real	0m1.371s
user	0m1.035s
sys	0m0.335s
```

.prettierignore
```diff
@@ -0,0 +1,55 @@
...
+/notes/
+
+# output of building various scripts that support the project
+/scripts/build/
+
+# these files are intentionally incomplete JavaScript files (they are parts of an Immediately Invoked Funciton
+# Expression (IIFE)). They act as a 'header' and 'footer' that get prepended and appended to the Stencil compiler.
+# Ignore them so Prettier doesn't fail and the 'prettier-ignore' pragma doesn't get put in the compiler output.
+/scripts/bundles/helpers/compiler-cjs-intro.js
+/scripts/bundles/helpers/compiler-cjs-outro.js
```
ltm, line 34: I suppose these should have a different extension as they're not valid JavaScript - but that's for another time.

ltm (approved):

rwaskiewicz:
I fixed up the commit messages, and did a `git diff` against the version that was approved prior to pushing. Waiting for CI to complete before I rebase and merge this

PR #2956, opened 2021-07-08
chore(tests): investigate karma tests don't run locally

rwaskiewicz:
## Pull request checklist

- [ ] Build (`npm run build`) was run locally ~and any changes were pushed~

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [x] Build related changes
- [ ] Other (please describe): N/A

Running `npm run test.prod` would fail locally. The reason for which was the subcommand `test.karma.prod` would fail.  This had a few implications:
- we couldn't run the our entire test suite locally
- we couldn't _release_ new versions of Stencil

## What is the new behavior?

- `npm ci` is now run as a part of the karma test scripts when running locally
  - [we do this in CI today, which is why is passed there](https://github.com/ionic-team/stencil/blob/ccf4c57a619598b6f13cf485747390ce7ab065f2/.github/workflows/main.yml#L63-L71)
- invokes the tests
  - this was [removed last year](https://github.com/ionic-team/stencil/commit/8686b4cab5c881aa01a783b30ffc53931e8a0dd3)
  - one test has been intentionally skipped, as its broken. I would rather 87 passing tests and one skipped one (with an item in the backlog to fix it) rather than none running if that's our only alternative 😉 .
  
- [ ] Yes
- [x] No

## Testing

I was able to run `npm run test.karma.prod` locally.


rwaskiewicz:
@johnjenkins thanks!

test/karma/test-app/slot-fallback/karma.spec.ts
```diff
@@ -1,7 +1,8 @@
 import { setupDomTests, waitForChanges } from '../util';
 
 
-describe('slot-fallback', () => {
+// TODO(STENCIL-18) Restore this test and fix the underlying issue.
+xdescribe('slot-fallback', () => {
```
rwaskiewicz, line 5: FWIW I normally prefer the `.skip` syntax here, but we're using Jasmine here, which doesn't have that in it's API

package.json
```diff
@@ -42,8 +42,8 @@
     "test.dist": "node scripts --validate-build",
     "test.end-to-end": "cd test/end-to-end && npm ci && npm test && npm run test.dist",
     "test.jest": "jest",
-    "test.karma": "cd test/karma && npm run karma",
-    "test.karma.prod": "cd test/karma && npm run karma.prod",
+    "test.karma": "cd test/karma && npm ci && npm run karma",
+    "test.karma.prod": "cd test/karma && npm ci && npm run karma.prod",
```
ltm, line 46: Should we be running `test.karma.prod` in CI as well (i.e. change this to `npm run test.karma.prod`)?
https://github.com/ionic-team/stencil/blob/ccf4c57a619598b6f13cf485747390ce7ab065f2/.github/workflows/main.yml#L71

rwaskiewicz: 👍 done in https://github.com/ionic-team/stencil/pull/2956/commits/4bc1011d76b038f51fad1c9ab6ae4effb2e06a72

test/karma/package.json
```diff
@@ -5,18 +5,18 @@
   "types": "test-dist/types/components.d.ts",
   "collection": "test-dist/collection/collection-manifest.json",
   "scripts": {
-    "build.app": "node ../../bin/stencil build --debug --es5",
+    "build.app": "npm run compile.test-app && node ../../bin/stencil build --debug --es5",
+    "compile.test-app": "node ../../node_modules/typescript/lib/tsc -p tsconfig.json",
```
rwaskiewicz, line 9: This was originally under a script named `tsc` - at first read I thought it was invoking the TS compiler directly, so I renamed it for clarity. This appears to only be directly used for the test-app, hence the name.  I did consider moving the `tsconfig.json` itself down to the `test-app/`, but decided against it as it's referenced in other places and felt out of scope of this user story

splitinfinities:
Could you squash each commits down into one? Afterward, looks great! 👍

rwaskiewicz:
@splitinfinities for sure, I normally do that at merge time, that way I'm not rewriting the git history until everyone's approved (that way if one person approves and the other asks for changes, we're all talking about the same lineage of commits)

ltm (approved):

splitinfinities (approved):
Looks good!

PR #2960, opened 2021-07-13
feat(cli): [STENCIL-12] Add Telemetry to Stencil CLI 

splitinfinities:
This is a chained PR, do not merge it!

Adds Telemetry options to the CLI. Try it out by running any of these commands in your linked project: 

`npx @stencil/core telemetry`
`npx @stencil/core telemetry on|off`

The debug flag will print the telemetry results, e.g:
`npx @stencil/core generate a-component --debug`

Satisfies ADR-1


src/cli/ionic-config.ts
```diff
@@ -0,0 +1,45 @@
+import { readJSON, writeJSON, mkdirp } from '@ionic/utils-fs';
```
ltm, line 1: I don't think you can import this without adding it to `dependencies`. Given our zero dependencies policy you may just need to copy the code from the ionic repo to the stencil repo.

splitinfinities:
Closing to point at the STENCIL-33 branch

src/cli/telemetry/ipc.ts
```diff
@@ -0,0 +1,89 @@
...
+}
+
+/**
+ * Receive and handle an IPC message.
+ *
+ * Assume minimal context and keep external dependencies to a minimum.
+ */
+export async function receive(msg: IPCMessage): Promise<void> {
+	debug('Received %O IPC message', msg.type);
+	console.log("WOOHOO")
```
ltm, line 48:
```suggestion
```

src/cli/ionic-config.ts
ltm: The same applies to `@ionic/utils-terminal` and `@ionic/utils-subprocess`.

PR #2963, opened 2021-07-14
feat(cli): [STENCIL-33] Writing and Reading the Ionic config file

splitinfinities:
Adds ability to allow writing and reading to the ~/.ionic/config.json file. When reading, if the file doesn't exist, readConfig will automatically create the file. 

Satisfies ADR-2

### DIFF
--- a/src/cli/ionic-config.ts
+++ b/src/cli/ionic-config.ts
@@ -33,11 +33,12 @@ export async function readConfig(): Promise<TelemetryConfig> {
 }
 
 export async function writeConfig(config: TelemetryConfig): Promise<void> {
-	await fs.mkdirp(path.dirname(DEFAULT_CONFIG_DIRECTORY)).then(async () => {
-		return await fs.writeJSON(DEFAULT_CONFIG_DIRECTORY, config, { spaces: '\t' });
-	}).catch(() => {
+	try {
+		await fs.mkdirp(path.dirname(DEFAULT_CONFIG_DIRECTORY))
+		await fs.writeJSON(DEFAULT_CONFIG_DIRECTORY, config, { spaces: '\t' });
+	} catch (error) {
 		console.error(`Stencil Telemetry: couldn't write configuration file to ${DEFAULT_CONFIG_DIRECTORY} - ${error}.`)
-	});
+	};
 }
 
 export async function updateConfig(newOptions: TelemetryConfig): Promise<void> {
--- a/src/cli/ionic-config.ts
+++ b/src/cli/ionic-config.ts
@@ -1,8 +1,8 @@
-import fs from 'fs-extra';
+import * as fs from 'fs';
 import * as os from 'os';
 import * as path from 'path';
 
-import { uuidv4 } from './telemetry/helpers';
+import { readJson, uuidv4 } from './telemetry/helpers';
 
 const CONFIG_FILE = 'config.json';
 export const DEFAULT_CONFIG_DIRECTORY = path.resolve(os.homedir(), '.ionic', CONFIG_FILE);
@@ -14,8 +14,7 @@ export interface TelemetryConfig {
 
 export async function readConfig(): Promise<TelemetryConfig> {
 	try {
-		
-		return await fs.readJson(DEFAULT_CONFIG_DIRECTORY);
+		return await readJson(DEFAULT_CONFIG_DIRECTORY);
 	} catch (e) {
 		if (e.code !== 'ENOENT') {
 			throw e;
@@ -34,8 +33,8 @@ export async function readConfig(): Promise<TelemetryConfig> {
 
 export async function writeConfig(config: TelemetryConfig): Promise<void> {
 	try {
-		await fs.mkdirp(path.dirname(DEFAULT_CONFIG_DIRECTORY))
-		await fs.writeJSON(DEFAULT_CONFIG_DIRECTORY, config, { spaces: '\t' });
+		await fs.promises.mkdir(path.dirname(DEFAULT_CONFIG_DIRECTORY), { recursive: true });
+		await fs.promises.writeFile(DEFAULT_CONFIG_DIRECTORY, JSON.stringify(config))
 	} catch (error) {
 		console.error(`Stencil Telemetry: couldn't write configuration file to ${DEFAULT_CONFIG_DIRECTORY} - ${error}.`)
 	};
--- a/src/cli/ionic-config.ts
+++ b/src/cli/ionic-config.ts
@@ -1,41 +1,40 @@
 import { getCompilerSystem } from './state/stencil-cli-config';
 import { readJson, uuidv4 } from './telemetry/helpers';
 
-const CONFIG_FILE = 'config.json';
-export const DEFAULT_CONFIG_DIRECTORY = (file: boolean = false) => {
-	return getCompilerSystem().resolvePath(`${getCompilerSystem().homeDir()}/.ionic${file ? "/" + CONFIG_FILE : ""}`);
-}
+export const DEFAULT_CONFIG = (file: boolean = false) => {
+  return getCompilerSystem().resolvePath(`${getCompilerSystem().homeDir()}/.ionic${file ? '/config.json' : ''}`);
+};
 
 export interface TelemetryConfig {
-	"telemetry.stencil"?: boolean,
-	"tokens.telemetry"?: string,
+  'telemetry.stencil'?: boolean;
+  'tokens.telemetry'?: string;
 }
 
 export async function readConfig(): Promise<TelemetryConfig> {
-	let config: TelemetryConfig = await readJson(DEFAULT_CONFIG_DIRECTORY(true));
+  let config: TelemetryConfig = await readJson(DEFAULT_CONFIG(true));
 
-	if (!config) {
-		config = {
-			"tokens.telemetry": uuidv4(),
-			"telemetry.stencil": true,
-		};
+  if (!config) {
+    config = {
+      'tokens.telemetry': uuidv4(),
+      'telemetry.stencil': true,
+    };
 
-		await writeConfig(config);
-	}
+    await writeConfig(config);
+  }
 
-	return config;
+  return config;
 }
 
 export async function writeConfig(config: TelemetryConfig): Promise<void> {
-	try {
-		await getCompilerSystem().createDir(DEFAULT_CONFIG_DIRECTORY(), { recursive: true });
-		await getCompilerSystem().writeFile(DEFAULT_CONFIG_DIRECTORY(true), JSON.stringify(config))
-	} catch (error) {
-		console.error(`Stencil Telemetry: couldn't write configuration file to ${DEFAULT_CONFIG_DIRECTORY(true)} - ${error}.`)
-	};
+  try {
+    await getCompilerSystem().createDir(DEFAULT_CONFIG(), { recursive: true });
+    await getCompilerSystem().writeFile(DEFAULT_CONFIG(true), JSON.stringify(config));
+  } catch (error) {
+    console.error(`Stencil Telemetry: couldn't write configuration file to ${DEFAULT_CONFIG(true)} - ${error}.`);
+  }
 }
 
 export async function updateConfig(newOptions: TelemetryConfig): Promise<void> {
-	const config = await readConfig();
-	await writeConfig(Object.assign(config, newOptions));
+  const config = await readConfig();
+  await writeConfig(Object.assign(config, newOptions));
 }
--- a/src/cli/ionic-config.ts
+++ b/src/cli/ionic-config.ts
@@ -1,0 +1,45 @@
+import { readJSON, writeJSON, mkdirp } from '@ionic/utils-fs';
+
+import * as os from 'os';
+import * as path from 'path';
+
+import { uuidv4 } from './telemetry/helpers';
+
+export const CONFIG_FILE = 'config.json';
+export const DEFAULT_CONFIG_DIRECTORY = path.resolve(os.homedir(), '.ionic', CONFIG_FILE);
+
+export interface SystemConfig {
+	"telemetry"?: boolean,
+	"telemetry.stencil"?: boolean,
+	"tokens.telemetry"?: string,
+}
+
+export async function readConfig(): Promise<SystemConfig> {
+	try {
+		return await readJSON(DEFAULT_CONFIG_DIRECTORY);
+	} catch (e) {
+		if (e.code !== 'ENOENT') {
+			throw e;
+		}
+
+		const config: SystemConfig = {
+			"tokens.telemetry": uuidv4(),
+			"telemetry": true,
+			"telemetry.stencil": true,
+		};
+
+		await writeConfig(config);
+
+		return config;
+	}
+}
+
+export async function writeConfig(config: SystemConfig): Promise<void> {
+	await mkdirp(path.dirname(DEFAULT_CONFIG_DIRECTORY));
+	await writeJSON(DEFAULT_CONFIG_DIRECTORY, config, { spaces: '\t' });
+}
+
+export async function updateConfig(newOptions: SystemConfig): Promise<void> {
+	const config = await readConfig();
+	await writeConfig(Object.assign(config, newOptions));
+}
--- a/src/cli/ionic-config.ts
+++ b/src/cli/ionic-config.ts
@@ -1,9 +1,11 @@
 import { getCompilerSystem } from './state/stencil-cli-config';
 import { readJson, uuidv4 } from './telemetry/helpers';
 
-export const DEFAULT_CONFIG = (file: boolean = false) => {
-  return getCompilerSystem().resolvePath(`${getCompilerSystem().homeDir()}/.ionic${file ? '/config.json' : ''}`);
-};
+export const default_config = () =>
+  getCompilerSystem().resolvePath(`${getCompilerSystem().homeDir()}/.ionic/config.json`);
+
+export const default_config_directory = () =>
+  getCompilerSystem().resolvePath(`${getCompilerSystem().homeDir()}/.ionic`);
 
 export interface TelemetryConfig {
   'telemetry.stencil'?: boolean;
@@ -11,7 +13,7 @@ export interface TelemetryConfig {
 }
 
 export async function readConfig(): Promise<TelemetryConfig> {
-  let config: TelemetryConfig = await readJson(DEFAULT_CONFIG(true));
+  let config: TelemetryConfig = await readJson(default_config());
 
   if (!config) {
     config = {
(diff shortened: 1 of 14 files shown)


src/cli/ionic-config.ts
```diff
@@ -0,0 +1,45 @@
...
+
+import * as os from 'os';
+import * as path from 'path';
+
+import { uuidv4 } from './telemetry/helpers';
+
+export const CONFIG_FILE = 'config.json';
+export const DEFAULT_CONFIG_DIRECTORY = path.resolve(os.homedir(), '.ionic', CONFIG_FILE);
+
+export interface SystemConfig {
```
rwaskiewicz, line 11: Consider renaming this to `TelemetryConfig`, as `SystemConfig` may be easily conflated with other parts of the 'system' (e.g. is this a compiler configuration, a build configuration, etc.)
```suggestion
export interface TelemetryConfig {
```

src/cli/ionic-config.ts
```diff
@@ -0,0 +1,45 @@
...
+	"telemetry"?: boolean,
```
rwaskiewicz, line 12: For my own edification, is this to enable/disable telemetry? Is this going to be shared across multiple project types?

src/cli/ionic-config.ts
```diff
@@ -0,0 +1,45 @@
...
+	"telemetry.stencil"?: boolean,
+	"tokens.telemetry"?: string,
```
rwaskiewicz, line 14: For my own understanding, is this a series of strings with a comma (or something similar) as a delimiter?

src/cli/ionic-config.ts
```diff
@@ -0,0 +1,45 @@
+import { readJSON, writeJSON, mkdirp } from '@ionic/utils-fs';
```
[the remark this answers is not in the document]
splitinfinities: I had a gut feeling that may be the case! 👍

src/cli/ionic-config.ts
```diff
@@ -0,0 +1,45 @@
...
+			throw e;
+		}
+
+		const config: SystemConfig = {
+			"tokens.telemetry": uuidv4(),
+			"telemetry": true,
+			"telemetry.stencil": true,
+		};
+
+		await writeConfig(config);
```
rwaskiewicz, line 31: Are there any error cases from writing a config file we need to watch out for here? e.g. if we don't have permission to write to the local FS? I'm not familiar with this module, apologies if this is a dumb question 😆

src/cli/ionic-config.ts
```diff
@@ -0,0 +1,45 @@
...
+
+		return config;
+	}
+}
+
+export async function writeConfig(config: SystemConfig): Promise<void> {
+	await mkdirp(path.dirname(DEFAULT_CONFIG_DIRECTORY));
```
rwaskiewicz, line 38: I think we  need to ensure two things here:
1. The call to `mkdirp` isn't rejected
2. If the call to `mkdirp` _is_ rejected, that we clean up after ourselves in the event that we clean up after ourselves. From the `mkdirp` [doc](https://github.com/isaacs/node-mkdirp#mkdirpdir-opts---promisestring--undefined)
> Promise rejects if any errors are encountered. Note that, in the case of promise rejection, some directories may have been created, as recursive directory creation is not an atomic operation.

However, the latter can tricky if we intend to reuse this directory for other projects, so perhaps only the first suggestion is required?

src/cli/telemetry/helpers.ts
```diff
@@ -0,0 +1,25 @@
...
+		// ignore
+	}
+
+	return null;
+};
+
+export const isInteractive = (): boolean => TERMINAL_INFO.tty && !TERMINAL_INFO.ci;
+
+export function uuidv4(): string {
+	return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, c => {
```
rwaskiewicz, line 19: Is there a strong need to roll our own UUID v4 generation? I suppose Stencil prefers a "zero dependency" approach, which is a good reason to avoid pulling in something like [UUID](https://www.npmjs.com/package/uuid). Do we have any documentation we can attach to this function that explains how we decided on this implementation?

rwaskiewicz (commented):
One overarching question - can we get these helper functions covered by (unit) tests please? LMK if you have any questions on that, although I hardly consider myself an expert 😉

src/cli/ionic-config.ts
```diff
@@ -0,0 +1,45 @@
+import { readJSON, writeJSON, mkdirp } from '@ionic/utils-fs';
+
+import * as os from 'os';
+import * as path from 'path';
+
+import { uuidv4 } from './telemetry/helpers';
+
+export const CONFIG_FILE = 'config.json';
+export const DEFAULT_CONFIG_DIRECTORY = path.resolve(os.homedir(), '.ionic', CONFIG_FILE);
```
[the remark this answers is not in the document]
splitinfinities: We can do this yes, originally I was considering logging out where this file is located via the `npx @stencil/core telemetry ` command.

src/cli/ionic-config.ts
splitinfinities: I'm adding one new prop to the ionic file - that's `telemetry.stencil` - so that a customer could enable telemetry for ionic CLI and disable telemetry for Stencil CLI. Originally when building this interface, I considered overwriting this key, or inferring the value for the `telemetry.stencil` key when first adding it. I went with a simpler approach where stencil only manages its prop, so I will nix the `telemetry` prop in this interface as its unused.

src/cli/ionic-config.ts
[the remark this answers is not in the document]
rwaskiewicz: Gotcha 👍 

IMO, we should follow [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) here - that way if/when we do decide to do that, we can make that change in the associated PR, which should make it a little easier to reason about why these constants are exposed

src/cli/ionic-config.ts
splitinfinities: `tokens.telemetry` is the result of the uuid function that gets called on Line 26 - it's just a plain string with some dashes. Looks kind of like "xxxxxxx-xxxxx-xxxx" (that's not an accurate format, I'm on a walk)

src/cli/ionic-config.ts
splitinfinities: I don't doubt that there are edge cases, I'm just as unfamiliar with the package as you :) I'll add some try catch blocks, but also this shouldn't error during any normal Stencil usage, only unless the customer specifically fires a telemetry command.

src/cli/ionic-config.ts
splitinfinities: I'll add recovery for option 1, which would also add messaging for option 2. FWIW, I copied this function directly from ionic-cli, and it's usage is nearly identical.

src/cli/telemetry/helpers.ts
splitinfinities: I copied this from Ionic-CLI and their approach to the UUID generation, in order to keep that synced up without adding dependencies.

src/cli/ionic-config.ts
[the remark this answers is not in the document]
splitinfinities: I ended up needing one of these for unit tests.

src/cli/ionic-config.ts
splitinfinities: Added error handling inside this function definition

src/cli/telemetry/helpers.ts
splitinfinities: Added!

splitinfinities:
@rwaskiewicz Added 6 tests for these new methods. 
Screen Shot 2021-07-14 at 12 22 33 PM

src/cli/telemetry/helpers.ts
ltm: FWIW, Node 14 added [crypto.randomUUID](https://nodejs.org/dist/latest-v14.x/docs/api/all.html#crypto_crypto_randomuuid_options).

src/cli/ionic-config.ts
```diff
@@ -0,0 +1,46 @@
+import fs from 'fs-extra';
```
ltm, line 1: I think `fs-extra` is a dependency that won't be available in the build. You might be able to use
```suggestion
import { promises as fs } from 'fs';
```

src/cli/test/ionic-config.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+	it('should create a file if it does not exist', async () => {
+		if (await fs.pathExists(DEFAULT_CONFIG_DIRECTORY)) {
+			fs.rmSync(DEFAULT_CONFIG_DIRECTORY);
+		}
+
+		expect(await fs.pathExists(DEFAULT_CONFIG_DIRECTORY)).toBe(false)
+
+		const config = await readConfig();
+
+		expect(typeof config).toBe("object");
```
rwaskiewicz, line 15: Same here, I don't think we strictly need these assertions on the type of an object

src/cli/telemetry/helpers.ts
rwaskiewicz: I take this back, we support Node 12 ATM

src/cli/ionic-config.ts
```diff
@@ -0,0 +1,46 @@
...
+import * as os from 'os';
+import * as path from 'path';
+
+import { uuidv4 } from './telemetry/helpers';
+
+const CONFIG_FILE = 'config.json';
+export const DEFAULT_CONFIG_DIRECTORY = path.resolve(os.homedir(), '.ionic', CONFIG_FILE);
```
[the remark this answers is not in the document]
splitinfinities: Since this is cut and pasted from Ionic CLI, I'd like to keep it as 1-to-1 as possible

src/cli/ionic-config.ts
splitinfinities: Well I'm glad I wrote those tests! I switched this out, worked great.

src/cli/telemetry/test/helpers.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+	});
+
+	it('should handle success correctly', async () => {
+		const result = await tryFn(async () => {
+			return true
+		})
+
+		expect(result).toBe(true)
+	});
+	it('handles returning false correctly', async () => {
```
rwaskiewicz, line 30: Non blocking: It'd be nice to have the spacing of the tests here be the same, since Prettier won't enforce that for us (I think)
```suggestion

	it('handles returning false correctly', async () => {
```

src/cli/telemetry/helpers.ts
```diff
@@ -0,0 +1,52 @@
...
+	try {
+		return await fn(...args);
+	} catch {
+		// ignore
+	}
+
+	return null;
+};
+
+export const isInteractive = (): boolean => TERMINAL_INFO.tty && !TERMINAL_INFO.ci;
```
rwaskiewicz, line 37: One final ask - can we get this fn under test?

src/cli/telemetry/helpers.ts
```diff
@@ -0,0 +1,52 @@
...
+	return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, c => {
+		const r = (Math.random() * 16) | 0;
+		const v = c == 'x' ? r : (r & 0x3) | 0x8;
+
+		return v.toString(16);
+	});
+}
+
+export async function readJson(path: string) {
+	const rawdata = await fs.readFileSync(path);
```
ltm, line 50: The `await` is unnecessary here:
```suggestion
	const rawdata = fs.readFileSync(path);
```

ltm: Alternatively, make it async using `fs.promises`:
```suggestion
	const rawdata = await fs.promises.readFileSync(path);
```

src/cli/telemetry/helpers.ts
```diff
@@ -0,0 +1,52 @@
...
+	return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, c => {
+		const r = (Math.random() * 16) | 0;
+		const v = c == 'x' ? r : (r & 0x3) | 0x8;
+
+		return v.toString(16);
+	});
+}
+
+export async function readJson(path: string) {
+	const rawdata = await fs.promises.readFileSync(path);
```
rwaskiewicz: CI is [currently failing](https://github.com/ionic-team/stencil/pull/2963/checks?check_run_id=3080055348)

> updateConfig › should edit a file
> 
>    TypeError: fs_1.default.promises.readFileSync is not a function


Did you mean:

```suggestion
	const rawdata = await fs.promises.readFile(path);
```

https://nodejs.org/api/fs.html#fs_fspromises_readfile_path_options

rwaskiewicz: NVM, looks like this got fixed while I was typing it 😆

src/cli/telemetry/helpers.ts
splitinfinities: Added!

rwaskiewicz (approved):

splitinfinities:
@rwaskiewicz and @ltm if you could take a peek at the new changes in 2dd693f regarding using the abstracted system calls that help us work in Deno or Node I'd really appreciate it! I'm going to bring these over to the STENCIL 12 PR in the meantime.

src/cli/telemetry/helpers.ts
```diff
@@ -0,0 +1,52 @@
...
+	return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, c => {
+		const r = (Math.random() * 16) | 0;
+		const v = c == 'x' ? r : (r & 0x3) | 0x8;
+
+		return v.toString(16);
+	});
+}
+
+export async function readJson(path: string) {
+	const file = await getCompilerSystem()?.readFile(path);
```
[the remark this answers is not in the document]
splitinfinities: We shouldn't need it, no.

src/cli/telemetry/test/helpers.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+		const pattern = new RegExp(/^[0-9A-F]{8}-[0-9A-F]{4}-4[0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}$/i);
+		const uuid = uuidv4()
+		expect(!!uuid.match(pattern)).toBe(true)
+	});
+
+});
+
+describe('isInteractive', () => {
+	it('should return false when tty is false', async () => {
+		TERMINAL_INFO = { ci: true, shell: true, tty: false, windows: true };
```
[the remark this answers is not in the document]
splitinfinities: I added a fixed test on the 12 branch because that's where the build was failing. Pulled that code into this branch.

src/cli/state/stencil-cli-config.ts
```diff
@@ -0,0 +1,107 @@
...
+	return StencilCLIConfig.getInstance(options);
+}
+
+export function getStencilCLIConfig(): StencilCLIConfig {
+	return StencilCLIConfig.getInstance();
+}
+
+export function getCompilerSystem(): CompilerSystem {
+	const stencilCLIConfig = getStencilCLIConfig();
+	return !!stencilCLIConfig && stencilCLIConfig?.sys
```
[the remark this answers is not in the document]
splitinfinities: This is an artifact of some protection I had once had, now we don't need it. Nixing here in a sec.

src/cli/state/stencil-cli-config.ts
```diff
@@ -0,0 +1,107 @@
...
+	private _coreCompiler: CoreCompiler;
+
+	private constructor (options: StencilCLIConfigArgs) {
+		this._args = options?.args || [];
+		this._logger = options?.logger;
+		this._sys = options?.sys;
+	}
+
+	public static getInstance(options?: StencilCLIConfigArgs): StencilCLIConfig {
+		if (!StencilCLIConfig.instance && !!options) {
```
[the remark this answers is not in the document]
splitinfinities: I don't think there's a need.

src/cli/test/ionic-config.spec.ts
```diff
@@ -0,0 +1,77 @@
+import { mockLogger } from '@stencil/core/testing';
+import { getCompilerSystem, initializeStencilCLIConfig } from '../state/stencil-cli-config';
+import { readConfig, writeConfig, updateConfig, DEFAULT_CONFIG_DIRECTORY } from '../ionic-config';
+import { createSystem } from '../../compiler/sys/stencil-sys';
+
+describe('readConfig', () => {
+
+	initializeStencilCLIConfig({
+		sys: createSystem(),
```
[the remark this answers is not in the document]
splitinfinities: I enabled Format on Save in VS Code last night, and pointed that to prettier, these files have been updated now.

src/cli/ionic-config.ts
```diff
@@ -1,17 +1,19 @@
 import { getCompilerSystem } from './state/stencil-cli-config';
 import { readJson, uuidv4 } from './telemetry/helpers';
 
-export const DEFAULT_CONFIG = (file: boolean = false) => {
-  return getCompilerSystem().resolvePath(`${getCompilerSystem().homeDir()}/.ionic${file ? '/config.json' : ''}`);
-};
+export const default_config = () =>
```
rwaskiewicz, line 4: Should these be `defaultConfig` and `defaultConfigDirectory`?

rwaskiewicz (approved):

src/cli/ionic-config.ts
splitinfinities: Patched!

src/cli/telemetry/helpers.ts
```diff
@@ -0,0 +1,51 @@
...
+};
+
+export const isInteractive = (object?: TerminalInfo): boolean => {
+  const terminalInfo =
+    object ||
+    Object.freeze({
+      tty: process.stdout.isTTY ? true : false,
+      ci:
+        ['CI', 'BUILD_ID', 'BUILD_NUMBER', 'BITBUCKET_COMMIT', 'CODEBUILD_BUILD_ARN'].filter(v => !!process.env[v])
+          .length > 0 || process.argv.includes('--ci'),
```
ltm, line 33: `process` is a Node built-in, so you'd have to access these through the compiler system. You should be able to replace `process.env[v]` with `CompilerSystem.getEnvironmentVar(v)`

splitinfinities: Ahh, hmm... I'm going to need to do a couple changes here in that case.

ltm (approved):

rwaskiewicz (approved):

PR #2964, opened 2021-07-14
feat(cli): [STENCIL-12] Adding Telemetry and CLI features. 

splitinfinities:
Adds Telemetry options to the CLI. Try it out by running any of these commands in your linked project:

`npx stencil telemetry`
`npx stencil telemetry on|off`

The debug flag will print the telemetry results, e.g:
`npx stencil generate a-component --debug --verbose`

Satisfies ADR-1

### DIFF
--- a/src/cli/ionic-config.ts
+++ b/src/cli/ionic-config.ts
@@ -26,16 +26,20 @@ export async function readConfig(): Promise<TelemetryConfig> {
   return config;
 }
 
-export async function writeConfig(config: TelemetryConfig): Promise<void> {
+export async function writeConfig(config: TelemetryConfig): Promise<boolean> {
+  let result = false;
   try {
     await getCompilerSystem().createDir(defaultConfigDirectory(), { recursive: true });
     await getCompilerSystem().writeFile(defaultConfig(), JSON.stringify(config));
+    result = true;
   } catch (error) {
     console.error(`Stencil Telemetry: couldn't write configuration file to ${defaultConfig()} - ${error}.`);
   }
+
+  return result;
 }
 
-export async function updateConfig(newOptions: TelemetryConfig): Promise<void> {
+export async function updateConfig(newOptions: TelemetryConfig): Promise<boolean> {
   const config = await readConfig();
-  await writeConfig(Object.assign(config, newOptions));
+  return await writeConfig(Object.assign(config, newOptions));
 }
--- a/src/cli/ionic-config.ts
+++ b/src/cli/ionic-config.ts
@@ -1,8 +1,10 @@
 import { getCompilerSystem } from './state/stencil-cli-config';
 import { readJson, uuidv4, UUID_REGEX } from './telemetry/helpers';
 
+export const isTest = () => process.env.JEST_WORKER_ID !== undefined
+
 export const defaultConfig = () =>
-  getCompilerSystem().resolvePath(`${getCompilerSystem().homeDir()}/.ionic/config.json`);
+  getCompilerSystem().resolvePath(`${getCompilerSystem().homeDir()}/.ionic/${isTest() ? "tmp-config.json" : "config.json"}`);
 
 export const defaultConfigDirectory = () => getCompilerSystem().resolvePath(`${getCompilerSystem().homeDir()}/.ionic`);
 
@@ -23,7 +25,7 @@ export async function readConfig(): Promise<TelemetryConfig> {
     await writeConfig(config);
   } else if (!!config && !config['tokens.telemetry'].match(UUID_REGEX)) {
     const newUuid = uuidv4();
-    await writeConfig(Object.assign(config, { 'tokens.telemetry': newUuid }));
+    await writeConfig({...config, 'tokens.telemetry': newUuid });
     config['tokens.telemetry'] = newUuid;
   }
 
--- a/src/cli/ionic-config.ts
+++ b/src/cli/ionic-config.ts
@@ -30,7 +30,7 @@ export async function writeConfig(config: TelemetryConfig): Promise<boolean> {
   let result = false;
   try {
     await getCompilerSystem().createDir(defaultConfigDirectory(), { recursive: true });
-    await getCompilerSystem().writeFile(defaultConfig(), JSON.stringify(config));
+    await getCompilerSystem().writeFile(defaultConfig(), JSON.stringify(config, null, 2));
     result = true;
   } catch (error) {
     console.error(`Stencil Telemetry: couldn't write configuration file to ${defaultConfig()} - ${error}.`);
--- a/src/cli/ionic-config.ts
+++ b/src/cli/ionic-config.ts
@@ -1,5 +1,5 @@
 import { getCompilerSystem } from './state/stencil-cli-config';
-import { readJson, uuidv4 } from './telemetry/helpers';
+import { readJson, uuidv4, UUID_REGEX } from './telemetry/helpers';
 
 export const defaultConfig = () =>
   getCompilerSystem().resolvePath(`${getCompilerSystem().homeDir()}/.ionic/config.json`);
@@ -21,6 +21,10 @@ export async function readConfig(): Promise<TelemetryConfig> {
     };
 
     await writeConfig(config);
+  } else if (!!config && !config['tokens.telemetry'].match(UUID_REGEX)) {
+    const newUuid = uuidv4();
+    await writeConfig(Object.assign(config, { 'tokens.telemetry': newUuid }));
+    config['tokens.telemetry'] = newUuid;
   }
 
   return config;
--- a/src/cli/run.ts
+++ b/src/cli/run.ts
@@ -39,16 +39,16 @@ export const run = async (init: CliInitOptions) => {
       sys.applyGlobalPatch(sys.getCurrentDirectory());
     }
 
-    if (task === 'help' || flags.help) {
-      taskHelp({ sys, logger });
-      return;
-    }
-
     // Update singleton with modifications
     stencilCLIConfig.logger = logger;
     stencilCLIConfig.task = task;
     stencilCLIConfig.sys = sys;
     stencilCLIConfig.flags = flags;
+
+    if (task === 'help' || flags.help) {
+      taskHelp();
+      return;
+    }
 
     startupLog(logger, task);
 
@@ -128,14 +128,6 @@ export const runTask = async (coreCompiler: CoreCompiler, config: Config, task: 
   config.outputTargets = config.outputTargets || [];
 
   switch (task) {
-    case 'telemetry':
-      await taskTelemetry();
-      break;
-
-    case 'help':
-      taskHelp(config);
-      break;
-
     case 'build':
       await taskBuild(coreCompiler, config);
       break;
@@ -149,12 +141,20 @@ export const runTask = async (coreCompiler: CoreCompiler, config: Config, task: 
       await taskGenerate(coreCompiler, config);
       break;
 
+    case 'help':
+      taskHelp();
+      break;
+
     case 'prerender':
       await taskPrerender(coreCompiler, config);
       break;
 
     case 'serve':
       await taskServe(config);
+      break;
+
+    case 'telemetry':
+      await taskTelemetry();
       break;
 
     case 'test':
@@ -167,7 +167,7 @@ export const runTask = async (coreCompiler: CoreCompiler, config: Config, task: 
 
     default:
       config.logger.error(`${config.logger.emoji('❌ ')}Invalid stencil command, please see the options below:`);
-      taskHelp(config);
+      taskHelp();
       return config.sys.exit(1);
   }
 };
--- a/src/cli/run.ts
+++ b/src/cli/run.ts
@@ -14,16 +14,20 @@ import { taskPrerender } from './task-prerender';
 import { taskServe } from './task-serve';
 import { taskTest } from './task-test';
 import { initializeStencilCLIConfig } from './state/stencil-cli-config';
+import { taskTelemetry } from './task-telemetry';
+
+import { telemetryAction } from './telemetry/telemetry';
 
 export const run = async (init: CliInitOptions) => {
   const { args, logger, sys } = init;
 
   // Initialize the singleton so we can use this throughout the lifecycle of the CLI.
-  const stencilCLIConfig = initializeStencilCLIConfig({ args, logger, sys});
+  const stencilCLIConfig = initializeStencilCLIConfig({ args, logger, sys });
 
   try {
     const flags = parseFlags(args, sys);
     const task = flags.task;
+
     if (flags.debug || flags.verbose) {
       logger.setLevel('debug');
     }
@@ -41,11 +45,6 @@ export const run = async (init: CliInitOptions) => {
     stencilCLIConfig.task = task;
     stencilCLIConfig.sys = sys;
     stencilCLIConfig.flags = flags;
-
-    if (task === 'help' || flags.help) {
-      taskHelp(sys, logger);
-      return;
-    }
 
     startupLog(logger, task);
 
@@ -79,7 +78,14 @@ export const run = async (init: CliInitOptions) => {
     loadedCompilerLog(sys, logger, flags, coreCompiler);
 
     if (task === 'info') {
-      taskInfo(coreCompiler, sys, logger);
+      await telemetryAction({
+        args,
+        sys,
+        logger,
+        action: async () => {
+          await taskInfo(coreCompiler, sys, logger);
+        },
+      });
       return;
     }
 
@@ -105,7 +111,15 @@ export const run = async (init: CliInitOptions) => {
 
     await sys.ensureResources({ rootDir: validated.config.rootDir, logger, dependencies: dependencies as any });
 
-    await runTask(coreCompiler, validated.config, task);
+    await telemetryAction({
+      args,
+      sys,
+      logger,
+      config: validated.config,
+      action: async () => {
+        await runTask(coreCompiler, validated.config, task);
+      },
+    });
   } catch (e) {
     if (!shouldIgnoreError(e)) {
       logger.error(`uncaught cli error: ${e}${logger.getLevel() === 'debug' ? e.stack : ''}`);
@@ -119,16 +133,20 @@ export const runTask = async (coreCompiler: CoreCompiler, config: Config, task: 
   config.outputTargets = config.outputTargets || [];
 
   switch (task) {
+    case 'telemetry':
+      await taskTelemetry(config);
+      break;
+
+    case 'help':
+      taskHelp(config);
+      break;
+
     case 'build':
       await taskBuild(coreCompiler, config);
       break;
 
     case 'docs':
       await taskDocs(coreCompiler, config);
-      break;
-
-    case 'help':
-      taskHelp(config.sys, config.logger);
       break;
 
     case 'generate':
@@ -154,7 +172,7 @@ export const runTask = async (coreCompiler: CoreCompiler, config: Config, task: 
 
     default:
       config.logger.error(`${config.logger.emoji('❌ ')}Invalid stencil command, please see the options below:`);
-      taskHelp(config.sys, config.logger);
+      taskHelp(config);
       return config.sys.exit(1);
   }
 };
--- a/src/cli/run.ts
+++ b/src/cli/run.ts
@@ -15,7 +15,6 @@ import { taskServe } from './task-serve';
 import { taskTest } from './task-test';
 import { initializeStencilCLIConfig } from './state/stencil-cli-config';
 import { taskTelemetry } from './task-telemetry';
-
 import { telemetryAction } from './telemetry/telemetry';
 
 export const run = async (init: CliInitOptions) => {
--- a/src/cli/run.ts
+++ b/src/cli/run.ts
@@ -40,6 +40,11 @@ export const run = async (init: CliInitOptions) => {
       sys.applyGlobalPatch(sys.getCurrentDirectory());
     }
 
+    if (task === 'help' || flags.help) {
+      taskHelp({ sys, logger });
+      return;
+    }
+
     // Update singleton with modifications
     stencilCLIConfig.logger = logger;
     stencilCLIConfig.task = task;
@@ -78,13 +83,8 @@ export const run = async (init: CliInitOptions) => {
     loadedCompilerLog(sys, logger, flags, coreCompiler);
 
     if (task === 'info') {
-      await telemetryAction({
-        args,
-        sys,
-        logger,
-        action: async () => {
-          await taskInfo(coreCompiler, sys, logger);
-        },
+      await telemetryAction(async () => {
+        await taskInfo(coreCompiler, sys, logger);
       });
       return;
     }
@@ -105,20 +105,16 @@ export const run = async (init: CliInitOptions) => {
       }
     }
 
+    stencilCLIConfig.validatedConfig = validated;
+
     if (isFunction(sys.applyGlobalPatch)) {
       sys.applyGlobalPatch(validated.config.rootDir);
     }
 
     await sys.ensureResources({ rootDir: validated.config.rootDir, logger, dependencies: dependencies as any });
 
-    await telemetryAction({
-      args,
-      sys,
-      logger,
-      config: validated.config,
-      action: async () => {
-        await runTask(coreCompiler, validated.config, task);
-      },
+    await telemetryAction(async () => {
+      await runTask(coreCompiler, validated.config, task);
     });
   } catch (e) {
     if (!shouldIgnoreError(e)) {
@@ -134,7 +130,7 @@ export const runTask = async (coreCompiler: CoreCompiler, config: Config, task: 
 
   switch (task) {
     case 'telemetry':
-      await taskTelemetry(config);
+      await taskTelemetry();
       break;
 
     case 'help':
--- a/src/cli/task-help.ts
+++ b/src/cli/task-help.ts
@@ -1,9 +1,9 @@
-import type { Config } from '../declarations';
+import { getCompilerSystem, getLogger } from './state/stencil-cli-config';
 import { taskTelemetry } from './task-telemetry';
 
-export const taskHelp = async (config: Config) => {
-  const logger = config.logger;
-  const sys = config.sys;
+export const taskHelp = async () => {
+  const logger = getLogger();
+  const sys = getCompilerSystem();
 
   const prompt = logger.dim(sys.details.platform === 'windows' ? '>' : '$');
 
--- a/src/cli/task-help.ts
+++ b/src/cli/task-help.ts
@@ -5,12 +5,12 @@ export const taskHelp = async (config: Config) => {
   const logger = config.logger;
   const sys = config.sys;
 
-  const p = logger.dim(sys.details.platform === 'windows' ? '>' : '$');
+  const prompt = logger.dim(sys.details.platform === 'windows' ? '>' : '$');
 
   console.log(`
   ${logger.bold('Build:')} ${logger.dim('Build components for development or production.')}
 
-    ${p} ${logger.green('stencil build [--dev] [--watch] [--prerender] [--debug]')}
+    ${prompt} ${logger.green('stencil build [--dev] [--watch] [--prerender] [--debug]')}
 
       ${logger.cyan('--dev')} ${logger.dim('.............')} Development build
       ${logger.cyan('--watch')} ${logger.dim('...........')} Rebuild when files update
@@ -25,7 +25,7 @@ export const taskHelp = async (config: Config) => {
 
   ${logger.bold('Test:')} ${logger.dim('Run unit and end-to-end tests.')}
 
-    ${p} ${logger.green('stencil test [--spec] [--e2e]')}
+    ${prompt} ${logger.green('stencil test [--spec] [--e2e]')}
 
       ${logger.cyan('--spec')} ${logger.dim('............')} Run unit tests with Jest
       ${logger.cyan('--e2e')} ${logger.dim('.............')} Run e2e tests with Puppeteer
@@ -33,19 +33,21 @@ export const taskHelp = async (config: Config) => {
 
   ${logger.bold('Generate:')} ${logger.dim('Bootstrap components.')}
 
-    ${p} ${logger.green('stencil generate')} or ${logger.green('stencil g')}
+    ${prompt} ${logger.green('stencil generate')} or ${logger.green('stencil g')}
 
 `);
-  
-  await taskTelemetry(config);
+
+  await taskTelemetry();
 
   console.log(`
   ${logger.bold('Examples:')}
 
-  ${p} ${logger.green('stencil build --dev --watch --serve')}
-  ${p} ${logger.green('stencil build --prerender')}
-  ${p} ${logger.green('stencil test --spec --e2e')}
-  ${p} ${logger.green('stencil generate')}
-  ${p} ${logger.green('stencil g my-component')}
-`)
+  
+  ${prompt} ${logger.green('stencil build --dev --watch --serve')}
+  ${prompt} ${logger.green('stencil build --prerender')}
+  ${prompt} ${logger.green('stencil test --spec --e2e')}
+  ${prompt} ${logger.green('stencil telemetry on')}
+  ${prompt} ${logger.green('stencil generate')}
+  ${prompt} ${logger.green('stencil g my-component')}
+`);
 };
--- a/src/cli/task-help.ts
+++ b/src/cli/task-help.ts
@@ -1,6 +1,10 @@
-import type { CompilerSystem, Logger } from '../declarations';
+import type { Config } from '../declarations';
+import { taskTelemetry } from './task-telemetry';
 
-export const taskHelp = (sys: CompilerSystem, logger: Logger) => {
+export const taskHelp = async (config: Config) => {
+  const logger = config.logger;
+  const sys = config.sys;
+
   const p = logger.dim(sys.details.platform === 'windows' ? '>' : '$');
 
   console.log(`
@@ -31,14 +35,17 @@ export const taskHelp = (sys: CompilerSystem, logger: Logger) => {
 
     ${p} ${logger.green('stencil generate')} or ${logger.green('stencil g')}
 
+`);
+  
+  await taskTelemetry(config);
 
+  console.log(`
   ${logger.bold('Examples:')}
 
-    ${p} ${logger.green('stencil build --dev --watch --serve')}
-    ${p} ${logger.green('stencil build --prerender')}
-    ${p} ${logger.green('stencil test --spec --e2e')}
-    ${p} ${logger.green('stencil generate')}
-    ${p} ${logger.green('stencil g my-component')}
-
-`);
+  ${p} ${logger.green('stencil build --dev --watch --serve')}
+  ${p} ${logger.green('stencil build --prerender')}
+  ${p} ${logger.green('stencil test --spec --e2e')}
+  ${p} ${logger.green('stencil generate')}
+  ${p} ${logger.green('stencil g my-component')}
+`)
 };
--- a/src/cli/task-telemetry.ts
+++ b/src/cli/task-telemetry.ts
@@ -12,7 +12,7 @@ export const taskTelemetry = async () => {
   const THANK_YOU = `Thank you for helping to make Stencil better! 💖`;
   const ENABLED_MESSAGE = `${logger.green('Enabled')}. ${THANK_YOU}\n\n`;
   const DISABLED_MESSAGE = `${logger.red('Disabled')}\n`;
-  let hasTelemetry = await checkTelemetry();
+  const hasTelemetry = await checkTelemetry();
 
   if (isEnabling) {
     const result = await enableTelemetry();
--- a/src/cli/task-telemetry.ts
+++ b/src/cli/task-telemetry.ts
@@ -11,7 +11,7 @@ export const taskTelemetry = async () => {
   )}`;
   const THANK_YOU = `Thank you for helping to make Stencil better! 💖`;
   const ENABLED_MESSAGE = `${logger.green('Enabled')}. ${THANK_YOU}\n\n`;
-  const DISABLED_MESSAGE = `${logger.red('Disabled')}\n`;
+  const DISABLED_MESSAGE = `${logger.red('Disabled')}\n\n`;
   const hasTelemetry = await checkTelemetry();
 
   if (isEnabling) {
--- a/src/cli/task-telemetry.ts
+++ b/src/cli/task-telemetry.ts
@@ -1,0 +1,36 @@
+import type { Config} from '../declarations';
+import { checkTelemetry, disableTelemetry, enableTelemetry } from './telemetry/telemetry';
+
+export const taskTelemetry = async (config: Config) => {
+  const p = config.logger.dim(config.sys.details.platform === 'windows' ? '>' : '$');
+  const logger = config.logger;
+  const isEnabling = config.flags.args.includes("on");
+  const isDisabling = config.flags.args.includes("off");
+  const INFORMATION = `Opt in or our of telemetry. Information about the data we collect is available on our website: ${logger.bold('https://stenciljs.com/telemetry')}`
+  const THANK_YOU = `Thank you for helping to make Stencil better! 💖`;
+  const ENABLED_MESSAGE = `${logger.green('Enabled')}. ${THANK_YOU}\n\n`;
+  const DISABLED_MESSAGE = `${logger.red('Disabled')}\n`;
+  let hasTelemetry = await checkTelemetry();
+
+  if (isEnabling) {
+    await enableTelemetry()
+    console.log(`\n  ${logger.bold('Telemetry is now ') + ENABLED_MESSAGE}`)
+    return;
+  }
+
+  if (isDisabling) {
+    await disableTelemetry()
+    console.log(`\n  ${logger.bold('Telemetry is now ') + DISABLED_MESSAGE}`)
+    return;
+  }
+
+  console.log(`  ${logger.bold('Telemetry:')} ${logger.dim(INFORMATION)}`)
+
+  console.log(`\n  ${logger.bold('Status')}: ${hasTelemetry ? ENABLED_MESSAGE : DISABLED_MESSAGE}`)
+
+  console.log(`    ${p} ${logger.green('stencil telemetry [off|on]')}
+
+        ${logger.cyan('off')} ${logger.dim('.............')} Disable sharing anonymous usage data
+        ${logger.cyan('on')} ${logger.dim('..............')} Enable sharing anonymous usage data
+  `);
+};
--- a/src/cli/task-telemetry.ts
+++ b/src/cli/task-telemetry.ts
@@ -1,34 +1,40 @@
-import type { Config} from '../declarations';
+import { getCompilerSystem, getLogger, getStencilCLIConfig } from './state/stencil-cli-config';
 import { checkTelemetry, disableTelemetry, enableTelemetry } from './telemetry/telemetry';
 
-export const taskTelemetry = async (config: Config) => {
-  const p = config.logger.dim(config.sys.details.platform === 'windows' ? '>' : '$');
-  const logger = config.logger;
-  const isEnabling = config.flags.args.includes("on");
-  const isDisabling = config.flags.args.includes("off");
-  const INFORMATION = `Opt in or our of telemetry. Information about the data we collect is available on our website: ${logger.bold('https://stenciljs.com/telemetry')}`
+export const taskTelemetry = async () => {
+  const logger = getLogger();
+  const prompt = logger.dim(getCompilerSystem().details.platform === 'windows' ? '>' : '$');
+  const isEnabling = getStencilCLIConfig().flags.args.includes('on');
+  const isDisabling = getStencilCLIConfig().flags.args.includes('off');
+  const INFORMATION = `Opt in or our of telemetry. Information about the data we collect is available on our website: ${logger.bold(
+    'https://stenciljs.com/telemetry',
+  )}`;
   const THANK_YOU = `Thank you for helping to make Stencil better! 💖`;
   const ENABLED_MESSAGE = `${logger.green('Enabled')}. ${THANK_YOU}\n\n`;
   const DISABLED_MESSAGE = `${logger.red('Disabled')}\n`;
   let hasTelemetry = await checkTelemetry();
 
   if (isEnabling) {
-    await enableTelemetry()
-    console.log(`\n  ${logger.bold('Telemetry is now ') + ENABLED_MESSAGE}`)
+    const result = await enableTelemetry();
+    result
+      ? console.log(`\n  ${logger.bold('Telemetry is now ') + ENABLED_MESSAGE}`)
+      : console.log(`Something went wrong when enabling Telemetry.`);
     return;
   }
 
   if (isDisabling) {
-    await disableTelemetry()
-    console.log(`\n  ${logger.bold('Telemetry is now ') + DISABLED_MESSAGE}`)
+    const result = await disableTelemetry();
+    result
+      ? console.log(`\n  ${logger.bold('Telemetry is now ') + DISABLED_MESSAGE}`)
+      : console.log(`Something went wrong when disabling Telemetry.`);
     return;
   }
 
-  console.log(`  ${logger.bold('Telemetry:')} ${logger.dim(INFORMATION)}`)
+  console.log(`  ${logger.bold('Telemetry:')} ${logger.dim(INFORMATION)}`);
 
-  console.log(`\n  ${logger.bold('Status')}: ${hasTelemetry ? ENABLED_MESSAGE : DISABLED_MESSAGE}`)
+  console.log(`\n  ${logger.bold('Status')}: ${hasTelemetry ? ENABLED_MESSAGE : DISABLED_MESSAGE}`);
 
-  console.log(`    ${p} ${logger.green('stencil telemetry [off|on]')}
+  console.log(`    ${prompt} ${logger.green('stencil telemetry [off|on]')}
 
         ${logger.cyan('off')} ${logger.dim('.............')} Disable sharing anonymous usage data
         ${logger.cyan('on')} ${logger.dim('..............')} Enable sharing anonymous usage data
--- a/src/cli/telemetry/helpers.ts
+++ b/src/cli/telemetry/helpers.ts
@@ -51,3 +51,11 @@ export async function readJson(path: string) {
   const file = await getCompilerSystem().readFile(path);
   return !!file && JSON.parse(file);
 }
+
+export function hasDebug() {
+  return process.argv.includes('--debug');
+}
+
+export function hasVerbose() {
+  return process.argv.includes('--verbose') && hasDebug();
+}
--- a/src/cli/telemetry/helpers.ts
+++ b/src/cli/telemetry/helpers.ts
@@ -37,6 +37,8 @@ export const isInteractive = (object?: TerminalInfo): boolean => {
   return terminalInfo.tty && !terminalInfo.ci;
 };
 
+export const UUID_REGEX = new RegExp(/^[0-9A-F]{8}-[0-9A-F]{4}-4[0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}$/i);
+
 // Plucked from https://github.com/ionic-team/capacitor/blob/b893a57aaaf3a16e13db9c33037a12f1a5ac92e0/cli/src/util/uuid.ts
 export function uuidv4(): string {
   return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, c => {
--- a/src/cli/telemetry/helpers.ts
+++ b/src/cli/telemetry/helpers.ts
@@ -53,9 +53,9 @@ export async function readJson(path: string) {
 }
 
 export function hasDebug() {
-  return process.argv.includes('--debug');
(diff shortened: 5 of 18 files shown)


src/cli/telemetry/ipc.ts
```diff
@@ -0,0 +1,89 @@
...
+}
+
+/**
+ * Receive and handle an IPC message.
+ *
+ * Assume minimal context and keep external dependencies to a minimum.
+ */
+export async function receive(msg: IPCMessage): Promise<void> {
+	debug('Received %O IPC message', msg.type);
+	console.log("WOOHOO")
```
ltm, line 48:
```suggestion
```

splitinfinities: Amended

src/cli/telemetry/telemetry.ts
```diff
@@ -0,0 +1,203 @@
...
+		const end = new Date();
+		duration = end.getTime() - start.getTime();
+	}
+	
+	const token = await getTelemetryToken();
+	const packages = await getInstalledPackages();
+	const targets = await getActiveTargets(config);
+
+	const data: TrackableData = {
+		token,
```
ltm, line 90: The service expects the telemetry token to be named `session_id`:
```suggestion
		session_id: token,
```

src/cli/telemetry/telemetry.ts
```diff
@@ -0,0 +1,203 @@
...
+		duration,
+		targets,
+		packages: packages || [],
```
ltm, line 93: I'm not 100% sure that we support arrays in the telemetry data. We'll have to test that.

src/cli/telemetry/telemetry.ts
```diff
@@ -0,0 +1,203 @@
...
+		arguments: flags.args,
+		ci: !!flags.ci,
```
ltm, line 95: Since we never send telemetry in CI this flag is presumably redundant.

src/cli/telemetry/telemetry.ts
```diff
@@ -0,0 +1,203 @@
...
+		stencil: coreCompiler.version,
+		system: `${sys.name} ${sys.version}`,
+		platform: `${details.platform} (${details.release})`,
+		compiler: sys.getCompilerExecutingPath(),
+		build: coreCompiler.buildId,
+		typescript: versions.typescript,
+		rollup: versions.rollup
+	};
+
+	if (await shouldTrack(data.ci)) {
```
ltm, line 106: Would it make sense to check this earlier in the function and return early?

src/cli/telemetry/telemetry.ts
```diff
@@ -0,0 +1,203 @@
...
+		token,
+		duration,
+		targets,
+		packages: packages || [],
+		arguments: flags.args,
+		ci: !!flags.ci,
+		task: flags.task,
+		stencil: coreCompiler.version,
+		system: `${sys.name} ${sys.version}`,
+		platform: `${details.platform} (${details.release})`,
```
ltm, line 99: `os_name` and `os_version` have special meaning in the backend, so I would separate these:

```suggestion
		os_name: details.platform,
		os_version: details.release,
```

src/cli/telemetry/telemetry.ts
```diff
@@ -0,0 +1,203 @@
...
+		if (target.type === "www") {
+			outputTargetSettings.serviceWorker = isObject(target?.serviceWorker)
+		}
+
+		return [target.type, outputTargetSettings]
+	});
+}
+
+async function getInstalledPackages() {
+	// Read package.json
```
ltm, line 129: Rather than reading the package.json it might be more useful to determine which versions are actually installed. You can get this info via e.g. `npm ls @stencil/core`, but there may be a better/more efficient way.

src/cli/telemetry/telemetry.ts
splitinfinities: If it doesn't work well, I cal swap to a comma separated string... I'll follow your lead on how best to handle it.

splitinfinities: Adjusted this to be an array of strings.

src/cli/telemetry/telemetry.ts
splitinfinities: I took a separate approach, let me know what you think.

src/cli/telemetry/helpers.ts
```diff
@@ -1,28 +1,10 @@
 import { getCompilerSystem } from '../state/stencil-cli-config';
 
-interface TerminalInfo {
-  /**
```
rwaskiewicz, line 4: FWIW, I think these JSDocs were fine to leave in  - I don't have definitive plans, but I'd like us to move to a 'better' documented codebase ('better' is subjective and can take on a variety of forms 😉 )

src/cli/telemetry/telemetry.ts
```diff
@@ -0,0 +1,293 @@
...
+  command: string;
+  arguments: string;
+  options: string;
+  duration: number;
+  error: string | null;
+  node_version: string;
+  // os: string;
+}
+
+export interface Metric<N extends string, D> {
```
rwaskiewicz, line 20: For the interfaces in this file, consider adding a JSDoc that gives a high level description of the interface. That way, when someone mouses over the interface, an editor's intellisense can provide that description to the dev, rather than making them do archaeology to figure out what objects of this shape/interface are

src/cli/telemetry/telemetry.ts
```diff
@@ -0,0 +1,293 @@
...
+    build: `${result.buildId}`,
+    typescript: versions.typescript,
+    rollup: versions.rollup,
+  };
+
+  await sendMetric('stencil_cli_command', data);
+  logger.debug(`${logger.blue('Telemetry')}: ${logger.gray(JSON.stringify(data))}`);
+}
+
+export async function telemetryAction(options: telemetryActionArgs) {
```
rwaskiewicz, line 104: Can we get these publicly exposed methods under test please?

src/cli/telemetry/telemetry.ts
```diff
@@ -0,0 +1,293 @@
...
+
+    return versions;
+  } catch (err) {
+    hasDebug() && console.error(err);
+    return [];
+  }
+}
+
+export async function shouldTrack(ci: boolean) {
+  return (await checkTelemetry()) && !ci && isInteractive();
```
rwaskiewicz, line 220: You can short circuit potentially more intensive tasks (reading user config) by reordering the conditional here
```suggestion
  return !ci && isInteractive() && (await checkTelemetry());
```
That said, I don't know how _much_ effort we'd save without actually measuring

src/cli/telemetry/telemetry.ts
```diff
@@ -0,0 +1,293 @@
...
+      : ionicPackages.map(([k, v]) => `${k}@${v}`);
+
+    return versions;
+  } catch (err) {
+    hasDebug() && console.error(err);
+    return [];
+  }
+}
+
+export async function shouldTrack(ci: boolean) {
```
rwaskiewicz, line 219: This function looks like it's only used in this file, so we can probably remove `export` here
```suggestion
async function shouldTrack(ci: boolean) {
```

src/cli/telemetry/telemetry.ts
```diff
@@ -0,0 +1,293 @@
...
+    }
+
+    return [target.type, outputTargetSettings];
+  });
+}
+
+async function getInstalledPackages() {
+  try {
+    // Read package.json and package-lock.json
+    const appRootDir = process.cwd();
```
rwaskiewicz, line 182: Is there an abstraction that allows to this work in Deno as well?

src/cli/telemetry/telemetry.ts
```diff
@@ -0,0 +1,293 @@
...
+      readJson,
+      getCompilerSystem().resolvePath(appRootDir + '/package.json'),
+    );
+    const packageLockJson: any = await tryFn(
+      readJson,
+      getCompilerSystem().resolvePath(appRootDir + '/package-lock.json'),
+    );
+
+    const packages = Object.entries({
+      ...packageJson.devDependencies,
```
rwaskiewicz, line 194: It really _shouldn't_ be, but what happens when `tryFn` returns `null` in the event of a read failure? Does this cause an `Error` trying to access a property on `undefined`?

I think that the type declaration on [this line](https://github.com/ionic-team/stencil/pull/2964/files#diff-2a84eb0343da5e51e060e0697c4c6fb67c2cc98b591ca94599ac984573787c3cR184) is tricking the compiler into thinking this can never be `null`, but I'm not sure that's the case.

I would recommend returning an empty array & logging the error in the event we cannot read their `package.json` or `package-lock.json`

src/cli/telemetry/telemetry.ts
```diff
@@ -0,0 +1,293 @@
...
+      ...packageJson.dependencies,
+    });
+
+    // Collect packages only in the stencil, ionic, or capacitor org's:
+    // https://www.npmjs.com/org/stencil
+    const ionicPackages = packages.filter(
+      ([k]) => k.startsWith('@stencil/') || k.startsWith('@ionic/') || k.startsWith('@capacitor/'),
+    );
+
+    // Original method, but cross reference with package-lock.json, don't consider yarn.
```
rwaskiewicz, line 204: I'm not sure this comment fully makes sense to me. Is it 'original' in comparison to the `filter` call above?

src/cli/telemetry/telemetry.ts
```diff
@@ -0,0 +1,293 @@
...
+    const versions = packageLockJson
+      ? ionicPackages.map(
+          ([k, v]) =>
+            `${k}@${packageLockJson.dependencies[k].version || packageLockJson.devDependencies[k].version || v}`,
```
rwaskiewicz, line 208: Again, this _shouldn't_ happen, but if for whatever reason either `dependencies` or `devDependencies` is not defined, consider:
```suggestion
            `${k}@${packageLockJson.dependencies?.[k].version || packageLockJson.devDependencies?.[k].version || v}`,

```

src/cli/task-telemetry.ts
```diff
@@ -0,0 +1,36 @@
...
+  const isEnabling = config.flags.args.includes("on");
+  const isDisabling = config.flags.args.includes("off");
+  const INFORMATION = `Opt in or our of telemetry. Information about the data we collect is available on our website: ${logger.bold('https://stenciljs.com/telemetry')}`
+  const THANK_YOU = `Thank you for helping to make Stencil better! 💖`;
+  const ENABLED_MESSAGE = `${logger.green('Enabled')}. ${THANK_YOU}\n\n`;
+  const DISABLED_MESSAGE = `${logger.red('Disabled')}\n`;
+  let hasTelemetry = await checkTelemetry();
+
+  if (isEnabling) {
+    await enableTelemetry()
```
rwaskiewicz, line 16: What happens if this fails? Right now, we only `console.error` the `Error` that occurred. As a result, we could fail to write the file, but report to the user we set the value correctly (erroneously).

Additionally, it may be worthwhile to report the value based on what's written in the config itself.  That introduces the possibility of us reporting our goof, but at least we'll always be reporting from a single source of truth

src/cli/task-telemetry.ts
```diff
@@ -0,0 +1,36 @@
...
+    console.log(`\n  ${logger.bold('Telemetry is now ') + ENABLED_MESSAGE}`)
+    return;
+  }
+
+  if (isDisabling) {
+    await disableTelemetry()
```
rwaskiewicz, line 22: What happens if this fails? Right now, we only `console.error` the `Error` that occurred. As a result, we could fail to write the file, but report to the user we set the value correctly (erroneously)

src/cli/telemetry/telemetry.ts
```diff
@@ -0,0 +1,293 @@
...
+    name,
+    timestamp: new Date().toISOString(),
+    source: 'stencil_cli',
+    value: data,
+  };
+
+  await sendTelemetry({ type: 'telemetry', message });
+}
+
+export async function sendTelemetry(data: any) {
```
rwaskiewicz, line 255: This function only appears to be used in this file. I think we can safely remove `export`:
```suggestion
async function sendTelemetry(data: any) {
```

rwaskiewicz:
@splitinfinities setting a reminder for both of us to reset the target branch on Monday. Have a great weekend!

src/cli/telemetry/telemetry.ts
```diff
@@ -0,0 +1,293 @@
...
+import { readConfig, updateConfig } from '../ionic-config';
+import { isObject } from '@utils';
+import { getCompilerSystem } from '../state/stencil-cli-config';
+
+export interface CommandMetricData {
+  // app_id: string;
+  command: string;
+  arguments: string;
+  options: string;
+  duration: number;
```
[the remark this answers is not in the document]
splitinfinities: I think this will work fine, I'll defer to Lars when he reviews this part!

src/cli/telemetry/telemetry.ts
```diff
@@ -0,0 +1,293 @@
...
+import { CompilerBuildResults, CompilerSystem, Config, Logger, PackageJsonData } from 'src/declarations';
+import { parseFlags } from '../parse-flags';
+import { loadCoreCompiler } from '../load-compiler';
+import { readConfig, updateConfig } from '../ionic-config';
+import { isObject } from '@utils';
+import { getCompilerSystem } from '../state/stencil-cli-config';
+
+export interface CommandMetricData {
+  // app_id: string;
+  command: string;
```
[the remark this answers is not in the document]
splitinfinities: This type was left unused, pulled from some other Ionic code, so I deleted it. I did however place this in the TrackableData interface.

src/cli/task-help.ts
```diff
@@ -31,14 +35,17 @@ export const taskHelp = (sys: CompilerSystem, logger: Logger) => {
...
-    ${p} ${logger.green('stencil generate')}
-    ${p} ${logger.green('stencil g my-component')}
-
-`);
+  ${p} ${logger.green('stencil build --dev --watch --serve')}
+  ${p} ${logger.green('stencil build --prerender')}
+  ${p} ${logger.green('stencil test --spec --e2e')}
+  ${p} ${logger.green('stencil generate')}
+  ${p} ${logger.green('stencil g my-component')}
+`)
```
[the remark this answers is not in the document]
splitinfinities: That's a great idea!

src/cli/task-telemetry.ts
```diff
@@ -0,0 +1,36 @@
+import type { Config} from '../declarations';
+import { checkTelemetry, disableTelemetry, enableTelemetry } from './telemetry/telemetry';
+
+export const taskTelemetry = async (config: Config) => {
+  const p = config.logger.dim(config.sys.details.platform === 'windows' ? '>' : '$');
```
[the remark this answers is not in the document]
splitinfinities: Switching these to "prompt"

src/cli/task-telemetry.ts
```diff
@@ -0,0 +1,36 @@
...
+  const logger = config.logger;
```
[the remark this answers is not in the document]
splitinfinities: Rewrote this to reference the singleton

src/cli/telemetry/helpers.ts
splitinfinities: Brought em back!

src/cli/telemetry/telemetry.ts
splitinfinities: I will have to determine that...

src/cli/telemetry/telemetry.ts
splitinfinities: Nixed

src/cli/task-telemetry.ts
splitinfinities: Added!

src/cli/telemetry/telemetry.ts
splitinfinities: I'm pretty sure I added some adequate error handling here. Let me know if you can think of anything else.

src/cli/telemetry/telemetry.ts
splitinfinities: Working on this rn .

src/cli/telemetry/telemetry.ts
splitinfinities: Patched up

src/cli/task-help.ts
```diff
@@ -1,12 +1,16 @@
-import type { CompilerSystem, Logger } from '../declarations';
+import type { Config } from '../declarations';
+import { taskTelemetry } from './task-telemetry';
 
-export const taskHelp = (sys: CompilerSystem, logger: Logger) => {
-  const p = logger.dim(sys.details.platform === 'windows' ? '>' : '$');
+export const taskHelp = async (config: Config) => {
```
rwaskiewicz, line 4: Is there a need to pass `Config` as an argument here, could we instead call `getCompilerSystem()` and `getLogger()` on `stencil-cli-config.ts`? What's the preferred interface to use for this singleton?

src/cli/task-telemetry.ts
```diff
@@ -0,0 +1,42 @@
...
+  const prompt = logger.dim(getCompilerSystem().details.platform === 'windows' ? '>' : '$');
+  const isEnabling = getStencilCLIConfig().flags.args.includes('on');
+  const isDisabling = getStencilCLIConfig().flags.args.includes('off');
+  const INFORMATION = `Opt in or our of telemetry. Information about the data we collect is available on our website: ${logger.bold(
+    'https://stenciljs.com/telemetry',
+  )}`;
+  const THANK_YOU = `Thank you for helping to make Stencil better! 💖`;
+  const ENABLED_MESSAGE = `${logger.green('Enabled')}. ${THANK_YOU}\n\n`;
+  const DISABLED_MESSAGE = `${logger.red('Disabled')}\n`;
+  let hasTelemetry = await checkTelemetry();
```
rwaskiewicz, line 15: This variable could be `const`, as we never reassign it
```suggestion
  const hasTelemetry = await checkTelemetry();
```

src/cli/task-telemetry.ts
```diff
@@ -0,0 +1,42 @@
...
+  const logger = getLogger();
+  const prompt = logger.dim(getCompilerSystem().details.platform === 'windows' ? '>' : '$');
+  const isEnabling = getStencilCLIConfig().flags.args.includes('on');
+  const isDisabling = getStencilCLIConfig().flags.args.includes('off');
+  const INFORMATION = `Opt in or our of telemetry. Information about the data we collect is available on our website: ${logger.bold(
+    'https://stenciljs.com/telemetry',
+  )}`;
+  const THANK_YOU = `Thank you for helping to make Stencil better! 💖`;
+  const ENABLED_MESSAGE = `${logger.green('Enabled')}. ${THANK_YOU}\n\n`;
+  const DISABLED_MESSAGE = `${logger.red('Disabled')}\n`;
```
rwaskiewicz, line 14: Double checking, is the double newline at the end of `ENABLED_MESSAGE` and the single newline at the end of `DISABLED_MESSAGE` intentionally different? Or should we make them uniform?

src/cli/telemetry/shouldTrack.ts
```diff
@@ -0,0 +1,11 @@
+import { isInteractive } from './helpers';
+import { checkTelemetry } from './telemetry';
+
+/**
+ * Used to determine if tracking should occur.
+ * @param ci boolean
+ * @returns boolean
```
rwaskiewicz, line 7: IMO, we can get the type information from the function header here.  I would suggest being a little more descriptive here:
```suggestion
 * @param ci whether or not the process is running in a Continuous Integration (CI) environment
 * @returns true if telemetry should be sent, false otherwise
```

src/cli/telemetry/helpers.ts
```diff
@@ -51,3 +51,11 @@ export async function readJson(path: string) {
...
   return !!file && JSON.parse(file);
 }
+
+export function hasDebug() {
+  return process.argv.includes('--debug');
+}
+
+export function hasVerbose() {
+  return process.argv.includes('--verbose') && hasDebug();
+}
```
rwaskiewicz, line 61: Can we get these functions under test please?

src/cli/telemetry/telemetry.ts
```diff
@@ -0,0 +1,304 @@
...
+export interface Metric {
+  name: string;
+  timestamp: string;
+  source: 'stencil_cli';
+  value: TrackableData;
+}
+
+/**
+ * Used for deeper understanding of how people are using certain output targets
+ **/
```
rwaskiewicz, line 20:
```suggestion
 */
```

src/cli/telemetry/telemetry.ts
```diff
@@ -0,0 +1,304 @@
...
+  rollup: string;
+  system: string;
+  build: string;
+  stencil: string;
+}
+
+/**
+ * Used to within taskBuild to provide the component_count property.
+ * @param result The results of a compiler build.
+ * @returns void
```
rwaskiewicz, line 55: I think we can omit this line if nothing is returned
```suggestion
```

src/cli/telemetry/telemetry.ts
```diff
@@ -0,0 +1,304 @@
...
+        console.debug('\nBad response from events service. Request body: %O', response.body.toString(), '\n');
+    }
+  } catch (e) {
+    hasVerbose() && console.debug('Telemetry request failed:', e);
+  }
+}
+
+/**
+ * Checks if telemetry is enabled
+ * @returns boolean
```
rwaskiewicz, line 285: See my comment above regarding having just the type in `@returns`. I think we should strive to be consistent here - we should decide _what_ we want to do for `@returns` and apply that uniformly here and to the `enableTelemetry` and `disableTelemetry` methods below

src/cli/task-telemetry.ts
```diff
@@ -0,0 +1,42 @@
+import { getCompilerSystem, getLogger, getStencilCLIConfig } from './state/stencil-cli-config';
+import { checkTelemetry, disableTelemetry, enableTelemetry } from './telemetry/telemetry';
+
+export const taskTelemetry = async () => {
+  const logger = getLogger();
+  const prompt = logger.dim(getCompilerSystem().details.platform === 'windows' ? '>' : '$');
+  const isEnabling = getStencilCLIConfig().flags.args.includes('on');
+  const isDisabling = getStencilCLIConfig().flags.args.includes('off');
```
rwaskiewicz, line 8: Similar to my comment above, I'm not sure what the appropriate interface to use is here - should the `logger` be pulled from `getStencilCLIConfig`?

src/cli/telemetry/telemetry.ts
```diff
@@ -0,0 +1,304 @@
...
+    timestamp: new Date().toISOString(),
+    source: 'stencil_cli',
+    value,
+  };
+
+  await sendTelemetry({ type: 'telemetry', message });
+}
+
+/**
+ * Used to read the ~/.ionic/config.json file's tokens.telemetry property.
```
rwaskiewicz, line 241: Food for thought - the read of ` ~/.ionic/config.json`, specifically the file name, is abstracted out to `readConfig`. Should the name/location of that file change, we may not remember to update this comment (which would make it stale/misinformation)

src/cli/telemetry/telemetry.ts
```diff
@@ -0,0 +1,304 @@
...
+export async function telemetryBuildFinishedAction(result: CompilerBuildResults) {
+  const { sys, flags, logger, validatedConfig } = getStencilCLIConfig();
+  const tracking = await shouldTrack(!!flags?.ci);
+
+  if (!tracking) {
+    return;
+  }
+
+  const details = sys.details;
+  const session_id = await getTelemetryToken();
```
rwaskiewicz, line 66: You may want to consider what happens if the user does something like open their config file and messes with the `tokens.telemetry` field. I think we're assuming that if the config file exists at the moment, `tokens.telemetry` exists and is valid

src/cli/telemetry/telemetry.ts
```diff
@@ -0,0 +1,304 @@
...
+
+  const data: TrackableData = {
+    session_id,
+    yarn: isUsingYarn(),
+    duration_ms: result.duration,
+    component_count,
+    targets,
+    packages: packages || [],
+    arguments: flags.args,
+    task: flags.task,
```
rwaskiewicz, line 80: Is it possible that `flags` can be `undefined`? I thought that's why we used optional chaining [here](https://github.com/ionic-team/stencil/pull/2964/files#diff-2a84eb0343da5e51e060e0697c4c6fb67c2cc98b591ca94599ac984573787c3cR59). Maybe I'm missing something (likely)? Otherwise I'd be concerned the compiler is missing something there 🤔

src/cli/telemetry/telemetry.ts
```diff
@@ -0,0 +1,304 @@
...
+
+    // They don't have a package.json for some reason? Eject button.
+    if (!!packageJson) {
+      return [];
+    }
+
+    const packages = Object.entries({
+      ...packageJson.devDependencies,
+      ...packageJson.dependencies,
+    }) as [string, string][];
```
rwaskiewicz, line 200: I don't know if we need this type assertion, as this call to `Object.entries` should give us a `[string, string][]` already (if you want to be declarative, there's nothing wrong with putting the type directly following `packages` like `const packages: [string, string][] = ...`)

src/cli/telemetry/telemetry.ts
```diff
@@ -0,0 +1,304 @@
...
+    );
+
+    // They don't have a package.json for some reason? Eject button.
+    if (!!packageJson) {
+      return [];
+    }
+
+    const packages = Object.entries({
+      ...packageJson.devDependencies,
+      ...packageJson.dependencies,
```
rwaskiewicz, line 199: Consider what happens in the user doesn't have a `dependencies` or `devDependencies` section, we'll end up with a runtime error trying to spread `undefined`

src/cli/telemetry/telemetry.ts
```diff
@@ -0,0 +1,304 @@
...
+    // https://www.npmjs.com/org/stencil
+    const ionicPackages = packages.filter(
+      ([k]) => k.startsWith('@stencil/') || k.startsWith('@ionic/') || k.startsWith('@capacitor/'),
+    );
+
+    const versions = packageLockJson
+      ? ionicPackages.map(
+          ([k, v]) =>
+            `${k}@${
+              (packageLockJson?.dependencies[k] as { version: string })?.version ??
```
rwaskiewicz, line 212: FWIW I don't think we need these type assertions here - since `packageLockJson` is on `any` type, the compiler _should_ be OK with these accesses

src/cli/telemetry/test/telemetry.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+    });
+
+    expect(spyShouldTrack).toHaveBeenCalled();
+
+    spyShouldTrack.mockRestore();
+  });
+});
+
+describe('checkTelemetry', () => {
+  it('will read from a file', async () => {
```
rwaskiewicz, line 70: I think this test does a little more than just read 🙂 It also toggles the status of telemetry

rwaskiewicz (changes requested):

src/cli/ionic-config.ts
```diff
@@ -26,16 +26,20 @@ export async function readConfig(): Promise<TelemetryConfig> {
...
   }
+
+  return result;
 }
 
-export async function updateConfig(newOptions: TelemetryConfig): Promise<void> {
+export async function updateConfig(newOptions: TelemetryConfig): Promise<boolean> {
   const config = await readConfig();
-  await writeConfig(Object.assign(config, newOptions));
+  return await writeConfig(Object.assign(config, newOptions));
```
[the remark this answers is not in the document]
splitinfinities: Oh, great point! Yes.

src/cli/task-help.ts
```diff
@@ -21,24 +25,29 @@ export const taskHelp = (sys: CompilerSystem, logger: Logger) => {
...
-    ${p} ${logger.green('stencil generate')}
-    ${p} ${logger.green('stencil g my-component')}
+  console.log(`
+  ${logger.bold('Examples:')}
 
+  
+  ${prompt} ${logger.green('stencil build --dev --watch --serve')}
+  ${prompt} ${logger.green('stencil build --prerender')}
+  ${prompt} ${logger.green('stencil test --spec --e2e')}
+  ${prompt} ${logger.green('stencil telemetry on')}
```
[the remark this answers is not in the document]
splitinfinities: I considered that, but these seem to be written as commands you can try by copy/paste. Does that make sense?

src/cli/task-help.ts
splitinfinities: I would love to migrate to use the singleton in more cases like what you point out - I just didn't want to change too much. I personally enjoy the ergonomics of function calls without any props, as you can tell from the code I added. 

For the singleton, my thought process was, as we come in and make modifications, let's refactor code to use the singleton more often. I'm happy to go farther, but want to keep this PR scoped closer to Telemetry than general cleanup and ergonomics.

src/cli/telemetry/helpers.ts
splitinfinities: Yes, plus I need to switch to the sys here actually.

src/cli/telemetry/telemetry.ts
splitinfinities: This works as intended where an array is returned despite an undefined spread. 

Screen Shot 2021-07-27 at 1 41 54 PM
 
Screen Shot 2021-07-27 at 1 41 29 PM

src/cli/task-telemetry.ts
splitinfinities: `getLogger()` does pull from the `getStencilCLIConfig()` singleton - it's a helper method to return that similar to `getCompilerSystem()`. I may also be misreading your question. Are you asking if `logger` should be stored on the singleton, as in philosophically?

src/cli/telemetry/telemetry.ts
splitinfinities: I didn't want to go further than what Ionic's other products do for telemetry. Are you asking me to validate the content of the tokens.telemetry just in case someone has edited the file? I can certainly do that if you would like.

src/cli/telemetry/telemetry.ts
splitinfinities: I've removed that optional chain.

src/cli/telemetry/helpers.ts
splitinfinities: Done

src/cli/telemetry/telemetry.ts
splitinfinities: Adde

src/cli/run.ts
```diff
@@ -42,11 +46,6 @@ export const run = async (init: CliInitOptions) => {
     stencilCLIConfig.sys = sys;
     stencilCLIConfig.flags = flags;
 
-    if (task === 'help' || flags.help) {
```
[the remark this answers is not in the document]
ltm: Did this actually get resolved? From a quick glance this block just got moved up a few lines. If it did get resolved, is there a reason it was moved? cc @rwaskiewicz

src/cli/task-help.ts
```diff
@@ -1,12 +1,16 @@
-import type { CompilerSystem, Logger } from '../declarations';
+import type { Config } from '../declarations';
+import { taskTelemetry } from './task-telemetry';
 
-export const taskHelp = (sys: CompilerSystem, logger: Logger) => {
-  const p = logger.dim(sys.details.platform === 'windows' ? '>' : '$');
+export const taskHelp = async (config: Config) => {
+  const logger = config.logger;
+  const sys = config.sys;
```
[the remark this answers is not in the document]
ltm: ...or even `export const taskHelp = async ({ logger, sys }: Config) => {`, but that's certainly a matter of preference.

src/sys/node/node-sys.ts
```diff
@@ -99,6 +99,19 @@ export function createNodeSys(c: { process?: any } = {}) {
       opts.window.Response = global.Response;
       opts.window.FetchError = (global as any).FetchError;
     },
+    fetch: (input: any, init: any) => {
+      const nodeFetch = require(path.join(__dirname, 'node-fetch.js'));
```
rwaskiewicz, line 103: Does it make sense to have a conditional on `global.fetch` like `applyPrerenderGlobalPatch` does? ([Link to function above this one](https://github.com/ionic-team/stencil/pull/2964/commits/d52a7d1134d70fe153dfa9840c313cfc7210eccb#diff-585e96e5364c622963a06a9d71c01a77ad8816bc84e97763dcefe7920176dd50R75)) - That way we don't need to redefine it, right?

src/compiler/sys/stencil-sys.ts
```diff
@@ -506,6 +505,8 @@ export const createSystem = (c?: { logger?: Logger }) => {
     return results;
   };
 
+  const fetch = global.fetch;
```
rwaskiewicz, line 508: For my own knowledge, should we be assigning `fetch` to `global` in our new `fetch` method [here](https://github.com/ionic-team/stencil/pull/2964/commits/d52a7d1134d70fe153dfa9840c313cfc7210eccb#diff-585e96e5364c622963a06a9d71c01a77ad8816bc84e97763dcefe7920176dd50R102)?

src/cli/telemetry/telemetry.ts
rwaskiewicz: Ah you're right! TI(re-)Learned!

src/cli/task-help.ts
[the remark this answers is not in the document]
rwaskiewicz: Ah, that's fair, I hadn't considered the easy copy/paste angle here. Feel free to leave as is

src/cli/run.ts
```diff
@@ -119,6 +128,14 @@ export const runTask = async (coreCompiler: CoreCompiler, config: Config, task:
   config.outputTargets = config.outputTargets || [];
 
   switch (task) {
+    case 'telemetry':
```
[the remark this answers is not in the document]
splitinfinities: Sorted!

src/cli/task-help.ts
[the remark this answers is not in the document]
splitinfinities: Reading Lar's feedback here, I decided to switch to the singleton reference here instead. Since I had already edited this task, I felt like it was a good idea to refactor the signature. I was just trying to avoid making that common place in the PR because I also edited the build task.

src/cli/task-help.ts
splitinfinities: See https://github.com/ionic-team/stencil/pull/2964#discussion_r678323780 - I switched this out.

src/cli/task-telemetry.ts
rwaskiewicz: To me it seems like there's two ways to get a a logger instance:
```
const loggerViaMethod = getLogger();
const loggerViaSingletonAccessor = getStencilCLIConfig().logger;
```
As a user of the singleton, it's not exactly clear to me if there's a preferred way to get the logger without reading the code that the former is a wrapper around the latter.  The ergonomics of pulling data off the singleton seems a little inconsistent to me - for some fields I need to call `getStencilCLIConfig()`, others have a helper function.  

That said, I'm not sure adding a helper function for each private method necessarily helps here. It gives us consistency/parity where there's a helper function for each private field that we expose via an accessor, but it doesn't answer the "which should I use?" question.

That's all a long winded way of saying "I think this is fine as is for now, because I don't have a great solution to offer up ATM". It's something we should be cognizant of as we evolve this class, but no action required ATM

src/cli/telemetry/telemetry.ts
rwaskiewicz: How about this, let's make that a follow on story - I think there's some discussion to be had there. IDK if there's any validation on the data type in our data store that would already reject the field if it was malformed. Would you mind adding something to our refinement spreadsheet for next Monday's session?

src/cli/telemetry/telemetry.ts
rwaskiewicz: Does that solve the issue though? We define `_flags` as:
```typescript
export default class StencilCLIConfig {
  static instance: StencilCLIConfig;

  private _args: string[];
  private _logger: Logger;
  private _sys: CompilerSystem;
  private _flags: ConfigFlags | undefined;
  private _task: string | undefined;
  private _coreCompiler: CoreCompiler | undefined;
  private _validatedConfig: LoadConfigResults | undefined;

  private constructor(options: StencilCLIConfigArgs) {
    this._args = options?.args || [];
    this._logger = options?.logger;
    this._sys = options?.sys;
    this._flags = options?.flags || undefined;
    this._validatedConfig = options?.validatedConfig || undefined;
  }

  public static getInstance(options?: StencilCLIConfigArgs): StencilCLIConfig {
    if (!StencilCLIConfig.instance) {
      StencilCLIConfig.instance = new StencilCLIConfig(options);
    }

    return StencilCLIConfig.instance;
  }

  public get flags() {
    return this._flags;
  }
  public set flags(flags: ConfigFlags) {
    this._flags = flags;
  }
```
so it's possible we called the ctor for the first time with an empty object literal
```typescript
const baz = getInstance({});
```
which sets `_flags_` to `undefined` right? Barring setting that somewhere else, this would be an unsafe access

src/cli/ionic-config.ts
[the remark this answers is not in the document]
rwaskiewicz: @splitinfinities did we get this under test?

src/sys/node/node-sys.ts
splitinfinities: I'm going to reference this here too: https://github.com/ionic-team/stencil/pull/2964#discussion_r678304924

Looking at this again... Initially I thought we could just call the `applyPrerenderGlobalPatch` function prior to returning the fetch method, but it's signature is designed for the dev server - see lines 81 and 85 where the dev server is hard coded in. Reviewing this, yes I think I should guard this so we don't mandate loading that polyfill code for fetch. 

I switched to a ternary on the first line of this function. That worked great in my test. Good eye here!

splitinfinities: See https://github.com/ionic-team/stencil/pull/2964/commits/5ca6854a933b5162906f2ea97aba464cff02a7e3 for the ternary

src/cli/task-telemetry.ts
splitinfinities: Generally you'll see me throughout this PR use the helper functions, until I need to update the value, at which point I'll use: 

`getStencilCLIConfig().logger = updatedLogger`

And the helper methods pretty much only try to help compensate line length. 

Don't worry, I'm not married to this interaction, but at a minimum the singleton definitely helps.

src/compiler/sys/stencil-sys.ts
splitinfinities: Yes, and that has been updated!

src/cli/telemetry/telemetry.ts
```diff
@@ -0,0 +1,301 @@
...
+  }
+}
+
+/**
+ * Checks if telemetry is enabled on this machine
+ * @returns true if telemetry is enabled, false otherwise
+ */
+export async function checkTelemetry(): Promise<boolean> {
+  const config = await readConfig();
+  return config['telemetry.stencil'];
```
ltm, line 284: If the user has an existing `~/.ionic/config.json` generated by the Ionic CLI, then `telemetry.stencil` will be missing and this function will return `undefined`.

```suggestion
  const config = await readConfig();
  if (config['telemetry.stencil'] === undefined) {
    config['telemetry.stencil'] = true;
    await writeConfig(config);
  }
  return config['telemetry.stencil'];
```

src/cli/telemetry/telemetry.ts
```diff
@@ -0,0 +1,301 @@
...
+  await sendTelemetry({ type: 'telemetry', message });
+}
+
+/**
+ * Used to read the config file's tokens.telemetry property.
+ * @returns string
+ */
+async function getTelemetryToken() {
+  const config = await readConfig();
+  return config['tokens.telemetry'];
```
ltm, line 241: Similar to `checkTelemetry()` this could end up returning undefined if the user has an existing Ionic config.
```suggestion
  const config = await readConfig();
  if (config['tokens.telemetry'] === undefined) {
    config['tokens.telemetry'] = uuidv4();
    await writeConfig(config);
  }
  return config['tokens.telemetry'];
```
If you add both suggestions we can remove the `if (!config)` block from `readConfig()` as well.

ltm (changes requested):

src/cli/telemetry/telemetry.ts
splitinfinities: @ltm Nice! I will add these and update that spot. Great catch! Thank you!

src/sys/node/node-sys.ts
splitinfinities: Okay I feel great about this, but revisiting https://github.com/ionic-team/stencil/pull/2964#discussion_r678304924, I think I may have to do something over there to consider fetch in the browser or in tests more appropriately.

src/cli/telemetry/telemetry.ts
```diff
@@ -0,0 +1,309 @@
...
+  await sendMetric('stencil_cli_command', data);
+  logger.debug(`${logger.blue('Telemetry')}: ${logger.gray(JSON.stringify(data))}`);
+
+  if (error) {
+    throw error;
+  }
+}
+
+function isUsingYarn() {
+  return getCompilerSystem().getEnvironmentVar('npm_execpath').includes('yarn');
```
ltm, line 157: This is throwing `TypeError: Cannot read property 'includes' of undefined` when the CLI is invoked directly.
```suggestion
  return getCompilerSystem().getEnvironmentVar('npm_execpath')?.includes('yarn') || false;
```

src/cli/telemetry/telemetry.ts
```diff
@@ -0,0 +1,309 @@
...
+
+  const details = sys.details;
+  const session_id = await getTelemetryToken();
+  const packages = await getInstalledPackages();
+  const targets = await getActiveTargets(validatedConfig.config);
+  const versions = getCoreCompiler()?.versions || { typescript: 'unknown', rollup: 'unknown' };
+  const component_count = Object.keys(result.componentGraph).length;
+
+  const data: TrackableData = {
+    session_id,
```
ltm, line 72: Sorry, I missed this earlier. The `session_id` should be a part of `Metric` and be set on the message in `sendMetric()`.
```suggestion
```

src/cli/telemetry/telemetry.ts
```diff
@@ -0,0 +1,309 @@
...
+import { isObject } from '@utils';
+import { getCompilerSystem, getCoreCompiler, getStencilCLIConfig } from '../state/stencil-cli-config';
+
+/**
+ * Used as the object sent to the server. Value is the data tracked.
+ */
+export interface Metric {
+  name: string;
+  timestamp: string;
+  source: 'stencil_cli';
```
ltm, line 14:
```suggestion
  timestamp: string;
  session_id: string;
  source: 'stencil_cli';
```

src/cli/telemetry/telemetry.ts
```diff
@@ -0,0 +1,309 @@
...
+ * The task to run in order to collect the duration data point.
+ */
+type TelemetryCallback = (...args: any[]) => void | Promise<void>;
+
+/**
+ * The model for the data that's tracked.
+ */
+export interface TrackableData {
+  yarn: boolean;
+  session_id: string;
```
ltm, line 35:
```suggestion
```

ltm (changes requested):

src/cli/telemetry/telemetry.ts
splitinfinities: Ohhh! Got you.

src/cli/ionic-config.ts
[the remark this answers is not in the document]
splitinfinities: Is this adequate? https://github.com/ionic-team/stencil/pull/2964/files#diff-a2d702e7c67979d0440b8322026a5249d8206bbeaf3e73f7c84df87cfdd2d07cR69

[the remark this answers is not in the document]
rwaskiewicz: I think this is fine ATM - normally I'd like us to get a little more granular there with respect to what we're testing (calling the `ionic-config` exposed functions directly) and splitting the tests out based on the assertion, but I think we can leave that for a future day

src/cli/ionic-config.ts
```diff
@@ -26,16 +26,20 @@ export async function readConfig(): Promise<TelemetryConfig> {
   return config;
 }
 
-export async function writeConfig(config: TelemetryConfig): Promise<void> {
+export async function writeConfig(config: TelemetryConfig): Promise<boolean> {
+  let result = false;
   try {
     await getCompilerSystem().createDir(defaultConfigDirectory(), { recursive: true });
     await getCompilerSystem().writeFile(defaultConfig(), JSON.stringify(config));
```
ltm, line 33: The [Ionic CLI](https://github.com/ionic-team/ionic-cli/blob/develop/packages/@ionic/cli-framework/src/lib/config.ts#L97) indents the config file with two spaces. We should match that for consistency:

```suggestion
    await getCompilerSystem().writeFile(defaultConfig(), JSON.stringify(config, null, 2));
```

src/cli/telemetry/telemetry.ts
```diff
@@ -0,0 +1,302 @@
...
+  return getCompilerSystem().getEnvironmentVar('npm_execpath')?.includes('yarn') || false;
+}
+
+async function getActiveTargets(config: Config): Promise<string[]> {
+  const result = (config.outputTargets as any[]).map(target => {
+    let type = target.type;
+
+    if (type === 'www' && (!!target?.serviceWorker || target.baseUrl !== '/')) {
+      type = `${type} (PWA/App)`;
+    }
```
ltm, line 154: Rather than overloading the output targets, I'd add a separate property (e.g. `has_app_target`) which you might be able to implement as:
```
function hasAppTarget(config: Config): boolean {
  return config.outputTargets.some(target => target.type === 'www' && (!!target.serviceWorker || target.baseUrl !== '/'));
}
```

splitinfinities: Ohhh, that's a great idea!

splitinfinities: Added it as `has_app_pwa_config` on TrackableData.

splitinfinities: See https://github.com/ionic-team/stencil/pull/2964/commits/c25196259c26085c4272bccd0f0625128ebccb08

src/cli/run.ts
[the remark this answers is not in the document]
splitinfinities: FYI I moved this back to where it originally was.

src/cli/telemetry/telemetry.ts
splitinfinities: Considering that use case you mentioned couldn't be possible as the code is written today, I don't think we'll experience that issue.

src/cli/test/ionic-config.spec.ts
```diff
@@ -2,6 +2,10 @@ import { mockLogger } from '@stencil/core/testing';
 import { getCompilerSystem, initializeStencilCLIConfig } from '../state/stencil-cli-config';
 import { readConfig, writeConfig, updateConfig, defaultConfig } from '../ionic-config';
 import { createSystem } from '../../compiler/sys/stencil-sys';
+import { UUID_REGEX } from '../telemetry/helpers';
+
+const UUID1 = '5588e0f0-02b5-4afa-8194-5d8f78683b36';
+const UUID2 = 'e5609819-5c24-4fa2-8817-e05ca10b8cae';
```
rwaskiewicz, line 8: IIUC, this will still write back to `~/.ionic/config.json` right? Do we have a story on the books to have this write to a test file & have the suite clean up after itself after the face??

src/cli/ionic-config.ts
```diff
@@ -21,6 +21,10 @@ export async function readConfig(): Promise<TelemetryConfig> {
     };
 
     await writeConfig(config);
+  } else if (!!config && !config['tokens.telemetry'].match(UUID_REGEX)) {
+    const newUuid = uuidv4();
+    await writeConfig(Object.assign(config, { 'tokens.telemetry': newUuid }));
```
rwaskiewicz, line 26: FYI/no action required - from a TypeScript typing perspective, there is a subtle difference between using `Object.assign` and the spread operator. `Object.assign` is typed such that the return type is the intersection of the two args:
```typescript
assign<T, U>(target: T, source: U): T & U;
```
if we were to mess up the name of the key in `U`, the compiler wouldn't catch the error:
```typescript
// compiles just fine
await writeConfig(Object.assign(config, { 'tokens.teleasdf': newUuid }));
```

whereas spreading the object will catch an error
```typescript
// errors! we can't assign to type `TelemetryConfig`
await writeConfig({ ...config, 'tokens.teleasdf': newUuid });
```

There are some subtle differences between the two (see ['Differences Versus Object.assign()'](https://thecodebarbarian.com/object-assign-vs-object-spread.html)) but I don't think they apply here

src/cli/telemetry/telemetry.ts
```diff
@@ -141,21 +144,18 @@ export async function telemetryAction(action?: TelemetryCallback) {
   }
 }
 
+function hasAppTarget(): boolean {
```
rwaskiewicz, line 147: Can we get this under test please?

rwaskiewicz (changes requested):

ltm (approved):

src/cli/test/ionic-config.spec.ts
splitinfinities: STENCIL-53!

src/cli/telemetry/telemetry.ts
splitinfinities: I was under the impression that you only wanted exported members while we're getting started on coverage. Is that your goal?

src/cli/ionic-config.ts
splitinfinities: Thank you. I did this because I wanted to modify the object in place, but the TypeScript side effect is news to me... Would you like to add `prefer-object-spread: error` to our ESLint rules? Open to either option.

rwaskiewicz: Maybe WRT `prefer-object-spread`, I need to start having that conversation with Lars this week about long term linting & formatting plans

src/cli/telemetry/telemetry.ts
splitinfinities: I'll being reorganizing the code to allow testing the object sent.

src/cli/telemetry/telemetry.ts
```diff
@@ -144,21 +102,55 @@ export async function telemetryAction(action?: TelemetryCallback) {
...
-async function getActiveTargets(): Promise<string[]> {
+export async function getActiveTargets(): Promise<string[]> {
   const result = getStencilCLIConfig().validatedConfig.config.outputTargets.map(t => t.type);
   return Array.from(new Set(result));
 }
 
+export const prepareData = async (duration_ms: number, component_count: number = undefined): Promise<TrackableData> => {
+  const { flags, sys } = getStencilCLIConfig();
+  const { typescript, rollup } = getCoreCompiler()?.versions || { typescript: 'unknown', rollup: 'unknown' };
+  const packages = await getInstalledPackages() || [];
```
rwaskiewicz, line 123: I _think_ this always returns an array, and don't think we need the `|| []` here

src/cli/telemetry/test/telemetry.spec.ts
```diff
@@ -76,3 +76,132 @@ describe('checkTelemetry', () => {
...
+    expect(telemetry.hasAppTarget()).toBe(true)
+  });
+
+  it('Result is correct when outputTargets contains www with serviceWorker', () => {
+    getStencilCLIConfig().validatedConfig = { config: { outputTargets: [{ type: "www", baseUrl: "https://example.com", serviceWorker: { swDest: "./tmp" } }] } } as LoadConfigResults
+    expect(telemetry.hasAppTarget()).toBe(true)
+  });
+});
+
+describe('hasAppTarget', () => {
```
rwaskiewicz, line 126: This outer level describe matches the one on line 80. Did you mean 'prepareData'?

src/cli/telemetry/test/telemetry.spec.ts
```diff
@@ -76,3 +76,132 @@ describe('checkTelemetry', () => {
...
+    getStencilCLIConfig().validatedConfig = { config: { outputTargets: [{ type: "www", baseUrl: "/" }] } } as LoadConfigResults;
+    expect(telemetry.hasAppTarget()).toBe(false)
+  });
+
+  it('Result is correct when outputTargets contains www with serviceWorker', () => {
+    getStencilCLIConfig().validatedConfig = { config: { outputTargets: [{ type: "www", serviceWorker: { swDest: "./tmp" }, }] } } as LoadConfigResults
+    expect(telemetry.hasAppTarget()).toBe(true)
+  });
+
+  it('Result is correct when outputTargets contains www with serviceWorker', () => {
```
rwaskiewicz, line 115: I think this test name needs to be updated, there's no service worker field set in this test

src/cli/telemetry/test/telemetry.spec.ts
```diff
@@ -76,3 +76,132 @@ describe('checkTelemetry', () => {
...
+    getStencilCLIConfig().validatedConfig = { config: { outputTargets: [{ type: "www", baseUrl: "https://example.com" }] } } as LoadConfigResults
+    expect(telemetry.hasAppTarget()).toBe(true)
+  });
+
+  it('Result is correct when outputTargets contains www with serviceWorker', () => {
```
rwaskiewicz, line 120: I think this test name needs to be updated, it mimics the two previous test names

rwaskiewicz (approved):
Approved with a few minor asks

src/cli/telemetry/test/telemetry.spec.ts
splitinfinities: I did!

Issue #2687, opened 2020-10-01
File names dont match with imports in JS files when building --prod


obrunopolo:
**Stencil version:**

```
 @stencil/core@2.0.3
```

**I'm submitting a:**

[x] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**

The app works well while using `stencil build --serve --dev --watch`, but once I build a production version, run a http server on the `www` folder and open it in the browser, it gets stuck on white screen on 50% of builds. On console I could see error 404, some resources were missing.

**Expected behavior:**

Production builds should work reliably.

**Steps to reproduce:**

Create a "Starter app" template and run `npm start` and `npm run build` a few times. After each production build, try to open on browser using some http server.

In my case I could see that index.html loads the first hashed js file successfully, but then it try to import 2 non-existing files.

Please note that this may not be related to the hash of file names, because I could reproduce this behavior with `hashFileNames: false` . In this case, it was importing a file with different content hash.

MarkChrisLevy:
I can confirm that issue. It is also not related to caching (both in browser and cli "--no-cache" setting).

obrunopolo:
@MarkChrisLevy that's correct. I also tryied to use `enableCache: false`  and still get the same error.

Just got this again today:

![image](https://user-images.githubusercontent.com/47830258/95240838-58940b80-07e3-11eb-9949-72e69c8c0407.png)

r3dDoX:
I'm having the same Problem since v2. Opened another issue couple weeks ago: #2671

Scan0815:
i can also reproduce this on 2.2.0 stenciljs

r3dDoX:
I am still able to reproduce this with 2.3.0

Scan0815:
@adamdbradley , 
I'm sorry to speak to you directly, but the error is really annoying, having to constantly delete the build folder in a running project. in version 1 it worked great.

MrAntix:
you can reproduce this with the starter app

``` 
npm init stencil 
npm i

npm run build => built index.html still has the app.css and app.js files in the head
npm run build => built index.html has compiled references except the nomodule one
```

kevinlandsberg:
Can reproduce the error in stencil 2.3.0
Always getting 404's for bundle file inside www

rohanrp:
[...]

Issue #2659, opened 2020-09-03
PWAs do not work offline with v2


sbannigan:
**Stencil version:**

```
 @stencil/core@2.0.3
```

**I'm submitting a:**

[x] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**

Stencil PWAs built with > v2 and a generated service worker don't work offline.

**Expected behavior:**

Stencil PWAs should function offline when using a generated service worker to cache the build files and fetch them from the cache automatically.

**Steps to reproduce:**

1. Generate a new stencil project with type `ionic-pwa` using `npm init stencil`
2. Enable the service worker by removing `serviceWorker: null` from the `www` output target
3. `npm install --save-dev workbox-build@4.3.1`
4. Build the project with `npm run build`
5. Either deploy the app or run the build directory on a local server
6. Using Chrome dev tools, check `Disable cache` on the network tab (to prevent using the disk cache, which should not be necessary if using a service worker) and also check `Offline` in Application / Service Workers. (You could also perform a real word test by installing the PWA on a phone or with Chrome and then manually turning off the network)
7. See that the app doesn't load at all (`app.esm.js` and `app.css` have 404 errors in the network tab)

Demo created using the above steps: https://sharp-goldstine-65dc8a.netlify.app/

**Other information:**

I believe this is happening because of some change from v1 -> v2 that changed how the build scripts are imported into the final `index.html`.

Here is what the app script tags look like from v2:
```html
  <script type="module" src="/build/app.esm.js" data-stencil data-resources-url="/build/"
    data-stencil-namespace="app"></script>
  <script nomodule="" src="/build/app.js" data-stencil></script>
  <link href="/build/app.css" rel="stylesheet">
```
and from v1:
```html
  <link rel="modulepreload" href="/build/p-be07d25c.js">
  <link rel="modulepreload" href="/build/p-177ede88.js">
  <link rel="modulepreload" href="/build/p-3a5bd1b8.js">
  <link rel="modulepreload" href="/build/p-58d13c5e.js">
  <link rel="modulepreload" href="/build/p-718ae824.js">
  <link rel="modulepreload" href="/build/p-9121f02b.js">
  <link rel="modulepreload" href="/build/p-9701a0dd.js">
  <script type="module" src="/build/p-be07d25c.js" data-stencil data-resources-url="/build/"
[...]

Issue #1631, opened 2019-06-12
Service worker does not include app.css in manifest


ficoos:
**Stencil version:**

```
@stencil/core@1.0.0 
```

**I'm submitting a:**

[ x] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**
When using service workers in development mode `app.css` is not included in the sw precache manifest.

**Expected behavior:**
`app.css` is included

**Steps to reproduce:**
`stencil build --dev --watch --serve --serviceWorker`

**Related code:**

```tsx
// insert any relevant code here
```

**Other information:**

When building in development you can see the the generate style and generate service worker stages happen concurrently. There might be a race between the sw globber and the style generation code

manucorporat:
why do you expect it included?

ficoos:
It's needed for the site to be available offline. Without it you don't have the global css available when you are offline. This makes testing the site's offline behavior on a dev server impossible.

manucorporat:
can you share your index.html?

michael42:
It seems that @stencil/core 1.14.0 explicitly ignores the main css and js files via `globIgnores`, which makes the resulting service worker rather pointless. It caches the HTML file, but then everything that file references is missing.

michael42:
I'll retract my statement. When using `href="/build/app.css"` and `src="/build/app.esm.js"`, these references will be rewritten into the `p-...` files that *are* cached by the ServiceWorker. I used relative references (`build/app.esm.js`), which are neither rewritten (in the `index.html`) nor cached by the ServiceWorker.

MrAntix:
removing the following lines from the built index.html file made it work for me

```
  <link href="/build/app.css" rel="stylesheet">
  <script type="module" src="/build/app.esm.js"></script>
  <script nomodule src="/build/app.js"></script>
```

splitinfinities:
Hey everyone, thank you for the patience. Would anyone happen to know if this behavior exists within v2? If we could update the repro case for these, I would appreciate it, and it would help us diagnose and solve this problem sooner than later.

rwaskiewicz:
Hey ionitron-bot, we didn't mean to close this. Bad bot!

rwaskiewicz:
[...]

PR #2970, opened 2021-07-20
fix (compiler) Fix race condition between writing app.esm.js and optimizeEsmImport

joewoodhouse:
Background
-----------

I've been investigating an issue where I was getting differing builds between my local build and my CI build (Netlify). 

Specifically I was seeing that locally the `<script type="module" src="/build/app.esm.js"></script>` tag was being re-written to the hashed version e.g. `<script type="module" src="/build/p-703da46a.js" data-stencil data-resources-url="/build/" data-stencil-namespace="app"></script>` - which is what I wanted. But on Netlify the app.esm.js src was remaining in the output index.html file, which was causing me service-worker/caching problems (see below).

Similar things have been reported https://github.com/ionic-team/stencil/issues/2659, https://github.com/ionic-team/stencil/issues/2687 and https://github.com/ionic-team/stencil/issues/1631.

Issue
-----

In the compiler, the `optimizeEsmImport` function is responsible for finding the app.esm.js file and replacing it with the hashed file name.

On stepping through the code I found that when `optimizeEsmImport` was being called, it was failing to find the content of the `entryPath` and so not replacing the script src attribute. 

Looking at `generateOutputTargets` I could see that `outputLazy` (which is responsible for writing the app.esm.js file) and `outputWww` (which writes the index.html, and calls `optimizeEsmImport`) were being run in a big Promise.all([]) so are basically racing against each other, with no guarantee that one will finish before the other (unless there is some other subtle system I haven't seen where one of those promises will wait for another).

So the fix was just to move the `outputWww` until after the Promise.all, in exactly the same way that `outputDocs` and `outputTypes` need to run afterwards.

Side Note - Service Workers
#####################

I'd just like to note that when the above behaviour does occur - i.e. you end up with the `app.esm.js` src in your final index.html, it's a total nightmare for PWAs and service workers. For starters the `app.esm.js` file is globIgnored by the service worker config, so the PWA is totally broken to start with in an offline scenario as the main entry module isn't (pre)cached. If you write a custom service worker that does precache the app.esm.js, that's better but there is no cache busting / hashing of that file, so as soon as you deploy a new build to e.g. Netlify (which wipes out all the previous .entry.js files) again your app is broken because your now old app.esm.js file is requesting files that no longer exist.

### DIFF
--- a/src/compiler/output-targets/index.ts
+++ b/src/compiler/output-targets/index.ts
@@ -32,12 +32,12 @@ export const generateOutputTargets = async (
     outputCustomElements(config, compilerCtx, buildCtx),
     outputHydrateScript(config, compilerCtx, buildCtx),
     outputLazyLoader(config, compilerCtx),
-    outputLazy(config, compilerCtx, buildCtx),
-    outputWww(config, compilerCtx, buildCtx),
+    outputLazy(config, compilerCtx, buildCtx)
   ]);
 
   // must run after all the other outputs
-  // since it validates files were created
+  // since it validates or uses files that were created
+  await outputWww(config, compilerCtx, buildCtx);
   await outputDocs(config, compilerCtx, buildCtx);
   await outputTypes(config, compilerCtx, buildCtx);
 
--- a/src/compiler/output-targets/index.ts
+++ b/src/compiler/output-targets/index.ts
@@ -32,7 +32,7 @@ export const generateOutputTargets = async (
     outputCustomElements(config, compilerCtx, buildCtx),
     outputHydrateScript(config, compilerCtx, buildCtx),
     outputLazyLoader(config, compilerCtx),
-    outputLazy(config, compilerCtx, buildCtx)
+    outputLazy(config, compilerCtx, buildCtx),
   ]);
 
   // must run after all the other outputs


joewoodhouse:
@rwaskiewicz any chance of a review on this?

rwaskiewicz:
Thanks @joewoodhouse! I think I have an idea as to what's happening here. To further investigate, could you please send a minimal reproduction? I understand this is happening in a CI environment and _that_ part may be hard to give us access to. Even a project where this happens would be super helpful. Thanks!

joewoodhouse:
Hi @rwaskiewicz 

Thanks for taking a look. Here's a basic repo that should create the issue
https://github.com/joewoodhouse/stencil-pr-2970

I think probably the reason partly comes down to the CI environment not having a `www` folder when the build happens, so perhaps something to do with caching? In that repo if you run `npm run build` you should get the failure (i.e. you'll see `app.esm.js` in `www/index.html`). However if you then run `npm run build.nodelete` it will work (the `app.esm.js` will be replaced with p-XXXXX.js or similar). 

Let me know if you need anything else.

joewoodhouse:
@rwaskiewicz just to flag I replied to this 12 days ago but it's still labelled as awaiting reply. Let me know if you need any more info.

alicewriteswrongs:
closing in favor of #4992

PR #2976, opened 2021-07-26
chore(docs): add git message instructions

rwaskiewicz:
- add additional details to CONTRIBUTING.md surrounding git commit
messages:
  - how/when to use a commit body
  - how to specify a breaking change in a PR
  - where to put JIRA story tracking information
- update/align header tags throughout the doc

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -37,15 +37,15 @@ Please see our [Contributor Code of Conduct](https://github.com/ionic-team/stenc
 4. Run `npm install` (make sure you have [node](https://nodejs.org/en/) and [npm](http://blog.npmjs.org/post/85484771375/how-to-install-npm) installed first)
 
 
-#### Updates
+### Updates
 
 1. Unit test. Unit test. Unit test. Please take a look at how other unit tests are written, and you can't write too many tests.
 2. If there is a `*.spec.ts` file located in the `test/` folder, update it to include a test for your change, if needed. If this file doesn't exist, please notify us.
 3. First run `npm run build`. Then run `npm run test` or `npm run test.watch` to make sure all tests are working, regardless if a test was added.
 
-#### Testing Changes Against a Project Locally
+### Testing Changes Against a Project Locally
 
-##### Testing with `npm link`:
+#### Testing with `npm link`:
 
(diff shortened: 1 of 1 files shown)


.github/CONTRIBUTING.md
```diff
@@ -94,11 +94,20 @@ Afterwards, to clean up:
 
 1. In the directory of your stencil project, run `npm install --save-dev stencil@<VERSION>` for the `<VERSION>` of Stencil core that was installed in your project prior to testing. 
 
-## Commit Message Format
+### Commit Message Format
 
-We have very precise rules over how our git commit messages should be formatted. This leads to readable messages that are easy to follow when looking through the project history. We also use the git commit messages to generate our changelog. (Ok you got us, it's basically Angular's commit message format).
+We strive to adhere to a consistent commit message format of:
```
ltm, line 99: Should we officially embrace [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/)?

rwaskiewicz: I strayed away here from referencing conventional commits, because they mark `scope` as optional. Turns out we don't use the scope explicitly when [we generate a changelog](https://github.com/ionic-team/stencil/blob/42cdeaec424fb053648a5ae97e611a7c58d69788/scripts/utils/release-utils.ts#L56) like I thought we did. Let me dig into that conventional commit package a little bit more - when I did the 2.7.0-0 dev release, some things didn't get populated in the `CHANGELOG.md` file in the root of this repo the way I thought they would

rwaskiewicz: @ltm I spent some time experimenting with our changelog generation scripts, and made a few updates to tailor the advice to the current behavior of the package we use.

I did update the verbiage to state we use the _Angular_ version of conventional commits, which is a little different in that:
- It does not use the '!' syntax for breaking changes (commits with this syntax won't get added to our `CHANGELOG` under the BREAKING CHANGES section today, this is a result of how the tooling is configured ATM)
- I added instructions on using the 'BREAKING CHANGE:' footer syntax. This _does_ get picked up in the BREAKING CHANGES section of the CHANGELOG
- I added an example that hopefully paints a picture of how a commit translates to entries on `CONTRIBUTING.md`

I also added `revert` to the acceptable types list, stealing that from the Angular folks as well.

splitinfinities (approved):

PR #2981, opened 2021-07-27
feat(rollup): configuration option for symlinks in rollup build

tcarrio:
## What 

Provides the option to override the default rules for `rollup` by specifying the `preserveSymlinks` configuration property.

In the case the option is not provided, this will continue to be undefined, and has no breaking changes.

## Fixes

This fixes various build tooling such as Bazel which make extensive use of symlinks for builds.

## Tests

Tests included provide runtime validation and type checking which requires all inputs possible on the `RollupInputOptions` and `RollupOutputOptions` be provided in the test cases.

Refs: #2840

### DIFF
--- a/src/compiler/bundle/bundle-output.ts
+++ b/src/compiler/bundle/bundle-output.ts
@@ -120,6 +120,7 @@ export const getRollupOptions = (
     treeshake: getTreeshakeOption(config, bundleOpts),
     inlineDynamicImports: bundleOpts.inlineDynamicImports,
     preserveEntrySignatures: bundleOpts.preserveEntrySignatures ?? 'strict',
+    preserveSymlinks: config.rollupConfig.inputOptions.preserveSymlinks ?? undefined,
 
     onwarn: createOnWarnFn(buildCtx.diagnostics),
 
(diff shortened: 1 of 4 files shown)


tcarrio:
This appears to be something that could be supported by making use of the `rollupPlugins` property in the stencil config.

Thanks to @loudmouth who provided more insight on this, you could use this in your `stencil.config.ts` to support symlink preservation:

```ts
  rollupPlugins: {
    // Plugins injected before rollupNodeResolve()
    before: [
      // This plugin is required to overwrite the `preserveSymlinks` rollup
      // config option so that Bazel can resolve files while running Stencil in
      // the Bazel sandbox.
      {
        name: 'overwrite-rollup-options',
        options: (options: RollupInputOptions) => {
          return {
            ...options,
            preserveSymlinks: true,
          };
        },
      },
    ],
  },
```

This has been tested with the latest version of Stencil to work for use cases requiring symlinks such as Bazel builds. This change is not required, but does simplify the process for use cases like Bazel. If you need a solution now though, the above snippet will provide what you need.

splitinfinities:
Hey! Thank for the PR. The team is going to investigate this because we suspect it would help improve Monorepo support among other things. We also want to understand if this could impact the compiler in other ways. 

Just wanted to give you a heads up that we have our eye on this!

PR #2986, opened 2021-08-02
chore(tests): Improve observability of unit tests

splitinfinities:
Add jest coverage reporters, add --coverage flag to test command, adding coverage reporters locally and ignore the generated files.


.gitignore
```diff
@@ -45,3 +45,4 @@ test/**/www/*
 test/**/hydrate/*
 .stencil
 /dependencies.json
+test-results/**
```
splitinfinities, line 48: For now, the results are produced and can be referred to locally.

jest.config.js
```diff
@@ -11,6 +11,12 @@ module.exports = {
     '@stencil/core/testing': '<rootDir>/testing/index.js',
     '@utils': '<rootDir>/src/utils',
   },
+  coverageDirectory: './test-results/',
+  coverageReporters: ["json", "lcov", "text", "clover", "json-summary"],
+  collectCoverageFrom: [
```
splitinfinities, line 16: For now, only `cli` and `utils` are collected for coverage. Without these lines, the `jest --coverage` command will take quite a while.

jest.config.js
```diff
@@ -11,6 +11,12 @@ module.exports = {
     '@stencil/core/testing': '<rootDir>/testing/index.js',
     '@utils': '<rootDir>/src/utils',
   },
+  coverageDirectory: './test-results/',
+  coverageReporters: ["json", "lcov", "text", "clover", "json-summary"],
```
splitinfinities, line 15: The first four are jest defaults - json-summary is something we could use in the future to more easily track quality over time.

jest.config.js
rwaskiewicz: Consider opening this up to all directories. IMO this doesn't provide a ton of value if we're trying to make this a team-wide initiative (if someone were to open a PR without any tests in any other directory we're back to square one). Rather consider creating either a new script in package.json for coverage and/or documenting that coverage can be determined by running `npm run test -- --coverage` (that double dash is important, it helps pipe `--coverage` to the `jest` command)

jest.config.js
rwaskiewicz: My thought here is that we follow [YAGNI](https://martinfowler.com/bliki/Yagni.html) and wait until we need it
```suggestion
  coverageReporters: ["json", "lcov", "text", "clover"],
```

rwaskiewicz (changes requested):

jest.config.js
rwaskiewicz: The tests aren't running as a part of the build process though 🤔 Did you mean CI time went up? Can you share the configuration you were using?

splitinfinities: Yeah! Just delete this line and run the test command. You'll see what I mean. There's an initial compilation process that Jest/Istanbul use to keep track of lines covered during the tests.

rwaskiewicz: I wonder if there's a configuration error with deleting `collectCoverageFrom`, as 
```
  collectCoverageFrom: [
    '<rootDir>/src/app-data/**/*.{js,jsx,ts,tsx}',
    '<rootDir>/src/app-globals/**/*.{js,jsx,ts,tsx}',
    '<rootDir>/src/cli/**/*.{js,jsx,ts,tsx}',
    '<rootDir>/src/compiler/**/*.{js,jsx,ts,tsx}',
    '<rootDir>/src/declarations/**/*.{js,jsx,ts,tsx}',
    '<rootDir>/src/dev-server/**/*.{js,jsx,ts,tsx}',
    '<rootDir>/src/hydrate/**/*.{js,jsx,ts,tsx}',
    '<rootDir>/src/internal/**/*.{js,jsx,ts,tsx}',
    '<rootDir>/src/mock-doc/**/*.{js,jsx,ts,tsx}',
    '<rootDir>/src/runtime/**/*.{js,jsx,ts,tsx}',
    '<rootDir>/src/screenshot/**/*.{js,jsx,ts,tsx}',
    '<rootDir>/src/sys/**/*.{js,jsx,ts,tsx}',
    '<rootDir>/src/testing/**/*.{js,jsx,ts,tsx}',
    '<rootDir>/src/utils/**/*.{js,jsx,ts,tsx}',
  ],
```
runs for me and doesn't seem to slow tests down too much. Are you sure you weren't accidentally picking up `node_modules`?

splitinfinities: 🤔 Could have been! Didn't even think about that.

splitinfinities: Added that chunk in 526763c

splitinfinities: Yeah that does work much better. Thank you.

jest.config.js
```diff
@@ -11,6 +11,24 @@ module.exports = {
     '@stencil/core/testing': '<rootDir>/testing/index.js',
     '@utils': '<rootDir>/src/utils',
   },
+  coverageDirectory: './test-results/',
```
rwaskiewicz, line 14: One non-blocking suggestion - I think `./coverage/` may be more apt here. We have a lot of `test*` directories today, naming this to something that helps differentiate it may make the top level directory of the repo a little more grok-able.

splitinfinities: Done!

package.json
```diff
@@ -49,7 +49,8 @@
     "test.testing": "node scripts/test/validate-testing.js",
     "test.sys.node": "echo test.sys.node",
     "test.prod": "npm run test.dist && npm run test.end-to-end && npm run test.jest && npm run test.karma && npm run test.sys.node && npm run test.testing",
-    "test.watch": "jest --watch"
+    "test.watch": "jest --watch",
+    "test.watchAll": "jest --watchAll --coverage"
```
rwaskiewicz, line 53: The convention for this file is to use dash-case instead of camelCase for multi-word scripts. Can we please do something like:
```suggestion
    "test.watch-all": "jest --watchAll --coverage"
```

rwaskiewicz (commented):
Works as expected, I have one small ask to modify a change introduced in the latest commit

rwaskiewicz (approved):

Issue #1392, opened 2019-02-27
CSP Nonce Support for generated files


MrMcGibblets:
**Stencil version:**
```
 @stencil/core@0.18.0
```

**I'm submitting a:**
[ ] bug report
[x] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**
Same as the closed issue - https://github.com/ionic-team/stencil/issues/496

When bundling we automatically create an inline style which violates the strict CSP usage

Whilst a workaround is to add each individual sha to the CSP policy it isn't ideal and not a nice dev experience.

In webpack this was achieved by adding a top-level nonce value which would then be added to all at bundle time. - https://github.com/styled-components/styled-components/issues/887

**Expected behavior:**
To be able to provide a nonce for styles (should be for all scripts, img etc...) that will be added to all build-generated scripts, styles etc... Allowing apps to be CSP ready out of the box.

**Steps to reproduce:**
Add CSP Policy for style-src 'self' and run app.

BorntraegerMarc:
Working in an organization this is a critical feature for us. It's quite a hazzle to add sha values for our styles.

Was there any internal discussion in the ionic team about this feature?

R0muald:
Is there any update from the Stencil team in regard of this feature? We also have some concern that limit some of our teams to move ahead with the stencilJs web-components used by our design system.

ptrot001:
Hi, Any update on this one? I can't believe nothing have been done yet to handle CSP in stencil libraries. Angular and react have added support for CSP in 2015-2016!

I'm also working in a big organization and CSP is a basic security requirement for us. It's a blocker right now!

Any update would be appreciated! 👍 @adamdbradley ?

Thanks

alexhhn:
Hi,
Any update on this?

sandersaspera:
We also need that, quite urgently

erwingeiger:
any updates about this? Also our organisation require to use a CSP.

sandersaspera:
> 
> 
> any updates about this? Also our organisation requires to use a CSP.

[...]

PR #2996, opened 2021-08-10
feat: possibility to add nonce attribute on script and style tags

erwingeiger:
This is about this topic: https://github.com/ionic-team/stencil/issues/1392
I added the possibility to add a nonce attribute on the <style> and <script> tags who are dynamically created by Stencil.
The value of the nonce attribute can be given with a global window variable.

### DIFF
--- a/src/client/client-patch-browser.ts
+++ b/src/client/client-patch-browser.ts
@@ -88,6 +88,7 @@ const patchDynamicImport = (base: string, orgScriptElm: HTMLScriptElement) => {
       if (!mod) {
         const script = doc.createElement('script');
         script.type = 'module';
+        if((window as any).NONCE_ID) script.setAttribute('nonce', (window as any).NONCE_ID);
         script.crossOrigin = orgScriptElm.crossOrigin;
         script.src = URL.createObjectURL(new Blob([`import * as m from '${url}'; window.${importFunctionName}.m = m;`], { type: 'application/javascript' }));
         mod = new Promise(resolve => {
--- a/src/client/polyfills/css-shim/load-link-styles.ts
+++ b/src/client/polyfills/css-shim/load-link-styles.ts
@@ -47,6 +47,8 @@ export function addGlobalLink(doc: Document, globalScopes: CSSScope[], linkElm: 
         styleEl.setAttribute('data-styles', '');
         styleEl.textContent = text;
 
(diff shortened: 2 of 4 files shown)


kliehm:
FYI, Firefox blocks all inline styles without a nonce. So basically Stencil cannot be used in production. Please prioritize this issue!

rwaskiewicz:
@kliehm Can you provide an example of Firefox blocking inline styles without a nonce? I'm able to get the following (although basic) HTML with inline CSS to render just fine in the following versions of Firefox:
- 95.0.2 (64-bit)
- 96.0b10 (64-bit)
```html
<!DOCTYPE html>
<html>
<body>

<h1 style="color:blue;text-align:center;">This is a heading</h1>
<p style="color:red;">This is a paragraph.</p>

</body>
</html> 
```

erwingeiger:
@rwaskiewicz It will go wrong where insertBefore is being used. Take this as an example.
Screenshot 2022-03-10 at 11 57 23

Maybe te best solution would be  what I created in this PR: https://github.com/ionic-team/stencil/pull/2996
We still use this solution but it's a 'hacky' way. Everytime we update Stencil core we have to manually add this functionality to create these nonce values.

Could you add this functionality in the core? A lot of people would be very pleased!

erwingeiger:
> Update.
> 
> After implementing the nonce, we still have problems with unsafe-eval. I've seen someone has a solution for this: #1287. Unfortunately the PR was closed because it was not supported in older browsers. Now we are a few years later. IE11 is not supported anymore. Is it possible to add it now?

What I said above  is solved because we don't support IE11 anymore.  Using the nonce is still necessary but with that we will not get any errors anymore

rwaskiewicz:
Initial support for CSP Nonces have been added in today's release of [Stencil 2.22.1](https://github.com/ionic-team/stencil/releases/tag/v2.22.1), with our official guidance on nonce usage with Stencil[ published here](https://stenciljs.com/docs/csp-nonce). As a result, I'm going to close this pull request. Thanks again for helping making Stencil better!

erwingeiger:
> Initial support for CSP Nonces have been added in today's release of [Stencil 2.22.1](https://github.com/ionic-team/stencil/releases/tag/v2.22.1), with our official guidance on nonce usage with Stencil[ published here](https://stenciljs.com/docs/csp-nonce). As a result, I'm going to close this pull request. Thanks again for helping making Stencil better!

That is great news @rwaskiewicz! Thank you for this.

PR #3001, opened 2021-08-16
Allow file change events to pass through to let the Copy task succeed

splitinfinities:
Allows the dev server to pick up changes during the www copy tasks, so consumers don't have to restart their dev server when developing. 

I've tested this experience on each starter repo, a monorepo, and framework. 

### Before this change: 

Set up a www output target that with a copy rule, e.g.

```
{
      type: 'www', 
      copy: [
        { src: 'images' }
      ]
}
```

Within that images directory, try to make file modifications - change the content, duplicate the file, delete the file. The contents should not be carried over during the process of the dev server's lifecycle. Commonly, you have to restart the dev server in order to make the copy task run.

### After this change: 

All file change events now get passed down to the tasks. Copy now works as expected - files get copied when needed. 

Internal ticket: STENCIL-29

### DIFF
--- a/src/sys/node/node-sys.ts
+++ b/src/sys/node/node-sys.ts
@@ -430,7 +430,7 @@ export function createNodeSys(c: { process?: any } = {}) {
         const tsFileWatcher = tsSysWatchDirectory(
           p,
           (fileName) => {
-            callback(normalizePath(fileName), null);
+            callback(normalizePath(fileName), 'fileUpdate');
           },
           recursive
         );


splitinfinities:
I believe, looking at the history of that file, that this is a regression that happened after we tried to improve our Monorepo support.

splitinfinities:
Also shoutout to @ltm for this fix - I just tested it.

rwaskiewicz (approved):

rwaskiewicz:
Before we merge this, @splitinfinities can you do me a favor and update the PR summary with the testing steps you took for historical purposes?

splitinfinities:
@rwaskiewicz Done!

ltm (approved):

PR #2892, opened 2021-04-23
feat(compiler): source maps

johnjenkins:
Hello!
This PR seeks to add source-maps to most stencil output - a feature that is long requested [here](https://github.com/ionic-team/stencil/issues/219), [here](https://github.com/ionic-team/stencil/issues/1255), [here](https://forum.ionicframework.com/t/how-to-set-up-sourcemaps-with-ionic/183113),  [here](https://github.com/ionic-team/stencil/issues/1650), [here](https://github.com/ionic-team/stencil/issues/1744) and proposed [here](https://stenciljs.com/blog/stencil-roadmap-fall-2019)

The changes simply seek to join the source-map dots (and therefore track the changes back to the original source) from the first typescript transforms and transpilation, through the rollup bundling and finally into and out of the optimisation.

I've addressed and tested most output targets: dist-lazy (cjs / esm / es5 varieties), dist-collection, dist-custom-elements and dist-custom-elements-bundle.

To keep things simple, there are no soureMap options - just a boolean.
Now setting `sourceMap: true` within `stencil.config.ts` will override typescript configuration options (sourceMap and inlineSources) and  generate source maps (external .js.map files with inlined sources) for all above outputs mentioned. 

Let me know if I've missed something or can improve anything. 

Many thanks!

fixes #1744
fixes #1650
fixes #219
fixes #1255

## Pull request checklist
 Please check if your PR fulfills the following requirements:
 
 * [x]  Tests for the changes have been added (for bug fixes / features)
 * [ ]  Docs have been reviewed and added / updated if needed (for bug fixes / features)
    * They've been added here
 * [x]  Build (`npm run build`) was run locally
 
 ## Pull request type
 Please check the type of change your PR introduces:
 
 * [ ]  Bugfix
 * [x]  Feature
 * [ ]  Code style update (formatting, renaming)
 * [ ]  Refactoring (no functional changes, no api changes)
 * [ ]  Build related changes
 * [ ]  Documentation content changes
 * [ ]  Other (please describe):
 
 ## What is the current behavior?
 There are no source maps
 
 ## What is the new behavior?
 As described above.
 
 ## Does this introduce a breaking change?
 * [ ]  Yes
 * [x]  No
 
 ## Manual Testing
 * Ran `npm build` and `npm link` from the root of the repo directory (on this branch)
 * Spun up a boilerplate Stencil app using the [Stencil CLI](https://stenciljs.com/docs/getting-started) `npm init stencil`
[...]

PR #3005, opened 2021-08-17
feat(compiler): consumer sourcemap support

rwaskiewicz:
This PR is an exact copy of #2892. After Prettier was added to the codebase, merge conflicts arose despite our best efforts. I took some time to clear them out on a separate branch. But a large majority of the work here is attributed to @johnjenkins, without whom we would not be nearly this far along. 

This branch continues that work, as the Stencil team continues to refine sourcemap work and prepare it to be merged into the main branch.


src/compiler/bundle/app-data-plugin.ts
```diff
@@ -66,7 +68,16 @@ export const appDataPlugin = (
         appendEnv(config, s);
         return s.toString();
       }
-      return null;
+      if (id !== config.globalScript) {
+        return null;
+      }
+
+      const module = compilerCtx.moduleMap.get(config.globalScript);
```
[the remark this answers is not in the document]
splitinfinities: AFAICT so far, This module map gets created much earlier in the lifecycle of the program. I learned this through my exercise through refactoring the compilerCtx into a singleton. Nothing sticks out to me, but if you search the code for instances of "moduleMap.set" you'll find the two places where this is called within the thread of the build process, then another in the client thread - must be to keep track of modules that must be updated for HMR. I haven't looked any deeper into that side. 

First, [the map is instantiated as a prop of the CompilerContext class](https://github.com/ionic-team/stencil/blob/master/src/compiler/build/compiler-ctx.ts#L33). 

Later, the map is updated in two places. [getModuleLegacy](https://github.com/ionic-team/stencil/blob/master/src/compiler/build/compiler-ctx.ts#L63) is responsible for that. It's also has a corresponding reset method that can be run on each module within the Map. 

From there it gets a little hairy to start linking for us all to understand, but this map is built up prior to builds, and this code that we're commenting on executed during a build. 

But to answer your question, this is available and has been populated at this point in time, yes.

src/compiler/bundle/app-data-plugin.ts
```diff
@@ -66,7 +68,16 @@ export const appDataPlugin = (
...
+      if (!module.sourceMapFileText) return { code: module.staticSourceFileText, map: null };
+
+      const sourceMap: d.SourceMap = JSON.parse(module.sourceMapFileText);
+      sourceMap.sources = sourceMap.sources.map((src) => basename(src));
+      return { code: module.staticSourceFileText, map: sourceMap };
```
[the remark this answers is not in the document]
splitinfinities: I think this is the signature of the [LoadResult](https://github.com/rollup/rollup/blob/69c099c9248d150fcb99c80c97b52adfda30fabc/src/rollup/types.d.ts#L253) as a [SourceDescription](https://github.com/rollup/rollup/blob/69c099c9248d150fcb99c80c97b52adfda30fabc/src/rollup/types.d.ts#L98)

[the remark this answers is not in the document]
splitinfinities: Similarly for a rollup plugin's transform method - [TransformResult](https://github.com/rollup/rollup/blob/69c099c9248d150fcb99c80c97b52adfda30fabc/src/rollup/types.d.ts#L261) as a partial Source Description

src/compiler/bundle/app-data-plugin.ts
```diff
@@ -66,7 +68,16 @@ export const appDataPlugin = (
         appendEnv(config, s);
         return s.toString();
       }
-      return null;
+      if (id !== config.globalScript) {
+        return null;
+      }
+
+      const module = compilerCtx.moduleMap.get(config.globalScript);
+      if (!module.sourceMapFileText) return { code: module.staticSourceFileText, map: null };
```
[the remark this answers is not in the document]
splitinfinities: I think this is a good idea imho.

src/compiler/build/compiler-ctx.ts
```diff
@@ -103,6 +103,8 @@ export const getModuleLegacy = (_config: d.Config, compilerCtx: d.CompilerCtx, s
       potentialCmpRefs: [],
       staticSourceFile: null,
       staticSourceFileText: '',
+      sourceMapPath: null,
```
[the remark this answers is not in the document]
splitinfinities: AFAIK, this method helps build the moduleMap, I don't understand yet why it's named with "legacy" yet, but generally this should help to build the object that will ultimately be a part of the moduleMap. 

Interestingly enough, this is in some way already on the buildCtx too, since buildCtx takes in the compilerCtx and uses that to do some internal stuff throughout the life of a build. So, this is technically available on the buildCtx, and may in fact be used by it too, but I don't yet know if it is for sure. At a cursory search through the codebase, it doesn't seem to be used. (I searched with .compilerCtx.moduleMap)

I assume it's set up this way to save on memory. If the module map was instantiated and populated for each build, there would be a lot of processing power and disk reads that would be prohibitive. So having it at a higher level cache and providing it down to the buildCtx should help mitigate that. 

I swear I'll get around to an ADR on compilerCtx and buildCtx lol. There's a lot to unravel about these two concepts.

src/compiler/bundle/app-data-plugin.ts
```diff
@@ -11,6 +11,8 @@ import {
   STENCIL_INTERNAL_HYDRATE_ID,
 } from './entry-alias-ids';
 import ts from 'typescript';
+import { basename } from 'path';
```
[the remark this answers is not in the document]
splitinfinities: I think we should use the abstraction here.

src/compiler/bundle/app-data-plugin.ts
```diff
@@ -52,7 +54,7 @@ export const appDataPlugin = (
       return null;
     },
 
-    load(id) {
+    load(id): d.RollupLoadHook {
```
splitinfinities, line 57: I've been thinking a lot on this file. 

I want to propose a couple refactoring items here that are not blocking, aren't worth adding to this PR, but worth talking about and adding a ticket to Jira if agreed upon.

splitinfinities: First, I think we should try to set a policy around our internal rollup plugins and their `load`/`transform` to always return either null, or the [SourceDescription](https://github.com/rollup/rollup/blob/69c099c9248d150fcb99c80c97b52adfda30fabc/src/rollup/types.d.ts#L98) object. That way, extending this code to allow source maps to other new files later down the line aren't a problem, but also if we ever want to pass along an AST without a source map, we can do just that. 

I don't necessarily agree about sometimes returning a string, or sometimes returning an object, and would love to see that as a policy for these two methods. 

I rewrote this function to be:
```typescript
    load(id) {
      const s = new MagicString(``);
      let result = null;

      if (id === STENCIL_APP_GLOBALS_ID) {
        appendGlobalScripts(globalScripts, s);
        result = {
          code: s.toString(),
          map: undefined,
        };
      } else if (id === STENCIL_APP_DATA_ID) {
        appendNamespace(config, s);
        appendBuildConditionals(config, build, s);
        appendEnv(config, s);

        result = {
          code: s.toString(),
          map: undefined,
        };
      } else if (id === config.globalScript) {
        const module = compilerCtx.moduleMap.get(config.globalScript);
        if (!module.sourceMapFileText) return { code: module.staticSourceFileText, map: null };

        const sourceMap: d.SourceMap = JSON.parse(module.sourceMapFileText);
        sourceMap.sources = sourceMap.sources.map((src) => basename(src));
        result = { code: module.staticSourceFileText, map: sourceMap };
      }

      return result;
    },
 ```

Comment (splitinfinities) [reply src/compiler/bundle/app-data-plugin.ts:57]: Second, and more contentious because I don't have a deeper understanding of the flow of data - I think in an effort to remove the if statements from the plugin, we should abstract this into three separate plugins. 
1. the STENCIL_APP_GLOBALS_ID plugin
2. the STENCIL_APP_DATA_ID plugin
3. the config.globalScript plugin

That way this file can be kept small and testable. 

Meaning, we'd have three plugins with these load functions: 

app-data-globals-plugin.ts
```typescript
    load(id) {
      const s = new MagicString(``);
      let result = null;

      if (id === STENCIL_APP_GLOBALS_ID) {
        appendGlobalScripts(globalScripts, s);
        result = {
          code: s.toString(),
          map: undefined,
        };
      } 
      
      return result;
    },
```

app-data-plugin.ts
```typescript
    load(id) {
      const s = new MagicString(``);
      let result = null;

      if (id === STENCIL_APP_DATA_ID) {
        appendNamespace(config, s);
        appendBuildConditionals(config, build, s);
        appendEnv(config, s);

        result = {
          code: s.toString(),
          map: undefined,
        };
      } 
      
      return result;
    },
```

app-data-globalScript-plugin.ts
```typescript
    load(id) {
      let result = null;

      if (id === config.globalScript) {
        const module = compilerCtx.moduleMap.get(config.globalScript);
        if (!module.sourceMapFileText) return { code: module.staticSourceFileText, map: null };

        const sourceMap: d.SourceMap = JSON.parse(module.sourceMapFileText);
        sourceMap.sources = sourceMap.sources.map((src) => basename(src));
        result = { 
          code: module.staticSourceFileText, 
          map: sourceMap 
        };
      } 
      
      return result;
    },
```

I recognize this is a subset of the content of the original file, so this is naive. But this plugin is overloaded without a doubt.

splitinfinities: For future reference - I think passing around an AST instead of a string may improve the performance of our consumers compilation. See https://github.com/rollup/rollup/issues/782

package.json
```diff
@@ -99,6 +99,7 @@
     "jest-environment-node": "^26.6.2",
     "listr": "^0.14.3",
     "magic-string": "^0.25.7",
+    "merge-source-map": "^1.1.0",
```
rwaskiewicz, line 102: I have a question out to see if this needs to be added to https://github.com/ionic-team/stencil/blob/925d4e924264df424c3519f4c0a91b22356a2ea6/scripts/license.ts#L9 or not

rwaskiewicz: NVM - got confirmation we do (need to add it) - https://github.com/ionic-team/stencil/pull/3016

rwaskiewicz: Done https://github.com/ionic-team/stencil/pull/3005/files#diff-a559fe1aaf8746e145121786b79b8b86b1f87b4da31f52039453bfb12b8370f8R21

src/compiler/bundle/app-data-plugin.ts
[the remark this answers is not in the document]
rwaskiewicz: I took a look into this this AM, turns out that we do our [own little polyfill of `path`](https://github.com/ionic-team/stencil/blob/master/src/compiler/sys/modules/path.ts) - there isn't a `node-sys` version of `path`. This is fine as-is

src/compiler/bundle/app-data-plugin.ts
```diff
@@ -66,31 +68,53 @@ export const appDataPlugin = (
...
 
-        return results.outputText;
+        if (config.sourceMap) {
+          const codeMap = codeMs.generateMap({
+            source: id,
+            file: id + '.map',
+            includeContent: true,
+            hires: true,
+          });
+          return { code: results.outputText, map: sourceMapMerge(codeMap, sourceMap) };
```
rwaskiewicz, line 114: @johnjenkins Do you recall why we want to merge the sourcemaps here?

I did some digging, at comparing the sourcemap generated by the TS compiler (left in the image below) and the one generated by MagicString (right in the image below):

Screen Shot 2021-09-21 at 2 05 02 PM

It seems the only difference is the name of `file` and `mappings`.

I _think_ per the [spec for sourcemaps](https://sourcemaps.info/spec.html#h.qz3o9nc69um5), `file` should point to `globals.js` in this situation (and should be easy enough to handle).

The `hires` (high resolution) mappings are what I'm intrigued by. I took at look at the [MagicString Docs](https://github.com/Rich-Harris/magic-string#sgeneratemap-options-)

> hether the mapping should be high-resolution. Hi-res mappings map every single character, meaning (for example) your devtools will always be able to pinpoint the exact location of function calls and so on. With lo-res mappings, devtools may only be able to identify the correct line - but they're quicker to generate and less bulky. If sourcemap 

Was the idea here to get sourcemaps with higher fidelity, then merge them back into the ones generated by TS?

johnjenkins: Hi @rwaskiewicz - yes the difference in mappings is the important thing ... the `file` is incorrect - or not as important at least.
We need to merge sourcemaps because we need to account for the extra import / export strings of code that can get added and to this end I needed to get creative!
1) Turn the incoming TS string to a MagicString - this will track the changes we make to the original TS and can generate a sourcemap tracking these changes.
2) Compile the TS using the `transpileModule` - this generates another sourcemap however the typescript compiler has no idea of the earlier amends made to the source TS therefore the sourcemap generated does not reflect these amends and is incorrect.
3) Merge the original sourcemap from MagicString and the new one from typescript - all changes made to the original source now being accounted for.

This is a little convoluted I agree - I was just working with what was there originally and appending strings of code to source input should be avoided in general for this reason.
For future devs - try to make these amends to sourcecode within TS plugins / AST to avoid this hinkiness.

Regarding the `hires` property - I simply applied it because the weird process described and corresponding merge would produce slightly incorrect sourcemaps and this fixed it. 
Attached is 2 screen caps. 1 with `hires` - each console points to the correct line from the original source. 2 without `hires` source lines highlighted can be out by 1
https://user-images.githubusercontent.com/5030133/134250837-edc67824-f668-4b85-8b34-4966017cc70c.mov
https://user-images.githubusercontent.com/5030133/134250827-634f7e57-5e24-451c-a353-21dde518489b.mov

src/compiler/app-core/bundle-app-core.ts
```diff
@@ -12,13 +12,14 @@ export const generateRollupOutput = async (
     return null;
   }
 
-  const { output } = await build.generate(options);
-  return output.map((chunk) => {
+  const { output }: { output: [OutputChunk, ...(OutputChunk | OutputAsset)[]] } = await build.generate(options);
+  return output.map((chunk: OutputChunk | OutputAsset) => {
```
splitinfinities, line 16: For my own understanding, what the difference between these two - Chunk and Asset?

src/compiler/bundle/app-data-plugin.ts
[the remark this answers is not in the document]
splitinfinities: Should we resolve this?

src/declarations/stencil-private.ts
```diff
@@ -346,6 +356,13 @@ export interface BundleEntryInputs {
   [entryKey: string]: string;
 }
 
+/**
+ * A note regarding Rollup types:
```
splitinfinities, line 360: @rwaskiewicz do you consider this Technical debt? Would we have a ticket in the backlog already if so?

splitinfinities (commented):
Looking good! Going to try this out in a couple local libraries and circle back.

src/compiler/bundle/app-data-plugin.ts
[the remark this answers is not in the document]
rwaskiewicz: Not yet, let me put up a PR with a few tweaks to this file

src/declarations/stencil-private.ts
rwaskiewicz: I don't think so, it's a pretty deep archtectural aspect of the system (I think, in so far as I can tell rn). I'll add something to the backlog as a low priority spike

rwaskiewicz: STENCIL-175

johnjenkins (approved):
LGTM

theo-staizen:
Beautiful work lads, just beautiful <3

Issue #2933, opened 2021-06-17
Attribute `referrerPolicy` is missing on `AnchorHTMLAttributes`


wavebeem:
**Stencil version:**

```
 @stencil/core@2.5.2
```

I looked at the code in GitHub and it doesn't appear to be in the latest version (2.6.0).

**I'm submitting a:**

[ ] bug report
[x] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**

The `referrerPolicy` attribute doesn't exist in JSX types, causing a failed build when using this attribute unless I use `// @ts-expect-error` to suppress the error.

**Expected behavior:**

The attribute `referrerPolicy` should exist for `JSXBase.AnchorHTMLAttributes` and be a string, causing there to be no TypeScript error when building the application.

**Steps to reproduce:**

1. Use the attribute `referrerPolicy` on an `<a>` tag and observe the compile failure.

**Related code:**

```tsx
// Type '{ href: string; target: string; referrerPolicy: string; }' is not
// assignable to type 'AnchorHTMLAttributes<HTMLAnchorElement>'.
//
//  Property 'referrerPolicy' does not exist on type
//  'AnchorHTMLAttributes<HTMLAnchorElement>'.ts(2322)
<a referrerPolicy="no-referrer-when-downgrade" />
```

**Other information:**

Alternately, the type of `referrerPolicy` could be an enumeration, as in `@types/react` https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L1993-L2014

This is a greater maintenance burden if more referrer policies are added in the future, but it should lead to a better developer experience in most cases, especially considering that the strings are quite long (e.g. "strict-origin-when-cross-origin"), and easy to misspell (e.g. "no-referer" instead of "no-referrer").

I can create a pull request for this, but I would like guidance on whether I should go with `string` or the previously mentioned enumeration from `@types/react` before creating it.

splitinfinities:
Hey there, thank you for the patience getting back to you. The new team is getting started and we're working through the backlog now.

This is a great idea! Thank you so much for submitting this. We don't quite yet have a timeline but I'll circle back with you soon. 

In the meantime, if anyone else can submit a PR, we'd happily take a look at it!

PR #3006, opened 2021-08-17
Add referrerPolicy to AnchorHTMLAttributes

wavebeem:
Add `referrerPolicy` to `AnchorHTMLAttributes`

Fixes https://github.com/ionic-team/stencil/issues/2933


rwaskiewicz (approved):
Thanks!

rwaskiewicz:
A quick note -  the team is considering this a non-breaking change, as this shouldn't affect most folks using these types. For those that are doing type manipulations that stripping the optional aspect of this interface (e.g. `type myType = Required<AnchorHTMLAttributes>`), and are not specifying the referrerPolicy today in any way, set `referrerPolicy` to `strict-origin-when-cross-origin`.

PR #3007, opened 2021-08-18
experiment(compiler): Rewrite compilerCtx to be a singleton

splitinfinities:
DO NOT REVIEW: it big and not done

As I've been working with the codebase, I've noticed I have to change multiple files often to complete some simple tasks - things like getting the current list of components that will be compiled over the course of the Telemetry PR. 

With the Telemetry work, I added a singleton to help to be able to access config and other helpful features quickly. 

This PR expands on that concept, introducing the `StencilCompilerContext` singleton, and rewriting each argument of a function within the Stencil codebase to refer to the singleton. 

This change is primarily intended as an ergonomics and code literacy change. It has observability consequences as well. It also has broader, non-explored-within-this-PR consequences, like allowing helper functions that could be imported as opposed to importing the singleton to call items on the function. (see example below)

We should be able to more effectively use compilerCtx throughout the lifecycle of the program, as well as have one central place to measure the performance impact of this concept over time. We can also identify common usages and write helper functions to minimize the written code:

e.g. allowing code such as: 

```ts
const someFunction = async (config: d.Config, buildCtx: d.BuildCtx, compilerCtx: d.compilerCtx, tsBuilder: ts.BuilderProgram) {
  //... logic
  compilerCtx.fs.writeFile("file.txt")
}
```

To be written more declaratively: 
```ts
import { writeFile } from "@utils";
const someFunction = async (config: d.Config, buildCtx: d.BuildCtx, tsBuilder: ts.BuilderProgram) {
  writeFile("file.txt")
}
```

---

This, admittedly, is a personal belief I follow, inherited from Sandy Metz - https://thoughtbot.com/blog/sandi-metz-rules-for-developers - Notably that no method should have more than 4 parameters. Mandating parameters that may increase infinitely to functions that often only pass them down one, two, or even three levels deep cause complexity and confusion. Most notably when the only purpose for passing the parameter down is to write a file from a result. 

This school of thought has helped me improve the readability of heavy logic and therefore, if adopted, could heighten our code literacy and velocity. 

This content will be placed in an ADR added to this PR if the direction gathers buy in.


src/compiler/build/build-finish.ts
```diff
@@ -18,6 +18,7 @@ export const buildFinish = async (buildCtx: d.BuildCtx) => {
 };
 
 export const buildAbort = (buildCtx: d.BuildCtx) => {
+  // Why is this using the buildCtx's config and compilerCtx? This feels strange. Why are they even duplicated into the buildCtx? Ergonomics?
```
splitinfinities, line 21: This is the only time that the `buildCtx.compilerCtx` is used - with the exception of the usage internal to the instance of the `buildCtx`. The singleton pattern solves for this. Feels bizarre to know that these properties are duplicated into the `buildCtx`, which gets recreated over and over again throughout the course of a `build --watch` lifecycle.

src/compiler/build/compiler-ctx.ts
```diff
@@ -17,7 +17,7 @@ export class CompilerContext implements d.CompilerCtx {
   activeFilesUpdated: string[] = [];
   activeDirsAdded: string[] = [];
   activeDirsDeleted: string[] = [];
-  addWatchDir: (path: string) => void = noop;
+  addWatchDir: (path: string, recursive: boolean) => void = noop;
```
splitinfinities, line 20: Found this bug while refactoring this. Apparently this was typed incorrectly, and downstream the signature didn't expect a second parameter. That's because the interface definition was divorced from the class definition. There may be more that will pop up this way too.

splitinfinities:
Closing bc after stewing on it, I don't think I agree with this idea.

PR #3009, opened 2021-08-18
chore(types) remove bespoke rollup types

rwaskiewicz:
this PR replaces our custom typings for rollup hooks with those
provided by rollup

since this is only a change to the typings, I did not do any functional testing, rather ensured the compiler still ran


splitinfinities (approved):
LGTM!

PR #3010, opened 2021-08-19
chore(docs): document the release process 

rwaskiewicz:
While going through the release process more thoroughly this morning, I did a fair amount of documentation of the code/fixed up some typings where I could. Rather than creating a Notion doc that would sit and rot, I decided to put this in the form of code comments that hopefully won't rot as quickly 😄 

Changes:
document the build process via JSDocs on all functions that are involved
with generating a release for Stencil. the intent here is that moving
forward, context for different aspects of the release process can be
quickly acquired rather than running through a mountain of code

Testing:
- Ran `npm run tsc.scripts` to verify the typings I edited are semantically correct
- Ran `npm run release.prepare` with the 'dry-run' option to best estimate going through the release process

### DIFF
--- a/scripts/build.ts
+++ b/scripts/build.ts
@@ -14,8 +14,14 @@ import { sysNode, sysNodeExternalBundles } from './bundles/sys-node';
 import { testing } from './bundles/testing';
 import { validateBuild } from './test/validate-build';
 import { rollup } from 'rollup';
+import type { RollupOptions } from 'rollup';
 
-export async function run(rootDir: string, args: string[]) {
+/**
+ * Runner for releasing a new version of Stencil
+ * @param rootDir the root directory of the Stencil repository
+ * @param args stringifed arguments that influence the release process
+ */
+export async function run(rootDir: string, args: ReadonlyArray<string>): Promise<void> {
   try {
     if (args.includes('--release')) {
       await release(rootDir, args);
@@ -34,7 +40,12 @@ export async function run(rootDir: string, args: string[]) {
   }
 }
 
-export async function createBuild(opts: BuildOptions) {
+/**
+ * Build the rollup configuration for each submodule of the project
+ * @param opts build options to be used as a part of the configuration generation
+ * @returns the rollup configurations used to build each of the project's major submodules
+ */
+export async function createBuild(opts: BuildOptions): Promise<readonly RollupOptions[]> {
   await Promise.all([
     emptyDir(opts.output.cliDir),
     emptyDir(opts.output.compilerDir),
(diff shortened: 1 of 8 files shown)


scripts/build.ts
```diff
@@ -14,8 +14,14 @@ import { sysNode, sysNodeExternalBundles } from './bundles/sys-node';
...
 import { rollup } from 'rollup';
+import type { RollupOptions } from 'rollup';
 
-export async function run(rootDir: string, args: string[]) {
+/**
+ * Runner for releasing a new version of Stencil
+ * @param rootDir the root directory of the Stencil repository
+ * @param args stringifed arguments that influence the release process
+ */
+export async function run(rootDir: string, args: ReadonlyArray<string>): Promise<void> {
```
rwaskiewicz, line 24: Generally, if I could use the readonly variant of arrays, I did. Arguments and the like probably shouldn't change while we're doing a release

scripts/license.ts
```diff
@@ -105,19 +112,29 @@ ${bundledDeps.map((l) => l.content).join('\n')}
   fs.writeFileSync(join(opts.buildDir, 'license-source.txt'), licenseSource.join('\n'));
 }
 
-function createBundledDeps(opts: BuildOptions, bundledDeps: BundledDep[], deps: string[]) {
-  if (Array.isArray(deps)) {
```
rwaskiewicz, line 109: I removed this `if` statement, as the typings _should_ prevent a non-array from being passed in

scripts/readme.md
```diff
@@ -1,15 +1,9 @@
-# Local Build and Testing
```
rwaskiewicz, line 1: This is already in the [CONTRIBUTING.md doc](https://github.com/ionic-team/stencil/blob/master/.github/CONTRIBUTING.md#testing-changes-against-a-project-locally)

scripts/release.ts
```diff
@@ -40,34 +45,41 @@ export async function release(rootDir: string, args: string[]) {
...
-        {
-          name: 'Dry Run',
-          value: getNewVersion(oldVersion, 'patch') + '-dryrun',
-        },
-        {
-          name: 'Other (specify)',
-          value: null,
-        },
-      ]),
-      filter: (input: any) => (isValidVersionInput(input) ? getNewVersion(oldVersion, input) : input),
```
rwaskiewicz, line 70: This felt a little iffy to me - I really didn't like that 'magic number 2' above, and this is how I though it'd best be refactored to remove that. LMK if you'd prefer I revert this

scripts/build.ts
splitinfinities: TIL!

scripts/release.ts
splitinfinities: lgtm

splitinfinities (approved):
This is looking great! Awesome work.

scripts/license.ts
```diff
@@ -181,47 +198,68 @@ function createBundledDepLicense(opts: BuildOptions, moduleId: string): BundledD
...
+ * @returns the formatted contributor information
+ */
+function getAuthor(contributor: any): string {
+  if (typeof contributor === 'string') {
+    return contributor;
   }
-  if (typeof c.name === 'string') {
-    if (typeof c.url === 'string') {
-      return `[${c.name}](${c.url})`;
+  if (typeof contributor.name === 'string') {
```
splitinfinities, line 242: I appreciate the clarity in renaming this variable.

PR #3012, opened 2021-08-20
fix: Solve issue where worker thread didn't have access to fetch

### LINKED ISSUE #3011: Stencil web API not transpiling correctly in 2.7.0
**Stencil version:**

```
 @stencil/core@2.7.0
```

**I'm submitting a:**

[x] bug report
[ ] feature request

**Current behavior:**
The `transpile` method of `@stencil/core/compiler/stencil.js` is no longer transpiling the source code given.
It seems to be returning the source code back as is, i.e the jsx/typescript is not compiled.

**Expected behavior:**
The transpile method should work the same as previous versions (2.6.0 or earlier).

**Steps to reproduce:**

See this fiddle: https://jsfiddle.net/dogoku/25Lstbfv/

**Related code:**

```tsx
// assuming "https://cdn.jsdelivr.net/npm/@stencil/core@2.7.0/compiler/stencil.min.js" was loaded

const src = `
  import { h } from "https://cdn.skypack.dev/@stencil/core";
  @Component({ tag: 'cmp-a' })
  export class CmpA {
    @Prop() name
    render() {
      return <h1>Hello {this.name}</h1>
    }
  }
`;
const result = window.stencil.transpileSync(src, { sourceMap: false });
console.log(result.code)
```
Using, 2.7.0, `result.code` is the same as `src`. Using 2.6.0, `result.code` should be: 

```js
import { defineCustomElement as __stencil_defineCustomElement } from "@stencil/core/internal/client";
import { h } from "https://cdn.skypack.dev/@stencil/core";
const CmpA = class extends HTMLElement {
  constructor() {
    super();
    this.__registerHost();
  }
  render() {
    return h("h1", null,
      "Hello ",
      this.name);
  }
};
__stencil_defineCustomElement(CmpA, [0, "cmp-a", {
    "name": [8]
  }]);
```

splitinfinities:
Closes #3011

The worker thread's in-memory system did not have the correct reference to fetch. 

How to test:
1. Check this branch out
2. Build the compiler with `npm run build`
3. Visit the `/test/browser-compile` directory. 
4. Run `npm i && npm start`
5. This app will display an app that runs the compiler in the browser. It lets you run the compiler with different settings. 

Originally, before this, the transpile results pane displayed empty and there were multiple errors in dev tools.

### DIFF
--- a/src/compiler/sys/stencil-sys.ts
+++ b/src/compiler/sys/stencil-sys.ts
@@ -505,7 +505,15 @@ export const createSystem = (c?: { logger?: Logger }) => {
     return results;
   };
 
-  const fetch = self?.fetch ?? global?.fetch ?? window?.fetch;
+  // Reference fetch from a worker thread, then a browser thread, then a node thread.
+  const fetch =
+    typeof self !== 'undefined'
+      ? self?.fetch
+      : typeof window !== 'undefined'
+      ? window?.fetch
+      : typeof global !== 'undefined'
+      ? global?.fetch
+      : undefined;
 
   const writeFile = async (p: string, data: string) => writeFileSync(p, data);
 
--- a/src/compiler/sys/stencil-sys.ts
+++ b/src/compiler/sys/stencil-sys.ts
@@ -506,13 +506,13 @@ export const createSystem = (c?: { logger?: Logger }) => {
   };
 
   /**
-  * `self` is the global namespace object used within a web worker.
-  * `window` is the browser's global namespace object (I reorganized this to check the reference on that second)
-  * `global` is Node's global namespace object. https://nodejs.org/api/globals.html#globals_global
-  *
-  * loading in this order should allow workers, which are most common, then browser, 
-  * then Node to grab the reference to fetch correctly. 
-  */
+   * `self` is the global namespace object used within a web worker.
+   * `window` is the browser's global namespace object (I reorganized this to check the reference on that second)
+   * `global` is Node's global namespace object. https://nodejs.org/api/globals.html#globals_global
+   *
+   * loading in this order should allow workers, which are most common, then browser,
+   * then Node to grab the reference to fetch correctly.
+   */
   const fetch =
     typeof self !== 'undefined'
       ? self?.fetch
--- a/src/compiler/sys/stencil-sys.ts
+++ b/src/compiler/sys/stencil-sys.ts
@@ -505,7 +505,14 @@ export const createSystem = (c?: { logger?: Logger }) => {
     return results;
   };
 
-  // Reference fetch from a worker thread, then a browser thread, then a node thread.
+  /**
+  * `self` is the global namespace object used within a web worker.
+  * `window` is the browser's global namespace object (I reorganized this to check the reference on that second)
+  * `global` is Node's global namespace object. https://nodejs.org/api/globals.html#globals_global
+  *
+  * loading in this order should allow workers, which are most common, then browser, 
+  * then Node to grab the reference to fetch correctly. 
+  */
   const fetch =
     typeof self !== 'undefined'
       ? self?.fetch
--- a/src/compiler/sys/stencil-sys.ts
+++ b/src/compiler/sys/stencil-sys.ts
@@ -505,7 +505,7 @@ export const createSystem = (c?: { logger?: Logger }) => {
     return results;
   };
 
-  const fetch = global.fetch;
+  const fetch = self?.fetch ?? global?.fetch ?? window?.fetch;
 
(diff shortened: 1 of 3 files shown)


src/compiler/sys/stencil-sys.ts
```diff
@@ -505,7 +505,7 @@ export const createSystem = (c?: { logger?: Logger }) => {
     return results;
   };
 
-  const fetch = global.fetch;
+  const fetch = self?.fetch ?? global?.fetch ?? window?.fetch;
```
splitinfinities, line 508: Truthfully, this can be executed in multiple spaces. So what I'm trying to do here is worker, then node, then window. I'm open to swapping the order here too.

test/browser-compile/src/components/app-root/app-root.tsx
```diff
@@ -57,7 +57,6 @@ export class AppRoot {
   }
 
   async compile() {
-    console.clear();
```
splitinfinities, line 60: When running this app, the console gets cleared after the worker loads, so the traspile problems weren't immediately obvious when I opened dev tools. 

PS: We should figure out how to automate this with Karma.

test/browser-compile/src/utils/css-template-plugin.ts
```diff
@@ -70,7 +70,7 @@ export const cssTemplatePlugin = {
   async load(id: string) {
     let code = styleImports.get(id.split('?')[0]);
     if (code) {
-      const results = await stencil.compile(code, {
+      const results = await stencil.transpile(code, {
```
splitinfinities, line 73: `compile` doesn't exist on the Stencil compiler API.

src/compiler/sys/stencil-sys.ts
rwaskiewicz: I couple questions/maybe asks here -

- How do `self` and `window` differ here?
- Can we add a comment here to explain how/why we chose this order?
- In what circumstances do we use the `fetch` on `self`, `global`, & `window`?

test/browser-compile/src/utils/css-template-plugin.ts
rwaskiewicz: Was this being masked by https://github.com/ionic-team/stencil/pull/3012#discussion_r693077180 clearing the console?

src/compiler/sys/stencil-sys.ts
splitinfinities: `self` is the global namespace object used within a [web worker](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers#web_workers_api).

`window` is the browser's global namespace object (I reorganized this to check the reference on that second)

`global` is Node's global namespace object. https://nodejs.org/api/globals.html#globals_global

I did switch these to do worker => browser => node

test/browser-compile/src/utils/css-template-plugin.ts
splitinfinities: This failed when running the start command in this directory, because compile didn't exist. It was a TypeScript error that was accurate, I'm assuming because it's been a while since this `browser-compile` project was started up. 

The comment you reference actually covered up errors in the worker, which if it wasn't here, would have immediately told me that how we get the fetch reference for the in-memory system was erroring.

src/compiler/sys/stencil-sys.ts
rwaskiewicz: 👍 I think that's what I'm looking for in the the ask for a comment. ATM, the comment is kind of repeating the code:
```typescript
  // Reference fetch from a worker thread, then a browser thread, then a node thread.
  const fetch =
    typeof self !== 'undefined'
      ? self?.fetch
      : typeof window !== 'undefined'
      ? window?.fetch
      : typeof global !== 'undefined'
      ? global?.fetch
      : undefined;
```

which doesn't do "future Will and Ryan" much good. Ideally we explain why we choose this order (why is window over global, and self over both of them?) and potentially the context they're used in (what you've got in the [previous comment](https://github.com/ionic-team/stencil/pull/3012#discussion_r693105559))

test/browser-compile/src/components/app-root/app-root.tsx
rwaskiewicz: Ah, I see now. It was clearing debugging statements in this component itself

And yes! Can you please put an item on the refinement sheet to discuss next week?

splitinfinities: Done!

src/compiler/sys/stencil-sys.ts
splitinfinities: Added!

rwaskiewicz (approved):

Issue #3013, opened 2021-08-22
ionic-pwa failing to start


adrm:
**Stencil version:**

```
 @stencil/core@2.7.0
```

**I'm submitting a:**

[x] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**

A new stencil 2.7 project will fail when trying to run npm start. The problem disappears if the version is changed back to 2.6. The resulting debug log is the following.

```
0 info it worked if it ends with ok
1 verbose cli [ '/usr/bin/node', '/usr/bin/npm', 'start' ]
2 info using npm@6.14.13
3 info using node@v14.17.0
4 verbose run-script [ 'prestart', 'start', 'poststart' ]
5 info lifecycle testproject@0.0.1~prestart: testproject@0.0.1
6 info lifecycle testproject@0.0.1~start: testproject@0.0.1
7 verbose lifecycle testproject@0.0.1~start: unsafe-perm in lifecycle true
8 verbose lifecycle testproject@0.0.1~start: PATH: /usr/lib/node_modules/npm/node_modules/npm-lifecycle/node-gyp-bin:/var/home/adrian/Proyectos/testproject/node_modules/.bin:/var/home/adrian/.deno/bin:/var/home/adrian/.cargo/bin:/var/home/adrian/.local/bin:/var/home/adrian/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
9 verbose lifecycle testproject@0.0.1~start: CWD: /var/home/adrian/Proyectos/testproject
10 silly lifecycle testproject@0.0.1~start: Args: [ '-c', 'stencil build --dev --watch --serve' ]
11 silly lifecycle testproject@0.0.1~start: Returned: code: 1  signal: null
12 info lifecycle testproject@0.0.1~start: Failed to exec start script
13 verbose stack Error: testproject@0.0.1 start: `stencil build --dev --watch --serve`
13 verbose stack Exit status 1
13 verbose stack     at EventEmitter.<anonymous> (/usr/lib/node_modules/npm/node_modules/npm-lifecycle/index.js:332:16)
13 verbose stack     at EventEmitter.emit (events.js:376:20)
13 verbose stack     at ChildProcess.<anonymous> (/usr/lib/node_modules/npm/node_modules/npm-lifecycle/lib/spawn.js:55:14)
[...]

PR #3014, opened 2021-08-23
fix(telemetry): handle malformed telemetry tokens 

rwaskiewicz:
Fixes: #3013 

**Summary:**

this fix handles the event that a user has an Ionic configuration file,
but does not have a 'tokens.telemetry' field in said file.

**Replication:**
With `master@HEAD`, modify your local `~/.ionic/config.json` in one of two ways
1. remove `'tokens.telemetry'` completely
2. set `'tokens.telemetry'` to undefined (or any other valid JSON)

Then, spin up a small Stencil project, and attempt to turn telemetry off/on:
 ```shell
$ cd <your_sandbox_dir>
$ npm init stencil
<create an ionic pwa app>
$ cd <your_app_name>
$ npx stencil telemetry on
[25:32.7]  @stencil/core
[25:32.8]  v2.7.0 🌟

[ ERROR ]  uncaught cli error: TypeError: Cannot read property 'match' of undefined
```

**Testing:**
With the HEAD of this branch, build it and install it in the PWA app you build
```shell
$ cd stencil
$ npm run build && npm pack
$ cd <your_sandbox_dir>/<your_app_name>
$ npm i <location_of_stencil_tarball>
$  ionic-pwa npx stencil telemetry on
[28:20.0]  @stencil/core
[28:20.1]  [LOCAL DEV] 🏆

Telemetry is now Enabled. Thank you for helping to make Stencil better! 💖
```

### DIFF
--- a/src/cli/ionic-config.ts
+++ b/src/cli/ionic-config.ts
@@ -15,6 +15,11 @@ export interface TelemetryConfig {
   'tokens.telemetry'?: string;
 }
 
+/**
+ * Reads an Ionic configuration file from disk, parses it, and performs any necessary corrections to it if ceratin
+ * values are deemed to be malformed
+ * @returns the config read from disk that has been potentially been updated
+ */
 export async function readConfig(): Promise<TelemetryConfig> {
   let config: TelemetryConfig = await readJson(defaultConfig());
 
@@ -25,7 +30,10 @@ export async function readConfig(): Promise<TelemetryConfig> {
     };
 
     await writeConfig(config);
-  } else if (!!config && !config['tokens.telemetry'].match(UUID_REGEX)) {
+  } else if (
+    !!config &&
+    (typeof config['tokens.telemetry'] !== 'string' || !config['tokens.telemetry'].match(UUID_REGEX))
+  ) {
     const newUuid = uuidv4();
     await writeConfig({ ...config, 'tokens.telemetry': newUuid });
     config['tokens.telemetry'] = newUuid;
--- a/src/cli/ionic-config.ts
+++ b/src/cli/ionic-config.ts
@@ -30,10 +30,7 @@ export async function readConfig(): Promise<TelemetryConfig> {
     };
 
     await writeConfig(config);
-  } else if (
-    !!config &&
-    (typeof config['tokens.telemetry'] !== 'string' || !config['tokens.telemetry'].match(UUID_REGEX))
-  ) {
+  } else if (!UUID_REGEX.test(config['tokens.telemetry'])) {
     const newUuid = uuidv4();
     await writeConfig({ ...config, 'tokens.telemetry': newUuid });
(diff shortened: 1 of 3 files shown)


src/cli/ionic-config.ts
```diff
@@ -25,7 +30,10 @@ export async function readConfig(): Promise<TelemetryConfig> {
     };
 
     await writeConfig(config);
-  } else if (!!config && !config['tokens.telemetry'].match(UUID_REGEX)) {
```
rwaskiewicz, line 28: This is the line that was failing. If the config file existed, but 'tokens.telemetry' either:
- existed and wasn't a string
- did not exist at all

we would call `match()` on "something" that didn't have it in it's prototype chain

src/cli/ionic-config.ts
```diff
@@ -25,7 +30,10 @@ export async function readConfig(): Promise<TelemetryConfig> {
...
+  } else if (
+    !!config &&
```
rwaskiewicz, line 34: @splitinfinities is there a reason I can't remove this check? Per [this condition](https://github.com/ionic-team/stencil/pull/3014/files#diff-e5b3f8482d62190b36ef5298593adece27d38744b93bb9d8f4f29fcfa2716965R26), this _should_ always return `true`

src/cli/ionic-config.ts
ltm: It should be safe to do the inverse and call `RegExp.test()` with "something" as an argument, i.e. `!UUID_REGEX.test(config['tokens.telemetry'])`.

rwaskiewicz: Ahhh good point - done in https://github.com/ionic-team/stencil/pull/3014/commits/fc90181e3c2a600a9efcdd8a87aaf77bf70628ea

ltm (approved):

src/cli/test/ionic-config.spec.ts
```diff
@@ -14,34 +14,56 @@ describe('readConfig', () => {
...
+    const result = await getCompilerSystem().stat(defaultConfig());
 
     expect(result.isFile).toBe(true);
 
     const config = await readConfig();
 
     expect(Object.keys(config).join()).toBe('telemetry.stencil,tokens.telemetry');
     expect(config['telemetry.stencil']).toBe(true);
-    expect(!!config['tokens.telemetry'].match(UUID_REGEX)).toBe(true);
+    expect(config['tokens.telemetry']).toMatch(UUID_REGEX);
```
splitinfinities, line 43: TIL!

splitinfinities (approved):
LGTM!

src/cli/ionic-config.ts
splitinfinities: I intended this to be a guard clause to protect against accessing config if it ended up not being defined. I think what you changed it to is sensible as well.

PR #3027, opened 2021-08-25
docs(rollup): Add an ADR + JSDoc to the bundle-output.ts

splitinfinities:
The goal of this ADR is to provide clarity around why the order of rollup plugins are executed during a build.  

This ADR will be a work in progress for a while, I imagine. The end goal of this is to have each rollup plugin documented, and the order of which communicated to the team. 

If you would like to contribute, please either comment your perspective of leave a PR pointing to the `adr-rollup-plugins-ordering` branch - that's highly encouraged.

### DIFF
--- a/docs/adr/0010-rollup-plugins-build-order.md
+++ b/docs/adr/0010-rollup-plugins-build-order.md
@@ -1,0 +1,31 @@
+# 10. Rollup plugins build order
+
+Date: 2021-08-25
+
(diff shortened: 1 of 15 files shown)


rwaskiewicz:
Since this PR has fallen out of sync with `main` and was a team effort that didn't take off, I'm going close this. Thanks again!

Issue #2926, opened 2021-06-04
CLI Flag --stats does nothing


pekayde:
**Stencil version:**
```
 @stencil/core@2.6.0
```
same issue already in
```
 @stencil/core@2.0.1
```

**I'm submitting a:**
[x] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**
The command `stencil build --stats` does not produce a stencil-stats.json as it is described here https://stenciljs.com/docs/cli

**Expected behavior:**
A file stencil-stats.json is created like described herehttps://stenciljs.com/docs/cli

**Steps to reproduce:**
Add `--stats` to your build script
Run `npm run build`

splitinfinities:
Hey there! Thanks for adding this. We don't have a timeframe, but we have identified this issue. I'll circle back around as we make progress on this.

rwaskiewicz:
I'm going to close this issue, as [v2.9.0 of Stencil](https://github.com/ionic-team/stencil/releases/tag/v2.9.0) resolves this issue. If it does not, please feel free to open another ticket!

PR #3030, opened 2021-08-26
fix: restore stats output target functionality.

splitinfinities:
## Pull request checklist

- [x] Prettier (`npm run format`) was run locally and passed

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [x] Build related changes

When the stats flag, or the stats output target was added to a consumers repo, the stencil-stats.json file was not generated. 

GitHub Issue Number: https://github.com/ionic-team/stencil/issues/2926
Jira: STENCIL-43

## What is the new behavior?

This file, when generated, provides a json file that folks can use in order to pull in what files were generated and how large those files were, as well as some of the configuration options passed to the compiler when it was invoked. 

- [ ] Yes
- [x] No

## Testing

Go check this app out: https://github.com/splitinfinities/stencil-build-stats

As well as this branch. Link them, and build. You'll see the stencil-stats.json file produce changes. 

## Other information

Stencil Site related PR: https://github.com/ionic-team/stencil-site/pull/773/files


src/declarations/stencil-private.ts
```diff
@@ -197,10 +197,16 @@ export interface RollupResults {
...
+  buildStats?: CompilerBuildStats;
   buildMessages: string[];
   bundleBuildCount: number;
   collections: Collection[];
   compilerCtx: CompilerCtx;
+  esmBrowserComponentBundle: BundleModule[];
+  esmComponentBundle: BundleModule[];
+  es5ComponentBundle: BundleModule[];
+  systemComponentBundle: BundleModule[];
+  commonJsComponentBundle: BundleModule[];
```
ltm, line 209: I was worried that adding these to the build context would increase the memory usage significantly. Luckily that is not the case. I ran 10 builds of ionic-framework using `master` and this PR. The increase is relatively minor:

| Build #     | `master`         | #3030            |
| ----------- | ---------------: | ---------------: |
| 1           | MEM: 682.1MB     | MEM: 658.6MB     |
| 2           | MEM: 743.8MB     | MEM: 677.3MB     |
| 3           | MEM: 546.5MB     | MEM: 757.6MB     |
| 4           | MEM: 727.2MB     | MEM: 697.9MB     |
| 5           | MEM: 731.8MB     | MEM: 658.1MB     |
| 6           | MEM: 730.2MB     | MEM: 656.2MB     |
| 7           | MEM: 738.5MB     | MEM: 749.1MB     |
| 8           | MEM: 726.2MB     | MEM: 655.7MB     |
| 9           | MEM: 709.2MB     | MEM: 684.9MB     |
| 10          | MEM: 663.2MB     | MEM: 617.7MB     |
| **Average** | **MEM: 681.3MB** | **MEM: 699.9MB** |

src/compiler/build/build-stats.ts
```diff
@@ -0,0 +1,157 @@
...
+    return {
+      // Should probably get the file size too.
+      key: bundle.entryKey,
+      components: bundle.cmps.map((c) => c.tagName),
+      bundleId: bundle.output.bundleId,
+      fileName: bundle.output.fileName,
+      imports: bundle.rollupResult.imports,
+      // code: bundle.rollupResult.code, // (use this to debug)
+      // Currently, this number is inaccurate vs what seems to be on disk.
+      byteSize: byteSize(bundle.rollupResult.code),
```
splitinfinities, line 94: @rwaskiewicz or @lars, any recs about how to get this number to be more accurate without reading the file from disk? Should I omit for the time being?

splitinfinities:
Note: I'm not quite sure how best to test this exclusively within Jest, considering I don't know how to mock components to be compiled as a part of the compilerCtx/buildCtx during the Jest process. If you have any ideas don't hesitate to point me in the right direction!

src/declarations/stencil-private.ts
rwaskiewicz: @ltm is there a process you used for this that we could doc/reproduce in the future? I could see this also being useful for build times as well

src/declarations/stencil-private.ts
```diff
@@ -199,10 +200,16 @@ export interface RollupResults {
...
+  buildStats?: CompilerBuildStats | { diagnostics: Diagnostic[] };
   buildMessages: string[];
   bundleBuildCount: number;
   collections: Collection[];
   compilerCtx: CompilerCtx;
+  esmBrowserComponentBundle: BundleModule[];
+  esmComponentBundle: BundleModule[];
+  es5ComponentBundle: BundleModule[];
+  systemComponentBundle: BundleModule[];
+  commonJsComponentBundle: BundleModule[];
```
rwaskiewicz, line 212: Consider marking these as `readonly`:
```suggestion
  esmBrowser: ReadonlyArray<CompilerBuildStatBundle>;
  esm: ReadonlyArray<CompilerBuildStatBundle>;
  es5: ReadonlyArray<CompilerBuildStatBundle>;
  system: ReadonlyArray<CompilerBuildStatBundle>;
  commonjs: ReadonlyArray<CompilerBuildStatBundle>;
```

rwaskiewicz: although that means you'd have to make some changes to `build-stats.ts`, which I added to that file as well

src/compiler/build/build-stats.ts
```diff
@@ -0,0 +1,177 @@
...
+  const statsTargets = config.outputTargets.filter((o) => o.type === 'stats') as d.OutputTargetStats[];
+
+  await Promise.all(
+    statsTargets.map((outputTarget) => {
+      config.sys.writeFile(outputTarget.file, JSON.stringify(data, null, 2));
+    })
+  );
+}
+
+function sanitizeBundlesForStats(bundleArray: d.BundleModule[]): d.CompilerBuildStatBundle[] {
```
rwaskiewicz, line 88: Consider making this return `ReadonlyArray<d.BundleModule>`
```suggestion
function sanitizeBundlesForStats(bundleArray: ReadonlyArray<d.BundleModule>): ReadonlyArray<d.CompilerBuildStatBundle> {
```

although this will require some changes to how `CompilerBuildStats` is defined, which I've suggested in the file as well

src/compiler/build/build-stats.ts
```diff
@@ -0,0 +1,177 @@
...
+}
+
+/**
+ *
+ * @param config
+ * @param buildCtx
+ * @returns
+ */
+export async function writeBuildStats(config: d.Config, data: d.CompilerBuildStats | { diagnostics: d.Diagnostic[] }) {
+  const statsTargets = config.outputTargets.filter((o) => o.type === 'stats') as d.OutputTargetStats[];
```
rwaskiewicz, line 79: Normally when we filter output targets, we do something like
```
  const outputTargets = config.outputTargets.filter(isOutputTargetDistCustomElementsBundle);
```
where `isOutputTargetDisCustomElementsBundle` is defined in [src/compiler/output-targets/output-utils.ts](https://github.com/ionic-team/stencil/blob/master/src/compiler/output-targets/output-utils.ts) as
```
export const isOutputTargetDistCustomElementsBundle = (
  o: d.OutputTarget
): o is d.OutputTargetDistCustomElementsBundle => o.type === DIST_CUSTOM_ELEMENTS_BUNDLE;
```
and `DIST_CUSTOM_ELEMENTS_BUNDLE` is a constant we define in that same file.  Is there a reason we don't do that here?

I think most of the machinery is already in place [here](https://github.com/ionic-team/stencil/blob/9fc76579a3a3e1127ba43b354a572ac40eda3770/src/compiler/output-targets/output-utils.ts#L95)

```suggestion
  const statsTargets = config.outputTargets.filter(isOutputTargetStats);
```
(you'll need to import `isOutputTargetStats` though)

src/compiler/build/build-stats.ts
```diff
@@ -0,0 +1,177 @@
...
+
+  await Promise.all(
+    statsTargets.map((outputTarget) => {
+      config.sys.writeFile(outputTarget.file, JSON.stringify(data, null, 2));
```
rwaskiewicz, line 83: Is it worth writing something out to the console should this fail to write 1+ files out if the returned object contains any `errors`?

src/compiler/build/build-stats.ts
```diff
@@ -0,0 +1,177 @@
...
+}
+
+function sanitizeBundlesForStats(bundleArray: d.BundleModule[]): d.CompilerBuildStatBundle[] {
+  if (!bundleArray) {
+    return [];
+  }
+
+  return bundleArray.map((bundle) => {
+    return {
+      // Should probably get the file size too.
```
rwaskiewicz, line 95: Is this different from `byteSize`?

src/compiler/build/build-stats.ts
rwaskiewicz: How inaccurate are we talking? Like an order of magnitude? or a few bytes?

rwaskiewicz (commented):
Do we have any plans to doc this as well? I noticed the checkbox in the PR summary was unchecked ATM

src/compiler/build/build-stats.ts
```diff
@@ -0,0 +1,177 @@
...
+        compiler: {
+          name: config.sys.name,
+          version: config.sys.version,
+        },
+        app: {
+          namespace: config.namespace,
+          fsNamespace: config.fsNamespace,
+          components: Object.keys(buildResults.componentGraph).length,
+          entries: Object.keys(buildResults.componentGraph).length,
+          bundles: buildResults.outputs.reduce((total, en) => (total += en.files.length), 0),
```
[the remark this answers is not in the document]
splitinfinities: Good catch!

src/compiler/output-targets/dist-lazy/lazy-output.ts
```diff
@@ -60,15 +60,13 @@ export const outputLazy = async (config: d.Config, compilerCtx: d.CompilerCtx, b
 
     const rollupBuild = await bundleOutput(config, compilerCtx, buildCtx, bundleOpts);
     if (rollupBuild != null) {
-      const [componentBundle] = await Promise.all([
```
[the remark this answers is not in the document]
splitinfinities: Lars generated some, but I'll write a node script to run a build 10 times to gather and report data.

src/compiler/build/build-stats.ts
splitinfinities: TIL! Thank you!

src/compiler/build/build-stats.ts
splitinfinities: I'll give it a shot, but the result of writeFile is typed as `any`... really wish that wasn't the case.

src/compiler/build/build-stats.ts
splitinfinities: No, it shouldn't be.

src/compiler/build/build-stats.ts
splitinfinities: Oh, duh. It's because the files are compressed. It's a symptom of uglifying the files being applied after the build. Ironically, there's also some larger discrepancy between esmBrowser and esm as a result of esm not being minified. Since esm is not compressed right now (there's another bug somewhere filed about this...) the esm file's byte sizes produced at this point are accurate. 

I think I'm going to rename this to "originalByteSize" instead. We should in the future figure out how to add that. Feels like an ordering problem. I'd hate to read files, as generating the stats may be _much_ slower.

splitinfinities:
@rwaskiewicz Added your feedback! Had a couple conversation starters in your comments. Also, re: docs, yes I'd love to add them, but I don't think they should yet block this from being merged. I'd like to collaborate with Anthony on the content of the page. Something for next sprint I imagine.

src/compiler/build/build-stats.ts
```diff
@@ -71,36 +71,39 @@ export function generateBuildStats(
...
+  const statsTargets = config.outputTargets.filter(isOutputTargetStats);
 
   await Promise.all(
-    statsTargets.map((outputTarget) => {
-      config.sys.writeFile(outputTarget.file, JSON.stringify(data, null, 2));
+    statsTargets.map(async (outputTarget) => {
+      const result = await config.sys.writeFile(outputTarget.file, JSON.stringify(data, null, 2));
+
+      if (result.error) {
+        config.logger.warn([`Stats failed to write file to ${outputTarget.file}: ${result.error}`]);
```
rwaskiewicz, line 86: Have we verified this won't print 
```
Stats failed to write file to [Object object]: [Object object]
```
?

src/compiler/build/build-stats.ts
```diff
@@ -1,12 +1,12 @@
...
 import type * as d from '../../declarations';
+import { isOutputTargetStats } from '../output-targets/output-utils';
 
 /**
  *
- * @param config
- * @param compilerCtx
- * @param buildCtx
- * @returns
+ * @param config the project build configuration
```
rwaskiewicz, line 7: Minor, the summary is missing in these JSDocs

rwaskiewicz:
@splitinfinities regarding doc: I don't think we can merge this without docs. I think that needs to be a part of our definition of done

src/compiler/build/build-stats.ts
splitinfinities: file should work fine, I removed the result.error so we can guard against that.

src/declarations/stencil-private.ts
```diff
@@ -196,13 +197,24 @@ export interface RollupResults {
   modules: RollupResultModule[];
 }
 
+export interface UpdatedBuildCtx {
```
rwaskiewicz, line 200: Can we rename this please? I don't love the idea of having an interface named `UpdatedBuildCtx` who isn't used across all the different build types that we have today (e.g. `custom-dist`)

src/compiler/build/test/build-stats.spec.ts
```diff
@@ -18,16 +18,20 @@ describe('generateBuildStats', () => {
...
-    const result = generateBuildStats(config, buildCtx);
+    const result: d.CompilerBuildStats = generateBuildStats(config, buildCtx);
 
     if (result.hasOwnProperty('timestamp')) {
       delete result.timestamp;
     }
 
+    if (result.hasOwnProperty('compiler') && result.compiler.hasOwnProperty('version')) {
+      delete result.compiler.version;
+    }
```
rwaskiewicz, line 29: What value was this producing that we had to delete it and remove it from the assertion below?

src/declarations/stencil-private.ts
splitinfinities: Could you recommend a better name?

src/compiler/build/test/build-stats.spec.ts
splitinfinities: The version that's built at the moment of the test's build (it appended the date and time to the version)

rwaskiewicz: Hmmm, what was it that we changed here that made this suddenly fail?

splitinfinities: It had been failing all along, actually, I just hadn't realized it because I didn't look into the build. My local build version wasn't failing because the timestamp hadn't changed.

splitinfinities:
Stencil Site's related PR: https://github.com/ionic-team/stencil-site/pull/773/files

src/compiler/output-targets/dist-lazy/lazy-output.ts
```diff
@@ -76,6 +76,8 @@ export const outputLazy = async (config: d.Config, compilerCtx: d.CompilerCtx, b
           buildCtx.es5ComponentBundle = result.buildCtx.es5ComponentBundle;
         } else if (result.name === 'esm-browser') {
           buildCtx.esmBrowserComponentBundle = result.buildCtx.esmBrowserComponentBundle;
+          buildCtx.buildResults = result.buildCtx.buildResults;
+          buildCtx.components = result.buildCtx.components;
```
rwaskiewicz, line 80: I'm not sure why we're doing this/what's going on here - can you explain this to me please?

src/compiler/build/test/build-stats.spec.ts
rwaskiewicz: I guess more concretely, what did we change in this PR to cause this to fail? Or are you saying it was failing since we turned tests back on in CI?

src/compiler/output-targets/dist-lazy/lazy-output.ts
splitinfinities: Yeah! The components and buildResults are passed on to the buildCtx within this file from the resultant builds. These needed to be set on the buildCtx here because the promises don't update the buildCtx any more, and the original stats code needs to understand the state of the updated buildCtx after the promises all run. I wrote it this way so that we could keep the parallelization of the builds.

src/compiler/build/test/build-stats.spec.ts
splitinfinities: I mean quite literally that all the tests in all of my previous commits failed on CI because the built compiler's version includes a timestamp, so I deleted the version that includes the timestamp.

rwaskiewicz: Gotcha. Arguably, we should mock timestamp out, but I think that may be out of scope for this PR. Would you mind creating a ticket and adding it to the backlog?

src/compiler/output-targets/dist-lazy/lazy-output.ts
rwaskiewicz: Does this affect the memory usage of Stencil in a negative way? I mean, I'm sure there's some cost associated with it, but do you have an idea of the order of magnitude?

splitinfinities: I suspect this comment still holds true, and may likely improve since I brought the parallelization back. 

> I was worried that adding these to the build context would increase the memory usage significantly. Luckily that is not the case. I ran 10 builds of ionic-framework using `master` and this PR. The increase is relatively minor:
> 
> Build #	`master`	#3030
> 1	MEM: 682.1MB	MEM: 658.6MB
> 2	MEM: 743.8MB	MEM: 677.3MB
> 3	MEM: 546.5MB	MEM: 757.6MB
> 4	MEM: 727.2MB	MEM: 697.9MB
> 5	MEM: 731.8MB	MEM: 658.1MB
> 6	MEM: 730.2MB	MEM: 656.2MB
> 7	MEM: 738.5MB	MEM: 749.1MB
> 8	MEM: 726.2MB	MEM: 655.7MB
> 9	MEM: 709.2MB	MEM: 684.9MB
> 10	MEM: 663.2MB	MEM: 617.7MB
> **Average**	**MEM: 681.3MB**	**MEM: 699.9MB**

src/compiler/build/test/build-stats.spec.ts
splitinfinities: STENCIL-137

rwaskiewicz (approved):

PR #3042, opened 2021-09-01
feat(markdown): improved export to markdown

Jagget:
This PR will improve the generation of `readme.md` and make the MD tables more universal.
Some MD editors do not respect  `\|` (e.g. JetBrains WebStorm)

```md
| Property | Attribute | Description    | Type                | Default |
| -------- | --------- | -------------- | ------------------- | ------- |
| `hello`  | `hello`   | This is a prop | `boolean \| string` | `false` |
| `hello`  | --        | This is a prop | `boolean \| string` | `false` |


| Property           | Attribute          | Description    | Type                               | Default            |
| ------------------ | ------------------ | -------------- | ---------------------------------- | ------------------ |
| <code>hello</code> | <code>hello</code> | This is a prop | <code>boolean &#124; string</code> | <code>false</code> |
| <code>hello</code> | --                 | This is a prop | <code>boolean &#124; string</code> | <code>false</code> |
```

### DIFF
--- a/src/compiler/docs/readme/docs-util.ts
+++ b/src/compiler/docs/readme/docs-util.ts
@@ -31,7 +31,7 @@ export class MarkdownTable {
 
 const escapeMarkdownTableColumn = (text: string) => {
   text = text.replace(/\r?\n/g, ' ');
-  text = text.replace(/\|/g, '\\|');
+  text = text.replace(/\|/g, '&#124;');
   return text;
 };
 
--- a/src/compiler/docs/readme/markdown-css-props.ts
+++ b/src/compiler/docs/readme/markdown-css-props.ts
@@ -14,7 +14,7 @@ export const stylesToMarkdown = (styles: d.JsonDocsStyle[]) => {
   table.addHeader(['Name', 'Description']);
 
   styles.forEach((style) => {
-    table.addRow([`\`${style.name}\``, style.docs]);
+    table.addRow([`<code>${style.name}</code>`, style.docs]);
   });
 
   content.push(...table.toMarkdown());
--- a/src/compiler/docs/readme/markdown-events.ts
+++ b/src/compiler/docs/readme/markdown-events.ts
@@ -15,7 +15,7 @@ export const eventsToMarkdown = (events: d.JsonDocsEvent[]) => {
   table.addHeader(['Event', 'Description', 'Type']);
 
   events.forEach((ev) => {
-    table.addRow([`\`${ev.event}\``, getDocsField(ev), `\`CustomEvent<${ev.detail}>\``]);
+    table.addRow([`<code>${ev.event}</code>`, getDocsField(ev), `<code>CustomEvent<${ev.detail}></code>`]);
   });
 
   content.push(...table.toMarkdown());
--- a/src/compiler/docs/readme/markdown-methods.ts
+++ b/src/compiler/docs/readme/markdown-methods.ts
@@ -11,7 +11,7 @@ export const methodsToMarkdown = (methods: d.JsonDocsMethod[]) => {
   content.push(``);
 
   methods.forEach((method) => {
-    content.push(`### \`${method.signature}\``);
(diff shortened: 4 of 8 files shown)


rwaskiewicz:
👋 

I took this for a test drive this morning, and I'm not sure I understand the underlying issue.  

What I did to test this out was:
1. Create a sample component project by running `npm init stencil` in a sandbox directory
    1. This created my component project with Stencil v2.8.0
2. Ran `npm run build`
    1. Markdown docs are generated by default for projects spun up with `npm run build`
3. Opened the project in 2 JetBrains products. The dark screen shot below is from IntelliJ 2021.2.1, the light one is from WebStorm 2021.2.1 using the default/built-in markdown renderer:
Screen Shot 2021-09-09 at 8 34 11 AM
Screen Shot 2021-09-09 at 8 34 21 AM

I didn't see anything that looked wrong immediately looking at the MD and the preview (well, the dark screenshot's table lines are a little light, but I think that's just the theme)

But, I wanted to compare the results to this branch to what's generally available today. So:
1. I pulled down this branch and built Stencil
2. Installed my locally built Stencil into the sample project
3. Opened up the project in the same JetBrains & WebStorm instances described above:
Screen Shot 2021-09-09 at 8 33 22 AM
Screen Shot 2021-09-09 at 8 33 37 AM

I'm not seeing a difference other that `<code>` over the usage of backticks (I know we're using the code point for pipes (`|`) as well). 

Can you please expound on the problem you're seeing and what this PR is solving? It's very possible I'm missing something here.

Jagget:
@rwaskiewicz thanks for digging in so thoroughly!
The point is when the Prop() has only a limited number of values.

Add this to the Component:

```ts
export type IEnvironmentType = 'local' | 'development' | 'staging' | 'preprod' | 'production' | null;
```

```ts
@Prop() env: IEnvironmentType = 'production';
```

And the result will not be so good:
image

In comparison with:
image

rwaskiewicz:
Ah, ok - I see now, I missed that detail in the PR summary's example. I'll dive back into this when I have a moment.

One additional question I had: Is there a particular reason we use `<code>` instead of backticks here? I couldn't find any definitive guidance on whether one should be used over the other.  I'm not opposed to using `<code>`, but it does clutter up the markdown file itself a little bit (IMO, I can see others preferring to read MD through a renderer all the time though)

Jagget:
@rwaskiewicz 
That's a good question! Actually, we have a whole chain of solutions here. We can't use `\|` to encode `|`, so we need to replace it with `&#124;` but we can't use `&#124;` inside the <code>``</code> ticks, so we need to replace those ticks with `<code>` tags.

rwaskiewicz:
Ah gotcha! We did a little more digging on this, particularly how markdown tables _should_ render.  Unfortunately, markdown doesn't have a well unified standard. The best we have is [CommonMark](https://commonmark.org/), which is an attempt to create a unified spec (and a 1.0 version has yet to emerge).  However, there is momentum behind this project has large projects have adopted it - GitHub, Stack Exchange, Qt, etc. As such, I'd like to follow that for Stencil's generated markdown.

Looking at the [standard proposal, example 200](https://github.github.com/gfm/#example-200):

> Include a pipe in a cell’s content by escaping it, including inside other inline spans:
```md
| f\|oo  |
| ------ |
| b `\|` az |
| b **\|** im |
```

which tells me Stencil actually may be conforming to this spec today.

Looking at JetBrains' Issue Tracker, this looks like a known issue on their end
- https://youtrack.jetbrains.com/issue/IDEA-213124
- https://youtrack.jetbrains.com/issue/IDEA-213125
- https://youtrack.jetbrains.com/issue/IDEA-222423
- https://youtrack.jetbrains.com/issue/IDEA-265099

So while I'd love for us to be able to merge this (especially as someone using IntelliJ) I don't think it's the best move for Stencil at this time.  I'm going to close this issue, but thank you so much for being so prompt and clear in your responses.  I seriously appreciate it 🙇

PR #3047, opened 2021-09-03
fix(runtime): textContent for scoped components with slots

rwaskiewicz:
## Pull request checklist

    - https://github.com/ionic-team/stencil-site/pull/767/files

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

GitHub Issue Number: https://github.com/ionic-team/stencil/issues/2801

The issue above points to a bug when using Stencil w/Maquette. The underlying cause is Stencil does not adhere to "the correct thing to do" when `textContent` is set, which Maquette is doing under the hood in the reproduction case when creating this node: 
```tsx
            h("calcite-label", {}, ["I am a text node and won't get properly relocated"]),
```

## What is the new behavior?

patch `textContent` for scoped components that use slots. because there is
no shadow dom, calling the setter/getter on `textContent` would
otherwise act on the bare HTML elements. use the stencil metadata to
relocate the text to the proper location

this functionality is hidden behind a feature flag. users may enable it
by setting `scopedSlotTextContentFix` to `true` in the `extras` section
of their Stencil configuration file. This field defaults to `false`
(keeping the same behavior that exists today) while this flag is
evaluated.

- [ ] Yes
- [X] No

## Testing

1. Pull down the reproduction in #2801, found at https://github.com/eriklharper/stencil-maquette-bug
2. Follow the reproduction steps in the issue to confirm the issue
3. Pull down this branch, and build Stencil `npm ci && npm run build.prod && npm pack`
4. In the reproduction, install the tarball that you generated in step 3 `npm i <PATH_TO_TARBALL>`
5. Update the `stencil.config.ts` file as follows:
```diff
+  extras: {
+    scopedSlotTextContentFix: true,
+  }
```
6. `npm start`, observe the text being slotted correctly:
Screen Shot 2021-09-03 at 3 50 47 PM

## Other information

this does _not_ bring this functionality over to the custom elements
build. this will be done in a future story (added to the team refinement sheet)

the second commit contains some JSDoc I created as I combed through the code. they are not strictly related/necessary for the fix, but something that were tangentially related enough for me to feel it was worth including here. I can put them in a separate PR if y'all feel it's not a good time/place to add them

### DIFF
--- a/src/runtime/dom-extras.ts
+++ b/src/runtime/dom-extras.ts
@@ -1,5 +1,6 @@
 import type * as d from '../declarations';
 import { BUILD } from '@app-data';
+import { NODE_TYPES } from '@stencil/core/mock-doc';
 import { CMP_FLAGS, HOST_FLAGS } from '@utils';
 import { PLATFORM_FLAGS } from './runtime-constants';
 import { plt, supportsShadow, getHostRef } from '@platform';
@@ -79,10 +80,12 @@ export const patchTextContent = (hostElementPrototype: HTMLElement, cmpMeta: d.C
         // get the 'default slot', which would be the first slot in a shadow tree (if we were using one), whose name is
         // the empty string
         const slotNode = getHostSlotNode(this.childNodes, '');
-        // when a slot node is found, the textContent will be found in the next sibling (text) node. only if we can
-        // find that text node should we try to pull a value from it
-        if (slotNode?.nextSibling) {
+        // when a slot node is found, the textContent _may_ be found in the next sibling (text) node, depending on how
+        // nodes were reordered during the vdom render. first try to get the text content from the sibling.
+        if (slotNode?.nextSibling?.nodeType === NODE_TYPES.TEXT_NODE) {
           return slotNode.nextSibling.textContent;
+        } else if (slotNode) {
+          return slotNode.textContent;
         } else {
           // fallback to the original implementation
           return this.__textContent;
@@ -93,10 +96,13 @@ export const patchTextContent = (hostElementPrototype: HTMLElement, cmpMeta: d.C
         // get the 'default slot', which would be the first slot in a shadow tree (if we were using one), whose name is
         // the empty string
         const slotNode = getHostSlotNode(this.childNodes, '');
-        // when a slot node is found, the textContent should be placed in the next sibling (text) node. only if we can
-        // find that text node should we try to assign a value to it it
-        if (slotNode?.nextSibling) {
+        // when a slot node is found, the textContent _may_ need to be placed in the next sibling (text) node,
+        // depending on how nodes were reordered during the vdom render. first try to set the text content on the
+        // sibling.
+        if (slotNode?.nextSibling?.nodeType === NODE_TYPES.TEXT_NODE) {
           slotNode.nextSibling.textContent = value;
+        } else if (slotNode) {
+          slotNode.textContent = value;
         } else {
           // we couldn't find a slot, but that doesn't mean that there isn't one. if this check ran before the DOM
           // loaded, we could have missed it. check for a content reference element on the scoped component and insert
--- a/src/runtime/dom-extras.ts
+++ b/src/runtime/dom-extras.ts
@@ -63,6 +63,55 @@ export const patchSlotAppendChild = (HostElementPrototype: any) => {
   };
 };
 
+/**
+ * Patches the text content of an unnamed slotted node inside a scoped component
+ * @param hostElementPrototype the `Element` to be patched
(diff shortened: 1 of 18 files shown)


rwaskiewicz:
Eagle eye! I did have your PR in mind when I was working on this 🙂 I definitely feel better equipped to start looking at that one in particular when I get the time after doing a piece of it myself.  I'm pushing this subset of work out now to offer up a fix for an enterprise customer, but that doesn't mean this can't be usurped by the implementation in #2938.  Just don't want you thinking we're ignoring it completely 😉

rwaskiewicz:
No worries - honestly after spending some time in the scoped-slot domain that PR's all the more impressive!

splitinfinities:
> this does not bring this functionality over to the custom elements
build. this will be done in a future story (added to the team refinement sheet)

Can you share a bit about this? Does some render work differ between output targets due to - I assume - Lazy Loading's extensive coupling? Can you see a path toward unifying that behavior?

rwaskiewicz:
> Can you share a bit about this? Does some render code differ between output targets due to - I assume - Lazy Loading's extensive coupling? Can you see a path toward unifying that behavior?

I think @johnjenkins is correct there (or, I came to the same conclusions laid out [here](https://github.com/ionic-team/stencil/pull/3047#issuecomment-912933088)).  

The [patch as defined in `dom-extras.ts`](https://github.com/ionic-team/stencil/pull/3047/files#diff-eb23d6e8894fa2c4119d3e85e5999853a3def0f6ebf6f4e7e214896f2d50f789R71) is agnostic of the output target that's being used.  In this PR, we're only applying it to [lazy loaded components](https://github.com/ionic-team/stencil/pull/3047/files#diff-391b4dd275b4968d6629df297b178c24ebb07597f843b3f1ce5a3d6f4de82e16R149).  

ATM, serving the custom elements bundle will result in the text node _NOT_ being placed in the correct place. So we should absolutely backport/unify this fix - I added this to the user story refinement sheet.  I didn't implement it here to keep this PR as simple as possible (and to get this fix out to the customer sooner). My thought here was that once we tackled some of the testing infrastructure work we have planned for this quarter, we'd be in a great place to backport this with tests 🙂

splitinfinities (approved):
Worked great from my local tests! :shipit:

src/runtime/dom-extras.ts
```diff
@@ -63,6 +63,55 @@ export const patchSlotAppendChild = (HostElementPrototype: any) => {
...
+    Object.defineProperty(hostElementPrototype, '__textContent', descriptor);
+
+    Object.defineProperty(hostElementPrototype, 'textContent', {
+      get(): string | null {
+        // get the 'default slot', which would be the first slot in a shadow tree (if we were using one), whose name is
+        // the empty string
+        const slotNode = getHostSlotNode(this.childNodes, '');
+        // when a slot node is found, the textContent will be found in the next sibling (text) node. only if we can
+        // find that text node should we try to pull a value from it
+        if (slotNode?.nextSibling) {
```
ltm, line 84: Why would this need to access the next sibling? Using `cmp-label` as an example, I assume `slotNode` and `slotNode.nextSibling` are the following: 
```
<Host>
  <label>
    <slot /> <!-- slotNode -->
    <div>Non-slotted text</div> <!-- slotNode.nextSibling -->
  </label>
</Host>
```

johnjenkins: it's because when not using `shadow: true` stencil tries to recreate the behaviour of slots using empty text nodes to preserve the slot position. Thus the next nodes after the slot text node, should be some slotted content.

rwaskiewicz: @ltm not exactly. Let's say we have a layout that looks something like:
```
<Host>
  <label>
    <slot />
    <div>Non-slot</div>
   </label>
</Host>
```
When we're doing slot relocation, calling `childNodes` on the label element yields:
Screen Shot 2021-09-08 at 1 59 45 PM
where `slotNode` is that first text node, and `slotNode.nextSibling` in this case is the second text node.

When we call `getHostSlotNode`, it finds the first text node, and assigns it to `slotNode` because that node has the Stencil marker '[s-sn]` (slot name) set to an empty string (which is what we wanted). 

However, in cases like this, the VDOM has placed the content in that _second_ text node. Attempting to place the value received when setting `textContent` in the first node will result in concatenation of that received value (from setting `textContent`) and what's in the second text node.  So, we check here for this case first.

However, you also presented a case internally where this isn't always the case. I've updated this PR with additonal tests and hopefully clearer comments. LMK what you think

rwaskiewicz:
Leaving a note for future us - we know this doesn't cover every edge case. We've marked this as experimental in the docs, and will likely be revisiting this in the future

PR #3064, opened 2021-09-16
fix(docs): Fix docs generation for method return values

tricki:
The `@returns` tags were ignored and instead `@return` was used, but the tag object wasn't properly converted to a string resulting in the output `[object Object]`.

## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

`@returns` was ignored and `@return` wasn't serialized, rendering it unusable.

GitHub Issue Number: N/A

## What is the new behavior?
`@returns` is properly output.

- [ ] Yes
- [x] No

Technically, you could say it's a breaking change but from a buggy state so it's very unlikely anyone has used `@return` before.

## Other information

**Before:**
image

**After:**
image


src/compiler/docs/generate-doc-data.ts
```diff
@@ -199,7 +199,10 @@ const getDocsMethods = (methods: d.ComponentCompilerMethod[]): d.JsonDocsMethod[
       name: member.name,
       returns: {
         type: member.complexType.return,
-        docs: member.docs.tags.filter((t) => t.name === 'return').join('\n'),
+        docs: member.docs.tags
+          .filter((t) => t.name === 'returns')
```
rwaskiewicz, line 203: I think we could make this a non-breaking (or at least backwards compatible) change by accepting either `@return` or `@returns` here.  Looking at [use JSDoc](https://jsdoc.app/tags-returns.html) site, the former is a synonym for the latter. FWIW, in a previous job my team and I used `@return` exclusively - I think supporting both would be OK.

This does open us up to situations where if we have _both_ `@return` and `@returns` or even duplicates:
```tsx
  /**
   * This is a description
   * @returns Nothing!
   * @return Nada!
   * @returns Not a thing!!!!!!
   */
  @Method()
  myTest() {

  }
```
would get generated as such:
```md
## Methods

### `myTest() => Promise<void>`

This is a description

#### Returns

Type: `Promise<void>`

Nothing!
Nada!
Not a thing!!!!!!
```
which is fine (IMO).

So, TL;DR, I propose we make this:
```suggestion
          .filter((t) => t.name === 'return' || t.name === 'returns')
```

rwaskiewicz (changes requested):
Thanks @tricki! I took your PR for a spin, and things are looking good. I had one small ask regarding backwards compatibility. Let me know what you think!

src/compiler/docs/generate-doc-data.ts
tricki: I didn't know `@return` was also allowed and see no reason not to allow it here, good catch. But note that it **is** technically a breaking change because before the output of `@return` was always `[object Object]`.

rwaskiewicz (approved):
Note to future me/team/community members. This is _technically_ a breaking change to the output, but is simultaneously a nice bug fix for doc generation. Since the output was previously unusable (it output `[Object object]`), I am erring on the side of this being added to the next minor release of Stencil

src/compiler/docs/generate-doc-data.ts
tricki: Happy to help. And I agree that this can go in a minor release because it's very unlikely to actually break something since it's obviously a bug and therefore shouldn't be relied upon.

Issue #3051, opened 2021-09-07
feat: Deno support


splitinfinities:
### Prequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/master/.github/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/master/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already include this feature request, without success.

### Describe the Feature Request

Deno is a runtime for JavaScript and TypeScript that has the added benefit of increased security as well as the ability to execute TypeScript code without a compilation step. 

### Describe the Use Case

Some use cases for Deno to be used are:
- For environments that need secure and controlled access. 
- ???

### Describe Preferred Solution

None as of yet. Have an idea? Pitch it! 

### Describe Alternatives

1. During Stencil's experimental support phase for Deno, Stencil built a system (which may be more aptly referred to as a platform-shim) in order to provide the necessary features to the compiler which. The things provided were functionality like making fetch requests, or writing to the file system. This is inarguably one way to write quality support for Deno. 

2. ???

Have an idea? Pitch it here!

PR #3067, opened 2021-09-16
refactor(deno): remove deno from codebase

rwaskiewicz:
## To Folks Who Would Like Deno to be Supported

Please 👍 the summary of the RFC located [here](https://github.com/ionic-team/stencil/issues/3051). Or leave a comment explaining your use case for Deno (but please do not leave a comment with just  '👍') 

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

We have a good amount of code for Deno, which we sunset recently. It doesn't actually function today.

GitHub Issue Number: N/A

## What is the new behavior?

Same as it does today 😆 This just rips Deno out of the Stencil repo

- [ ] Yes
- [x] No - we've talked about this as a team. Given the fact it does not work today, we do not consider this removal to be a breaking change

## Testing

In addition to CI based tests:

- Built a tarball with this version of Stencil locally, and was able to use it successfully in a new app spun up with `npm init stencil`
- Smoke tested the output of the `test/browser-compiler`


src/compiler/sys/environment.ts
```diff
@@ -1,16 +1,13 @@
...
-
 export const IS_NODE_ENV =
-  !IS_DENO_ENV &&
   typeof global !== 'undefined' &&
   typeof require === 'function' &&
   !!global.process &&
   typeof __filename === 'string' &&
   (!(global as any as Window).origin || typeof (global as any as Window).origin !== 'string');
 
-export const OS_PLATFORM = IS_NODE_ENV ? process.platform : IS_DENO_ENV ? Deno.build.os : '';
```
rwaskiewicz, line 11: `Deno.build.os` could return the string `"windows"` which is why we can safely remove the check for it in the line below (in fact, the compiler yells at us because `"windows"` is not a valid `process.plaform` value

src/compiler/sys/modules/os.ts
```diff
@@ -2,7 +2,7 @@ import { OS_PLATFORM } from '../environment';
 
 export const EOL = '\n';
 
-export const platform = () => (OS_PLATFORM === 'windows' ? 'win32' : OS_PLATFORM);
```
rwaskiewicz, line 5: We're able to do this as a result of [this comment's context](https://github.com/ionic-team/stencil/pull/3067/files#r710419635)

bartlomieju:
Hi @rwaskiewicz! I'm one of Deno maintainers. I was just about to open a PR updating vendored Deno dependencies (namely `deno.land/std@0.63.0` to `deno.land/std@0.108.0`) and found this PR.

I was very impressed how you solved Deno compatiblity using `sys` pattern and I wanted to use the same pattern in other projects to provide Deno compatibility. FYI we've significantly improved our Node compatibility layer since you started using it. 

I'm very curious to hear your thoughts about the subject and problems that you faced, and to know if there's anything we could do to help out with those problems.

rwaskiewicz:
Hey @bartlomieju! 

Thanks! Although I can’t take credit for the sys abstraction, it’s a certainly a handy one :slightly_smiling_face:

With regards to Deno support in Stencil, we found that our Deno support (which had been marked as experimental by a previous maintainer) wasn’t being actively used. We polled our community, and weren’t able to surface any active users of Stencil + Deno.  Additionally, when we tried to understand the current state of the functionality, we realized that it had broken some time ago, and hadn’t been noticed. 

The new team behind Stencil is relatively small, and is still getting ramped up on all of its inner workings.  While we’d like to support as much as possible, we realized we have to be judicious about the functionality that we support.  Due to the lack of usage, we felt that refurbishing Deno support was not going to serve the needs of our community.

With all of that being said, if there is community desire to bring Deno support back, we’re absolutely up for it.  That sys abstraction makes adding runtimes rather easy,  Also, if you’re interested to hear more/talk further, I’m happy to do so!

ltm (approved):

bartlomieju:
@rwaskiewicz thanks for the response! 

> With regards to Deno support in Stencil, we found that our Deno support (which had been marked as experimental by a previous maintainer) wasn’t being actively used. We polled our community, and weren’t able to surface any active users of Stencil + Deno. Additionally, when we tried to understand the current state of the functionality, we realized that it had broken some time ago, and hadn’t been noticed.

> The new team behind Stencil is relatively small, and is still getting ramped up on all of its inner workings. While we’d like to support as much as possible, we realized we have to be judicious about the functionality that we support. Due to the lack of usage, we felt that refurbishing Deno support was not going to serve the needs of our community.

That makes perfect sense and I completely understand this situation.

> With all of that being said, if there is community desire to bring Deno support back, we’re absolutely up for it. That sys abstraction makes adding runtimes rather easy, Also, if you’re interested to hear more/talk further, I’m happy to do so!

That sounds great; in parallel we're ramping out efforts to make codebases supporting Node and Deno at the same time easier to maintain and hopefully in the coming weeks we'll be able to provide the solution that would lower the maintenance burden.

PR #3081, opened 2021-09-27
refactor(telemetry) relocating helpers into the @utils sub-module

splitinfinities:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

GitHub Issue Number: N/A

## What is the new behavior?

- The helpers introduced in the Telemetry PR should have been usable from within the `@util`'s submodule. Hopefully, these can be even more reused throughout the codebase. 

- [ ] Yes
- [x] No


src/utils/has-flags.ts
```diff
@@ -0,0 +1,19 @@
+import type * as d from '../declarations';
+
+/**
+ * Does the command have the debug flag?
+ * @param config The config passed into the Stencil command
+ * @returns true if --debug has been passed, otherwise false
+ */
+export function hasDebug(config: d.Config) {
+  return config.flags.debug;
```
rwaskiewicz, line 9: IIRC, we were making some assumptions around `flags` being truthy in the original implementation.  If we move this helper to be generally used by others, I don't think that assumption holds.  We should protect ourselves with a small tweak:
```suggestion
  return config.flags?.debug;
```

Can we add a test for this case?

src/utils/has-flags.ts
```diff
@@ -0,0 +1,19 @@
...
+}
+
+/**
+ * Does the command have the verbose and debug flags?
+ * @param config The config passed into the Stencil command
+ * @returns true if both --debug and --verbose have been passed, otherwise false
+ */
+export function hasVerbose(config: d.Config) {
+  return config.flags.verbose && hasDebug(config);
```
rwaskiewicz, line 18: IIRC, we were making some assumptions around `flags` being truthy in the original implementation.  If we move this helper to be generally used by others, I don't think that assumption holds.  We should protect ourselves with a small tweak:
```suggestion
  return config.flags?.verbose && hasDebug(config);
```

Can we add a test for this case?

src/utils/is-interactive.ts
```diff
@@ -0,0 +1,17 @@
+import type * as d from '../declarations';
+
+export declare const TERMINAL_INFO: d.TerminalInfo;
+
+export const isInteractive = (sys: d.CompilerSystem, config: d.Config, object?: d.TerminalInfo): boolean => {
```
rwaskiewicz, line 5: Since we're splitting this out, can we add a JSDoc for this please?

src/utils/test/has-flags.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+    const config = {
+      flags: {
+        debug: false,
+        verbose: false,
+      },
+    };
+
+    expect(hasDebug(config)).toBe(false);
+  });
+});
```
rwaskiewicz, line 25: To coincide with my test suggestion above:
```suggestion

  it('returns false when flags are not on the config', () => {
    const config = {};

    expect(hasDebug(config)).toBe(false);
  });
});
```

src/utils/test/has-flags.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+    const config = {
+      flags: {
+        debug: false,
+        verbose: false,
+      },
+    };
+
+    expect(hasVerbose(config)).toBe(false);
+  });
+});
```
rwaskiewicz, line 60: To coincide with my testing suggestion above:
```suggestion

  it('returns false when flags are not on the config', () => {
    const config = {};

    expect(hasVerbose(config)).toBe(false);
  });
});
```

src/utils/try-fn.ts
```diff
@@ -0,0 +1,9 @@
+export const tryFn = async <T extends (...args: any[]) => Promise<R>, R>(fn: T, ...args: any[]): Promise<R | null> => {
```
rwaskiewicz, line 1: Since we're pulling this out to a helper, can we JSDoc this function please?

src/utils/uuid.ts
```diff
@@ -0,0 +1,11 @@
+export const UUID_REGEX = new RegExp(/^[0-9A-F]{8}-[0-9A-F]{4}-4[0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}$/i);
+
+// Plucked from https://github.com/ionic-team/capacitor/blob/b893a57aaaf3a16e13db9c33037a12f1a5ac92e0/cli/src/util/uuid.ts
+export function uuidv4(): string {
```
rwaskiewicz, line 4: Since we're pulling this out to a helper, can we JSDoc this function please?

src/utils/uuid.ts
```diff
@@ -0,0 +1,11 @@
+export const UUID_REGEX = new RegExp(/^[0-9A-F]{8}-[0-9A-F]{4}-4[0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}$/i);
```
rwaskiewicz, line 1: Nit: Can we rename this to `UUID_V4_REGEX`?

src/utils/read-json.ts
```diff
@@ -0,0 +1,12 @@
...
+
+/**
+ * Reads and parses a JSON file from the given `path`
+ * @param sys The system where the command is invoked
+ * @param path the path on the file system to read and parse
+ * @returns the parsed JSON
+ */
+export async function readJson(sys: d.CompilerSystem, path: string): Promise<any> {
+  const file = await sys.readFile(path);
+  return !!file && JSON.parse(file);
```
rwaskiewicz, line 11: Something to consider as we pull this out - our use case is more general. In other words, although we wrote this for telemetry, other consumers may need to change this in order to make it work for them, which can have an affect on the original (telemetry) use case.

An example of this is we don't handle when/if `JSON.parse` throws. We made some assumptions around this in the original use case, which may not hold going forward.

No action required ATM, but something to be cognizant of/think about as a semi-rhetorical question: "If someone were to change this to re-`throw` an error from `JSON.parse`, how would my tests catch it? How would telemetry be affected?

rwaskiewicz (changes requested):

rwaskiewicz:
Closing, as this change  has fallen out of sync with `main`, and I don't believe is as necessary as we originally thought. Thanks again!

PR #3082, opened 2021-09-27
feat(telemetry) adding yarn 1 support, sanitizing data pre-flight

splitinfinities:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

We're unable to understand how Yarn users work with Stencil today. 
GitHub Issue Number: N/A

## What is the new behavior?

- This PR allows the yarn.lock file to be read. 
- Sanitizes two data points pre-flight - the system version (Node/Deno) and the packages array. 

- [ ] Yes
- [x] No


src/cli/telemetry/telemetry.ts
```diff
@@ -124,10 +124,12 @@ export const prepareData = async (
     component_count,
     targets,
     packages,
+    packages_no_versions,
     arguments: config.flags.args,
     task: config.flags.task,
     stencil,
-    system,
+    system: system,
```
rwaskiewicz, line 131: Non-blocking, any reason this can't stay as it is today?
```suggestion
    system,
```

src/cli/telemetry/telemetry.ts
```diff
@@ -170,20 +175,67 @@ async function getInstalledPackages(sys: d.CompilerSystem, config: d.Config): Pr
...
-          ([k, v]) =>
-            `${k}@${packageLockJson?.dependencies[k]?.version ?? packageLockJson?.devDependencies[k]?.version ?? v}`
-        )
-      : ionicPackages.map(([k, v]) => `${k}@${v}`);
+    try {
+      if (!isUsingYarn(sys)) {
+        packages = await npmPackages(sys, ionicPackages);
+      } else if (isUsingYarn(sys)) {
+        packages = await yarnPackages(sys, ionicPackages);
+      } else {
```
rwaskiewicz, line 183: Doesn't `isUsingYarn` always return a boolean? I'm not sure I see the scenario where this `else` block is entered (since we test both `isUsingYarn` in the `else if` condition and it's negation in the `if` condition clause)

In fact, I think this could be refactored as such
```
// NOTE to not be committed: I pulled this out to keep our `try` block tiny
const isUsingYarn = isUsingYarn(sys);
try {
  packages = isUsingYarn ? await yarnPackages(sys, ionicPackages) : await npmPackages(sys, ionicPackages);
} catch (e) {
   packages = ionicPackages.map(([k, v]) => `${k}@${v.replace('^', '')}`);
}

File: package.json [hunk 3]
```diff
@@ -79,6 +79,8 @@
     "@types/sizzle": "^2.3.2",
     "@types/webpack": "^4.41.26",
     "@types/ws": "^7.4.0",
+    "@types/yarnpkg__lockfile": "^1.1.5",
+    "@yarnpkg/lockfile": "^1.1.0",
```
rwaskiewicz, line 83: Can you add `@yarnpkg/lockfile` to https://github.com/ionic-team/stencil/blob/66d0476bb8acff5225926cc50769db16a5fd8ed3/scripts/license.ts#L12

src/cli/telemetry/telemetry.ts
```diff
@@ -170,20 +175,67 @@ async function getInstalledPackages(sys: d.CompilerSystem, config: d.Config): Pr
...
+
+/**
+ * Visits the yarn lock file to find the exact versions that are installed
+ * @param sys The system where the command is invoked
+ * @param ionicPackages a list of the found packages matching `@stencil`, `@capacitor`, or `@ionic` from the package.json file.
+ * @returns an array of strings of all the packages and their versions.
+ */
+async function yarnPackages(sys: d.CompilerSystem, ionicPackages: [string, string][]) {
+  const appRootDir = sys.getCurrentDirectory();
+  const yarnLock = sys.readFileSync(appRootDir + '/yarn.lock');
```
rwaskiewicz, line 224: We should use `path.join` (or the sys equivalent) here, to make the path delimiter OS-agonstic

src/cli/telemetry/telemetry.ts
```diff
@@ -170,20 +175,67 @@ async function getInstalledPackages(sys: d.CompilerSystem, config: d.Config): Pr
...
 
+/**
+ * Visits the npm lock file to find the exact versions that are installed
+ * @param sys The system where the command is invoked
+ * @param ionicPackages a list of the found packages matching `@stencil`, `@capacitor`, or `@ionic` from the package.json file.
+ * @returns an array of strings of all the packages and their versions.
+ */
+async function npmPackages(sys: d.CompilerSystem, ionicPackages: [string, string][]) {
+  const appRootDir = sys.getCurrentDirectory();
+  const packageLockJson: any = await tryFn(readJson, sys, sys.resolvePath(appRootDir + '/package-lock.json'));
```
rwaskiewicz, line 207: We should use `path.join` (or the sys equivalent) here, to make the path delimiter OS-agonstic

src/cli/telemetry/telemetry.ts
```diff
@@ -170,20 +175,67 @@ async function getInstalledPackages(sys: d.CompilerSystem, config: d.Config): Pr
...
+
+  return ionicPackages.map(([k, v]) => {
+    const identifiedVersion = `${k}@${v}`;
+    let version = packageLockJson.object[identifiedVersion]?.version;
+    version = version.includes('undefined') ? sanitizeDeclaredVersion(identifiedVersion) : version;
+    return `${k}@${version}`;
+  });
+}
+
+function sanitizeDeclaredVersion(version: string) {
```
rwaskiewicz, line 235: Can we get a JSDoc on this comment to explain what it's doing?

rwaskiewicz (changes requested):
I took a first pass on this, but didn't get around to testing it going much deeper than the surface level. LMK what you think

src/cli/telemetry/telemetry.ts
```diff
@@ -170,20 +176,67 @@ async function getInstalledPackages(sys: d.CompilerSystem, config: d.Config): Pr
...
+ * @returns an array of strings of all the packages and their versions.
+ */
+async function yarnPackages(sys: d.CompilerSystem, ionicPackages: [string, string][]) {
+  const appRootDir = sys.getCurrentDirectory();
+  const yarnLock = sys.readFileSync(sys.resolvePath(appRootDir + '/yarn.lock'));
+  const packageLockJson = sys.parseYarnLockFile(yarnLock);
+
+  return ionicPackages.map(([k, v]) => {
+    const identifiedVersion = `${k}@${v}`;
+    let version = packageLockJson.object[identifiedVersion]?.version;
```
rwaskiewicz, line 224: Since `object` is of type `any`, can we handle the case where it is `undefined`/`null`?

src/cli/telemetry/telemetry.ts
rwaskiewicz:
```suggestion
function sanitizeDeclaredVersion(version: string): string {
```

src/cli/telemetry/telemetry.ts
```diff
@@ -170,20 +176,67 @@ async function getInstalledPackages(sys: d.CompilerSystem, config: d.Config): Pr
...
+  });
+}
+
+/**
+ * Visits the yarn lock file to find the exact versions that are installed
+ * @param sys The system where the command is invoked
+ * @param ionicPackages a list of the found packages matching `@stencil`, `@capacitor`, or `@ionic` from the package.json file.
+ * @returns an array of strings of all the packages and their versions.
+ */
+async function yarnPackages(sys: d.CompilerSystem, ionicPackages: [string, string][]) {
```
rwaskiewicz, line 217:
```suggestion
async function yarnPackages(sys: d.CompilerSystem, ionicPackages: [string, string][]): Promise<string[]> {
```

src/cli/telemetry/telemetry.ts
```diff
@@ -170,20 +176,67 @@ async function getInstalledPackages(sys: d.CompilerSystem, config: d.Config): Pr
...
   }
 }
 
+/**
+ * Visits the npm lock file to find the exact versions that are installed
+ * @param sys The system where the command is invoked
+ * @param ionicPackages a list of the found packages matching `@stencil`, `@capacitor`, or `@ionic` from the package.json file.
+ * @returns an array of strings of all the packages and their versions.
+ */
+async function npmPackages(sys: d.CompilerSystem, ionicPackages: [string, string][]) {
```
rwaskiewicz, line 200:
```suggestion
async function npmPackages(sys: d.CompilerSystem, ionicPackages: [string, string][]): Promise<string[]> {
```

rwaskiewicz (changes requested):
Tested on Windows + MacOS, looking good 😎 

Left a few comments/nitpicks around typings, but nothing major

package.json
splitinfinities: Truthfully? /shrug

splitinfinities: It's a part of the compiler process, since anything within the node sys gets bundled up and shipped. Similar to how we bundle TypeScript or Rollup.

src/cli/telemetry/telemetry.ts
```diff
@@ -170,20 +176,67 @@ async function getInstalledPackages(sys: d.CompilerSystem, config: d.Config): Pr
...
     );
 
-    packages = packageLockJson
-      ? ionicPackages.map(
-          ([k, v]) =>
-            `${k}@${packageLockJson?.dependencies[k]?.version ?? packageLockJson?.devDependencies[k]?.version ?? v}`
-        )
-      : ionicPackages.map(([k, v]) => `${k}@${v}`);
+    try {
+      packages = yarn ? await yarnPackages(sys, ionicPackages) : await npmPackages(sys, ionicPackages);
```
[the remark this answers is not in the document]
splitinfinities: If the lock files cannot be read (doesn't exist/incorrect file structure), that's when these should throw.

src/cli/telemetry/telemetry.ts
```diff
@@ -170,20 +176,67 @@ async function getInstalledPackages(sys: d.CompilerSystem, config: d.Config): Pr
...
+/**
+ * Visits the yarn lock file to find the exact versions that are installed
+ * @param sys The system where the command is invoked
+ * @param ionicPackages a list of the found packages matching `@stencil`, `@capacitor`, or `@ionic` from the package.json file.
+ * @returns an array of strings of all the packages and their versions.
+ */
+async function yarnPackages(sys: d.CompilerSystem, ionicPackages: [string, string][]) {
+  const appRootDir = sys.getCurrentDirectory();
+  const yarnLock = sys.readFileSync(sys.resolvePath(appRootDir + '/yarn.lock'));
+  const packageLockJson = sys.parseYarnLockFile(yarnLock);
```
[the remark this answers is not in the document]
splitinfinities: If reading fails, it throws and resolves back to the parent try/catch block. All other paths are truthy as you say.

src/cli/telemetry/telemetry.ts
splitinfinities: We would handle that above, within the getInstalledPackages try/catch block.

src/cli/telemetry/telemetry.ts
```diff
@@ -170,20 +176,67 @@ async function getInstalledPackages(sys: d.CompilerSystem, config: d.Config): Pr
...
+}
+
+/**
+ * This function is used for fallback purposes, where an npm or yarn lock file doesn't exist in the consumers directory.
+ * This will strip away ^ and ~ from the declared package versions in a package.json.
+ * @param version the raw semver pattern identifier version string
+ * @returns a cleaned up representation without any qualifiers
+ */
+function sanitizeDeclaredVersion(version: string) {
+  return version.replace(/[*^~]/g, '');
```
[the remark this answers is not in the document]
splitinfinities: Yes, that was the intent

src/cli/telemetry/telemetry.ts
```diff
@@ -170,20 +176,67 @@ async function getInstalledPackages(sys: d.CompilerSystem, config: d.Config): Pr
...
+    const identifiedVersion = `${k}@${v}`;
+    let version = yarnLockYml.object[identifiedVersion]?.version;
+    version = version.includes('undefined') ? sanitizeDeclaredVersion(identifiedVersion) : version;
+    return `${k}@${version}`;
+  });
+}
+
+/**
+ * This function is used for fallback purposes, where an npm or yarn lock file doesn't exist in the consumers directory.
+ * This will strip away ^ and ~ from the declared package versions in a package.json.
```
rwaskiewicz, line 232:
```suggestion
 * This will strip away '*', '^' and '~' from the declared package versions in a package.json.
```

rwaskiewicz (approved):

splitinfinities:
Note that yarn 2+ are not supported

Issue #3101, opened 2021-10-12
bug: dist-custom-elements does not work with ssr


liamdebeasi:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/.github/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

2.9.0

### Current Behavior

When building a Stencil project in an SSR environment, the compilation fails with an `HTMLElement is not defined`. This is because the custom elements directly extend `HTMLElement`, but `HTMLElement` is not available in an SSR environment.

So for example, `ion-accordion` component in Ionic 6 looks like this:

```js
const Accordion = class extends HTMLElement { ... }
```

As a workaround, we could have the classes get extended from a special class that works in SSR:

```js
const HTMLElementSSR = (
  typeof HTMLElement !== 'undefined'
  ? HTMLElement
  : class {});

const Accordion = class extends HTMLElementSSR { ... }
```

### Expected Behavior

I would expect the build to not fail.

### Steps to Reproduce

1. Clone https://github.com/liamdebeasi/ionic-next-js. This is a Next.js app running Ionic React 6
2. Run `npm install`.
3. Run `npm run build`. Observe that the build gives an error: `ReferenceError: HTMLElement is not defined`. Opening the source of the error should reference an Ionic component that directly extends `HTMLElement`.

### Code Reproduction URL

https://github.com/liamdebeasi/ionic-next-js

### Additional Information

Related Ionic issue: https://github.com/ionic-team/ionic-framework/issues/24026

rwaskiewicz:
I've confirmed this has been a bug since Stencil@v2.0.0 (at the very least). We'll get this ingested into our issue tracker

rwaskiewicz:
@willmartian's PR, #3131 has been released in [release of v2.11.0](https://github.com/ionic-team/stencil/releases/tag/v2.11.0) of Stencil, which happened ~30 minutes ago. Should anyone find a new instance of it (this bug), please open a new issue!

PR #3131, opened 2021-11-03
fix(dist-custom-elements): add ssr checks

willmartian:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

GitHub Issue Number: #3101

## What is the new behavior?

- The generated `defineCustomElement` function now checks if `customElements` is defined.
- Components in the custom elements build now check if `HTMLElement` is defined before extending from it:
```js
const MyElement extends from (
  typeof HTMLElement !== 'undefined'
  ? HTMLElement
  : class {});
```

- [ ] Yes
- [x] No

## Other information

This fix is a prerequisite for https://github.com/ionic-team/ionic-framework/pull/24163


rwaskiewicz (commented):
The reason I think these tests are failing is rooted in a declaration ordering error. If you `npm ci && npm run build && pushd test/karma && npm run karma.prod`, we see the tests fail (but that much you knew 🙃)

After doing that, open up `www/custom-elements-delegates-focus/main.js` (from the `test/karma` dir) and you'll see:
```javascript
let CustomElementsDelegatesFocus$1 = class extends HTMLElementSSR {
  constructor() {
    super();
    this.__registerHost();
    this.__attachShadow();
  }
  render() {
    return (h(Host, null, h("input", null)));
  }
  static get delegatesFocus() { return true; }
  static get style() { return sharedDelegatesFocusCss; }
};
var HTMLElementSSR = (typeof HTMLElement !== "undefined" ? HTMLElement : class  {
});
``` 
If I reorder those two statements:
```diff
+ var HTMLElementSSR = (typeof HTMLElement !== "undefined" ? HTMLElement : class  {
+ });
let CustomElementsDelegatesFocus$1 = class extends HTMLElementSSR {
  constructor() {
    super();
    this.__registerHost();
    this.__attachShadow();
  }
  render() {
    return (h(Host, null, h("input", null)));
  }
  static get delegatesFocus() { return true; }
  static get style() { return sharedDelegatesFocusCss; }
};
- var HTMLElementSSR = (typeof HTMLElement !== "undefined" ? HTMLElement : class  {
- });
```
that test passes, so it looks like that aspect of the PR at least is rooted in these types of ordering issues.

I haven't given this PR too much other focus, but I think that might get you going again 🏃

willmartian:
Hi @rwaskiewicz, when I test this build in Ionic Framework, I am receiving this error in our Vue e2e tests: `Uncaught DOMException: Failed to construct 'CustomElement': The result must not have attributes.` 

The same error is described here: https://stackoverflow.com/questions/43836886/failed-to-construct-customelement-error-when-javascript-file-is-placed-in-head

I don't believe this error is being caused by any of the changes in this PR, but this PR fixes earlier points of failure in those same e2e tests (`customElements` is not defined). So I am not getting to this error in our tests unless testing this PR's build. 

The stackoverflow link above expands on the error, stating that it originates from accessing attributes or children in the custom element's constructor, which should not happen as per spec: https://html.spec.whatwg.org/multipage/custom-elements.html#custom-element-conformance

I am still familiarizing myself with the Stencil source and so I need some assistance verifying, do we follow the spec regarding the above?

I see that the generated custom element for `ion-button` for example, makes calls to `this.__registerHost` and `this.__attachShadow`. Do either of those functions potentially access attributes/children in the constructor?

rwaskiewicz (approved):

PR #3132, opened 2021-11-05
refactor(karma): relocate www output target karma tests

rwaskiewicz:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

this commit is a part of the STENCIL-48 effort. it does not bring full/flexibile
support for custom elements just yet, but i figured it be best to get this refactor
landed/out of the way first

this commit is a part of a larger effort to provide additional capacity
to stencil's karma tests for running multiple output targets.
specifically, this commit focuses on moving the existing test root from
the output of the www output target to a sub-directory.

GitHub Issue Number: N/A

## What is the new behavior?

Previously, karma tests generated a directory structure like so:
```
karma/
  test-www/  <-- Location of test output
```

This commit creates the following structure when applied:
```
karma/
  test-output/  <-- Each subdirectory contains its output target's results
    test-custom-elements/
    test-dist/
    test-www/
```

this commit also reorders some commands in the karma directory
package.json to aid in readability of scripts

- [ ] Yes
- [x] No - But developers should remove the old testing output directory once this commit lands. We could keep it in `gitignore`, but I didn't love that idea as it keeps old test files around and can make issues harder to diagnose (in theory, open to doing it anyway)

## Testing

E2E tests have been run


.prettierignore
```diff
@@ -37,11 +37,12 @@ node_modules/
 coverage/
 
 # output from compiling Stencil projects for testing purposes
-test/**/dist/
+test/browser-compile/
+test/karma/test-sibling/
```
rwaskiewicz, line 41: I realized we were accidentally running prettier on these two output dirs. We didn't notice because CI runs Prettier in a separate task and in the past (when it wasn't a separate task), we'd run it before we generated this output.  So, this only affects local runs of Prettier. Can pull out into a different commit if desired

test/karma/karma.config.js
```diff
@@ -94,16 +98,21 @@ module.exports = function (config) {
...
       // 'test-app/prerender-test/karma.spec.ts',
-      'test-app/**/*.spec.ts',
-      'test-app/util.ts',
-      'test-app/assets/angular.min.js',
-      { pattern: 'www/**/*', watched: false, included: false, served: true, nocache: true, type: 'module' },
+      'test-app/**/*.spec.ts', // tells karma these are tests we need to serve & run
+      'test-app/util.ts', // used by 'www' output target tests to load components
+      'test-app/assets/angular.min.js', // used by a 'www' output target test
+      {
+        pattern: path.join(WWW_OUT_DIR, '/**/*'),
```
rwaskiewicz, line 105: Only this value changed, but because I pushed it over the line length limit, we see some churn 😢

test/karma/package.json
```diff
@@ -1,26 +1,28 @@
...
+  "types": "test-output/test-dist/types/components.d.ts",
+  "collection": "test-output/test-dist/collection/collection-manifest.json",
   "scripts": {
+    "clean": "rm -rf test-output tmp-compiled-tests",
+    "start": "node ../../bin/stencil build --dev --watch --serve --es5",
     "build.app": "npm run compile.test-app && node ../../bin/stencil build --debug --es5",
     "compile.test-app": "node ../../node_modules/typescript/lib/tsc -p tsconfig.json",
-    "build.prerender": "node ../../bin/stencil build --config test-prerender/stencil.config-prerender.ts --prerender --debug && node test-prerender/no-script-build.js",
-    "build.sibling": "node ../../bin/stencil build --config test-sibling/stencil.config.ts --debug",
+    "build.custom-elements": "webpack-cli --config test-app/custom-elements-output-webpack/webpack.config.js && webpack-cli --config test-app/custom-elements-output-tag-class-different/webpack.config.js && webpack-cli --config test-app/custom-elements-delegates-focus/webpack.config.js",
```
rwaskiewicz, line 12: I still need to come up with a more elegant solution here, but splitting this command out seemed like a good first step

test/karma/package.json
```diff
@@ -1,26 +1,28 @@
...
     "build.invisible-prehydration": "npm run build.invisible-prehydration.default && npm run build.invisible-prehydration.true && npm run build.invisible-prehydration.false",
     "build.invisible-prehydration.default": "node ../../bin/stencil build --config test-invisible-prehydration/stencil.config.ts --debug",
     "build.invisible-prehydration.true": "node ../../bin/stencil build --config test-invisible-prehydration/stencil.invisiblePrehydrationTrue.config.ts --debug",
     "build.invisible-prehydration.false": "node ../../bin/stencil build --config test-invisible-prehydration/stencil.invisiblePrehydrationFalse.config.ts --debug",
+    "build.prerender": "node ../../bin/stencil build --config test-prerender/stencil.config-prerender.ts --prerender --debug && node test-prerender/no-script-build.js",
```
rwaskiewicz, line 17: Just alphabetizing/reordering here

test/karma/test-prerender/stencil.config-prerender.ts
```diff
@@ -1,4 +1,5 @@
 import { Config } from '../../../internal';
+const { WWW_OUT_DIR } = require('../constants');
```
splitinfinities, line 2: Is this supposed to use the node require function vs the es6 import syntax? Not blocking but dunno how portable this may be.

splitinfinities (approved):
This one looks great, curious how the constants are consumed with the method of node vs es6 module loading.

test/karma/test-prerender/stencil.config-prerender.ts
rwaskiewicz: ATM our Karma infrastructure doesn't support proper ESModules/syntax across the board.  We have a few different avenues to build/run/view tests, so it felt best to keep it simple in this PR and not open that can of worms.

I'm not too worried about portability here, as these tests are meant to be run on a machine where a dev has the codebase pulled down (or CI).

PR #3140, opened 2021-11-11
Extend the System abstraction to support framework wrappers

splitinfinities:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

While working on the Angular bindings and applying the binding to a couple starter projects, as well as long-lived projects, I discovered that the bindings weren't relying on the abstracted system layer. This would cause me to need to install nodePolyfills/rollup-plugin-node-polyfills during compilation. This causes some pretty huge confusion points as the onboarding of a binding then causes folks to poke around at docs in order to understand the issue. 

GitHub Issue Number: N/A

## What is the new behavior?

This PR adds 6 new items to the Compiler system. 

- `EOL` - the end of line character for the platform the system is running on.
- `joinPaths` - the path.join option 
- `isAbsolutePath` - a function to determine if a string represents an absolute path
- `basename` - returns the last part of a path ([link](https://nodejs.org/api/path.html#pathbasenamepath-ext))
- `dirname` - returns the directory name of a path ([link](https://nodejs.org/api/path.html#pathdirnamepath)) 
- `relative` - returns the relative path from one directory to another ([link](https://nodejs.org/api/path.html#pathrelativefrom-to))

- [ ] Yes
- [x] No

## Testing

I NPM linked this in the output target's repo, and in the starter libraries & my long-lived repo. I then rewrote parts of the Angular output target code to leverage to updated system, instead of importing from `path`, `os`, and more. I will have a corresponding PR for the output targets repo shortly.

### DIFF
--- a/src/compiler/sys/stencil-sys.ts
+++ b/src/compiler/sys/stencil-sys.ts
@@ -15,7 +15,7 @@ import type {
   Logger,
 } from '../../declarations';
 import platformPath from 'path-browserify';
-import { basename, dirname, join } from 'path';
+import path, { basename, dirname, join, relative } from 'path';
 import * as process from 'process';
 import * as os from 'os';
 import { buildEvents } from '../events';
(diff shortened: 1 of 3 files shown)


willmartian:
Hi @splitinfinities, since we are just importing and exporting again, what is the benefit of an end user using the `path` methods exported here vs importing `path` in their own project?

Issue #3147, opened 2021-11-19
bug: invalid error callstacks and coverage reports in tests for components.


paladin80:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/.github/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

2.10.0 - 2.11.0

### Current Behavior

When running tests in stencil, any error callstacks and coverage report don't match the source code.

### Expected Behavior

Error callstacks should point exactly to the line of TS code, where the error happened. Jest coverage report files should properly highlight source files.

Expected behavior was fullfilled in older stencil versions.

### Steps to Reproduce

For call stacks:
1. npm init stencil, choose component. Or just download the stencil-component-starter.
2. npm install
3. Open utils.ts and add a throwing code to line 2. Example: '1'.d();
4. npm test
5. Note the error output of tests: TypeError: "1".d is not a function 
   at Object.format (src/utils/utils.ts:308:7)
   at Object.<anonymous> (src/utils/utils.spec.ts:15:20)
6. The callstack points to line 308 of the utils.ts, but this file has less than 10 lines of code.
7. The callstack points to line 15 of utils.spec.ts, which is an empty line.

For coverage:
1. As in previous steps, install stencil-component-starter.
2. Open utils.ts and add a false if statement with some code:
  if (1 == 2) {
    first = middle;
  }
3. Run stencil test --spec --coverage
4. Note that created coverage report has invalid highlighting for the utils.ts
![tests](https://user-images.githubusercontent.com/1074406/142637662-fd5c4aff-fe67-4303-89b8-85722aa805d9.png)

### Code Reproduction URL

https://github.com/paladin80/stencil-component-starter

### Additional Information

I think that generated source maps don't work in tests.

smndhm:
I'm having the same issue.

paladin80:
The issue appeared in the version 2.10.0. And is still present in 2.11.0

Jest cache is preventing to clearly discover the problem. So please remove the "jest" folder from your %tmp% directory when testing the versions.

paladin80:
OS used: Windows 10 x64
Node used: 14.18.1 and 17.1.0 (same result)

[...]

PR #3163, opened 2021-12-01
fix(sourcemaps): do not encode inline sourcemaps

rwaskiewicz:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

this commit fixes a regression originally introduced in #3100, where
inline sourcemaps that are data URI's are erroneously encoded using
RFC-3986. this causes the sourcemaps for tests to not correctly map the
transpiled output back to the original source code.

GitHub Issue Number: https://github.com/ionic-team/stencil/issues/3147

## What is the new behavior?

do not URI encode the data prefix+mime type

- [ ] Yes
- [X] No

## Testing

To test, I spun up 2 component libraries using `npm init stencil`:
1. `with-sourcemap-issue`, which is the standard output one would see from the Stencil CLI (with v2.11.0 installed)
2. `with-local-stencil-build`, which differs from the above only in name and has a Stencil build with this branch installed on it (`npm ci && npm run build && npm pack` to generated the tarball, then installed with `npm i <PATH_TO_TAR>`)

From there, I ran through the two test cases described in the issue summary of #3147, described below

### 'For Call Stacks'

Reproduction setup (in both projects):
1. Open `utils.ts` and add a throwing code to line 2. Example: '1'.d();
```diff
export function format(first: string, middle: string, last: string): string {
+ '1'.d();
  return (first || '') + (middle ? ` ${middle}` : '') + (last ? ` ${last}` : '');
}

```
2. `npm test -- --coverage --verbose --no-cache`

Outcomes:
In `with-sourcemap-issue`, the sourcemap does not map back to the correct line, which should be "(src/utils/utils.ts:2:7)" (where I added the line). In `with-local-stencil-build`, the sourcemap maps back to the correct line. The results are shown below, where `with-sourcemap-issue` is on the left, and `with-local-stencil-build` is on the right

Screen Shot 2021-12-01 at 8 57 37 AM

### 'For Coverage'

Reproduction setup (in both projects):
1. Open `utils.ts` and add a branching expression that will never be hit
```diff
export function format(first: string, middle: string, last: string): string {
+  if (1==2) {
+    first = middle;
+  }
  return (first || '') + (middle ? ` ${middle}` : '') + (last ? ` ${last}` : '');
}

```
2. `npm test -- --coverage --verbose --no-cache`

Outcomes:
In `with-sourcemap-issue`, the sourcemap does not map back to the correct line when opening the coverage report. (the coverage reports can be found in the `coverage/` dir in the root of each project). In `with-local-stencil-build`, the sourcemap maps back to the correct line in the coverage report. The results are shown below, where `with-sourcemap-issue` is on the left, and `with-local-stencil-build` is on the right.

Screen Shot 2021-12-01 at 9 02 54 AM

Note how the uncovered line number column is incorrect for the following code (which is the same piece of code in each project) on the left, and correct on the right
Screen Shot 2021-12-01 at 9 01 04 AM
Screen Shot 2021-12-01 at 9 01 24 AM

## Other information

Shoutout to @johnjenkins for their work in debugging this! https://github.com/ionic-team/stencil/issues/3147#issuecomment-983147710

### DIFF
--- a/src/compiler/transpile.ts
+++ b/src/compiler/transpile.ts
@@ -1,5 +1,5 @@
 import { TranspileOptions, TranspileResults, Config, TransformOptions, TransformCssToEsmInput } from '../declarations';
-import { catchError, getSourceMappingUrlLinker, isString } from '@utils';
+import { catchError, isString, JS_SOURCE_MAPPING_URL_LINKER } from '@utils';
 import { getPublicCompilerMeta } from './transformers/add-component-meta-static';
 import { getTranspileCssConfig, getTranspileConfig, getTranspileResults } from './config/transpile-options';
 import { patchTypescript } from './sys/typescript/typescript-sys';
@@ -74,10 +74,12 @@ const transpileCode = (
         mapObject.sources = [transpileOpts.file];
         delete mapObject.sourceRoot;
 
+        // generate a data URI for the inlined sourcemap. because we do the base-64 encoding ourselves and we do not
+        // want to encode the prefix & MIME type, build the sourcemap URI by hand
         const mapBase64 = Buffer.from(JSON.stringify(mapObject), 'utf8').toString('base64');
         const sourceMapInlined = `data:application/json;charset=utf-8;base64,` + mapBase64;
         const sourceMapComment = results.code.lastIndexOf('//#');
-        results.code = results.code.slice(0, sourceMapComment) + getSourceMappingUrlLinker(sourceMapInlined);
+        results.code = results.code.slice(0, sourceMapComment) + JS_SOURCE_MAPPING_URL_LINKER + sourceMapInlined;
       } catch (e) {
         console.error(e);
       }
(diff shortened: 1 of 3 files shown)


rwaskiewicz:
Moving this to a draft while I add a few more tests

rwaskiewicz:
Ready to go! Did a little more testing around weird file names like `util+foo.ts` and `utili().ts` to make sure sourcemaps work there as well (since the name of the file is included in the sourcemap, we need to handle that and now do)

ltm (approved):

rwaskiewicz:
This fix has been included in the [v2.12.0 release](https://github.com/ionic-team/stencil/releases/tag/v2.12.0).

PMudra:
Thanks for this fix!

I am not 100% sure about this, but maybe this change also fixed my issues with `toMatchInlineSnapshot` in my tests. I have been using v2.11.0 and I haven't been able to initialize and update inline snapshots. Upgrading to v2.12.1 fixed that issue.

Issue #3136, opened 2021-11-08
RFC: Deprecating dist-custom-elements-bundle


splitinfinities:
# Premise

In helping the Ionic Framework team the launch of v6, we've made updates to the Framework Bindings (React, Angular, and Vue) which have better bundling capabilities by using the `dist-custom-elements` output target. Through this process, we've discovered that we really didn't understand why we were maintaining two output targets - `dist-custom-elements` and `dist-custom-elements-bundle`. 

As of [Stencil v2.12.0](https://github.com/ionic-team/stencil/releases/tag/v2.12.0), if you are using `dist-custom-elements-bundle`, you will begin to see a deprecation message for the `dist-custom-elements-bundle` output target which asks you to use `dist-custom-elements` instead. In an upcoming major version of Stencil, the `dist-custom-elements-bundle` will be removed. The deprecation message will have a link to this issue, so that we can keep feedback organized. 

If you use the `dist-custom-elements-bundle` output target, and have a unique reason that the `dist-custom-elements` output target will not solve, we'd love it if you shared your use case here. 

Thank you so much for using Stencil! 

# 2022.01.06 - Update

Hey folks, @rwaskiewicz here. I'm working on reviewing the needs/concerns y'all have brought up, great feedback thus far! I wanted to post a high level update in the PR summary (although I'm sure I'll have conversations with a few of you in threads).

## Typings

A few of you have mentioned that typings aren't being generated for the `dist-custom-elements` output target. Although it's strange, I can confirm that that is indeed the current behavior of Stencil.  I've got some additional research work scheduled around this topic, but assuming that doesn't throw up any red flags, we'll be adding typings to `dist-custom-elements`.

## ESM Support

I'll be reaching out in GH here to get a better idea of what libraries/frameworks are blocking adoption in your projects. Jest is certainly a big one that I'm aware of, but I want to get a better sense of the situation from y'all.

## Externalizing Dependencies

I need to do a little more research/thinking about this before making any kind of statement on where we'll go with it

## Re-exporting components via a common `index.js` file

Similar to externalizing dependencies, I want to do a little testing around this. I don't think it would be hard per se, but I wanted to be cautious about overpromising specific solutions to (very valid) use cases.

MarkChrisLevy:
[...]

PR #3167, opened 2021-12-03
refactor(dist-custom-elements-bundle): add deprecation warning

willmartian:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

`dist-custom-elements-bundle` is being deprecated in an upcoming major release of Stencil. See discussion in the following GitHub issue. Currently, no warning is shown to the user communicating this.

GitHub Issue Number: #3136

## What is the new behavior?

If `dist-custom-elements-bundle` is found in a Stencil config, a warning is fired when the build process begins that states: `dist-custom-elements-bundle is deprecated and will be removed in a future version. Use "dist-custom-elements" instead. If "dist-custom-elements" does not meet your needs, please add a comment to https://github.com/ionic-team/stencil/issues/3136.`

- [ ] Yes
- [x] No

## Testing

A npm package tarball was generated from this branch (npm ci && npm run build && npm pack). It was then installed in a fresh Stencil project (`npx init stencil`). Said project was built both with and without `dist-custom-elements-bundle` being present in the output targets config to verify that the warning only appeared when the output target is present.

### DIFF
--- a/src/compiler/config/outputs/index.ts
+++ b/src/compiler/config/outputs/index.ts
@@ -1,6 +1,6 @@
 import type * as d from '../../../declarations';
-import { buildError } from '@utils';
-import { VALID_TYPES } from '../../output-targets/output-utils';
+import { buildError, buildWarn } from '@utils';
+import { DIST_CUSTOM_ELEMENTS_BUNDLE, VALID_TYPES } from '../../output-targets/output-utils';
 import { validateCollection } from './validate-collection';
 import { validateCustomElement } from './validate-custom-element';
 import { validateCustomOutput } from './validate-custom-output';
@@ -22,6 +22,9 @@ export const validateOutputTargets = (config: d.Config, diagnostics: d.Diagnosti
       err.messageText = `Invalid outputTarget type "${
         outputTarget.type
       }". Valid outputTarget types include: ${VALID_TYPES.map((t) => `"${t}"`).join(', ')}`;
+    } else if (outputTarget.type === DIST_CUSTOM_ELEMENTS_BUNDLE) {
+      const warning = buildWarn(diagnostics);
+      warning.messageText = `dist-custom-elements-bundle is deprecated and will be removed in a future version. Use "dist-custom-elements" instead. If "dist-custom-elements" does not meet your needs, please add a comment to https://github.com/ionic-team/stencil/issues/3136.`;
     }
   });
 
--- a/src/compiler/config/test/validate-config.spec.ts
+++ b/src/compiler/config/test/validate-config.spec.ts
@@ -312,6 +312,16 @@ describe('validation', () => {
     expect(validated.diagnostics).toHaveLength(1);
   });
 
+  it('should warn when dist-custom-elements-bundle is found', () => {
+    userConfig.outputTargets = [
+      {
+        type: 'dist-custom-elements-bundle',
+      } as any,
+    ];
+    const validated = validateConfig(userConfig);
+    expect(validated.diagnostics).toHaveLength(1);
+  });
+
   it('should default outputTargets with www', () => {
     const { config } = validateConfig(userConfig);
     expect(config.outputTargets.some((o) => o.type === 'www')).toBe(true);


src/compiler/config/test/validate-config.spec.ts
```diff
@@ -312,6 +312,16 @@ describe('validation', () => {
     expect(validated.diagnostics).toHaveLength(1);
   });
 
+  it('should warn when dist-custom-elements-bundle is found', () => {
+    userConfig.outputTargets = [
+      {
+        type: 'dist-custom-elements-bundle',
+      } as any,
```
rwaskiewicz, line 319: Do we need this `as any` type assertion? I think this object literal is a [valid Output Target on its own](https://github.com/CODE-AND-MORE/stencil-core-build/blob/9219856097b4f34bfe1125327c675bbef4b5b343/internal/stencil-public-compiler.d.ts#L1740), since all other properties for the `OutputTargetDistCustomElementsBundle` are optional
```
export interface OutputTargetDistCustomElementsBundle extends OutputTargetBaseNext {
  type: 'dist-custom-elements-bundle';
  empty?: boolean;
  externalRuntime?: boolean;
  copy?: CopyTask[];
  inlineDynamicImports?: boolean;
  includeGlobalScripts?: boolean;
  minify?: boolean;
}
// where 'OutputTargetBaseNext' is:
export interface OutputTargetBaseNext {
  type: string;
  dir?: string;
}
```
I'd suggest removing it if we don't need it:
```suggestion
      },
```

src/compiler/config/test/validate-config.spec.ts
```diff
@@ -312,6 +312,16 @@ describe('validation', () => {
...
+    ];
+    const validated = validateConfig(userConfig);
+    expect(validated.diagnostics).toHaveLength(1);
```
rwaskiewicz, line 322: I like this, and since we verified that `validated.diagnostics` has a value at index 0, what're your thoughts on verifying the message as well?
```suggestion
    expect(validated.diagnostics).toHaveLength(1);
    expect(validated.diagnostics[0].messageText).toBe(
      'dist-custom-elements-bundle is deprecated and will be removed in a future version. Use "dist-custom-elements" instead. If "dist-custom-elements" does not meet your needs, please add a comment to https://github.com/ionic-team/stencil/issues/3136.'
    );
```

src/compiler/config/outputs/index.ts
```diff
@@ -22,6 +22,9 @@ export const validateOutputTargets = (config: d.Config, diagnostics: d.Diagnosti
       err.messageText = `Invalid outputTarget type "${
         outputTarget.type
       }". Valid outputTarget types include: ${VALID_TYPES.map((t) => `"${t}"`).join(', ')}`;
+    } else if (outputTarget.type === DIST_CUSTOM_ELEMENTS_BUNDLE) {
+      const warning = buildWarn(diagnostics);
```
rwaskiewicz, line 26: Can we add a `TODO` for the eventual removal of this code (when we remove the output target)? It looks like we have a ticket already defined for that work, so all I think we need is to add something to the effect of:
```suggestion
      // TODO(STENCIL-260): Remove this check when the 'dist-custom-elements-bundle' is removed
      const warning = buildWarn(diagnostics);
```

rwaskiewicz (changes requested):
Left a few comments/suggestions, LMK what you think.

@splitinfinities PTAL and verify this meets your acceptance criteria surrounding running (namely with our [Compiler/CLI APIs](https://stenciljs.com/docs/compiler-api#transpile))

splitinfinities:
Checked this branch out, linked it to a component library and both within `npm run build` and `npm run build --watch`, the CLI produced a warning prior to the devServer process beginning as expected per AC. Passing the watch flag also only generates this warning once throughout the lifecycle of the `--watch` command, as anticipated. 

Screen Shot 2021-12-03 at 3 49 44 PM

I have no AC for this warning to surface into a Browser's Console during the lifecycle of a watch command. I have no AC for this warning to fire off during an invoked command via [Stencil's Core Compiler API](https://stenciljs.com/docs/compiler-api). 

So yeah, this looks great to me! :shipit:

src/compiler/config/outputs/index.ts
```diff
@@ -22,6 +22,9 @@ export const validateOutputTargets = (config: d.Config, diagnostics: d.Diagnosti
...
+      warning.messageText = `dist-custom-elements-bundle is deprecated and will be removed in a future version. Use "dist-custom-elements" instead. If "dist-custom-elements" does not meet your needs, please add a comment to https://github.com/ionic-team/stencil/issues/3136.`;
```
rwaskiewicz, line 27: Should we specify this will only occur in a _major_ version?

```suggestion
      warning.messageText = `dist-custom-elements-bundle is deprecated and will be removed in a future major version. Use "dist-custom-elements" instead. If "dist-custom-elements" does not meet your needs, please add a comment to https://github.com/ionic-team/stencil/issues/3136.`;
```

CC @splitinfinities

willmartian: Good catch 👍

src/compiler/config/test/validate-config.spec.ts
willmartian: Nope, I copied from the test above it and missed that. Thanks 👍

src/compiler/config/test/validate-config.spec.ts
willmartian: Yes, good point. Updating with the changed warning message from latest commit:

```suggestion
    expect(validated.diagnostics).toHaveLength(1);
    expect(validated.diagnostics[0].messageText).toBe(
      'dist-custom-elements-bundle is deprecated and will be removed in a future major version release. Use "dist-custom-elements" instead. If "dist-custom-elements" does not meet your needs, please add a comment to https://github.com/ionic-team/stencil/issues/3136.'
    );
```

willmartian (commented):
Thanks for review; made appropriate changes.

rwaskiewicz (approved):

rwaskiewicz:
This fix has been included in the [v2.12.0 release](https://github.com/ionic-team/stencil/releases/tag/v2.12.0).

PR #3171, opened 2021-12-06
refactor(jest): add internal support for jest 27

rwaskiewicz:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

GitHub Issue Number: N/A

## What is the new behavior?

this commit adds support for stencil to run jest 27 for its own unit
tests. although it does not enable jest 27 for consumers of stencil, it
is the first step in the process that future work shall build upon.

update `jest`, `jest-cli` and `jest-environment-node` all to v27. jest
v27 switched many defaults that would otherwise break stencil tests. to
support jest v27:
- explicitly use the 'jest-jasmine2' test runner in our own internal
  configuration. moving to 'jest-circus' will occur in STENCIL-307 and
  is not on the critical path for supporting Jest 27
- create a `getVmContext` function in jest environment setup, a new
  field that is required in v27
- remove unneeded `async` keyword from telemetry tests, which were
  causing jest to fail

in addition to the changes above, several changes were required to be
made to our `jest-preprocessor`, which is used both by consumers of
stencil in addition to internal stencil tests:

Between Jest v23 (the latest when this file was originally authored) and
v27, the function signatures of `process` and `getCacheKey` have
changed, most ostensibly between v26 and v27. for the purposes of this
PR, we attempt to maintain backwards compatibility to the best of our
ability, which leads to extra type checking. it is expected that this be
removed in a future major version of Stencil.

The context of `this` has changed between v26 and v27 when code coverage
is generated for files that do not have tests associated with them. in
the original implementation, running jest with the '--coverage' flag
would cause a message to be printed to STDERR for _every file_ that
didn't have test coverage. To circumvent this, cached instances of
`tsconfig.json#options` and its stringified version have been hoisted to
no longer require being referenced through `this`.

An integer used to bust the Jest cache (previously just the number '6')
has been refactored out to a constant, `CACHE_BUSTER` with a JSDoc to
explain its usage and when it should be updated

- [ ] Yes
- [x] No

## Testing

_Almost_ all Stencil tests pass in this branch. One test was explicitly disabled. The reason for which being is that between Jest v26 and v27, CLI args for the library are no longer publicly exposed. This is a known issue with how Jest + Stencil integrate with one another, and the work was previously planned to be done in STENCIL-71. To avoid adding additional work/scope to this PR, the test is temporarily disabled. *

In addition, this library was tested in the Ionic Framework by building this branch (`npm ci && npm run build && npm pack`) and performing the following in order:

### Ionic Core Tests

- the created tarball was installed in `ionic-framework/core`
- `npm run test` was run to invoke spec and e2e tests, all of which passed
- `npm run test.screenshot` was *not* run to invoke screenshot tests as I get failures using an older version of Puppeteer on Framework on an M1 mac
- `npm run test.treeshake` was run


willmartian (approved):
Looks good, nice work. Followed the test directions in the PR description and everything seems to be working as intended.

Additionally, I tested `npm run test.screenshot` in `src/core` locally and received no test runner errors. The screenshot tests themselves did fail, however, but this is consistent with `main` on Ionic Framework and not related to this PR as far as I can tell.

rwaskiewicz:
Thanks @willmartian - part of that on my end may be related to running a newer version of Puppeteer (due to some compatibility issues with M1 machines), much appreciated for double checking 🙏

Issue #3170, opened 2021-12-05
bug: Unknown compiler option 'useUnknownInCatchVariables'.


zardoy:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/.github/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

v2.11.0

### Current Behavior

Fails to start with this option present in tsconfig.

### Expected Behavior

Not fail.

### Steps to Reproduce

1. Grab https://github.com/ionic-team/stencil-component-starter.
2. Add this to compilerOptions of tsconfig:

```ts
"useUnknownInCatchVariables": true,
```

### Code Reproduction URL

https://github.com/ionic-team/stencil-component-starter

### Additional Information

For context: I can't really just remove this flag, since its in my shared tsconfig (I'm `extend`ing from it).
I don't really understand how the compiler works without installing typescript as dependency. I can only see you have locked TS version in `package.json`, but why?

rwaskiewicz:
Hey @zardoy 👋 

The reason this is occurring is that Stencil bundles the TypeScript compiler version for its users.  The reason for this is that we need to make certain guarantees around typings, how TypeScript defines internal data structures when we compile web components and more.  Because we have such strong guarantees that we need to meet, we internally pin the TypeScript version to a specific version. Currently, we pin TypeScript to v4.3

It looks like [`useUnknownInCatchVariables` was introduced in v4.4](https://www.typescriptlang.org/tsconfig#useUnknownInCatchVariables), which explains the issue you've explained here.  At this time, it doesn't look like there's an easy way to suppress that message. We're going to moving Stencil to v4.4 in the next few months. In the mean time, I would suggest not using the shared tsconfig you mentioned if at all possible. I know that can be tough for a few reasons, but unfortunately that's the best workaround I can offer at this time.   Once Stencil has been released using v4.4 of TypeScript, this issue should be resolved

zardoy:
I personally think it is a typescript failure as it doesn't provide a way to reset `compilerOptions` in tsconfig.
However, is there a way to make Stencil print a warning instead of crashing at startup? If there is no easy way of achieving that, I'll just wait for the next release.

[...]

PR #3205, opened 2022-01-13
feat(typescript): typescript 4.5 support

rwaskiewicz:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

We currently support v4.3 of TypeScript

GitHub Issue Numbers:
- Relates to https://github.com/ionic-team/stencil/issues/3170

## What is the new behavior?

This PR adds support for TypeScript 4.5. I originally intended to only bump Stencil to v4.4, but there were some changes to a private TypeScript API Stencil uses (and shouldn't, more on that later) in TS v4.4 that were reverted in v4.5 (again, more on that later).  Below is a summary of each commit by name (should I rebase this and mess up the SHA links)

<details><summary>chore(dependencies): upgrade typescript to v4.4.4</summary>

Version bump us to v4.4.4 - this was before I realized we'd have a problem with 4.4. This commit will later be superseded by the upgrade to 4.5.
</details>
<details><summary>update client browser patching</summary>

update client-patch-browser.ts patching of `performance.mark()` &
`performance.measure()` to be ignored by the type checker. this patch
allows stencil to support Safari 10; since we don't have access to a
non-existent performance buffer and changing the return value for either
of these functions could be considered a breaking change, defer altering
the (non-compliant) functions until stencil drops support for Safari 10

</details>
<details><summary>remove non-compliant window.location.reload() arguments </summary>

`reload` has no parameter, although Firefox supports a non-standard
'forceGet' boolean parameter for bypassing its cache. for all other
browsers, this is otherwise ignored. these arguments are used in the
context of stencil's dev-server (IE are only used at dev time). given
these two statements, the risk of a truly breaking change is deemed
minimal here

</details>
<details><summary>update fn signature of mock-doc's customElementRegistry#whenDefined </summary>

set the return type of `whenDefined` to be
`Promise<CustomElementConstructor>` over `Promise<void>` to meet new
type checking expectations with TS v4.4. this leads to a small
functional change in that the return value for this function when the
internal registry contains an entry for a given `tagName`, the function
now returns the constructor associated with the tag name, rather than
void.

</details>
<details><summary>update mock-doc performance function signatures</summary>

update the function signatures of mock-doc's performance implementation
by declaring the return type on `mark()` and `measure()` to align with
the `Performance` interface, but _continue to return undefined_,
ignorning the type checker errors.

stencil's stubs for these functions assume access to underlying browser
api's (namely a performance buffer). given that older browsers may not
support this functionality and implementating a mock version of these
functions falls well outside the scope of a typescript update, simply
ignore the errors.

we take this approach, rather than:
```typescript
// @ts-ignore
mark(): void { /* no-op */ }
```
as stencil can (and does) instantiate an instance of `MockPerformance`
and attempts to assign it to variable(s) that are of type
`Performance`. this circumvents the type checker (although stencil was
lying to it already) to prevent additional `// @ts-ignore` pragmas

this may result in typing errors for consumers, however the actual
values shall remain the same.

</details>
<details><summary>window - add ts-ignore to handle mock-window & window type diffs</summary>

In various parts of the codebase, Window is used as a typing when the actual concrete type is a MockWindow - this leads to some tricky spots where when the TS compiler sees these as two distinct, unrelated types, calling a MockWindow method on an object typed as Window leads to compiler errors/adding @ts-ignore pragmas to the code.  Created a ticket for us to revisit this

</details>
<details><summary>useUnknownInCatchVariables </summary>

enable `useUnknownInCatchVariables` in all TSConfig files, with the
exception of the main `tsconfig.json`. Although flipping that bit was
originally going to be a part of the TS 4.4 upgrade, our decision to
upgrade from 4.3 -> 4.5 directly increased scope a bit. That and the
fact there are ~90 places where we want to rework the code, I feel more
comfortable splitting that off into a separate effort

</details>
<details><summary>[PAUSE] Stop current line of work, look into TS 4.5 level of effort</summary>

i'm concerned about the `matchFile` call and how the [method changes](https://github.com/microsoft/TypeScript/pull/44710/files#diff-1516c8349f7a625a2e4a2aa60f6bbe84e4b1a499128e8705d3087d893e01d367) for
v4.4, then is [reverted in v4.5](https://github.com/microsoft/TypeScript/pull/46787). based on the conversations/issues in GH
surrounding these changes, I'm strongly considering moving directly to
v4.5. pause here to investigate those changes

**TO BE CLEAR** - Stencil needs to move off this API. I have a story in the backlog for us to do this

</details>
<details><summary>TS 4.5</summary>

update karma-typescript with monounity/karma-typescript@f4cdc43 - this is necessary due to internal changes made in TypeScript 4.5.

update TypeScript to v4.5.4

remove the temporary hack during the TS v4.4 upgrade for `matchFile`
call to succeed.

update single unit test that was failing following the upgrade.

update typescript import statements that were broken as a result of the
upgrade to handle `AssertClause`s. move typescript calls to use factory
methods

</details>

- [ ] Yes
- [x] No - I'm following the same guidance we have for previous upgrades here. We treat TS upgrades as 'minor'/'non-breaking', just as the TS team does.

- [TypeScript 4.4 breaking changes](https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#breaking-changes)
- [TypeScript 4.5 breaking changes](https://devblogs.microsoft.com/typescript/announcing-typescript-4-5/#breaking-changes)

Because there's a little bit of a chance there could be a less than smooth upgrade here, I wanna hold off on adding this to the Stencil v2.13 release, and slate this for v2.14

## Testing

In addition to unit/karma tests helping us out here, I also installed this branch's generated tarball (`npm run build && npm pack`) in Ionic Framework's core module, then built Framework/ran its tests.  For better or worse, the source of most breaking changes would come from the TS upgrade itself (rather than changes we made to accommodate the upgrade, since there are a few `ts-ignore` pragmas in here)

### DIFF
--- a/src/mock-doc/custom-element-registry.ts
+++ b/src/mock-doc/custom-element-registry.ts
@@ -81,14 +81,14 @@ export class MockCustomElementRegistry implements CustomElementRegistry {
     }
   }
 
-  whenDefined(tagName: string) {
+  whenDefined(tagName: string): Promise<CustomElementConstructor> {
     tagName = tagName.toLowerCase();
 
     if (this.__registry != null && this.__registry.has(tagName) === true) {
-      return Promise.resolve();
+      return this.__registry.get(tagName).cstr;
     }
 
-    return new Promise<void>((resolve) => {
+    return new Promise<CustomElementConstructor>((resolve) => {
       if (this.__whenDefined == null) {
         this.__whenDefined = new Map();
       }
--- a/src/mock-doc/custom-element-registry.ts
+++ b/src/mock-doc/custom-element-registry.ts
@@ -85,7 +85,7 @@ export class MockCustomElementRegistry implements CustomElementRegistry {
     tagName = tagName.toLowerCase();
 
     if (this.__registry != null && this.__registry.has(tagName) === true) {
-      return this.__registry.get(tagName).cstr;
+      return Promise.resolve<CustomElementConstructor>(this.__registry.get(tagName).cstr);
     }
 
     return new Promise<CustomElementConstructor>((resolve) => {
--- a/package-lock.json
+++ b/package-lock.json
@@ -80,7 +80,7 @@
         "sizzle": "^2.3.6",
         "terser": "5.6.1",
         "tslib": "^2.1.0",
-        "typescript": "4.3.5",
+        "typescript": "^4.4.4",
         "webpack": "^4.46.0",
         "ws": "7.4.6"
       },
@@ -11239,9 +11239,9 @@
       }
     },
     "node_modules/typescript": {
-      "version": "4.3.5",
-      "resolved": "https://registry.npmjs.org/typescript/-/typescript-4.3.5.tgz",
-      "integrity": "sha512-DqQgihaQ9cUrskJo9kIyW/+g0Vxsk8cDtZ52a3NGh0YNTfpUSArXSohyUGnvbPazEPLu398C0UxmKSOrPumUzA==",
+      "version": "4.4.4",
+      "resolved": "https://registry.npmjs.org/typescript/-/typescript-4.4.4.tgz",
+      "integrity": "sha512-DqGhF5IKoBl8WNf8C1gu8q0xZSInh9j1kJJMqT3a94w1JzVaBU4EXOSMrz9yDqMT0xt3selp83fuFMQ0uzv6qA==",
       "dev": true,
       "bin": {
         "tsc": "bin/tsc",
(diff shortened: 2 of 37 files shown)


src/mock-doc/custom-element-registry.ts
```diff
@@ -81,14 +81,14 @@ export class MockCustomElementRegistry implements CustomElementRegistry {
     }
   }
 
-  whenDefined(tagName: string) {
+  whenDefined(tagName: string): Promise<CustomElementConstructor> {
     tagName = tagName.toLowerCase();
 
     if (this.__registry != null && this.__registry.has(tagName) === true) {
-      return Promise.resolve();
+      return this.__registry.get(tagName).cstr;
```
ltm, line 88: I'm not completely sure what the type of `cstr` is, but assuming it's `CustomElementConstructor` I would expect this to still return a promise:

```suggestion
      return Promise.resolve<CustomElementConstructor>(this.__registry.get(tagName).cstr);
```

rwaskiewicz: Good call - technically it may not be a promise. Updated in https://github.com/ionic-team/stencil/pull/3205/commits/b365de1984f4e948f5acee6f5da0ba1e61508fd2

ltm (approved):

PR #3211, opened 2022-01-20
chore(typescript): enable useUnknownInCatchVariables in core

rwaskiewicz:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Other (please describe): TypeScript compiler changes

Errors declared in `catch` blocks are implicitly typed as `any`

GitHub Issue Number: N/A

## What is the new behavior?

this commit enables the 'useUnknownInCatchVariables' flag in the core
compiler's tsconfig.json file. the remainder of the changes related
directly to errors emitted by the TypeScript compiler as a result of
this change.

the most common change in this commit is updating a `catch` block that
consists of a single call to stencil's `catchError` function. because
`catchError` may theoretically accept a value or object that does not
conform to the `Error` interface, we annotate many of the error
variables declared in the `catch` block as type `any`, in a good-faith
effort not to lose any useful information that we would otherwise turn
into a `Diagnostic`. Using a type guard or other type narrowing
mechanism would be feasible if any value that is not of type `Error`
were logged in some other way. However in an attempt to not pollute the
codebase with two different logging paths, we accept the ambiguity of
`catchError` and catching errors of type `any`  for the time being.

- [ ] Yes
- [x] No

## Testing

The TS Compiler was run over the codebase, to ensure no type-safety issues.  Admittedly it's not the most comprehensive, but covers _most_ of the cases throughout this patch.

## Other information

**Do not merge please. Relies on** #3205

### DIFF
--- a/src/compiler/sys/resolve/resolve-module-sync.ts
+++ b/src/compiler/sys/resolve/resolve-module-sync.ts
@@ -139,9 +139,11 @@ export const createCustomResolverSync = (
       const fsFilePath = normalizeFsPath(p);
       try {
         return sys.realpathSync(fsFilePath);
-      } catch (realpathErr) {
-        if (realpathErr.code !== 'ENOENT') {
-          throw realpathErr;
+      } catch (realpathErr: unknown) {
+        if (isErrnoException(realpathErr)) {
+          if (realpathErr.code !== 'ENOENT') {
+            throw realpathErr;
+          }
         }
       }
       return fsFilePath;
@@ -150,3 +152,15 @@ export const createCustomResolverSync = (
     extensions: exts,
   } as any;
 };
+
+/**
+ * Type guard to determine if an Error is an instance of `ErrnoException`. For the purposes of this type guard, we
+ * must ensure that the `code` field is present. This type guard was written with the `ErrnoException` definition from
+ * https://github.com/DefinitelyTyped/DefinitelyTyped/blob/d121716ed123957f6a86f8985eb013fcaddab345/types/node/globals.d.ts#L183-L188
+ * in mind.
+ * @param err the entity to check the type of
+ * @return true if the provided value is an instance of `ErrnoException`, `false` otherwise
+ */
+function isErrnoException(err: unknown): err is NodeJS.ErrnoException {
+  return err instanceof Error && err.hasOwnProperty('code');
+}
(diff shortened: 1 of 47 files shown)


src/compiler/sys/resolve/resolve-module-sync.ts
```diff
@@ -139,9 +139,12 @@ export const createCustomResolverSync = (
       const fsFilePath = normalizeFsPath(p);
       try {
         return sys.realpathSync(fsFilePath);
-      } catch (realpathErr) {
-        if (realpathErr.code !== 'ENOENT') {
-          throw realpathErr;
+      } catch (realpathErr: unknown) {
+        if (realpathErr instanceof Object && 'code' in realpathErr) {
+          // @ts-ignore: we've determined 'code' is in the prototype chain, but TS isn't happy about the access
```
ltm, line 144: FYI: This is because `in` isn't a type guard: microsoft/TypeScript#21732. One option would be to implement a type guard function for [`NodeJS.ErrnoException`]():

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/globals.d.ts#L183-L188

src/compiler/sys/resolve/resolve-module-sync.ts
```diff
@@ -139,9 +139,12 @@ export const createCustomResolverSync = (
       const fsFilePath = normalizeFsPath(p);
       try {
         return sys.realpathSync(fsFilePath);
```
ltm, line 141: Unrelated to your changes... does this even throw an exception? According to the interface of `sys.realpathSync()` it returns a `CompilerSystemRealpathResults` and doesn't throw an exception:
https://github.com/ionic-team/stencil/blob/b00585fb643332eb9b1c188e8e23a8e8a4ec742d/src/declarations/stencil-public-compiler.ts#L1032-L1035

rwaskiewicz: It shouldn't, although I think the problem here is that we can't guarantee implementations on that interface _won't_ throw. Perhaps this is just a bit of defensive programming?

src/compiler/sys/resolve/resolve-module-sync.ts
rwaskiewicz: Good call, impl in f0fb959c7..b88604bad

Issue #3251, opened 2022-02-23
bug: upgrade to v2.14.0 breaks jest tests using esm


johncrim:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/.github/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

2.14.0

### Current Behavior

Tests using external `jest.config.js` file (necessary to transform .mjs and .js files using ES modules) work fine using @stencil/core 2.13.0 . Upgrading to 2.14.0 results in errors like:

```
    C:\src\github\johncrim\repro-stencil-jest\libs\util\util.js:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,jest){import { Observable, NEVER } from 'rxjs';
                                                                                      ^^^^^^

    SyntaxError: Cannot use import statement outside a module

      1 | import { Component, Prop, h, Element } from '@stencil/core';
      2 | import { format } from '../../utils/utils';
    > 3 | import { resizeObservable } from '../../../libs/util/util';
        | ^
      4 | import { Subscription } from 'rxjs';
      5 |
      6 | @Component({

      at Runtime.createScriptFromCode (node_modules/jest-runtime/build/index.js:1728:14)
      at Object.<anonymous> (src/components/my-component/my-component.tsx:3:1)
```

This occurs in tests with dependencies on external ESM files, which have dependencies on other external ESM files.

The external jest config adds support for processing .mjs files, and that's about it:

```js
# jest.config.js
module.exports = {
  preset: '@stencil/core/testing',
  testRegex: '/src/.*\\.spec\\.(ts|tsx|js)$',
  testRunner: 'jest-jasmine2',
  transform: {
    '^.+\\.(ts|tsx|jsx|css|mjs)$': '@stencil/core/testing/jest-preprocessor.js',
  }
};
```

### Expected Behavior

No breaking change in test running when upgrading to @stencil.core 2.14.0

### Steps to Reproduce

```bash
git clone git@github.com:johncrim/repro-stencil-jest.git
npm install
npm run jest
```

Yields this error:

```
    C:\src\github\johncrim\repro-stencil-jest\libs\util\util.mjs:1
[...]

Issue #3230, opened 2022-02-02
feat: Support ES modules in jest tests


johncrim:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/.github/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already include this feature request, without success.

### Describe the Feature Request

Stencil tests should be able to run using Jest ES Module support, instead of transpiling all ESM modules to CJS. This needs to be done to improve test perf, and because ESM is preferred in all modern projects that are used in browsers.

Right now, stencil jest support hard-codes CJS output: https://github.com/ionic-team/stencil/blob/main/src/testing/test-transpile.ts

### Describe the Use Case

When I run stencil tests, I should have the option to run them in Jest ESM mode.

The lack of this feature causes bugs and wasted dev effort, eg: https://github.com/ionic-team/stencil/issues/2178

Performance should be improved b/c .mjs files (most dependencies) will no longer need to be transpiled. Dev cycle speed is important!

In addition, there are behavioral differences between CJS and ESM, and it's valuable for the test environment to be as similar as possible to the runtime environment.

### Describe Preferred Solution

A config option exists for running tests in Jest ESM mode.

### Additional Information

https://kulshekhar.github.io/ts-jest/docs/next/guides/esm-support/

Related issues:
* https://github.com/ionic-team/stencil/issues/2608
* https://github.com/ionic-team/stencil/issues/2178

I'm willing to work on a PR for this - looks like it shouldn't be too hard.

rwaskiewicz:
Hey @johncrim 👋 

I agree that Stencil should support ES Modules when using Jest.  To be honest, I had not been considering this support until Jest's ESM support came out of it's 'experimental' phase.  But if you're interested, I say go for it!

johncrim:
Thanks @rwaskiewicz . I certainly understand waiting until the experimental phase ends - I can't fault you for that, though I don't know how long it will remain in experimental. 

It's working acceptably with ts-jest and jest + angular (though still a little bumpy). I just migrated our angular jest tests to Jest ESM and would like to tackle this as long as it's not too involved (from a code review it looks pretty straightforward, but I could certainly be wrong about that).

johncrim:
[...]

PR #3256, opened 2022-02-24
feat: Jest ESM support

johncrim:
Note up front: I don't believe this is quite ready to merge, open issues/questions noted below. I'm posting this PR for initial review/evaluation.

## Pull request checklist

## Pull request type

- [x] Bugfix
- [x] Feature
- [ ] Build related changes

GitHub Issue Number: #3230 

## What is the new behavior?

Jest tests are run using ES modules if the following changes are made:
1. Execute node with `--experimental-vm-modules`, e.g. `node --experimental-vm-modules node_modules/jest/bin/jest.js` or `NODE_OPTIONS=--experimental-vm-modules npx jest` etc.
    I also like to include `--no-warnings` to eliminate the experimental feature warning on startup.
2. Add to Stencil config: `testing.useESModules: true`

Alternatively, if jest is run directly (not through the stencil CLI), the following `jest.config.cjs` can be used to run stencil tests using ES modules:
```js
module.exports = {
  preset: '@stencil/core/testing',

  globals: {
    stencil: {
      testing: {
        useESModules: true
      }
    }
  },
  extensionsToTreatAsEsm: ['.ts', '.tsx', '.jsx'],
};
```

In addition, a bug is fixed: If `--runInBand` is passed in the CLI, `--max-workers=${config.maxConcurrentWorkers}` is no longer set. I had to fix this in order to debug my work on this feature. The behavior was incorrect before (one of `--runInBand` or `--max-workers` wins, b/c `--runInBand` means no external worker processes), and I added tests to validate it.

I can break that out if you really want me to...

- [ ] Yes
- [x] No

Changes should not be breaking - the previous code paths are effectively unchanged, so Jest tests using CJS continue to run as before.

## Testing

I added some new unit tests, which only validate the config generation. I also added a new test sub-project `test/jest-esm`. To run it:

```
npm run build
cd test/jest-esm
npm run jest ##Runs jest via the jest.config.cjs
npm run jest -- --clear-cache
npm run test ## Runs tests via stencil CLI
```

I didn't see a way to loop this project into the CI build; eg the `test/jest-spec-runner` isn't called in the github workflow.

## Open issues

1. Documentation. I added js docs to the new config values, but that's probably not sufficient. I didn't see an obvious place to add docs.

2. More tests would be nice - eg running some jest + jest ESM tests through the whole stack, but I didn't see a pattern to follow.

4. [Jest expect extensions are currently borken when using ESM and `@jest/globals`](https://github.com/facebook/jest/issues/12267). That's why there's this ugly expect:

```ts
(expect(page.root) as unknown as jest.JestMatchers<HTMLElement>).toEqualHtml
```

TLDR: The return type of `@jest/globals expect()` is not the same as `jest.JestMatchers<T>`. So anyone using Jest ESM with expect extensions will have to work around the typing issue until this issue is fixed. Note that this applies to anyone using Jest ESM, it is not stencil specific, but stencil developers likely use expect extensions.

I think the last issue is tolerable (and looks like it should be fixed soon) - since this is opt-in, I think it's reasonable to provide this as a preview feature while Jest improves their ESM support.

## Other

I validated that these changes provide a fix for #3251, if projects affected by #3251 are willing to switch to using ES modules while testing. Since #3251 only appears due to ESM dependencies, I would think this would be attractive.


johncrim:
I added logging to help if `useESModules` is true, but `node --experimental-vm-modules` was not set:

```
❯ stencil test --spec
[31:01.5]  @stencil/core
[31:01.8]  [LOCAL DEV]
[31:01.9]  testing spec files
[31:02.0]  jest args: --spec --max-workers=8

[ ERROR ]  node must be run with --experimental-vm-modules for stencil.config testing.useESModules: true

... other errors
```

christian-bromann:
@johncrim thanks for raising the PR and supporting Stencil.

The team has decided not to move forward with this. We are planning to build a Jest preset for Jest v30 and up that will be published as a standalone module e.g. `@stencil/jest`. We will also likely not going to support multiple Jest versions anymore moving on. A lot of changes made here will be helpful in this effort and we will make sure to include you in the development process moving forward.

Again, thanks a lot for everything!

johncrim:
Thanks for the update @christian-bromann . As long as ESM is handled well, I'm happy.

PR #3286, opened 2022-03-17
feat(runtime): event emitter accepts event specific custom event

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

The interface for event emitter that all Stencil events are typed to does not allow overriding the signature of the `CustomEvent` type.

Considering this pattern:
```ts
@Event() myEvent: EventEmitter<void>;

doThing() {
  this.myEvent.emit();
}

@Method()
async complete() {}
```

When implementing this component, the user needs to get access to the element reference, to execute the `complete()` method after the event is fired.

```ts
<my-component onMyEvent={ev => ev.target.complete()} />
```

This code results in a typing error, since the type signature of `ev` is `CustomEvent<void>` and does not include the element target. 

GitHub Issue Number: N/A

## What is the new behavior?

Add a secondary generic to the `EventEmitter` interface, that allows Stencil custom events to specify a different type signature for the `CustomEvent` (must extend `CustomEvent`).

Taking the same example from above, it becomes:
```ts
interface MyEventCustomEvent extends CustomEvent {
  target: HTMLMyComponentElement;
}

@Event() myEvent: EventEmitter<void, MyEventCustomEvent>;
```

This will make the generated types include our HTML element, making this valid syntax/type safe:

```ts
<my-component onMyEvent={ev => ev.target.complete()} />
```

`ev.target` = `HTMLMyComponentElement`

- [ ] Yes
- [x] No

### DIFF
--- a/src/declarations/stencil-public-runtime.ts
+++ b/src/declarations/stencil-public-runtime.ts
@@ -454,8 +454,8 @@ export interface ComponentInterface {
 }
 
 // General types important to applications using stencil built components
-export interface EventEmitter<T = any> {
-  emit: (data?: T) => CustomEvent<T>;
+export interface EventEmitter<T = any, EventCustomEvent extends CustomEvent<T> = CustomEvent<T>> {
+  emit: (data?: T) => EventCustomEvent;
 }
 
 export interface RafCallback {


rwaskiewicz:
Thanks @sean-perkins! Can you do me a favor and add a few tests to the `test/karma/test-app` directory for me so that if we merge this, we don't regress on it in the future? Our instructions for doing so can be found [here](https://github.com/ionic-team/stencil/blob/main/test/karma/README.md#creating-a-new-test).  I think we'd want three tests (one directory for each) where we exercise:
- the example that you have in the PR summary there as a base case of sorts
- a case where the first generic of `EventEmitter` is a concrete type (something other than `any`/`void`/`unknown`) to make sure the `on*` typings are generated correctly
  - it's OK if this test is basically the same as the first case, only with that first generic being non-void being the only change. The compiler should just be able to yell at us if there's any typing errors
- We should probably add a test where we use `@Listen` for good measure

Once that's all fine and good, the last thing we'll want to do is update the documentation for Events. I haven't gotten around to revamping those just yet, but would like to get something in there about this second generic argument. The page we'll be updating can be found [here](https://github.com/ionic-team/stencil-site/blob/master/src/docs/components/events.md), and the instructions for testing out your changes can be found in the [Stencil site's contributing guidelines](https://github.com/ionic-team/stencil-site/blob/master/CONTRIBUTING.md#setup). 

LMK if you have any questions!

sean-perkins:
Closing in favor of: https://github.com/ionic-team/stencil/pull/3296

3296 resolves the true underlying issue.

PR #3302, opened 2022-03-25
feat(node): drop node 12 support 

rwaskiewicz:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Node 12 will hit End of Life on 2022.04.30. In preparation for that date,
we are adding a PR to the **v3 branch**. This is a **breaking change**
that we will not be propagating back to Stencil V2.X. 

This commit doesn't intend to set any expectations around v3's release
date. I have some bandwidth on hack days, and wanted to get a jump on
this work. The intent is to merge it into the v3 branch and evaluate other
changes to Stencil (either from `main` or directly onto the v3 branch)
against this series of changes.

GitHub Issue Number: N/A

## What is the new behavior?

This PR is split into several commits, each of which updates a specific part of the project to support Node 14+. For the most part, this has more emphasis on disallowing Node 12 than it does for updating code to run Node 14+. Node's list of deprecated APIs was consulted for version 14+, and any code changes/future TODOs have been added as a part of the PR.

It's probably best that this PR be reviewed commit by commit, so that way the intent of each change is clear (when compared with the git commit msg).

- [X] Yes
- [ ] No

## Testing

CI has been run. We've consulted with the Framework team on this in the past. When we start offering alpha/beta releases, we'll be sure to run them against the Framework team's repo in some way/shape/form


src/sys/node/node-sys.ts
```diff
@@ -344,6 +344,9 @@ export function createNodeSys(c: { process?: any } = {}) {
       return new Promise((resolve) => {
         const recursive = !!(opts && opts.recursive);
         if (recursive) {
+          // TODO(STENCIL-410): In a future version of Node, `force: true` will be required in the options argument. At
+          // the time of this writing, Stencil's Node typings do not support this option.
+          // https://nodejs.org/docs/latest-v16.x/api/deprecations.html#dep0147-fsrmdirpath--recursive-true-
```
alicewriteswrongs, line 349: does this mean that if running on node >16 a call to `sys.removeDir` will throw an error relating to lacking the `force: true` param? If so, I think the impact may be minimal / nonexistent — a quick grep didn't turn up any places where we actually call the function, although I easily could have missed one

package.json
```diff
@@ -130,7 +130,7 @@
     "ws": "7.4.6"
   },
   "engines": {
-    "node": ">=12.10.0",
+    "node": ">=14.10.0",
     "npm": ">=6.0.0"
```
alicewriteswrongs, line 134: kinda funny to have both the `volta` config and `engines` - out of curiosity, do you know why we're specifying `engines`? the only context where I've had to set it before was to let Heroku know what node version a project should use

alicewriteswrongs (approved):
this all looks good to me! just had a few secondary questions

alicewriteswrongs:
also I tested out installing this commit into a test component repo (using `npm pack`) and, as expected, I got an error message w/ node 12 and a looming deprecation warning for node 14.5

package.json
rwaskiewicz: `engines` is for the users of Stencil, where the volta config is for folks developing/hacking on the Stencil compiler itself.

`engines` [doc](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#engines) handles everything but the version of npm/node we should be using for running `npm install`, running npm scripts, etc in the repo: 

> You can specify the version of node that your stuff works on (example omitted)
> And, like with dependencies, if you don't specify the version (or if you specify "*" as the version), then any version of node will do.
> You can also use the "engines" field to specify which versions of npm are capable of properly installing your program. For example: [omitted]

It also isn't strictly enforced:

> Unless the user has set the engine-strict config flag, this field is advisory only and will only produce warnings when your package is installed as a dependency.

Volta on the other hand, is strictly enforced when working on Stencil, although it tries to be as transparent/out of the way as possible.

src/sys/node/node-sys.ts
rwaskiewicz: I'm not sure, as the documentation for Node isn't abundantly clear:

> In future versions of Node.js, recursive option will be ignored for fs.rmdir, fs.rmdirSync, and fs.promises.rmdir.

> Use fs.rm(path, { recursive: true, force: true }), fs.rmSync(path, { recursive: true, force: true }) or fs.promises.rm(path, { recursive: true, force: true }) instead.

It only states that the runtime deprecation began to be fully enforced (whatever _that_ means) in Node v16.  I think we'll just need to keep an eye on this as newer versions of Node are release (and we deprecate older ones)

rwaskiewicz:
I'm going to merge this - the checks for Node 12 aren't going to pass, because they aren't gonna run (that's the point of this whole PR 🙃). Everything else is passing though 🤞

PR #3314, opened 2022-04-04
Ap/remove some any

alicewriteswrongs:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

GitHub Issue Number: N/A

We have a few `any` annotations in `src/cli/parse-flags.ts` which would be nice to get rid of.

## What is the new behavior?

- this refactors an object into 3 different string enums, which makes it possible to strongly type the keys for an object (we're using the enum members as keys). This lets us remove an `as any` from a few places. It shouldn't result in any runtime changes.

- [ ] Yes
- [x] No

## Testing

Just made sure the tests are passing locally.


src/cli/parse-flags.ts
```diff
@@ -160,48 +160,61 @@ const configCase = (prop: string) => {
...
+  'docs-json',
+  'emulate',
+  'log-level',
+  'root',
+  'screenshot-connector',
+] as const;
+
+type AliasMap = Partial<
+  Record<typeof BOOLEAN_ARG_OPTS[number] | typeof NUMBER_ARG_OPTS[number] | typeof STRING_ARG_OPTS[number], string>
+>;
```
rwaskiewicz, line 211: Minor, we might be able to make this a little easier to understand for folks not used to how indexed types work for arrays here at the cost of a little more verbosity:

```suggestion
type ArrayValuesAsKeys<T extends ReadonlyArray<string>> = T[number];
type OptionKeys =
  | ArrayValuesAsKeys<typeof BOOLEAN_ARG_OPTS>
  | ArrayValuesAsKeys<typeof NUMBER_ARG_OPTS>
  | ArrayValuesAsKeys<typeof STRING_ARG_OPTS>;
type AliasMap = Partial<Record<OptionKeys, string>>;
```

src/cli/parse-flags.ts
```diff
@@ -42,8 +42,8 @@ export const parseFlags = (args: string[], sys?: CompilerSystem): ConfigFlags =>
 };
 
 const parseArgs = (flags: any, args: string[], knownArgs: string[]) => {
```
rwaskiewicz, line 44: Although this function is relatively self-explanatory based on it's name, can we add a JSDoc to it while we're in here?
```suggestion
/**
 * A short description of the function, ideally 1-2 sentences
 * @param flags a short description of this argument
 * @param args  a short description of this argument 
 * @param knownArgs a short description of this argument
 */
const parseArgs = (flags: any, args: string[], knownArgs: string[]) => {
```

rwaskiewicz (changes requested):
All in all looks good to me! Two small asks - LMK what you think!

src/cli/parse-flags.ts
alicewriteswrongs: yes, that's definitely a little easier to read and figure out!

src/cli/parse-flags.ts
rwaskiewicz: Agreed, that caught my eye too - probably better for another day

rwaskiewicz (approved):

PR #3315, opened 2022-04-05
test(prop): add unit tests to prop type generation 

rwaskiewicz:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

The functionality in [`generate-prop-types`](https://github.com/ionic-team/stencil/blob/8feaf0816731d8dc1d880cecf6a432526295d65a/src/compiler/types/generate-prop-types.ts#L1), which is used in the generation of type declaration files for class members decorated with `@Prop()`, is currently not unit tested. There are however, [end to end tests](https://github.com/ionic-team/stencil/blob/8feaf0816731d8dc1d880cecf6a432526295d65a/src/compiler/transformers/test/parse-props.spec.ts#L3) for this functionality.  We add unit tests here to minimize an upcoming bug fix PR down in size and under proper unit testing

GitHub Issue Number: N/A

## What is the new behavior?

This PR adds unit tests for `generate-prop-types`

Secondarily, it adds a new stubbing mechanism for Stencil tests.  I intend to write up an Architectural Design Record (ADR) if we choose to adopt this long term, but in short it allows us to create stubs for Stencil types. It focuses on:
1. Being type safe - stubs provide reasonable defaults for all keys in an object literal-like type. Stencil objects can be complex/confusing to test authors (especially if they've never seen a type before) - by providing reasonable defaults, we can guide folks to using properly filled out stubs and away from using type assertions
2. Overrideable - stubs should allow test authors to provide an override key/value for any field in a type (while maintaining type safety)
3. Composable - Stubs can be combined to create more complex types
4. Reusable - Rather than copy-pasting a code snippet to generate an entity, we have one central source of truth per entity.

Please take a look at how these stubs are used in the context of this PR and let me know what you think.

Note: the stubs are intentionally be committed in the same directory as the files that use them currently. if this pattern is more widely adopted by the codebase/team, we will move these stubs out to a more common location

- [ ] Yes
- [x] No

## Testing

- Unit tests pass (no new functionality there)
- Verified `stub` files do not appear in Jest output (we do not attempt to run them as tests, nor do they appear in the code coverage output)
- Verified `stub` files do not appear in the tarball generated using `npm pack`

### DIFF
--- a/src/compiler/types/tests/ComponentCompilerMeta.stub.ts
+++ b/src/compiler/types/tests/ComponentCompilerMeta.stub.ts
@@ -1,0 +1,23 @@
+import * as d from '@stencil/core/declarations';
+
+/**
+ * Generates a stub {@link ComponentCompilerMeta}. This function uses sensible defaults for the initial stub. However,
+ * any field in the object may be overridden via the `overrides` argument.
+ * @param overrides a partial implementation of `ComponentCompilerMeta`. Any provided fields will override the
+ * defaults provided by this function.
+ * @returns the stubbed `ComponentCompilerMeta`
+ */
+export const stubComponentCompilerMeta = (
+  overrides: Partial<d.ComponentCompilerMeta> = {}
+): d.ComponentCompilerMeta => {
+  // TODO(STENCIL-378): Continue to build out default stub, remove the type assertion on `default`
+  const defaults: d.ComponentCompilerMeta = {
+    events: [],
+    methods: [],
+    properties: [],
+    sourceFilePath: '/some/stubbed/path/my-component.tsx',
+    virtualProperties: [],
+  } as d.ComponentCompilerMeta;
+
+  return { ...defaults, ...overrides };
+};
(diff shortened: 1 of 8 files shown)


src/compiler/types/tests/ComponentCompilerMeta.stub.ts
```diff
@@ -0,0 +1,23 @@
...
+ * Generates a stub {@link ComponentCompilerMeta}. This function uses sensible defaults for the initial stub. However,
+ * any field in the object may be overridden via the `overrides` argument.
+ * @param overrides a partial implementation of `ComponentCompilerMeta`. Any provided fields will override the
+ * defaults provided by this function.
+ * @returns the stubbed `ComponentCompilerMeta`
+ */
+export const stubComponentCompilerMeta = (
+  overrides: Partial<d.ComponentCompilerMeta> = {}
+): d.ComponentCompilerMeta => {
+  // TODO(STENCIL-378): Continue to build out default stub, remove the type assertion on `default`
```
rwaskiewicz, line 13: This is a pretty complex type - I decided to add what I need here, and we'll continue to build this out. STENCIL-378 has been in the backlog for a few weeks, so _technically_ I made progress there 😉

PR #3316, opened 2022-04-05
test(methods): add unit tests to method type generation

rwaskiewicz:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

The functionality in [`generate-method-types`](https://github.com/ionic-team/stencil/blob/8feaf0816731d8dc1d880cecf6a432526295d65a/src/compiler/types/generate-method-types.ts#L1), which is used in the generation of type declaration files for class members decorated with `@Method()`, is currently not unit tested. There are however, [end to end tests](https://github.com/ionic-team/stencil/blob/8feaf0816731d8dc1d880cecf6a432526295d65a/src/compiler/transformers/test/parse-methods.spec.ts#L3) for this functionality.  We add unit tests here to minimize an upcoming bug fix PR down in size and under proper unit testing

GitHub Issue Number: N/A

## What is the new behavior?

this commit adds unit testing to the generation of `@Method` decorated
class members - specifically how we generate type information that will
be used in a type declaration file (`d.ts` file).

this commit changes the interface of the internal `generateMethodTypes`
function, requiring that `ComponentCompilerMeta` be passed to the
function rather than just the `ComponentCompilerMethod` metadata found
on the former object. this change is for a future effort where we will
be using additional information from the component metadata to avoid
naming collisions in type declaration file generation. although the
author would have liked to have made this interface change then, adding
the change now simplifies the future pr.

this pr continues to build on the stub usage explained in #3315 

- [ ] Yes
- [x] No

## Testing

Existing + new unit tests pass. Integration + karma tests pass

## Other information

This commit relies on #3315, please do not merge until the parent has landed

### DIFF
--- a/src/compiler/types/generate-method-types.ts
+++ b/src/compiler/types/generate-method-types.ts
@@ -7,14 +7,12 @@ import { getTextDocs } from '@utils';
  * @returns the generated type metadata
  */
 export const generateMethodTypes = (cmpMeta: d.ComponentCompilerMeta): d.TypeInfo => {
-  return [
-    ...cmpMeta.methods.map((cmpMethod) => ({
-      name: cmpMethod.name,
-      type: cmpMethod.complexType.signature,
-      optional: false,
-      required: false,
-      internal: cmpMethod.internal,
-      jsdoc: getTextDocs(cmpMethod.docs),
-    })),
-  ];
+  return cmpMeta.methods.map((cmpMethod) => ({
+    name: cmpMethod.name,
+    type: cmpMethod.complexType.signature,
+    optional: false,
+    required: false,
+    internal: cmpMethod.internal,
+    jsdoc: getTextDocs(cmpMethod.docs),
+  }));
 };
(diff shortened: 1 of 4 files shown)


src/compiler/types/generate-method-types.ts
```diff
@@ -1,13 +1,20 @@
...
-    jsdoc: getTextDocs(cmpMethod.docs),
-  }));
+/**
+ * Generates the individual event types for all @Method() decorated events in a component
+ * @param cmpMeta component runtime metadata for a single component
+ * @returns the generated type metadata
+ */
+export const generateMethodTypes = (cmpMeta: d.ComponentCompilerMeta): d.TypeInfo => {
+  return [
+    ...cmpMeta.methods.map((cmpMethod) => ({
```
ltm, line 11: Is there a benefit to using the spread operator here? The `map()` method already [creates a new array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map).

ltm (approved):

src/compiler/types/generate-method-types.ts
rwaskiewicz: Good point - removed in https://github.com/ionic-team/stencil/pull/3316/commits/1f6dc411506f09e1544ffc16eec0e67c8ee6147f

PR #3318, opened 2022-04-05
refactor(types): add clarity to type declaration file generation

rwaskiewicz:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

GitHub Issue Number: N/A

## What is the new behavior?

This PR is split into three commits (I'm gonna refer to them relative to one another in case the SHAs get updated, so I don't have to update the summary):

### Commit 1
this commit adds jsdoc to various internal/private stencil types for
finding typescript types and avoiding naming collisions. the purpose of
this commit is to keep refactoring/chores of this nature separate from
core logic changes - this allows this commit to be made separate from
existing pull requests

### Commit 2
this commit refactors the `update-import-refs` helper module by adding
jsdoc and renaming existing variables in the name of improving clarity.
this module utilizes multiple data structures that cover similar aspects
of the type resolution process, but can be difficult to keep a mental
model of. the intent of this refactor is to clarify the intent/purpose
of this code.

### Commit 3

this commit add jsdoc to the function that helps generate type
information for a component.

- [ ] Yes
- [x] No

## Testing

- Commit 1 is all JSDoc and one typing name update, no testing should be needed here
- Commit 2 adds JSDoc, updates local variable names, and adds a few additional TypeScript types. We are going to rely on the compiler/existing tests to catch an errors here
- Commit 3 is all JSDoc, no testing should be needed here

## Other information

This PR is part of a larger fix for avoiding naming collisions in `.d.ts` files Stencil generates. However, I've split this off to shave ~100 LoC off the main PR. This can be landed on its own, but I'll probably reference it in the main PR for folks less familiar with this section of the codebase

### DIFF
--- a/src/compiler/types/update-import-refs.ts
+++ b/src/compiler/types/update-import-refs.ts
@@ -3,7 +3,7 @@ import { dirname, resolve } from 'path';
 
 /**
  * Find all referenced types by a component and add them to the `importDataObj` parameter
- * @param importDataObj key/value of type import file, each value is an array of imported types
+ * @param importDataObj an output parameter that contains the imported types seen thus far by the compiler
  * @param typeCounts a map of seen types and the number of times the type has been seen
  * @param cmp the metadata associated with the component whose types are being inspected
  * @param filePath the path of the component file
--- a/src/compiler/types/update-import-refs.ts
+++ b/src/compiler/types/update-import-refs.ts
@@ -4,75 +4,104 @@ import { dirname, resolve } from 'path';
 /**
  * Find all referenced types by a component and add them to the `importDataObj` parameter
  * @param importDataObj key/value of type import file, each value is an array of imported types
- * @param allTypes an output parameter containing a map of seen types and the number of times the type has been seen
+ * @param typeCounts a map of seen types and the number of times the type has been seen
  * @param cmp the metadata associated with the component whose types are being inspected
  * @param filePath the path of the component file
  * @returns the updated import data
  */
 export const updateReferenceTypeImports = (
   importDataObj: d.TypesImportData,
-  allTypes: Map<string, number>,
+  typeCounts: Map<string, number>,
   cmp: d.ComponentCompilerMeta,
   filePath: string
 ): d.TypesImportData => {
-  const updateImportReferences = updateImportReferenceFactory(allTypes, filePath);
+  const updateImportReferences = updateImportReferenceFactory(typeCounts, filePath);
 
   return [...cmp.properties, ...cmp.events, ...cmp.methods]
-    .filter((cmpProp) => cmpProp.complexType && cmpProp.complexType.references)
-    .reduce((obj, cmpProp) => {
-      return updateImportReferences(obj, cmpProp.complexType.references);
+    .filter(
+      (cmpProp: d.ComponentCompilerProperty | d.ComponentCompilerEvent | d.ComponentCompilerMethod) =>
+        cmpProp.complexType && cmpProp.complexType.references
+    )
+    .reduce((typesImportData: d.TypesImportData, cmpProp) => {
+      return updateImportReferences(typesImportData, cmpProp.complexType.references);
     }, importDataObj);
 };
 
-const updateImportReferenceFactory = (allTypes: Map<string, number>, filePath: string) => {
+/**
+ * Describes a function that updates type import references for a file
+ * @param existingTypeImportData locally/imported/globally used type names. This data structure keeps track of the name
+ * of the type that is used in a file, and an alias that Stencil may have generated for the name to avoid collisions.
+ * @param typesReferences type references from a single file
+ * @returns updated type import data
+ */
+type ImportReferenceUpdater = (
+  existingTypeImportData: d.TypesImportData,
+  typeReferences: { [key: string]: d.ComponentCompilerTypeReference }
+) => d.TypesImportData;
+
+/**
+ * Factory function to create an `ImportReferenceUpdater` instance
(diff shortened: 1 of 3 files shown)


src/compiler/types/update-import-refs.ts
```diff
@@ -4,75 +4,104 @@ import { dirname, resolve } from 'path';
 /**
  * Find all referenced types by a component and add them to the `importDataObj` parameter
  * @param importDataObj key/value of type import file, each value is an array of imported types
```
alicewriteswrongs, line 6: Just reading through it looks like we're going to modify the `importDataObj` object that's passed in and then eventually return the same object. Would it make sense to clarify here in the JSDoc that that's what's going on? Just thinking that if I see a type signature like `func: (x: Foo, y: Bar) => Foo` I would tend to assume that the return `Foo` is going to be a new object (i.e. if I `let foo: Foo = { ... }; let otherFoo = func(foo);` then `foo === bar` should be `false`) but we might want to clarify that the return value is going to be the same object with new things added to it.

I guess this is a roundabout way of saying that I'm not sure what our attitude towards immutability is in Stencil - I generally find it is helpful to have immutable contracts between modules or other larger-scale 'interchanges' in the code, since then I don't have to know the implementation of a function defined in module `B` that I'm calling in module `A` to know that it's not going to mutate things that I pass in in an unexpected way, but at the same time often some well-scoped shared, mutable state is the simplest way to accomplish something in JS so it's not necessarily helpful to have a dogma of 'no mutation ever'.

alicewriteswrongs, line 6: Just reading through it looks like we're going to modify the `importDataObj` object that's passed in and then eventually return the same object. Would it make sense to clarify here in the JSDoc that that's what's going on? Just thinking that if I see a type signature like `func: (x: Foo, y: Bar) => Foo` I would tend to assume that the return `Foo` is going to be a new object (i.e. if I `let foo: Foo = { ... }; let otherFoo = func(foo);` then `foo === bar` should be `false`) but we might want to clarify that the return value is going to be the same object with new things added to it.

I guess this is a roundabout way of saying that I'm not sure what our attitude towards immutability is in Stencil - I generally find it is helpful to have immutable contracts between modules or other larger-scale 'interchanges' in the code, since then I don't have to know the implementation of a function defined in module `B` that I'm calling in module `A` to know that it's not going to mutate things that I pass in in an unexpected way, but at the same time often some well-scoped shared, mutable state is the simplest way to accomplish something in JS so it's not necessarily helpful to have a dogma of 'no mutation ever'.

alicewriteswrongs (changes requested):
the renaming, comments, etc all look good to me! I just had questions about a point in one JSDoc and was wondering about `{ [key: string]: any }` vs `Record<string, any>`.

src/declarations/stencil-private.ts
```diff
@@ -850,12 +850,32 @@ export interface ComponentCompilerPropertyComplexType {
...
+ *
+ * Each key in this record is intended to be the names of the types used by a component. However, this is not enforced
+ * by the type system (I.E. any string can be used as a key).
+ *
+ * Note any key can be a user defined type or a TypeScript standard type.
+ */
 export interface ComponentCompilerTypeReferences {
-  [key: string]: ComponentCompilerTypeReference;
+  [typeName: string]: ComponentCompilerTypeReference;
 }
```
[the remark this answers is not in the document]
rwaskiewicz: I think it can go either way - with a `Record` type, it's cleaner looking (IMO), but comes at the cost of hiding some of the underlying detail. With `Record<string, ComponentCompilerTypeReference>` (or any types really) we don't have the opportunity to name the key in the index signature.  We _could_ argue that we 'lose' some data that way for consumers.  However, we now have JSDoc on this type, so whether or not we're _really_ losing anything probably comes down to how various editors/plugins show the data from the language server.

FWIW this is what JetBrains shows me (there's no index signature at all)
Screen Shot 2022-04-18 at 12 03 24 PM.

The only other thing I can think of is that moving this to a `Record` would require to change this from an `interface` to a `type`:
```
export type ComponentCompilerTypeReferences = Record<string, ComponentCompilerTypeReference>;
```

AFAICT, we don't export this type anywhere (either directly or indirectly), and it's defined in `stencil-private.ts` (so nobody should be using it other than us). That minimizes the chances of anyone relying on the existing `ComponentCompilerTypeReferences` def as an interface - by moving this from an `interface` to a `type`, we remove the possibility of type merging/augmenting `ComponentCompilerTypeReference` automagically.

With that in mind, I think I'd be happy making the change. Worst case scenario, its an easy enough revert. 

93a06af3b

src/compiler/types/update-import-refs.ts
rwaskiewicz: 👍 updated in e2ebcf32c

As far as immutability goes...I don't think we can derive a hard and fast rule to uniformly apply to everything.  In general I agree with your point above about it not being helpful to be overly dogmatic here.  The problem that I think we will need to be aware of is that immutability can lead to a larger memory footprint.  When we handle data structures of certain sizes, that has the capability to crash the compiler (or more specifcially, the Node runtime runs out of memory, and _that_ crashes the compiler).  I don't think that's going to be a huge problem here, but have seen it lead to issues elsewhere in the codebase

alicewriteswrongs: that makes sense, and going the immutable route does generally lead to a good deal of overhead w/ things like immer and whatnot. I'm for what gives us good performance!

alicewriteswrongs (approved):
I think this all looks good!

PR #3331, opened 2022-04-14
chore(github): update github action versions

rwaskiewicz:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes

We're using an older version of the setup-node github action in our jobs, and are using a less secure versioning scheme for both setup-node and checkout actions

GitHub Issue Number: N/A

## What is the new behavior?

this commit updates the versions of the github actions used by the
browserstack and main workflows:

- checkout is pinned to the v2.4.0 release
- setup-node is upgraded to the v3.1.1 release

in this commit we move to using SHAs instead of the semver string per
the guidance of [github](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions)

> Pinning an action to a full length commit SHA is currently the only
> way to use an action as an immutable release. Pinning to a particular
> SHA helps mitigate the risk of a bad actor adding a backdoor to the
> action's repository, as they would need to generate a SHA-1 collision
> for a valid Git object payload.

although we _probably_ can inherently trust github actions that are
authored by github, better safe than sorry.

- [ ] Yes
- [x] No

## Testing

Tailed the logs from CI to ensure that everything worked as expected

We use the `pull_request_target` trigger instead of `pull_request` for our BrowserStack job. With `pull_request_target` triggers, the versions of GitHub Actions that are used (checkout and setup-node) are pulled from the _target_ branch, (e.g. `main`) rather than the branch associated with the pull request.  To ensure that the changes to the BrowserStack job were working correctly, I temporarily switched our usage of `pull_request_target` to `pull_request` in https://github.com/ionic-team/stencil/pull/3331/commits/db660a1d5b6e0f2bec18a9fa7d66c08ce94c6d1f. This causes a new BrowserStack job to run with the updates.  I used the output from this job to verify my changes, which can be found [here](https://github.com/ionic-team/stencil/runs/6024275166?check_suite_focus=true).

### DIFF
--- a/.github/workflows/browserstack.yml
+++ b/.github/workflows/browserstack.yml
@@ -3,8 +3,8 @@ name: BrowserStack
 on:
   pull_request_target:
     branches:
-      - main
-      - v3.0.0-dev
+      - 'main'
+      - 'v3.0.0-dev'
   push:
     branches:
       - 'main'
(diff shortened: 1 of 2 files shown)


alicewriteswrongs (approved):
all looks straightforward enough to me!

PR #3332, opened 2022-04-14
chore(github): split up github actions

rwaskiewicz:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes

Our CI system was stood up pretty quickly to get things 'working'
while simultaneously trying to learn/deliver value to Stencil. While
we've made incremental improvements/additions to it over the last
9 months or so, it's become unwieldy/a time sink when it comes to
landing code.  In particular, the issues that this PR will seek to resolve
are:

- No 'retry' granularity when a flaky e2e test fails. If a test is flaky and fails,
we must rerun the build/unit tests/e2e tests for a specific version of node +
operating system
- Slowness - builds have been creeping into the ~20 min mark. Stencil isn't 
a small project, but it's not so big that CI should take 20 minutes.

GitHub Issue Number: N/A

## What is the new behavior?

this commit reworks the 'main' github workflow to address the two
points above. it uses the concept of composite actions to allow us to
parallelize and categorize our tests in new ways.

previously, we would run the build, unit tests, e2e tests, etc all in one
job, then run one job for each OS/Node version combination:
Screen Shot 2022-04-15 at 11 33 21 AM

now, we run a job for the build and format separately. upon completion of
the build job, we run the analysis tests, unit tests, and e2e tests for each
OS/Node version in a separate job. this results in more simultaneous runners
(which we appear to be allowed under the GH Free plan for OSS) and a faster 
CI run.
Screen Shot 2022-04-15 at 11 32 46 AM

this change no longer builds stencil for each OS/Node version, as doing so is
not representative of how stencil is built today. this saves us additional time
in the pipeline and brings it closer to reality.

this change removes the unused `sys.node` script, which simply echoed 'sys.node'
to the terminal.

this change also removes `npm pack`ing stencil and pushing the tarball to github
actions as a separate step, as this was not used.

this change also fixes a small bug in the e2e tests, where dependencies were
not guaranteed to be installed before running the tests

- [ ] Yes
- [x] No

## Testing

I've run CI, _a lot_ to test various part of this rework. The problem (IMO)
is that the changes can't be validated without careful documentation reading
and pushing configuration changes.  So I pushed (a lot). Please please please
ask questions if you have any about a particular line/configuration

## Other information

I'm going to need to update GitHub settings immediately after this lands to mark the new CI as required to land PRs (and remove the old ones). I can't do this ahead of time, otherwise that'd block other PRs from landing in the interim

I focused only on the main workflow here - I did not attempt to refactor our BrowserStack workflow in this commit/PR to limit scope. I don't believe there's much room/benefit from a speed perspective to be had here by doing a similar action there.


test/package.json
```diff
@@ -7,7 +7,7 @@
     "analysis.build-and-analyze": "npm run build && npm run analysis",
     "build": "npm run build.browser-compile && npm run build.hello-world && npm run build.hello-vdom && npm run build.todo && npm run build.end-to-end && npm run build.ionic",
     "build.browser-compile": "cd ./browser-compile && npm run build",
-    "build.end-to-end": "cd ./end-to-end && npm run build",
+    "build.end-to-end": "cd ./end-to-end && npm ci && npm run build",
```
rwaskiewicz, line 10: In the old CI, we would have installed these from a previous, unrelated step. This ensures we always have these deps

.github/workflows/main.yml
```diff
@@ -1,4 +1,4 @@
-name: Build & Test
+name: CI
```
rwaskiewicz, line 1: This isn't a _great_ name. But the only place it shows up is here:
Screen Shot 2022-04-15 at 11 47 27 AM

Where I wanted something short+sweet to focus on the job that actually failed

.github/workflows/README.md
```diff
@@ -0,0 +1,104 @@
+# Stencil Continuous Integration (CI)
```
rwaskiewicz, line 1: This readme is intentionally high level. I didn't wanna regurgitate the details of every job/step here. If there's something y'all have questions about lmk. An inline comment somewhere may be necessary

.github/workflows/README.md
```diff
@@ -0,0 +1,104 @@
...
+It is responsible for:
+- Building Stencil (and Validating the Build)
+- Running Unit, E2E, etc. Tests against the Build
+- Verifying that the Codebase is Properly Formatted
+
+Each of these tasks are codified as _jobs_, and are listed in `main.yml`.
+Most of the jobs above are contingent on the 'build step' finishing (otherwise there would be nothing to test).
+The diagram below displays the dependencies between each job.
+
+```mermaid
```
rwaskiewicz, line 38: Use 'display rich diff' in GH to see the 🧜‍♀️ diagram

.github/workflows/actions/download-archive/action.yml
```diff
@@ -0,0 +1,20 @@
...
+    description: 'name of the archive to decompress'
+runs:
+  using: 'composite'
+  steps:
+    - uses: actions/download-artifact@fb598a63ae348fa914e94cd0ff38f362e927b741 # v3.0.0
+      with:
+        name: ${{ inputs.name }}
+        path: ${{ inputs.path }}
+
+    - name: Exract Archive
```
alicewriteswrongs, line 18:
```suggestion
    - name: Extract Archive
```

.github/workflows/actions/get-core-dependencies/action.yml
```diff
@@ -0,0 +1,24 @@
+name: 'Get Core Dependencies'
+description: 'initializes core dependencies after repo has been pulled'
+runs:
+  using: composite
+  steps:
+    - name: Use Node 16
+      uses: actions/setup-node@56337c425554a6be30cdef71bf441f15be286854 # v3.1.1
+      with:
+        node-version: 16
+        cache: 'npm'
```
alicewriteswrongs, line 10: I want to check my understanding here - does this mean that the dependency install is always going to happen on node 16? I wonder if we could run into an issue where a `postInstall` script runs correctly on node 16 but would error on node 14 - possibly this isn't really something to worry about though

.github/workflows/main.yml
```diff
@@ -9,75 +9,63 @@ on:
...
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
-  build_and_test:
+  build_core:
+    name: Build Stencil
+    runs-on: 'ubuntu-latest'
+    steps:
+      - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # v2.4.0
```
alicewriteswrongs, line 21: a small idea I had: each of the jobs that we define here in `main.yml` starts with the same `actions/checkout` bit, and then within each of their separate yml files (e.g. `build-core/action.yml` for this one) has to start with the `get-core-dependencies` action.

Would it be helpful for us to move the `checkout` action into the `get-core-dependencies` action, possibly renaming it to something like `checkout-and-install` (or some better name than that) ? I was just thinking that if we did that then we'd avoid having to put this SHA in four different places, and could just put it in one place instead.

alicewriteswrongs (changes requested):
total run time definitely is lower, and I think this is much more nicely factored overall (and will make re-running failed builds easier)

just had a few comments / questions reading though!

.github/workflows/main.yml
```diff
@@ -9,75 +9,63 @@ on:
...
+      - uses: ./.github/workflows/actions/build-core
+
+  format:
+    name: Format Check
+    runs-on: 'ubuntu-latest'
+    steps:
+      - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # v2.4.0
+      - uses: ./.github/workflows/actions/format
```
ltm, line 29: The GitHub log viewer doesn't play well with composite actions and makes the output a lot less comprehensible. It probably comes down to personal preference, but I'd rather have one large workflow with each step clearly visible in the log. The composite actions are useful when reused in multiple steps.

Screen Shot 2022-04-15 at 1 26 08 PM

Screen Shot 2022-04-15 at 1 26 02 PM

.github/workflows/main.yml
```diff
@@ -9,75 +9,63 @@ on:
...
+
+  unit_test:
+    needs: [build_core]
     strategy:
+      fail-fast: false
       matrix:
-        node: ['12', '14', '16']
-        os: ['ubuntu-latest', 'windows-latest']
+        node: [ '12', '14', '16' ]
+        os: [ 'ubuntu-latest', 'windows-latest' ]
```
ltm, line 37: Prettier uses no bracket spacing in this case:
```suggestion
        node: ['12', '14', '16']
        os: ['ubuntu-latest', 'windows-latest']
```

.github/workflows/actions/get-core-dependencies/action.yml
```diff
@@ -0,0 +1,24 @@
...
+
+    - name: Cache Root Node Modules
+      id: get-stencil-root-node-modules-cache
+      uses: actions/cache@48af2dc4a9e8278b89d7fa154b955c30c6aaab09 # v3.0.2
+      env:
+        cache-name: stencil-root-node-modules
+      with:
+        key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('./package-lock.json') }}
+        path: ./node_modules
```
ltm, line 19: Note that `actions/cache` [recommends](https://github.com/actions/cache/blob/main/examples.md#node---npm) not caching `node_modules`.

.github/workflows/actions/build-core/action.yml
```diff
@@ -0,0 +1,28 @@
...
+    - name: Validate Testing
+      run: npm run test.testing
+      shell: bash
+
+    - name: Upload Build Artifacts
+      uses: ./.github/workflows/actions/upload-archive
+      with:
+        name: stencil-core
+        output: stencil-core-build.zip
+        paths: build cli compiler dependencies.json dev-server internal mock-doc scripts screenshot sys testing
```
ltm, line 27: I'm not sure if it's feasible, but if the artifact is generated using `npm pack` instead we don't have to keep this list in sync with `package.json:files` and it would ensure that the tests run against the to-be-published package (or at least close it).

.github/workflows/actions/get-core-dependencies/action.yml
alicewriteswrongs: based on https://github.com/actions/setup-node#caching-packages-dependencies it looks like we should already be getting some dependency caching in our `actions/setup-node` use above (b/c `cache: 'npm'`) - so we might be able to just inline that in each of the testing jobs

.github/workflows/README.md
```diff
@@ -0,0 +1,104 @@
+# Stencil Continuous Integration (CI)
+
+Continuous integration (CI) is an important aspect of any project, and is used to verify and validate the changes to the
+codebase work as intended, do not introduce regressions (bugs), and adhere to coding standards (e.g. formatting rules).
```
alicewriteswrongs, line 4:
```suggestion
codebase work as intended, to avoid introducing regressions (bugs), and to adhere to coding standards (e.g. formatting rules).
```

alicewriteswrongs (changes requested):
little copy-edit

.github/workflows/actions/build-core/action.yml
rwaskiewicz: That's fair, I'm gonna create a "throwaway PR" to test this (so that I don't spam folks by verifying this and other changes) later this week and will report back

.github/workflows/actions/get-core-dependencies/action.yml
rwaskiewicz: >  does this mean that the dependency install is always going to happen on node 16?

Yeah, that's correct. At the moment we do the install/build/publish all from Node 16 (manually, although we have plans to change that), so I considered the risk of a `postInstall` script running differently in a separate version of Node to be minimal.

.github/workflows/actions/get-core-dependencies/action.yml
rwaskiewicz: I'll play around with this a bit in a "throwaway PR" (as to not spam y'all) and report back. I agree we probably shouldn't cache `node_modules/` - I think where I want this to move is that if we're running on Linux/Mac we cache `~/.npm` and if we're on windows we cache the equivalent dir.  I'll play with it

.github/workflows/main.yml
rwaskiewicz: > Would it be helpful for us to move the checkout action into the get-core-dependencies action

I had the same idea, the problem is that putting that in a separate composite action/job will run the checkout on another VM, so we won't have the code, actions, etc. we need to run a job. GitHub actually errors here asking if we checked out the code or not when we move this 😢 

But let me double check that in a "throwaway PR" - I was moving fast and breaking things during JEDI days, let me double check my understanding and I'll get back to you on this

.github/workflows/main.yml
rwaskiewicz: That's fair, I tend to think it's all the same log at the end of the day. IMO I prefer the speed at the cost of readability here.  Maybe there's something that can be done here with >1 workflow and dependent workflows. That may give us a good balance if we can get each test (analysis, unit, e2e) in their own workflow. Let me play with it in a "throwaway PR" - I tried on JEDI days but couldn't get it working the way I wanted (at all 😆)

.github/workflows/main.yml
rwaskiewicz: Note to self: I should get prettier running in this dir (on YAML)

.github/workflows/README.md
rwaskiewicz: 👍 fixed in https://github.com/ionic-team/stencil/pull/3332/commits/8f85fe5d6465af68c7989ae9d4ade23d80beb2df

.github/workflows/actions/download-archive/action.yml
rwaskiewicz: Bah - clearly my spell check isn't working 😢 fixed in https://github.com/ionic-team/stencil/pull/3332/commits/b0eb5d8c92f70a5d63f51dcf991a8ba0cea7fe00. Good catch!

.github/workflows/main.yml
rwaskiewicz: Fixed in https://github.com/ionic-team/stencil/pull/3332/commits/25ee864ff645ab91b9f4a7fabab1178ddeee1a26

.github/workflows/actions/get-core-dependencies/action.yml
alicewriteswrongs: that makes sense 👍

.github/workflows/main.yml
rwaskiewicz: I double checked this last night and confirmed it's a bit of a "chicken and egg" problem. It'd be nice to abstract the checkout action into the `get-core-dependencies` job, but to get that job, we'd need to checkout the repo first 🙃 For now I think we'll need to leave this as is (the good news is that at least we should be able to ctrl-f/replace instances of the SHA when it comes time to update this)

.github/workflows/actions/get-core-dependencies/action.yml
rwaskiewicz: Fixed in https://github.com/ionic-team/stencil/pull/3332/commits/3bb286922bc1131e480ca37fb053d38365689b3c - I had to add an extra step for windows to ensure that we got the correct `APP_DATA` dir

.github/workflows/actions/build-core/action.yml
rwaskiewicz: I tried this with a `mv \`npm pack\` some-tar-name.tar` - I ran into problems a few problems here:
- `npm pack` packed everything, including `node_modules/`. This wasn't a huge time sink (although there was some), but it definitely spammed the logs when I unpacked it in a subsequent step
- When unpacking `npm pack`, GHA would crash. No message/logs (to be fair, I didn't enable extra debugging in the repo), just an exit code of 1.  

I may poke at this a little more so we don't have to update these paths, but the early experiments weren't super encouraging 😆

.github/workflows/main.yml
alicewriteswrongs: sounds straightforward enough!

.github/workflows/actions/get-core-dependencies/action.yml
```diff
@@ -0,0 +1,40 @@
...
+runs:
+  using: composite
+  steps:
+    - name: Use Node 16
+      uses: actions/setup-node@56337c425554a6be30cdef71bf441f15be286854 # v3.1.1
+      with:
+        node-version: 16
+        cache: 'npm'
+
+    - name: NPM Cache (Unix)
```
alicewriteswrongs, line 12: is this step (and the windows-specific step below) necessary? In the test output it looks like a caching operation is happening twice when this job is running:

Screen Shot 2022-04-20 at 9 50 08 AM

In the documentation for the `setup-node` action it says that caching the global package cache for npm between builds should be supported by setting `cache: 'npm'` - am I missing something here, or are we possibly duplicating that here?

alicewriteswrongs (changes requested):
just had a question about caching

.github/workflows/actions/get-core-dependencies/action.yml
rwaskiewicz: Nope - good call, that shouldn't be needed anymore. I ported the caching of `node_modules/` to caching `~/.npm` and forgot that setup-node does that for us 😝

.github/workflows/actions/build-core/action.yml
rwaskiewicz: Unfortunately I think we have to remain on this implementation - I couldn't get everything packed up the way I wanted to that didn't involve one list of directories or another to maintain :-(

.github/workflows/actions/get-core-dependencies/action.yml
rwaskiewicz: I moved the impl to reusable workflows in https://github.com/ionic-team/stencil/pull/3332/commits/d7a0795dae7bbda3211e7061847e947914445158. I'll write something up about the trade-offs in the morning

rwaskiewicz:
@ltm @alicewriteswrongs 

In https://github.com/ionic-team/stencil/pull/3332/commits/3bb286922bc1131e480ca37fb053d38365689b3c, I moved us from using composite actions to using [reusable workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows).  

This ought to give us a better logging experience in the GH UI. Previously, all the steps that occurred in a single composite action would look something like this:
Screen Shot 2022-04-22 at 8 11 53 AM

Now, it should look something like this:
Screen Shot 2022-04-22 at 8 23 47 AM

Regardless of whether we use composite actions or reusable workflows, the raw log that is served by GH remains the same. By moving us to composite workflows, the gain is that the individual steps that were abstracted away into an action can be inspected one-by-one in the UI.  Although our CI logs are pretty small ATM, this should help with small spot checking of the logs/investigating failures

It's probably worth calling out there are a few[ limitations to using reusable workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows#limitations):

> 1. Reusable workflows can't call other reusable workflows.
> 2. Any environment variables set in an env context defined at the workflow level in the caller workflow are not propagated to the called workflow.
> 3. The strategy property is not supported in any job that calls a reusable workflow.

At this time, our CI is designed such that we don't run into any of these issues:

1. We have a 'main' yaml file that acts as a caller of all reusable workflows
2. We don't use any env vars that need to propagate at this time
3. We work around this by using the strategy property in the called workflows, rather than in the caller workflow (main)

☝️ something we should keep in mind as we think through the growth of this system. Although I'm not sure _what_ that looks like exactly, but these are decisions that we'll make today that affect the ones we can make tomorrow 😆 (or we just rewrite it all again, that's why pencils have erasers, right?)

alicewriteswrongs:
Thanks for the explanation of reusable workflows vs composite actions and the limitations of each. I also found this blog post which was (somewhat) helpful in understanding the differences / limitations: https://github.blog/2022-02-10-using-reusable-workflows-github-actions/

A lot of terminology going on with gh actions!

Am I right in understanding that limitation no. 3 (jobs that use `strategy` cannot call reusable workflows) is the reason we have to keep the `get-core-dependencies` and `(download|upload)-archive` jobs as composite actions instead of being able to switch them over to reusable workflows?

I don't get why using `strategy` and calling a reusable workflow in the same isn't supported, it seems kind of unfortunate. 

I think it is helpful to separate out the different test procedures as reusable workflows so they can be written in separate YAML files without bundling together all the steps in the gh actions UI. Even though they are reusable workflows we aren't really re-using any of them, but are only using the composite actions in multiple places. So it seems like we could simply have the test jobs and whatnot all get written into the `main.yml` file directly if we wanted without changing too much. 

I have a question I just thought of which might be something to consider here. Our test reusable actions currently all look the same with the exception of the particular `package.json` script which is called at the end. What if we just had a `run-test` reusable action which took a test command parameter instead? Something like this:

```yaml
name: Test

on:
  workflow_call:
    inputs:
      test_command:
        required: true

jobs:
  test:
    name: (${{ matrix.os }}.${{ matrix.node }})
    strategy:
      fail-fast: false
      matrix:
        node: [ '12', '14', '16' ]
        os: [ 'ubuntu-latest', 'windows-latest' ]
    runs-on: ${{ matrix.os }}
    steps:
      - name: Checkout Code
        uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # v2.4.0

      - name: Get Core Dependencies
        uses: ./.github/workflows/actions/get-core-dependencies

      - name: Download Build Archive
        uses: ./.github/workflows/actions/download-archive
        with:
          name: stencil-core
          path: .
          filename: stencil-core-build.zip

      - name: Run Test Command
        run: npm run ${{ inputs.test_command }}
        shell: bash
```

If we saved that in `.github/workflows/run-test.yml` then I think in `main.yml` our tests could each look something like


```yaml
  analysis_tests:
    name: Analysis Tests
    needs: [ build_core ]
    uses: ./.github/workflows/run-test.yml
    with:
      test_command: "test.analysis"
```

the advantage of that kind of approach would be just that then our build matrix and node versions get declared in one place, instead of being duplicated across the testing reusable workflows.

Does that make sense, and am I missing any reason why that wouldn't work?

Also — it's entirely possible that we should stop yak shaving on gh actions and merge what you have here! So I think if you think so, feel free to say so 😆

rwaskiewicz:
@alicewriteswrongs 

> Am I right in understanding that limitation no. 3 (jobs that use strategy cannot call reusable workflows) is the reason we have to keep the get-core-dependencies and (download|upload)-archive jobs as composite actions instead of being able to switch them over to reusable workflows?

Yep 👍 

> I have a question I just thought of which might be something to consider here. Our test reusable actions currently all look the same with the exception of the particular package.json script which is called at the end. What if we just had a run-test reusable action which took a test command parameter instead? Something like this:

I do think this would work! The reason I didn't take this a step further is that I'm not 100% sure what the 'next step' in optimizing CI runs look like.  As a concrete example the `test-analysis.yml` workflow calls `npm run test.analysis` in the root `package.json`. That acts as a proxy to calling `analysis.build-and-analyze` in `test/package.json`.

Expanding looks something like this:
```
"analysis.build-and-analyze": "npm run build && npm run analysis",

# expand 'npm run build' and 'npm run analysis' (note the analysis script is skipped until STENCIL-389 is done)
"analysis.build-and-analyze": "npm run build.browser-compile && /
    npm run build.hello-world && /
    npm run build.hello-vdom && /
    npm run build.todo && /
    npm run build.end-to-end && /
    npm run build.ionic && /
    exit 0 && node ./.scripts/analysis.js",
```

I don't _love_ the fact we're building all of these packages in sequence. There's probably a lot of room to parallelize them (and drop the time of this step quite a bit).  But that work felt like it was out of scope for this PR - I opted to keep these workflows separate was to give us some time to experiment with different strategies for iterating on each task (unit tests, end-to-end tests, analysis tests, etc.).  If we find that we have the chance to reconcile all of these into a single reusable composite action after we've spent some more time working on these, I'll all for it!

alicewriteswrongs (approved):
I think we can merge this and get things to a much better place than they are today and then look for further optimizations down the road 👍

rwaskiewicz:
Merging 'as admin' - CI is passing, but the checks required no longer exist. I'm going to change that in GH after this lands

PR #3333, opened 2022-04-14
test(compiler): remove strictNullChecks errors from validate-dist.ts

alicewriteswrongs:
this makes changes necessary to fix the errors from compiling w/
`strictNullChecks` set to `true` in
src/compiler/config/outputs/validate-dist.ts.

## Pull request checklist

Please check if your PR fulfills the following requirements:
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [x] Build (`npm run build`) was run locally and any changes were pushed
- [x] Unit tests (`npm test`) were run locally and passed
- [x] E2E Tests (`npm run test.karma.prod`) were run locally and passed
- [x] Prettier (`npm run prettier`) was run locally and passed

## Pull request type

 

 

Please check the type of change your PR introduces:
- [ ] Bugfix
- [ ] Feature
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

## What is the current behavior? What is the new behavior?

On the main branch turning on `strictNullChecks` produces 2229 errors. With the changes on this branch we get 2182, so we've fixed 47. Gotta start somewhere!

## Does this introduce a breaking change?

- [ ] Yes
- [ ] No

## Testing

Tests should all pass! The typechecker too.

You can also run `npx tsc --strictNullChecks` locally and verify that you get fewer errors than on `main`. You could also change that value in `tsconfig.json` temporarily and check that you don't get any compile errors in the file when opening it in your editor, or check the output to make sure there aren't any errors in `validate-dist.ts`.

PR #3335, opened 2022-04-15
test(config): remove strictNullChecks errors from compiler/config

alicewriteswrongs:
This is another PR for removing _some_ strictNullChecks errors, in
particular in the code related to loading and validating the `Config`.

This PR proposes a change to how the `Config` loading / validation
pipeline works. Instead of dealing the whole way through with a single
`Config` type, this puts a boundary in place which is crossed in the
`validateConfig` function between an `UnvalidatedConfig` and a `Config`.

How does this work? Basically `Config` is the type that we expect to be
able to pass around everywhere else in the codebase without issue, while
`UnvalidatedConfig` is a type which is around to specifically denote the
fact that the object we're dealing with is sort of a 'config in
progress' and it cannot yet be used freely throughout the compiler.

`UnvalidatedConfig` is implemented w/ a new type I added called `Loose`,
which looks like this:

```ts
type Loose<T extends Object> = Record<string, any> & Partial<T>
```

`UnvalidatedConfig` looks like this:

```ts
type UnvalidatedConfig = Loose<Config>
```

This amounts to 1) making all properties on `Config` optional (via
`Partial`) and 2) allowing access to properties which are _not_ defined
on `Config`.

We need the former because this opens the door to later changing the
typing of properties on `Config` to _not_ all be optional (as they
currently are). This will be a big help towards turning
`strictNullChecks` on, since doing so right now results in a literal ton
of errors around property access on `Config` objects—consider, for
instance, that currently `Config.sys` is an optional property, so _every
time_ the `.sys` property is accessed we'll get an error with
`strictNullChecks`.

We need the latter change because with `strictNullChecks` if you do
something like the following you'll get a type error:

```ts
interface Foo {
    bar: string
}

let obj: Foo = {
    bar: "hey!"
}

let otherPropValue = obj["asdfasdf"]
```

TypeScript here will (and should!) throw an error for `otherPropValue`
because the index type on `Foo` is `"bar"` and not `string`. The
`Record<string, any> &` bit in the definition of `Loose` lets us access
properties not defined on `Config` in our validation code (for an
example of this see the `setBoolean` function in
`/src/compiler/config/config-utils.ts`) without giving up on types in
`Config` entirely. What do I mean by that? Basically just that if you do
this:

```ts
let config: UnvalidatedConfig = someFunctionThatProducesConfig();
```

then the type of `config["somePropertyWeDontDefine"]` will be `any`, and
we can do things we need to with such properties within the validation
code, _but_ if I access `config.sys` the type will still be
`CompilerSystem`. So for this reason I think typing something as
`Loose<T>` is superior to just doing `Record<string, any>` or `Object`
or (shudders) `any`.

This commit just gets us started with this sort of 'lifecycle' for
Config objects (they start out as `UnvalidatedConfig`s, get passed
through `validateConfig` and come out as full `Config` objects) and more
work will be needed to get all of the config validation and loading code
on board. Once we do that we can safely start removing optional
properties on `Config` itself, since we'll have assurance from the
compiler that anything marked `Config` has already been fully validated
and should always have, e.g., a `.sys` property, a `buildDir` property,
etc. This will make it much easier for us to turn `strictNullChecks` on
without having to sprinkle the codebase with (literally hundreds of)
optional property accesses (`?.`).

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes

I was working a bit yesterday and today on what would be involved in turning on `strictNullChecks` for typescript. I noticed that the `Config` object in particular is not in a good position to comply with `strictNullChecks`, since almost every property defined on it is optional. This means that w/ `strictNullChecks` almost every time we access a property on it we'd need to add a `?.` or `if (config.sys === undefined)` or something like that.

## What is the new behavior?

I am proposing here a change to how this object is typed, specifically by adding a new type called `UnvalidatedConfig` that we use throughout the configuration loading / validation process. Then our `validateConfig` function has it's signature changed, from

```ts
function validateConfig(config: Config): Config
```

to

```ts
function validateConfig(confing: UnvalidatedConfig): Config
```

This will give us some confidence that any time we see a `Config` we know it's already good to go, and we could start to remove all the optional typings on it so that we'd be able to turn `strictNullChecks` on.

This change also brings us from 2229 `strictNullChecks` errors on `main` to 2185 errors, which isn't a ton _but_ I think introducing the `UnvalidatedConfig` change will allow us to more strictly type the `Config` object in the future and that will I think facilitate getting rid of a lot more errors.

- [ ] Yes
- [x] No

## Testing

I ran the typechecking a bunch, ran the tests a whole bunch, and tried it out in a local test component I have. I am fairly confident that these changes are basically confined to the type system and changing `null` to `undefined` in a few places, but it's definitely something to check out thoroughly!


src/compiler/config/config-utils.ts
```diff
@@ -21,64 +44,31 @@ export const setBooleanConfig = (config: any, configName: string, flagName: stri
...
 
-  if (typeof config[userConfigName] === 'boolean') {
+  if (isBoolean(config[userConfigName])) {
     config[configName] = config[userConfigName];
   } else {
     config[configName] = defaultValue;
   }
 };
 
-export const setNumberConfig = (config: any, configName: string, _flagName: string, defaultValue: number) => {
```
alicewriteswrongs, line 31: I removed a few functions here because they were commited in 2018 and I couldn't find any uses of them

src/compiler/config/config-utils.ts
```diff
@@ -21,64 +44,31 @@ export const setBooleanConfig = (config: any, configName: string, flagName: stri
...
+ * Find any possibly mis-capitalized configuration names on the config, logging
+ * a little warning for the user to let them know. This lets us recover values
+ * set under (a subset of) improperly spelled configs and automatically hoist
+ * them into the config under the right key.
+ *
+ * @param config d.Config
+ * @param correctConfigName the configuration name that we're checking for right now
+ * @returns a string container a mis-capitalized config name found on the
+ * config object, if any.
+ */
```
alicewriteswrongs, line 64: this, like all other JSDoc comments I'm adding, reflects _my belief_ of what this function does - very possibly I am wrong!

src/compiler/config/validate-config.ts
```diff
@@ -132,8 +145,11 @@ export const validateConfig = (userConfig?: Config) => {
     return arr;
   }, [] as RegExp[]);
 
+  // this is well justified I promise :)
+  let validatedConfig: Config = config as any;
```
alicewriteswrongs, line 149: this is, afaict, unfortunately necessary for now. The thing to do is to refactor all of this validation so that we do something like:

```ts
const obj = {
  flags: validateFlags(userConfig),
  outputTargets: validateOutputTargets(userConfig),
  ...
}
```

basically 1) refactoring the 'sub-validators' to take an `UnvalidatedConfig` (and whatever else they need) and return the `Config` prop they're concerned with and 2) then building up an object based on that which conforms to the `Config` interface. This can then be returned from the function and we'll have full type-checking of everything going on to convert an `UnvalidatedConfig` to a `ValidatedConfig`.

src/compiler/config/validate-hydrated.ts
```diff
@@ -1,9 +1,9 @@
-import { Config, HydratedFlag } from '../../declarations';
+import { HydratedFlag, UnvalidatedConfig } from '../../declarations';
 import { isString } from '@utils';
 
-export const validateHydrated = (config: Config) => {
-  if (config.hydratedFlag === null || config.hydratedFlag === false) {
-    return null;
+export const validateHydrated = (config: UnvalidatedConfig) => {
+  if (config.hydratedFlag === undefined || config.hydratedFlag === null || config.hydratedFlag === false) {
+    return undefined;
```
alicewriteswrongs, line 6: this was returning `null` but `config.hydratedFlag` is typed `undefined | HydratedFlag`, so I changed the early return to `undefined`

src/compiler/config/validate-config.ts
rwaskiewicz: Hmmm with `strictNullChecks: true` set, I'm not seeing any errors if I revert this change (diff below just in case).  Do we still need this change?

```diff
-  // this is well justified I promise :)
-  let validatedConfig: Config = config as any;
-
   return {
-    config: validatedConfig,
+    config,
     diagnostics,
   };
```

src/compiler/config/validate-dev-server.ts
```diff
@@ -28,24 +32,23 @@ export const validateDevServer = (config: d.Config, diagnostics: d.Diagnostic[])
 
   devServer.address = devServer.address.split('/')[0];
 
-  let addressPort: number;
   const addressSplit = devServer.address.split(':');
   if (addressSplit.length > 1) {
     if (!isNaN(addressSplit[1] as any)) {
       devServer.address = addressSplit[0];
-      addressPort = parseInt(addressSplit[1], 10);
+      devServer.port = parseInt(addressSplit[1], 10);
```
rwaskiewicz, line 39: Is it possible that we're accidentally prematurely overwriting this value if [`flags.port` is a number](https://github.com/ionic-team/stencil/pull/3335/files#diff-1b29641703a279611b48a9e91ce3ead0b080bc5a91df7b6347483b0231601257R43). It looks like if the previous implementation, `addressSplit[1]` took precedence over `flags.port`. Here, it looks like we've flipped that

src/compiler/config/validate-dev-server.ts
```diff
@@ -3,21 +3,25 @@ import { buildError, isBoolean, isNumber, isString, normalizePath } from '@utils
...
   if ((config.devServer === null || (config.devServer as any)) === false) {
-    return null;
+    return undefined;
   }
 
   const flags = config.flags;
   const devServer = { ...config.devServer };
 
-  if (isString(flags.address)) {
+  if (flags && flags.address && isString(flags.address)) {
```
rwaskiewicz, line 17: I wonder if it makes sense to update the declaration of `flags` on line 14 to default to an object literal like so:

```diff
- const flags = config.flags;
+ const flags = config.flags || {};
```

That way, we don't need to do checks such as this:
```suggestion
  if (flags.address && isString(flags.address)) {
```

or do optional chaining on every usage of `flags` throughout this function.

Thoughts?

src/compiler/config/validate-testing.ts
```diff
@@ -4,20 +4,20 @@ import { isAbsolute, join, basename, dirname } from 'path';
...
-export const validateTesting = (config: d.Config, diagnostics: d.Diagnostic[]) => {
+export const validateTesting = (config: d.UnvalidatedConfig, diagnostics: d.Diagnostic[]) => {
   const testing = (config.testing = Object.assign({}, config.testing || {}));
 
   if (!config.flags || (!config.flags.e2e && !config.flags.spec)) {
     return;
   }
 
-  let configPathDir = config.configPath;
+  let configPathDir = config.configPath!;
```
rwaskiewicz, line 14: For my own understanding, how do we know that `configPath` exists on `config` at this point in the function?

src/utils/helpers.ts
```diff
@@ -82,10 +82,11 @@ export const pluck = (obj: { [key: string]: any }, keys: string[]) => {
 
 export const isBoolean = (v: any): v is boolean => typeof v === 'boolean';
 export const isDefined = (v: any) => v !== null && v !== undefined;
-export const isUndefined = (v: any) => v === null || v === undefined;
+export const isUndefined = (v: any): v is null | undefined => v === null || v === undefined;
 export const isFunction = (v: any): v is Function => typeof v === 'function';
 export const isNumber = (v: any): v is boolean => typeof v === 'number';
```
rwaskiewicz, line 87: Not related to this PR, but I think there's a tiny typo in this guard:
```suggestion
export const isNumber = (v: any): v is number => typeof v === 'number';
```

rwaskiewicz (changes requested):

src/compiler/config/config-utils.ts
alicewriteswrongs: excellent! deleting code is the best kind of refactoring

src/utils/helpers.ts
alicewriteswrongs: 😱

alicewriteswrongs:
@rwaskiewicz one higher-level / "conceptual" question I was wondering about after your comment [here](https://github.com/ionic-team/stencil/pull/3333#discussion_r853450738) - do we want to have a `Config` type with all the optional properties, and have `validateConfig` have a signature like `Loose<Config> -> ValidatedConfig` where `ValidatedConfig` is maybe for right now basically `type ValidatedConfig = Config` but could eventually become something like `type ValidatedConfig = Required<Config>`?

Right now in this PR, which I put together without reflecting _super deeply_ on this stuff I have added an `UnvalidatedConfig` type which is `Loose<Config>`. It doesn't seem to me like we'd want to have Stencil users be exposed to an `UnvalidatedConfig` type since that would be a weird and confusing change from how they've been using the `Config` type until now - which makes me lean towards introducing a `ValidatedConfig` type instead which is only used internally within the compiler.

I'm not sure - what are your thoughts?

src/compiler/config/validate-testing.ts
alicewriteswrongs: Basically we don't (right now) have a type-level assurance that it's not null, but if you look at the `loadConfig` function you'll see that we basically try to read in the configuration file at `configPath` before we call the `validate-config` function with the output, and there is an early return if something goes wrong with reading the config. So basically by the time we're calling `validate-config` (and then subsequently `validate-testing`) we already know that this is present because we've been able to read a file at this path.

This is my understanding of how this works anyway!

src/compiler/config/validate-config.ts
alicewriteswrongs: yeah I was able to delete it without problems - I think I had the `as` cast there when I was experimenting with making some of the properties on `Config` required

src/compiler/config/config-utils.ts
```diff
@@ -1,17 +1,40 @@
...
+ *    on the config.
+ *
+ * @param config the config that we want to update
+ * @param configName the key we're setting on the config
+ * @param flagName either the name of a ConfigFlag prop we want to hoist up or null
+ * @param defaultValue the default value we should set!
+ */
+export const setBooleanConfig = <K extends keyof d.Config>(
+  config: d.UnvalidatedConfig,
+  configName: K,
```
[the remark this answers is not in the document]
alicewriteswrongs: nice! that does work

src/declarations/stencil-public-compiler.ts
```diff
@@ -2120,8 +2154,9 @@ export interface LoadConfigInit {
   initTsConfig?: boolean;
 }
 
+//
```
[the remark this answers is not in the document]
alicewriteswrongs: sure, more JSDoc is always good :)

src/compiler/config/validate-config.ts
```diff
@@ -12,17 +12,30 @@ import { validateRollupConfig } from './validate-rollup-config';
...
+} => {
   const config = Object.assign({}, userConfig || {}); // not positive it's json safe
   const diagnostics: Diagnostic[] = [];
 
   // copy flags (we know it'll be json safe)
   config.flags = JSON.parse(JSON.stringify(config.flags || {}));
 
   // default devMode false
-  if (config.flags.prod) {
+  if (config?.flags?.prod) {
```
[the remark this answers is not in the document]
alicewriteswrongs: yes, we can remove that (and on line 38) without getting any additional `strictNullChecks` error

src/compiler/config/validate-config.ts
```diff
@@ -48,7 +61,7 @@ export const validateConfig = (userConfig?: Config) => {
   setBooleanConfig(config, 'sourceMap', null, typeof config.sourceMap === 'undefined' ? false : config.sourceMap);
   setBooleanConfig(config, 'watch', 'watch', false);
   setBooleanConfig(config, 'buildDocs', 'docs', !config.devMode);
-  setBooleanConfig(config, 'buildDist', 'esm', !config.devMode || config.buildEs5);
+  setBooleanConfig(config, 'buildDist', null, !config.devMode || config.buildEs5);
```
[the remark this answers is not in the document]
alicewriteswrongs: I wasn't sure about that - this argument is supposed to be of type `keyof d.ConfigFlags` and `esm` is not a property on `ConfigFlags` so, assuming that `ConfigFlags` objects actually conform to that interface there shouldn't ever be a value under the `"esm"` key to pull out - I changed it for that reason, but it could be the case that we should keep it any, if we wanted to I think we'd need to add an `esm: boolean` prop to `ConfigFlags`

src/compiler/config/validate-dev-server.ts
alicewriteswrongs: I think you're right - I moved the `if (isNumber(flags?.port)) { ... }` block up above this `addressSplit` business so that we'll start with `flags.port` and then overwrite it with `addressSplit[1]` if it makes sense to do so

rwaskiewicz (changes requested):
Looking good! Mostly just a few optional chaining comments

src/compiler/config/validate-dev-server.ts
alicewriteswrongs: ok I went and re-did both the logic here and then also added a bunch of tests that passed with the original behavior (I filled in a few gaps with testing too, and de-duped some identical tests using `it.each` while I was in there).

Hopefully the new version 1) keeps the same behavior and 2) is easier to read and follow!

alicewriteswrongs:
@rwaskiewicz I think comments should all be addressed and this is ready for another review now (I updated it last week but forgot to ping you — sorry!)

src/compiler/config/test/validate-dev-server.spec.ts
```diff
@@ -24,23 +24,14 @@ describe('validateDevServer', () => {
...
-    expect(config.devServer.address).toBe('localhost');
-  });
-
-  it('should remove trailing / from address', () => {
-    inputConfig.devServer.address = 'localhost/';
-    const { config } = validateConfig(inputConfig);
-    expect(config.devServer.address).toBe('localhost');
-  });
+  it.each(['https://localhost', 'http://localhost', 'https://localhost/', 'http://localhost/', 'localhost/'])(
+    'should remove extraneious stuff from addres %p',
```
rwaskiewicz, line 28:
```suggestion
    'should remove extraneous stuff from address %p',
```

rwaskiewicz (approved):

rwaskiewicz (approved):

Issue #2859, opened 2021-03-19
Feature Request: Name space type names as needed rather than change them


kensodemann:
**Stencil version:**

```
 @stencil/core@2.0.1
```

**I'm submitting a:**

[ ] bug report
[x] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**
When components import types specific to them, sometimes exposed in public APIs, if any of these types happen to have a duplicate name, the generated types will rename conflicting imports successfully, but will only refer to one type in all component APIs that use the duplicate-named type.

For example, the following two components have a local same-named type `Name` used in a public method (boilerplate component code omitted):

```
// my-person.tsx
import { Name } from './resources';

@Method()
saySomething(name: Name): void {
  console.log(name);
  }

// my-pupper.tsx
import { Name } from './resources';

@Method()
saySomething(name: Name): void {
  console.log(name);
}
```

Which produces the following in components.d.ts

```
import { Name } from "./components/my-person/resources";
import { Name as Name1 } from "./components/my-pupper/resources";
export namespace Components {
  interface MyPerson {
    "saySomething": (name: Name) => Promise<void>;
  }
  interface MyPupper {
    "saySomething": (name: Name) => Promise<void>;
  }
}
```

**Expected behavior:**
Can the compiler use namespacing, inline the types or something else to avoid having to rename the types unnecessarily?

**Related code:**
Repo for repro case: https://github.com/jcfranco/stencil-duplicate-types-not-assigned-correctly-in-component-typings

PR #3337, opened 2022-04-18
fix(types): components.d.ts type resolution for duplicate types

rwaskiewicz:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

See the linked GH Issue below

GitHub Issue Number: https://github.com/ionic-team/stencil/issues/2859

## What is the new behavior?

this commit updates the type resolution of types used by components when
generating a `components.d.ts` file. types that have been deduplicated
are now piped to functions used to generate the types for class members
that are decorated with `@Prop()`, `@Event()`, and `@Method()`, and used
when generating the types for each.

- [ ] Yes
- [x] No

## Testing

Unit tests have been written + passed

Manually we can verify this against the reproduction repo:

First, let's get the baseline setup:
```
git clone https://github.com/jcfranco/stencil-duplicate-types-not-assigned-correctly-in-component-typings sample
cd sample
npm i && git commit -am 'baseline after npm install'
npm i @stencil/core@latest
git commit -am 'install latest stencil version, 2.15.0'
npm run build
git commit -am 'build with stencil v2.15.0'
git add *
git commit -m 'build with stencil v2.15.0'
```

In the generated `src/components.d.ts` file, we can confirm the bug:
```ts
import { Name } from "./components/my-person/resources";
import { Name as Name1 } from "./components/my-pupper/resources";
export namespace Components {
    interface MyComponent {
        "saySomething": (name: Name) => Promise<void>;
    }
    interface MyPupper {
        "saySomething": (name: Name) => Promise<void>;
    }
}
```
Note how `MyPupper`'s `saySomething` method uses `Name`, not `Name1`.

By checking out this branch, running `npm ci && npm build && npm pack`, we can generated a tarball with the changes included in this PR, we can install the tarball in the reproduction repo + rebuild
```
npm i <PATH_TO_TAR>
npm run build
```
which generates:
```ts
import { Name } from "./components/my-person/resources";
import { Name as Name1 } from "./components/my-pupper/resources";
export namespace Components {
    interface MyComponent {
        "saySomething": (name: Name) => Promise<void>;
    }
    interface MyPupper {
        "saySomething": (name: Name1) => Promise<void>;
    }
}
```

### DIFF
--- a/src/compiler/types/stencil-types.ts
+++ b/src/compiler/types/stencil-types.ts
@@ -1,5 +1,5 @@
 import type * as d from '../../declarations';
-import { dirname, join, relative } from 'path';
+import { dirname, join, relative, resolve } from 'path';
 import { isOutputTargetDistTypes } from '../output-targets/output-utils';
 import { normalizePath } from '@utils';
 
@@ -28,6 +28,81 @@ export const updateStencilTypesImports = (typesDir: string, dtsFilePath: string,
     dtsContent = dtsContent.replace(/(from\s*(:?'|"))@stencil\/core('|")/g, `$1${coreDtsPath}$2`);
   }
   return dtsContent;
+};
+
+/**
+ * Utility for ensuring that naming collisions do not appear in type declaration files for a component's class members
+ * decorated with @Prop, @Event, and @Method
+ * @param typeReferences all type names used by a component class member
+ * @param typeImportData locally/imported/globally used type names, which may be used to prevent naming collisions
+ * @param sourceFilePath the path to the source file of a component using the type being inspected
+ * @param initialType the name of the type that may be updated
+ * @returns the updated type name, which may be the same as the initial type name provided as an argument to this
+ * function
+ */
+export const updateTypeIdentifierNames = (
+  typeReferences: d.ComponentCompilerTypeReferences,
+  typeImportData: d.TypesImportData,
+  sourceFilePath: string,
+  initialType: string
+): string => {
+  let currentTypeName = initialType;
+
+  // iterate over each of the type references, as there may be >1 reference to inspect
+  for (let typeReference of Object.values(typeReferences)) {
+    const importResolvedFile = getTypeImportPath(typeReference.path, sourceFilePath);
+
+    if (!typeImportData.hasOwnProperty(importResolvedFile)) {
+      continue;
+    }
+
+    for (let typesImportDatumElement of typeImportData[importResolvedFile]) {
+      currentTypeName = updateTypeName(currentTypeName, typesImportDatumElement);
+    }
+  }
+  return currentTypeName;
+};
+
+/**
+ * Determine the path of a given type reference, relative to the path of a source file
+ * @param importResolvedFile the path to the file containing the resolve type. may be absolute or relative
+ * @param sourceFilePath the component source file path to resolve against
+ * @returns the path of the type import
+ */
+const getTypeImportPath = (importResolvedFile: string, sourceFilePath: string): string => {
+  const isPathRelative = importResolvedFile && importResolvedFile.startsWith('.');
+  if (isPathRelative) {
+    importResolvedFile = resolve(dirname(sourceFilePath), importResolvedFile);
+  }
+
+  return importResolvedFile;
+};
+
+/**
+ * Determine whether the string representation of a type should be replaced with an alias
+ * @param currentTypeName the current string representation of a type
+ * @param typeAlias a type member and a potential different name associated with the type member
+ * @returns the updated string representation of a type. If the type is not updated, the original type name is returned
+ */
+const updateTypeName = (currentTypeName: string, typeAlias: d.TypesMemberNameData): string => {
+  if (!typeAlias.importName) {
+    return currentTypeName;
+  }
+
+  // TODO(STENCIL-419): Update this functionality to no longer use a regex
+  // negative lookahead specifying that quotes that designate a string in JavaScript cannot follow some expression
+  const endingStrChar = '(?!("|\'|`))';
+  /**
+   * A regular expression that looks at type names along a [word boundary](https://www.regular-expressions.info/wordboundaries.html).
+   * This is used as the best approximation for replacing type collisions, as this stage of compilation has only
+   * 'flattened' type information in the form of a String.
+   *
+   * This regex should be expected to capture types that are found in generics, unions, intersections, etc., but not
+   * those in string literals. We do not check for a starting quote (" | ' | `) here as some browsers do not support
+   * negative lookbehind. This works "well enough" until STENCIL-419 is completed.
+   */
+  const typeNameRegex = new RegExp(`${typeAlias.localName}\\b${endingStrChar}`, 'g');
+  return currentTypeName.replace(typeNameRegex, typeAlias.importName);
 };
 
 /**
--- a/src/compiler/types/stencil-types.ts
+++ b/src/compiler/types/stencil-types.ts
@@ -69,7 +69,7 @@ export const updateTypeIdentifierNames = (
  * @param sourceFilePath the component source file path to resolve against
  * @returns the path of the type import
  */
-const getTypeImportPath = (importResolvedFile: string, sourceFilePath: string): string => {
+const getTypeImportPath = (importResolvedFile: string | undefined, sourceFilePath: string): string => {
   const isPathRelative = importResolvedFile && importResolvedFile.startsWith('.');
   if (isPathRelative) {
     importResolvedFile = resolve(dirname(sourceFilePath), importResolvedFile);
--- a/src/compiler/types/tests/stencil-types.spec.ts
+++ b/src/compiler/types/tests/stencil-types.spec.ts
@@ -1,0 +1,298 @@
+import * as d from '@stencil/core/declarations';
+import path from 'path';
+import { stubComponentCompilerMeta } from './ComponentCompilerMeta.stub';
+import { stubComponentCompilerTypeReference } from './ComponentCompilerTypeReference.stub';
+import { stubTypesImportData } from './TypesImportData.stub';
+import { updateTypeIdentifierNames } from '../stencil-types';
+
+describe('stencil-types', () => {
+  describe('updateTypeMemberNames', () => {
+    let dirnameSpy: jest.SpyInstance<ReturnType<typeof path.dirname>, Parameters<typeof path.dirname>>;
+    let resolveSpy: jest.SpyInstance<ReturnType<typeof path.resolve>, Parameters<typeof path.resolve>>;
+
+    beforeEach(() => {
+      dirnameSpy = jest.spyOn(path, 'dirname');
+      dirnameSpy.mockImplementation((path: string) => path);
+
+      resolveSpy = jest.spyOn(path, 'resolve');
+      resolveSpy.mockImplementation((...pathSegments: string[]) => pathSegments.pop());
+    });
+
+    afterEach(() => {
+      dirnameSpy.mockRestore();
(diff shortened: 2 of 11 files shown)


alicewriteswrongs:
just tested this out and was able to 1) repro the issue with the repo provided and 2) confirm that this PR does fix it. reading through the code now!

src/compiler/types/stencil-types.ts
```diff
@@ -30,6 +30,81 @@ export const updateStencilTypesImports = (typesDir: string, dtsFilePath: string,
...
+
+  // iterate over each of the type references, as there may be >1 reference to inspect
+  for (let typeReference of Object.values(typeReferences)) {
+    const importResolvedFile = getTypeImportPath(typeReference.path, sourceFilePath);
+
+    if (!typeImportData.hasOwnProperty(importResolvedFile)) {
+      continue;
+    }
+
+    for (let typesImportDatumElement of typeImportData[importResolvedFile]) {
```
alicewriteswrongs, line 59: understanding check: here we are iterating through all the type import information for _all previously processed components_ and running the name update for the name we're looking at right now (`initialType`) in order to ensure it doesn't clash with anything already imported.

Am I right that because we check all of them, we'll be able to handle a clash with _n_ other type names, not just one? Something like

```ts
import { Name } from './moduleA'
import { Name as Name1 } from './moduleB'
import { Name as Name2 } from './moduleC'
```

It seems like as we iterate through to update an import like `import { Name } from './moduleD'` we'd find the info about importing `Name` and update our identifier to `Name1`, then find the info for `Name as Name1` and update our identifier to `Name2`, and so until we settle on adding a line like

```ts
import { Name as Name3 } from './moduleD'
```

in order to not clash with any of the previous imports.

Is that correct?

src/compiler/types/stencil-types.ts
```diff
@@ -30,6 +30,81 @@ export const updateStencilTypesImports = (typesDir: string, dtsFilePath: string,
...
+      currentTypeName = updateTypeName(currentTypeName, typesImportDatumElement);
+    }
+  }
+  return currentTypeName;
```
alicewriteswrongs, line 63: one thing I'm wondering - this function is going to be called for every component prop, every event, and so on, and then for each of those we're doing two nested for loops. Are we worried about the performance impact of doing that? Do you think it's worthwhile / necessary to do a little measurement to see if there's an impact? When I see a `for ... of` within a `for ... of` and then realize that this function is itself called within a loop I wonder, but I don't feel like I have enough context to know how the things we're looping through are going to be. So possibly not a concern, but wanted to flag it to see your thoughts and make sure I'm understanding what's going on.

alicewriteswrongs, line 63: one thing I'm wondering - this function is going to be called for every component prop, every event, and so on, and then for each of those we're doing two nested for loops. Are we worried about the performance impact of doing that? Do you think it's worthwhile / necessary to do a little measurement to see if there's an impact? When I see a `for ... of` within a `for ... of` and then realize that this function is itself called within a loop I wonder, but I don't feel like I have enough context to know how the things we're looping through are going to be. So possibly not a concern, but wanted to flag it to see your thoughts and make sure I'm understanding what's going on.

alicewriteswrongs (commented):
I think that looking through everything pretty much looks good, I noticed like just a few really small things that we might want to change but otherwise things look relatively straightforward.

One thing that I want to do is sort of check my understanding of what's going on here to make sure I'm getting what we're changing / how we're fixing the issue with duplicate type names.

So here's my understanding of where this change intervenes in the Stencil compilation process and what we're doing.

### What's the problem?

When we're compiling a stencil project one of the things we output is a `components.d.ts` file which includes type declarations for all the components in the project.[^1]

Since this file contains type information pulled from all of our separate component source files, which can be thought of as separate type namespaces or contexts, there is a possibility of a name clash if:

1. two different modules import different types from elsewhere and either one renames the import to clash with the name in the other module, they both rename the import to the same thing, or something along those lines
2. two modules each export a type with the same name which are each in turn imported in (separate) component files[^2].

If we hit either of these scenarios it will lead to a situation in which we're referencing the wrong type. I don't think this will have a runtime effect but any typescript project which is importing the components and trying to use them will have errors.

### How do we fix it?

So the way to fix this is that when we are merging together the type information from our separate component source files we need to keep track of the global imports and type references in order to check that a name that we're going to use in a particular component's portion of the `components.d.ts` file _has not already been used by another component_. If it has been, we need to do something like `Name -> Name1` to ensure that the second component's type declaration does not erroneously reference a type that was pulled in to `components.d.ts` by another component.

Making this work requires edits to the function responsible for getting type information for a component (`generateComponentTypes`) and the sub-functions to which it delegates (`generatePropTypes`, `generateEventTypes`, and so on). The edit we need to make to each of these little functions is pretty much the same - they all need to be able to look at the `TypesImportData` object that we update w/ everything we're going to be importing into `components.d.ts` in order to see if the name of a type imported by the component in question needs to be updated to not clash with other imports.

This work is done in the `updateTypeIdentifierNames` function based on information already returned from the `updateReferenceTypeImports` function.

`updateReferenceTypeImports` basically puts together a map from filepaths (of source files) to an array of all the types that are found within them. The objects in that array (of type `TypesMemberNameData`) have two properties, `localName` and `importName`. Local name is what they are called in their home modules, and `importName` is a disambiguated name (like `Name1` in the repro example) that will be safe to import in our mushed-together `components.d.ts` namespace.

`updateReferenceTypeImports` will basically take that `TypesImportData` object that we get back from `updateReferenceTypeImports` and the `ComponentCompilerTypeReferences` for a given component. Then it looks through the reference types and, for each one:

1. figures out what file path the reference is from
2. looks at `TypesImportData` in order to get a new, `components.d.ts` namespace-safe name for the reference in question, since `TypesImportData` is already set up that way.
5. Create a regular expression to update imports for a possibly-clashing type to a non-clashing type.

Then of course we run those regexes and the imports get fixed. Then all this stuff gets returned back to the `generateComponentTypesFile` in the same way it did before this change.

Anyway — sorry for the essay! I was just thinking that it's hard to review something like this if I don't know what's going on, so I wanted to really make sure I understand what we're trying to do, and how this fixes the issue we are seeing. Let me know where I'm missing things, or if there's just any other context that would be helpful for understanding what's going on here.


[^1]: It's a secondary question w/r/t this PR but I'm not clear on why we need to generate this file ourselves and why TypeScript can't do so. I feel like I get _what we're doing_ but some part of the _why_ I haven't discovered yet.
[^2]: I believe this was the scenario in the reproduction repo.

src/compiler/types/stencil-types.ts
rwaskiewicz: > Are we worried about the performance impact of doing that?

A little, yes. Specifically, I'm concerned about the cost of these two lines in `updateTypeName`:
```ts
  const typeNameRegex = new RegExp(`${typeAlias.localName}\\b${endingStrChar}`, 'g');
  return currentTypeName.replace(typeNameRegex, typeAlias.importName);
```
where we build this regex and run `replace` globally on `currentTypeName`. I tried to add as many cases where we can avoid this work as possible (if there's no alias to replace a collision with, if we don't see any collisions at all, etc.), but we still see something like `O(m * n * p)` performance where:
- `m` is the size of `ComponentCompilerTypeReferences` 
- `n` is the size of `TypesImportData` 
- `p` is the length of the string type of the collision (maybe, the perf implications may be worse here with allocating memory for a new string)

My hope is the `m` and `n` are relatively small since we evaluate this on a per component (and therefore per file) basis. `p` is a bit trickier to optimize ATM, and is going to require some rework that I planned in STENCIL-418+STENCIL-419 (we'd need to move some of this cost to allocating/storing the actual types instead of flattening them to strings).

src/compiler/types/tests/stencil-types.spec.ts
```diff
@@ -0,0 +1,298 @@
+import * as d from '@stencil/core/declarations';
+import path from 'path';
+import { stubComponentCompilerMeta } from './ComponentCompilerMeta.stub';
+import { stubComponentCompilerTypeReference } from './ComponentCompilerTypeReference.stub';
+import { stubTypesImportData } from './TypesImportData.stub';
+import { updateTypeIdentifierNames } from '../stencil-types';
+
+describe('stencil-types', () => {
```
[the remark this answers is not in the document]
rwaskiewicz: 🎉 🎉 🎉

rwaskiewicz:
@alicewriteswrongs 

You're understanding of what's going on here is spot on. No worries for the essay! 😆 

One thing that's probably worth clarifying:

> Since this file contains type information pulled from all of our separate component source files, which can be thought of as separate type namespaces or contexts, there is a possibility of a name clash if:

 > 1. two different modules import different types from elsewhere and either one renames the import to clash with the name in the other module, they both rename the import to the same thing, or something along those lines

This is correct, although Stencil doesn't handle aliasing in import statements correctly 100% of the time. I added a reproduction case/bug to the backlog as STENCIL-429 for us to work through cases where `import { SomeType as YetAnotherType } from './my-types';` isn't always handled correctly

src/compiler/types/stencil-types.ts
alicewriteswrongs: Ok that all makes sense, I was thinking that given that we're going through a component at a time it shouldn't be a _huge_ issue, but if stencil users have some enormous components with 30 or 40 imports we might see some real impact there.

Incidentally, this seems like exactly the type of situation where some automated performance testing against a large stencil project would be super handy!

alicewriteswrongs:
The thought occurred to me, what if in `ComponentA.tsx` I do `import { Name as MyName } from './libA'` and then in `ComponentB.tsx` I do `import { Name as MyName } from './libB'`, since it looked like the code was not set up to necessarily handle that sort of collision. Well, something to fix in the future I guess!

src/compiler/types/tests/stencil-types.spec.ts
```diff
@@ -0,0 +1,298 @@
...
+       * This helper asserts that the generated type matches T'.
+       *
+       * This helper is meant to be used to test that transformation occurs based on the format/type of `basePath`
+       * provided (e.g. relative vs absolute).
+       *
+       * @param basePath the path to use when running the type resolution
+       * @param initialType the original type found in a class member
+       * @param expectedType the type that is expected to be generated
+       */
+      const testTypeTransformForPath = (basePath: string, initialType: string, expectedType: string): void => {
```
alicewriteswrongs, line 139: sort of a minor thing, but it might be slightly more readable to move this function earlier in the block, up above where it's used, so the reader is encouraged to at least glance at it before reading the tests (I was thinking the same thing for `testTypeIsTransformed` below)

alicewriteswrongs (approved):
only really have one take-it-or-leave-it comment, other than that I think everything looks good!

src/compiler/types/tests/stencil-types.spec.ts
rwaskiewicz: I think it could go either way - I generally prefer to order functions in the order they're called, but that's more of a subjective opinion (at least in this scenario in JavaScript where order doesn't affect the execution).  My vote for now would be to leave it as it for now.  In the future I'd like to get ESLint set up in the codebase, and then we can let it warn us when we violate rules (we'll still have to decide/agree on the rules we run though 😉)

alicewriteswrongs: I did check earlier today out of curiousity and adding eslint w/ basic rules doesn't produce _that_ many errors:

Screen Shot 2022-04-22 at 1 16 33 PM

most of them are 'hey you used `any`, don't do that!'

PR #3349, opened 2022-04-29
test(bundler): add component library for testing

rwaskiewicz:
## Important

~This is going to be a part of a chain of PRs. Once all of them are up/approved, I'll merge the child-most PR into its parent, that parent into its own parent, etc. At the end, we shall have only this PR which contains all of its children, and we'll commit it as one large commit.~

~As a result, this doesn't have all of the functionality necessary for STENCIL-338 to be complete yet - that'll be incrementally added as we merge the child PRs~

This PRs has all commits. Once documentation is approved, we'll rebase & squash it to merge it

## Pull request checklist

- [x] Docs have been reviewed and added / updated if needed (for bug fixes / features)- https://github.com/ionic-team/stencil-site/pull/876

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Other (please describe): Testing infrastructure for a bug fix

Stencil does not support component libraries being built with the `dist` output target,
then used in an application using a bundler such vite, or parcel

GitHub Issue Number: N/A

## What is the new behavior?

add a stencil component library to be used in tests that verify
applications that consume the library and are bundled with vite, parcel,
etc.

this commit does not introduce any tests themselves, but rather is
boilerplate code that will be used in future commits (the intent is to
stack multiple pull requests atop this one)

the second commit wires the bundler tests to github actions. these tests are
kept in a new reusable workflow that can run in parallel with existing
analysis, unit and e2e tests

- [ ] Yes
- [x] No

## Testing

This commit itself doesn't have much to test. It's going to be used as a part of downstream commits, which exercise it more thoroughly.  I did run the npm scripts that are associated with this PR to verify they worked as intended/am tailing the logs to verify github actions work correctly

## Other Information

Since this is a part of a chain, please do not use GH to 'Update Branch' - I'll take care of that manually (and be careful not to rebase mid-review cycle)


.github/workflows/test-bundlers.yml
```diff
@@ -0,0 +1,28 @@
...
+
+on:
+  workflow_call:
+    # Make this a reusable workflow, no value needed
+    # https://docs.github.com/en/actions/using-workflows/reusing-workflows
+
+jobs:
+  bundler_tests:
+    name: Verify Bundlers
+    runs-on: 'ubuntu-latest'
```
alicewriteswrongs, line 11: just wondering - is it total overkill / not necessary to test this across the `node` and `os` matrix like we are doing for the e2e tests and whatnot?

alicewriteswrongs (approved):
this all looks straightforward enough! just had one small clarifying question

.github/workflows/test-bundlers.yml
rwaskiewicz: I think so - although we need to validate runtime behavior (the component is lazily loaded correctly), I don't think it's pertinent that we test this across all Node versions/OS's.

colinbate:
Great to see this functionality get merged in. Will we see a release cut soon with this included?

rwaskiewicz:
@colinbate This feature was a part of the v2.16.0 release, which we cut today.  To get started using it, please see the documentation around the [experimentalimportinjection](https://stenciljs.com/docs/config-extras#experimentalimportinjection) flag.  After setting this flag to `true` in your `stencil.config.ts` file and recompiling your project, you will be able to take advantage of the changes in this PR

TRMW:
@rwaskiewicz Thanks for adding this! We are considering using this in a production app using Vite, but are a little concerned that the feature will be removed in the future, give the `experimental` prefix. Do you have a sense for how committed Stencil is to maintaining this feature? Is there a timeline for when/how it would be made _not_ experimental?

rwaskiewicz:
@TRMW the experimental prefix is there so we can try to warn folks around some sharp edges they may find in using this functionality.  While I can't promise we'll support this functionality until the end of time, I don't have any intent on removing it in the future.

TRMW:
@rwaskiewicz Great, thanks for the context! Is the plan to remove the experimental prefix once the "sharp edges" you referenced are no longer present?

rwaskiewicz:
@TRMW - yes, although IIRC there were less of them then we thought we'd find. It's a bit of a chicken & egg problem - the more people using a feature like this will probably reveal more issues (since folks work on projects with different setups/configurations), but the 'experimental' naming can deter them from adopting it (making it harder to find those edge cases).

We're probably in a place where we can drop the 'experimental' part of the flag, although that naming won't go away completely until we release a new major version of Stencil (since it is technically a breaking change).  In the mean time, we could probably update the code to accept both `experimentalImportInjection` and `importInjection` 🤔

wilcoxmd:
👋 Hi @rwaskiewicz, I work with @TRMW. Great to hear that there have been few issues with import injection, and that it seems ready to remove the 'experimental' prefix! Was the primary concern bundle size increases, or are there other things we should be aware of when using this flag? 

> In the mean time, we could probably update the code to accept both `experimentalImportInjection` and `importInjection` 

I think this would make sense to do if you're ready to drop the prefix. It's not a breaking change, and clearly signals intent to support the feature.

rwaskiewicz:
@wilcoxmd 👋 

> Was the primary concern bundle size increases, or are there other things we should be aware of when using this flag?

A little of both. We knew bundle sizes would increase a bit, and as far as any other things, none that we know of at this time.  While we don't _think_ there should be any sharp edges moving forward, it's tough to say for certain given the number of tools/libraries folks use to build their projects and how they may interact with one another.

wilcoxmd:
@rwaskiewicz Thanks, that all makes sense and is great to know. I appreciate the extra context around this and your quick responses. 🙏

PR #3356, opened 2022-05-05
add strictNullChecks CI check

alicewriteswrongs:
## Pull request checklist

## Pull request type

This is basically a CI / tooling PR which adds a CI action that runs `tsc` with `--strictNullchecks` on both the branch for a given PR and on `main` and reports via a comment on the PR 1) differences in error count between the two and 2) some statistics about errors on the PR (which files have the most, etc).

The idea is that this will firstly discourage us from checking in code which _adds_ null check errors, and then also encourage us to make changes which decrease the number of errors. We currently have 2188, which while kind of a lot is not _so_ many, and a few dozen PRs which each clear up 50-100 or so will get us to where we want to be. So the hope is that by increasing visibility with this CI action we'll keep on improving things in that regard.

This check is intended to be removed in the future when we fix all the errors and are then in a position to turn `strictNullChecks` on all the time. We could also use it for enabling other compiler options which increase the error count as well.

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Other (please describe): CI!

GitHub Issue Number: N/A

## What is the new behavior?

Now PRs will have a check run which 1) gets the errors on `main` and 2) on the PR and then prints a report out in a comment (check out this PR for an example of what that looks like).

Here's what it looks like if you decrease the error count:

Screen Shot 2022-05-06 at 12 22 50 PM

If you _increase_ it:

Screen Shot 2022-05-06 at 12 32 49 PM

and, finally, if it stays the same:

Screen Shot 2022-05-06 at 12 29 40 PM

**NOTE** This branch is set up right now to _both_ set the [workflow job summary](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#adding-a-job-summary) and also comment on the PR with the same Markdown output. The workflow output thingy is nice but I don't think it's nearly as visible as a comment on the PR, so if we want this to be something we're actively reminded of fairly consistently then I think we might want to keep both.

If we decide we want it to be more quiet then the job summary is definitely the more elegant option.

- [ ] Yes
- [x] No

### DIFF
--- a/.github/workflows/strict-null-checks.yml
+++ b/.github/workflows/strict-null-checks.yml
@@ -24,4 +24,4 @@ jobs:
         uses: ./.github/workflows/actions/get-core-dependencies
 
       - name: Run Typescript compiler
-        uses: npx tsc --noEmit --strictNullChecks --pretty false
+        run: npx tsc --noEmit --strictNullChecks --pretty false
--- a/.github/workflows/strict-null-checks.yml
+++ b/.github/workflows/strict-null-checks.yml
@@ -57,27 +57,27 @@ jobs:
       - name: Set action output
         run: GITHUB_STEP_SUMMARY=$(node scripts/strict-null-checks-ci.js)
 
-      - name: Set comment body
-        id: set-comment-body
-        run: |
-          body=$(node scripts/strict-null-checks-ci.js)
-          body="${body//'%'/'%25'}"
-          body="${body//$'\n'/'%0A'}"
-          body="${body//$'\r'/'%0D'}"
-          echo ::set-output name=body::$body
+#       - name: Set comment body
+#         id: set-comment-body
+#         run: |
+#           body=$(node scripts/strict-null-checks-ci.js)
+#           body="${body//'%'/'%25'}"
+#           body="${body//$'\n'/'%0A'}"
+#           body="${body//$'\r'/'%0D'}"
+#           echo ::set-output name=body::$body
 
-      - name: Find Comment
-        uses: peter-evans/find-comment@v2
-        id: fc
-        with:
-          issue-number: ${{ github.event.pull_request.number }}
-          comment-author: 'github-actions[bot]'
-          body-includes: '### `--strictNullChecks` error report'
+#       - name: Find Comment
+#         uses: peter-evans/find-comment@v2
+#         id: fc
+#         with:
+#           issue-number: ${{ github.event.pull_request.number }}
+#           comment-author: 'github-actions[bot]'
+#           body-includes: '### `--strictNullChecks` error report'
 
-      - name: Create or update comment
-        uses: peter-evans/create-or-update-comment@v2
-        with:
-          comment-id: ${{ steps.fc.outputs.comment-id }}
-          issue-number: ${{ github.event.pull_request.number }}
-          body: ${{ steps.set-comment-body.outputs.body }}
-          edit-mode: replace
+#       - name: Create or update comment
+#         uses: peter-evans/create-or-update-comment@v2
+#         with:
+#           comment-id: ${{ steps.fc.outputs.comment-id }}
+#           issue-number: ${{ github.event.pull_request.number }}
+#           body: ${{ steps.set-comment-body.outputs.body }}
+#           edit-mode: replace
--- a/.github/workflows/strict-null-checks.yml
+++ b/.github/workflows/strict-null-checks.yml
@@ -3,7 +3,7 @@ name: Strict Null Checks Burndown
 # from typescript's strictNullChecks option. Once we get to 0
 # we can turn that option on all the time and delete this job!
 #
-# TODO (STENCIL-446): Remove this workflow once `strictNullChecks` is enabled
+# TODO(STENCIL-446): Remove this workflow once `strictNullChecks` is enabled
 
 on:
   pull_request:
--- a/.github/workflows/strict-null-checks.yml
+++ b/.github/workflows/strict-null-checks.yml
@@ -55,7 +55,7 @@ jobs:
           name: null_errors_main
 
       - name: Set action output
-        run: $GITHUB_STEP_SUMMARY=$(node scripts/strict-null-checks-ci.js)
+        run: GITHUB_STEP_SUMMARY=$(node scripts/strict-null-checks-ci.js)
 
       - name: Set comment body
         id: set-comment-body
--- a/.github/workflows/strict-null-checks.yml
+++ b/.github/workflows/strict-null-checks.yml
@@ -24,7 +24,7 @@ jobs:
         uses: ./.github/workflows/actions/get-core-dependencies
 
       - name: Run Typescript compiler
-        run: npx tsc --noEmit --strictNullChecks --pretty false | npx tsc-output-parser > errors.json
+        run: npm run tsc.strictNullChecksReport
 
       - name: Set comment body
         id: set-comment-body
--- a/.github/workflows/strict-null-checks.yml
+++ b/.github/workflows/strict-null-checks.yml
@@ -17,8 +17,6 @@ jobs:
       # there, save the result
       - name: Checkout main
         uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # v2.4.0
-        with:
-          ref: main
 
       - name: Get Core Dependencies
         uses: ./.github/workflows/actions/get-core-dependencies
--- a/.github/workflows/strict-null-checks.yml
+++ b/.github/workflows/strict-null-checks.yml
@@ -24,4 +24,29 @@ jobs:
         uses: ./.github/workflows/actions/get-core-dependencies
 
       - name: Run Typescript compiler
-        run: npx tsc --noEmit --strictNullChecks --pretty false
+        run: npx tsc --noEmit --strictNullChecks --pretty false | npx tsc-output-parser > errors.json
+
+      - name: Set comment body
+        id: set-comment-body
+        run: |
+          body=$(node scripts/strict-null-checks-ci.js)
+          body="${body//'%'/'%25'}"
+          body="${body//$'\n'/'%0A'}"
+          body="${body//$'\r'/'%0D'}"
+          echo ::set-output name=body::$body
+
+      - name: Find Comment
+        uses: peter-evans/find-comment@v2
+        id: fc
+        with:
+          issue-number: ${{ github.event.pull_request.number }}
+          comment-author: 'github-actions[bot]'
+          body-includes: '### `--strictNullChecks` error report"'
+
+      - name: Create or update comment
+        uses: peter-evans/create-or-update-comment@v2
+        with:
+          comment-id: ${{ steps.fc.outputs.comment-id }}
+          issue-number: ${{ github.event.pull_request.number }}
+          body: ${{ steps.set-comment-body.outputs.body }}
+          edit-mode: replace
--- a/.github/workflows/strict-null-checks.yml
+++ b/.github/workflows/strict-null-checks.yml
@@ -21,10 +21,8 @@ jobs:
       - name: Get Core Dependencies
         uses: ./.github/workflows/actions/get-core-dependencies
 
-      # this is only necessary while tsc-output-parser isn't a dependency in main
-      # once it is, we can introduce a PR to remove this step
       - name: Install tsc-output-parser
-        run: npm install @aivenio/tsc-output-parser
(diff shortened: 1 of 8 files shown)


alicewriteswrongs:
test comment

.github/workflows/strict-null-checks.yml
```diff
@@ -0,0 +1,80 @@
+name: Strict Null Checks Burndown
+# this is a temporary workflow while we burn down the errors
+# from typescript's strictNullChecks option. Once we get to 0
+# we can turn that option on all the time and delete this job!
+
+on:
+  pull_request:
```
alicewriteswrongs, line 7: I wrote these jobs in a separate file (instead of in `main.yml`) because we have that file running on

```yml
on:
  push:
    branches:
      - 'main'
      - 'v3.0.0-dev'
  pull_request:
    branches:
      - '**'
```

for this workflow I think we only want it to run on a PR, since we need to have a PR on which to write our little comment report out.

.github/workflows/strict-null-checks.yml
```diff
@@ -0,0 +1,80 @@
...
+        with:
+          name: null_errors_main
+
+      - name: Set comment body
+        id: set-comment-body
+        run: |
+          body=$(node scripts/strict-null-checks-ci.js)
+          body="${body//'%'/'%25'}"
+          body="${body//$'\n'/'%0A'}"
+          body="${body//$'\r'/'%0D'}"
```
alicewriteswrongs, line 63: see https://github.com/peter-evans/create-or-update-comment#setting-the-comment-body-from-a-file

package.json
```diff
@@ -56,6 +56,7 @@
     "test.watch-all": "jest --watchAll --coverage"
   },
   "devDependencies": {
+    "@aivenio/tsc-output-parser": "^2.1.1",
```
alicewriteswrongs, line 59: I added this as a dev dependency -- however, we could simply install this package manually in the action if we want to avoid increasing the number of dependencies

scripts/strict-null-checks-ci.js
```diff
@@ -0,0 +1,170 @@
+/**
```
alicewriteswrongs, line 1: what is the best way to deal with a script like this? I wrote it in JS to avoid have to deal with some sort of build step before being able to just run it in the action, but we could write it in typescript instead and either execute it with `ts-node` or do a quick `tsc` on it.

My hope is that this is something that only sticks around for a few months before we're able to delete it, so I'm inclined to keep things lightweight and uncomplicated

alicewriteswrongs (commented):
just some things I wanted to highlight

scripts/strict-null-checks-ci.js
```diff
@@ -0,0 +1,170 @@
...
+const fs = require('fs');
+
+/**
+ * Load JSON data, formatted by `tsc-output-parser`
+ *
+ * In the GH Actions workflow we write two json files, one for the PR
+ * branch and one for main (so that we can compare them).
+ *
+ */
+const prData = JSON.parse(String(fs.readFileSync('./null_errors_pr.json')));
```
rwaskiewicz, line 15: Should we wrap the `fs.readFileSync` calls here in a try/catch in the event that reading the file(s) fails? I suppose that if we fail to read the file, the workflow _should_ fail (and it's non-blocking on the PR landing).  I'm just not sure what the DX is like if we hit a failure here.  Do you know if the stacktrace makes it clear enough which file read failed?

scripts/strict-null-checks-ci.js
```diff
@@ -0,0 +1,170 @@
...
+const mainData = JSON.parse(String(fs.readFileSync('./null_errors_main.json')));
+
+const errorsOnMain = mainData.length;
+const errorsOnPR = prData.length;
+
+/**
+ * Build up an object which maps error codes (like `TS2339`) to
+ * a `Set` of all the messages we see for that code
```
rwaskiewicz, line 23: Minor, can we add `@param` & `@returns` jsdoc for this files functions?

scripts/strict-null-checks-ci.js
```diff
@@ -0,0 +1,170 @@
...
+
+  return errorCodeMessageMap;
+};
+
+const errorCodeMessages = getErrorCodeMessages();
+
+/**
+ * Build a map which just counts the entries in an array
+ */
+const countArrayEntries = (xs) => {
```
rwaskiewicz, line 42: Minor, can we give this parameter a more descriptive name than `xs`? I'm not sure what it's supposed to represent at first glance

scripts/strict-null-checks-ci.js
```diff
@@ -0,0 +1,170 @@
...
+      .forEach(([path, errorCount]) => {
+        out.push(tableRow(path, errorCount));
+      });
+  })
+);
+
+lines.push('');
+
+// then we can do something similar with errors, building a table with the
+// TypeScript error code, the number of occurences, and a collapsed section
```
rwaskiewicz, line 141:
```suggestion
// TypeScript error code, the number of occurrences, and a collapsed section
```

scripts/strict-null-checks-ci.js
rwaskiewicz: The `scripts` directory normally gets transpiled as a part of the `npm run build` command (we do this on every build to ensure the tooling in `scripts` is always up to date).  If we made this a `.ts` file, it'd get compiled out to `scripts/build/`. That does present a bit of a chicken/egg problem, since we're doing this step outside of the `build` step in CI to get that parallelism.

I think there are 4 options here:

## 1) Make this file a `.ts` file, compile the `scripts` directory as a part of this workflow

Compiling the `scripts` directory is a relatively quick process (compared to type checking the compiler & running rollup on everything).  We could run tsc across the directory (`node scripts --prepare`), then point the job to run `scripts/build/strict-null-checks-ci.js`

This would add a little length of time to the job - the output on my machine for `time node scripts --prepare) is ~5 seconds, and would probably longer in GH's hosted runners (especially the Windows ones 🙃). But, it'd give us type safety "for free".

## 2) Use ts-node

I don't see why this _couldn't_ work. We'd probably need to `npm install` it as a part of the job and make sure its executable - off the top of my head that's the most work we'd have to do.  I'd prefer option 1 over this to keep things 'simple' (i.e. not spin up another way we transpile/run things).

## 3) Use `// @ts-check` directive

We could add the `// @ts-check` directive to the top of the file, and annotate it with types.  It's "nice" in that we don't have to set up any compilation steps, but it's a bit tedious to do and the type checking isn't nearly as strong as using tsc (and editor dependent insofar as how 'loud' you make TS yell at you).  I added an example patch of what this might look like (I had to make it a .txt file, because you can't upload `.patch` files to GH in comments 🤣) - [example-patch.txt](https://github.com/ionic-team/stencil/files/8679049/example-patch.txt)

## 4) Do Nothing/Leave it as is

Probably my least favorite option. My personal preference would be to get type safety from the get-go.

.github/workflows/strict-null-checks.yml
rwaskiewicz: 👍 I like it!

.github/workflows/strict-null-checks.yml
```diff
@@ -0,0 +1,83 @@
...
+      - name: Find Comment
+        uses: peter-evans/find-comment@v2
+        id: fc
+        with:
+          issue-number: ${{ github.event.pull_request.number }}
+          comment-author: 'github-actions[bot]'
+          body-includes: '### `--strictNullChecks` error report'
+
+      - name: Create or update comment
+        uses: peter-evans/create-or-update-comment@v2
```
rwaskiewicz, line 78:
```suggestion
        uses: peter-evans/create-or-update-comment@c9fcb64660bc90ec1cc535646af190c992007c32 # v2.0.0
```

rwaskiewicz (changes requested):
Looks good - most of my comments are small nits to keep things consistent with the rest of the codebase/CI.

package.json
alicewriteswrongs: that makes sense, I'll go ahead and make that change

.github/workflows/strict-null-checks.yml
```diff
@@ -0,0 +1,83 @@
...
+      # this is only necessary while tsc-output-parser isn't a dependency in main
+      # once it is, we can introduce a PR to remove this step
+      - name: Install tsc-output-parser
+        run: npm install @aivenio/tsc-output-parser
+
+      - name: Run Typescript compiler and generate JSON-formatted error file (main branch)
+        run: npx tsc --strictNullChecks --noEmit --pretty false | npx tsc-output-parser > null_errors_main.json
+
+      - name: Upload null_errors_main.json
+        uses: actions/upload-artifact@v3
```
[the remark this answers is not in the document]
alicewriteswrongs: yeah good catch, I meant to make this consistent with our usage in other workflows but it slipped my mind

.github/workflows/strict-null-checks.yml
```diff
@@ -0,0 +1,83 @@
...
+
+      - name: Download error file for main branch
+        uses: actions/download-artifact@v3
+        with:
+          name: null_errors_main
+
+      - name: Set action output
+        run: node scripts/strict-null-checks-ci.js > $GITHUB_STEP_SUMMARY
+
+      - name: Set comment body
```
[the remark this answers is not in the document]
alicewriteswrongs: sure, that makes sense

scripts/strict-null-checks-ci.js
alicewriteswrongs: old habit from Haskell! the idea is if you have a type `x` then the plural of that is `xs`, and `x[]` is something like the plural of `x`...

I'll change it to `arr` or something like that

.github/workflows/strict-null-checks.yml
```diff
@@ -0,0 +1,90 @@
+name: Strict Null Checks Burndown
+# this is a temporary workflow while we burn down the errors
+# from typescript's strictNullChecks option. Once we get to 0
+# we can turn that option on all the time and delete this job!
+#
+# TODO (STENCIL-446): Remove this workflow once `strictNullChecks` is enabled
```
rwaskiewicz, line 6: Nit - to keep the codebase greppable with the same format:
```suggestion
# TODO(STENCIL-446): Remove this workflow once `strictNullChecks` is enabled
```

scripts/strict-null-checks-ci.js
rwaskiewicz: 👍 I'm OK with that/leaving it as is for now in this case

scripts/strict-null-checks-ci.ts
```diff
@@ -0,0 +1,247 @@
+/**
+ * TODO (STENCIL-446): Remove this script once `strictNullChecks` is enabled
```
rwaskiewicz, line 2:
```suggestion
 * TODO(STENCIL-446): Remove this script once `strictNullChecks` is enabled
```

scripts/strict-null-checks-ci.ts
```diff
@@ -0,0 +1,247 @@
...
+         * see {@link https://github.com/microsoft/TypeScript/blob/main/src/compiler/diagnosticMessages.json}
+         */
+        errorString: string; // t
+      };
+    };
+    message: {
+      type: 'Message';
+      /**
+       * this is the actual, concrete error message for a given error, so it includes
+       * concrete informationa about the actual type error encountered (for instance
```
rwaskiewicz, line 54:
```suggestion
       * concrete information about the actual type error encountered (for instance
```

scripts/strict-null-checks-ci.ts
```diff
@@ -0,0 +1,247 @@
...
+    };
+    tsError: {
+      type: 'TsError';
+      value: {
+        type: 'error';
+        /**
+         * These strings are TS error codes like `TS2339`
+         * see {@link https://github.com/microsoft/TypeScript/blob/main/src/compiler/diagnosticMessages.json}
+         */
+        errorString: string; // t
```
rwaskiewicz, line 47: Is 't' meant to stand for something, or is this a dangling comment?

rwaskiewicz (changes requested):
@alicewriteswrongs Thanks! A few nitpicks on the updates, then we should be good to :shipit:

scripts/strict-null-checks-ci.ts
alicewriteswrongs: oops!

rwaskiewicz (approved):

PR #3367, opened 2022-05-12
refactor(tests): flesh out stubComponentCompilerMeta

alicewriteswrongs:
This adds all the fields to our `stubComponentCompilerMeta` function in
order to return a valid `ComponentCompilerMeta` object. The fields all
have (what I believe to be) sensible defaults.

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

GitHub Issue Number: N/A

The `stubComponentCompilerMeta` function that we already had kicking around isn't quite fully fleshed out yet. When working on STENCIL-332 I got to a point where for testing I needed to be able to create and modify some well-formed `ComponentCompilerMeta` objects, so I thought I might as well put up a PR to fill out this stub.

## What is the new behavior?

- this now includes all of the fields defined on the interface with (what I believe to be) a doable default value
- I also changed every place that was annotated with the corresponding Jira ticket to use the function - there may be more places where we could use this though!

- [ ] Yes
- [x] No


src/compiler/types/tests/ComponentCompilerMeta.stub.ts
```diff
@@ -9,17 +9,85 @@ import * as d from '@stencil/core/declarations';
...
-    sourceFilePath: '/some/stubbed/path/my-component.tsx',
-    tagName: 'stub-cmp',
-    virtualProperties: [],
-  } as d.ComponentCompilerMeta;
-
-  return { ...defaults, ...overrides };
-};
+): d.ComponentCompilerMeta => ({
+  assetsDirs: [],
+  componentClassName: '',
```
rwaskiewicz, line 14: Since we have a `tagName` of ['stub-cmp'](https://github.com/ionic-team/stencil/pull/3367/files#diff-e18c71115908470b2236d446a239b6524130d3a443cb5c672e66287fd1719f6aR89) it might make sense to give this a name like 'StubCmp'
```suggestion
  componentClassName: 'StubCmp',
```

FWIW, the tag name and the class name don't _have_ to line up in a normal Stencil project, but it might make sense here to minimize the cognitive overhead if they were different

src/compiler/types/tests/ComponentCompilerMeta.stub.ts
```diff
@@ -9,17 +9,85 @@ import * as d from '@stencil/core/declarations';
...
+  jsFilePath: '/some/stubbed/path/my-other-component.js',
+  legacyConnect: [],
+  legacyContext: [],
+  listeners: [],
+  methods: [],
+  potentialCmpRefs: [],
+  properties: [],
+  shadowDelegatesFocus: false,
+  sourceFilePath: '/some/stubbed/path/my-component.tsx',
+  sourceMapPath: '/some/stubbed/path/my-other-component.map',
```
rwaskiewicz, line 85: Should the source map file line up with the source?

```suggestion
  sourceMapPath: '/some/stubbed/path/my-component.js.map',
```

src/compiler/types/tests/ComponentCompilerMeta.stub.ts
```diff
@@ -9,17 +9,85 @@ import * as d from '@stencil/core/declarations';
...
+  hasVdomXlink: false,
+  hasWatchCallback: false,
+  htmlAttrNames: [],
+  htmlParts: [],
+  htmlTagNames: [],
+  internal: false,
+  isCollectionDependency: false,
+  isPlain: false,
+  isUpdateable: false,
+  jsFilePath: '/some/stubbed/path/my-other-component.js',
```
rwaskiewicz, line 76: I _think_ this should line up with the source file name:

```suggestion
  jsFilePath: '/some/stubbed/path/my-component.js',
```

src/compiler/transformers/test/add-component-meta-proxy.spec.ts
```diff
@@ -50,53 +51,33 @@ describe('add-component-meta-proxy', () => {
     });
 
     it('returns a call expression', () => {
-      const result: ts.CallExpression = createAnonymousClassMetadataProxy(
-        // TODO(STENCIL-378): Replace with a getMockComponentCompilerMeta() call
```
rwaskiewicz, line 54: 🔥

rwaskiewicz (changes requested):
Just a few data suggestions for the mock

src/compiler/types/tests/ComponentCompilerMeta.stub.ts
alicewriteswrongs: probably!

alicewriteswrongs:
For a, um, _large number_ of these properties I basically just copied over the interface code and did things like `s/boolean;/false,/g`, `s/string;/"",/g`, so there may well be some inappropriate default values in there!

rwaskiewicz:
@alicewriteswrongs I think that's fine - we may end up creating a second one of these at some point with more realistic data if we find we're building a stub and overriding more fields than not (especially if there are patterns to those overrides as we apply them).  I imagine we'll do that for some of the stubs (if not this one) - IMO it'll depend on the different scenarios we (unit) test and the level of duplication we start to see as we use these more and more.  For now the 'baseline' here is pretty similar to what we had today, seems like a good starting point to me 👍

rwaskiewicz (approved):

PR #3369, opened 2022-05-18
Ap/eslint

alicewriteswrongs:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

GitHub Issue Number: N/A

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No


package.json
```diff
@@ -92,6 +95,8 @@
     "core-js-builder": "~3.6.5",
     "css": "^3.0.0",
     "dts-bundle-generator": "~5.3.0",
+    "eslint": "^8.13.0",
+    "eslint-config-prettier": "^8.5.0",
```
rwaskiewicz, line 99: IIRC, this has to be added to the `.eslintrc.js` file to turn those rules off that conflict with Prettier.  It's probably a no-op right now, since `no-unused-vars` doesn't conflict with Prettier.

Specifically, I think we need to do something [like this](https://github.com/prettier/eslint-config-prettier#installation) in `.eslintrc.js`:
```diff
module.exports = {
  root: true,
  parser: '@typescript-eslint/parser',
  plugins: [
    '@typescript-eslint',
  ], 
+ extends: {
+   // prettier must be last in this list to ensure that it's not overridden by other configs
+   prettier,
+ }
  rules: {
    "@typescript-eslint/no-unused-vars": [2]
  }
};
```

(FWIW, I'm a big fan of commenting the heck out of lint config files to try to make the 'whys' ridiculously apparent)

.eslintrc.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+  root: true,
+  parser: '@typescript-eslint/parser',
+  plugins: [
+    '@typescript-eslint',
+  ],
+  rules: {
+    "@typescript-eslint/no-unused-vars": [2]
```
rwaskiewicz, line 8: Nit - can we use "error" instead of 2? Otherwise one has to understand ESLint's configuration to understand what a '2' is.
```suggestion
    "@typescript-eslint/no-unused-vars": ["error"]
```

.eslintrc.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
```
rwaskiewicz, line 1: Out of curiosity, does running Prettier in the repo catch this file?

rwaskiewicz (changes requested):
A few additional thoughts/questions

1. Did you test out any of the rules [options](@typescript-eslint/no-unused-vars)? Are there any we should/shouldn't adopt here?

2. I think we'll want to wire this up as an NPM script sooner than later. Do you think it's best we do that now, or do so in a subsequent PR?

3. Does the CI check that we'll add for unused functions act separately from a CI check that runs lint? (I.E. are they two processes or one?)

package.json
alicewriteswrongs: yeah, I think you're right that it's a good idea to set that up now. When we have no rules set up it doesn't make much difference, but why not make sure it's set up from the beginning!

.eslintrc.js
alicewriteswrongs: currently no, if we made this change to the `package.json` script then we'd include it:

```diff
diff --git a/package.json b/package.json
index d89d75738..047720d17 100644
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "rollup.watch": "rollup --watch --config",
     "license": "node scripts --license",
     "prettier": "npm run prettier.base -- --write",
-    "prettier.base": "prettier \"./({bin,scripts,src,test}/**/*.{ts,tsx,js,jsx})|bin/stencil\"",
+    "prettier.base": "prettier \"./({bin,scripts,src,test}/**/*.{ts,tsx,js,jsx})|*.js|bin/stencil\"",
     "prettier.dry-run": "npm run prettier.base -- --list-different",
     "test": "jest --coverage",
     "test.analysis": "cd test && npm run analysis.build-and-analyze",
```

there might be a more elegant way to do that!

doing so would also format our `jest.config.js` and `rollup.config.js` files, resulting in pretty minimal changes:

```diff
diff --git a/jest.config.js b/jest.config.js
index 2071bc01a..095f25159 100644
--- a/jest.config.js
+++ b/jest.config.js
@@ -57,5 +57,5 @@ module.exports = {
   testRegex: '/(src|scripts)/.*\\.spec\\.(ts|tsx|js)$',
   // TODO(STENCIL-307): Move away from Jasmine runner for internal Stencil tests, which involves re-working environment
   // setup
-  testRunner: 'jest-jasmine2'
+  testRunner: 'jest-jasmine2',
 };
diff --git a/rollup.config.js b/rollup.config.js
index 9a3d58d01..db6708815 100644
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,7 +1,7 @@
 const { createBuild } = require('./scripts/build/build.js');
 const { getOptions } = require('./scripts/build/utils/options.js');
 
-module.exports = function(args) {
+module.exports = function (args) {
   const opts = getOptions(__dirname, {
     isProd: !!args['config-prod'],
     isCI: !!args['config-ci'],
```

I think we might want to do that, just for consistency

rwaskiewicz: Yeah I like that - let's go that route!

alicewriteswrongs:
questions!

1. I was planning to add the `"argsIgnorePattern"` option ([doc](https://eslint.org/docs/rules/no-unused-vars#argsignorepattern)) in order to jibe with the typical idiom that `_foo` means that I meant for that argument to be unused but stick around — I didn't put it in this PR initially just, idk, because, but I just added it now :)
2. My thought was that it would make sense to add a `package.json` script when we're ready to add `eslint must not fail` to the CI checks that we want to enforce. The way I envisioned that working was that once we've fixed all of the `no-unused-vars` errors we can now enshrine not allowing unused variables as a new requirement for code to be checked in going forward, and then we can slowly expand the roster of eslint rules we add like that until we're ready to do a PR where we switch over to the recommended rulesets (`'eslint:recommended'` and `'plugin:@typescript-eslint/recommended'`).

   That said, I just checked and after setting the `"argsIgnorePattern"` option we are sitting at only 21 violations of `no-unused-vars`, so it may be just easier to fix those now and add passing `eslint src/**/*.ts src/**/*.tsx` to CI now - what do you think?

Does that kind of make sense?

alicewriteswrongs:
I think we want `"varsIgnorePattern": "^h$"` set as well

alicewriteswrongs:
@rwaskiewicz I just went ahead and fixed all of the unused vars errors that crop up when we have ignoring `^_` in params and ignoring `(h|Fragment)` set in the rule config, and then also added a quick `package.json` script. If you think that looks good I think we could add a check that `eslint` as-configured has to pass to CI now. We could add another job (probably would add a decent bit of overhead though) or add it to the job where we run `prettier` right now, maybe renaming it `"Lint n' Format"` or something along those lines

package.json
```diff
@@ -54,7 +54,8 @@
     "test.testing": "node scripts/test/validate-testing.js",
     "test.prod": "npm run test.dist && npm run test.end-to-end && npm run test.jest && npm run test.karma && npm run test.sys.node && npm run test.testing && npm run test.analysis",
     "test.watch": "jest --watch",
-    "test.watch-all": "jest --watchAll --coverage"
+    "test.watch-all": "jest --watchAll --coverage",
+    "lint": "eslint src/*.ts src/**/*.ts src/**/*.tsx"
```
rwaskiewicz, line 58: Nit - can we keep this list of scripts alphabetized by their name/key?

src/compiler/sys/tests/stencil-sys.spec.ts
```diff
@@ -70,7 +70,7 @@ describe('stencil system', () => {
     expect(newStat.isFile).toBe(true);
 
     const content = await sys.readFile('/dir/file-new');
-    expect('content').toBe('content');
```
rwaskiewicz, line 73: 😬

src/declarations/stencil-private.ts
```diff
@@ -2124,6 +2124,7 @@ export interface Url {
 
 declare global {
   namespace jest {
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
rwaskiewicz, line 2127: Nit - similar to `// @ts-ignore` comments, can we add the reason as to why we have to disable a rule? That way folks reading the code in the future understand why the `// eslint-disable*` directive is there.  In ESLint, separating the reason from the directive itself is done with 2+ `-`'s:

```diff
    // eslint-disable-next-line @typescript-eslint/no-unused-vars -- the reason we're disabling this rule
```

rwaskiewicz (changes requested):
> I was planning to add the "argsIgnorePattern" option ([doc](https://eslint.org/docs/rules/no-unused-vars#argsignorepattern)) in order to jibe with the typical idiom that _foo means that I meant for that argument to be unused but stick around — I didn't put it in this PR initially just, idk, because, but I just added it now :)

This LGTM! Added a small comment about revising ignoring `h | Fragment`.

> My thought was that it would make sense to add a package.json script when we're ready to add eslint must not fail to the CI checks that we want to enforce. The way I envisioned that working was that once we've fixed all of the no-unused-vars errors we can now enshrine not allowing unused variables as a new requirement for code to be checked in going forward, and then we can slowly expand the roster of eslint rules we add like that until we're ready to do a PR where we switch over to the recommended rulesets ('eslint:recommended' and 'plugin:@typescript-eslint/recommended').

> That said, I just checked and after setting the "argsIgnorePattern" option we are sitting at only 21 violations of no-unused-vars, so it may be just easier to fix those now and add passing eslint src/**/*.ts src/**/*.tsx to CI now - what do you think?

> I think we want "varsIgnorePattern": "^(h|Fragment)$" set as well

All makes sense to me 👍 
I have wondered if we would ever want/get to using `eslint:recommended` - I haven't looked at the default ruleset in a while now, but had a "feeling" that there might be something in there that clashes with something that's fundamental to how Stencil is written.  But we might as well keep picking rules off the recommended list like you said - and if we can hit parity with it, all the better 🎉 

> We could add another job (probably would add a decent bit of overhead though) or add it to the job where we run prettier right now, maybe renaming it "Lint n' Format" or something along those lines

Yeah, combining lint + format makes sense to me! Let's do that now so we don't regress on any of the work you've done here

src/compiler/sys/tests/stencil-sys.spec.ts
alicewriteswrongs: yeah I think I audibly said "ouch" when I saw that 😬

package.json
rwaskiewicz: Well, I guess I can't read - cause I thought this was already alphabetized 🙃 thanks!

src/compiler/sys/tests/stencil-sys.spec.ts
rwaskiewicz: Some floating point numbers too, just to be safe

alicewriteswrongs:
@rwaskiewicz ok I think feedback should all be addressed now

rwaskiewicz (approved):
👍 lgtm!

PR #3375, opened 2022-05-23
remove unused exports on output utils

alicewriteswrongs:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Other (please describe): removing unused exports

GitHub Issue Number: N/A

## What is the new behavior?

This removes unused exports from `src/compiler/output-targets/output-utils.ts`, just to move the needle a little bit on our unused exports tech debt burndown. It also fixes some strict null checks compiler errors that were a bit incidental to this, and makes a few (small) code changes in the output-utils file.

- [ ] Yes
- [x] No

## Testing

Tests are passing, could be good to validate some of this just by building a compiler and running it on some example stuff.

### DIFF
--- a/src/compiler/output-targets/output-utils.ts
+++ b/src/compiler/output-targets/output-utils.ts
@@ -129,9 +129,11 @@ type ValidConfigOutputTarget = typeof VALID_CONFIG_OUTPUT_TARGETS[number];
  * @returns whether or not the targetType is a valid, configurable output target.
  */
 export function isValidConfigOutputTarget(targetType: string): targetType is ValidConfigOutputTarget {
-  // unfortunately `includes` is typed on `Array<T>` as `(el: T):
+  // unfortunately `includes` is typed on `ReadonlyArray<T>` as `(el: T):
   // boolean` so a `string` cannot be passed to `includes` on a
   // `ReadonlyArray` 😢 thus we `as any`
+  //
+  // see microsoft/TypeScript#31018 for some discussion of this
   return VALID_CONFIG_OUTPUT_TARGETS.includes(targetType as any);
 }
 
--- a/src/compiler/output-targets/output-utils.ts
+++ b/src/compiler/output-targets/output-utils.ts
@@ -124,6 +124,9 @@ type ValidConfigOutputTarget = typeof VALID_CONFIG_OUTPUT_TARGETS[number];
 
 /**
  * Check whether a given output target is a valid one to be set in a Stencil config
+ *
+ * @param targetType the type which we want to check
+ * @returns whether or not the targetType is a valid, configurable output target.
  */
 export function isValidConfigOutputTarget(targetType: string): targetType is ValidConfigOutputTarget {
   // unfortunately `includes` is typed on `Array<T>` as `(el: T):
--- a/src/compiler/output-targets/output-utils.ts
+++ b/src/compiler/output-targets/output-utils.ts
@@ -14,36 +14,6 @@ export const relativeImport = (pathFrom: string, pathTo: string, ext?: string, a
   return normalizePath(`${relativePath}/${basename(pathTo, ext)}`);
 };
 
-export const getDistEsmDir = (outputTarget: d.OutputTargetDist, sourceTarget?: d.SourceTarget) =>
-  join(outputTarget.buildDir, 'esm', sourceTarget || '');
-
-export const getDistEsmComponentsDir = (outputTarget: d.OutputTargetDist, sourceTarget: d.SourceTarget) =>
-  join(getDistEsmDir(outputTarget, sourceTarget), 'build');
-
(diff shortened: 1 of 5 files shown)


src/compiler/output-targets/output-utils.ts
```diff
@@ -153,6 +117,18 @@ export const VALID_TYPES = [
...
+
+type ValidConfigOutputTarget = typeof VALID_CONFIG_OUTPUT_TARGETS[number];
+
+/**
+ * Check whether a given output target is a valid one to be set in a Stencil config
+ */
+export function isValidConfigOutputTarget(targetType: string): targetType is ValidConfigOutputTarget {
+  // unfortunately `includes` is typed on `Array<T>` as `(el: T):
+  // boolean` so a `string` cannot be passed to `includes` on a
+  // `ReadonlyArray` 😢 thus we `as any`
```
alicewriteswrongs, line 131: This is a bit of an annoying thing with typescript. If you have a `ReadonlyArray` like this:

```ts
const xs = ["foo", "bar"] as const;
```

then `type arrEntry = xs[number]` will be of type `"foo" | "bar"`, which is great! But unfortunately, array methods are also all typed like this. You of course want this for things like `Array.prototype.map`, that should be something like `map<T>: (x: "foo" | "bar") => T`, but we don't really want this for `Array.prototype.includes`. The problem is that `ReadonlyArray.prototype.includes` also has the type of the array entries, so it's something like

```ts
interface ReadOnlyArray {
  includes: (x: "foo" | "bar") => boolean
}
```

I think I'd prefer 

```ts
interface ReadOnlyArray {
  includes: (x: string) => x is "foo" | "bar"
}
```

since narrowing the type of the argument for `.includes` means that I can't take an arbitrary string and test if it's found in a `ReadonlyArray` I've declared 😢 

I found an issue on the typescript repo talking about this and discussing various workarounds: https://github.com/microsoft/TypeScript/issues/31018

Anywhoo - to work around this I added this typeguard function here which, while it has an internal type hole (`as any`), does let us have a way to check for membership within a `ReadonlyArray` for an arbitrary string. Not sure this is the best solution though!

src/compiler/output-targets/output-utils.ts
```diff
@@ -99,40 +69,34 @@ export const isOutputTargetDistTypes = (o: d.OutputTarget): o is d.OutputTargetD
...
+export const WWW = 'www';
+
+/**
+ * Valid output targets to specify in a Stencil config.
+ *
+ * Note that there are some output targets (e.g. `DIST_TYPES`) which are programmatically
+ * set as output targets by the compiler when other output targets (in that case `DIST`) are set, but which are _not_ supported in a Stencil config. This is enforced in the output
+ * target validation code.
+ */
+export const VALID_CONFIG_OUTPUT_TARGETS = [
```
alicewriteswrongs, line 100: I renamed this because I didn't think `VALID_TYPES` was a very good name, especially for an identifier that is exported outside of the context of a single module (I think even for an unexported name I'd want something more descriptive but 🤷‍♀️ )

src/compiler/output-targets/output-utils.ts
```diff
@@ -99,40 +69,35 @@ export const isOutputTargetDistTypes = (o: d.OutputTarget): o is d.OutputTargetD
...
-export const DIST = `dist`;
-export const DIST_COLLECTION = `dist-collection`;
-export const DIST_CUSTOM_ELEMENTS = `dist-custom-elements`;
-export const DIST_CUSTOM_ELEMENTS_BUNDLE = `dist-custom-elements-bundle`;
-
-export const DIST_TYPES = `dist-types`;
-export const DIST_HYDRATE_SCRIPT = `dist-hydrate-script`;
-export const DIST_LAZY = `dist-lazy`;
-export const DIST_LAZY_LOADER = `dist-lazy-loader`;
+export const CUSTOM = 'custom';
```
alicewriteswrongs, line 74: I didn't know why these were all templates so I just changed them to be strings

src/compiler/output-targets/output-utils.ts
```diff
@@ -14,36 +14,6 @@ export const relativeImport = (pathFrom: string, pathTo: string, ext?: string, a
   return normalizePath(`${relativePath}/${basename(pathTo, ext)}`);
 };
 
-export const getDistEsmDir = (outputTarget: d.OutputTargetDist, sourceTarget?: d.SourceTarget) =>
```
alicewriteswrongs, line 17: I couldn't find any uses of these helper functions, and builds and tests pass without them so I believe it is safe to delete them.

alicewriteswrongs (commented):
pointing out a few things

rwaskiewicz (changes requested):

PR #3383, opened 2022-05-26
typecheck a subset of jest tests

alicewriteswrongs:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

GitHub Issue Number: N/A

Currently our tests aren't checked! This means that we'll only fix type errors in test files sort of incidentally, if they pop up as errors in someone's editor while they're working on tests.

## What is the new behavior?

This adds a subset of the jest tests (those which have a `.spec.ts` extension) to the `include` option in `tsconfig.json`, so that we'll typecheck those files and catch any type errors therein.

Some notes:

- I added a `// @ts-nocheck` to files which are currently skipped since the changes necessary to get the typechecker passing is _basically_ equivalent to the work required to get those tests passing again, and I didn't want to make the PR larger than necessary
- I tried to avoid `as Foo` as much as possible but in a few places it was the easiest way to get rid of an error and I thought it wasn't unreasonable, so there are a few scattered around
- this change leads to a big increase in `strictNullChecks` errors 😢 

There are a few other things noted on specific lines below.

- [ ] Yes
- [x] No

## Testing

The changes are basically all in spec files, so I think as long as the build is passing that should be good!

### DIFF
--- a/src/cli/telemetry/test/telemetry.spec.ts
+++ b/src/cli/telemetry/test/telemetry.spec.ts
@@ -3,14 +3,16 @@ import * as telemetry from '../telemetry';
 import * as shouldTrack from '../shouldTrack';
 import { createSystem } from '../../../compiler/sys/stencil-sys';
 import { mockLogger } from '@stencil/core/testing';
+import * as coreCompiler from '@stencil/core/compiler';
 
 describe('telemetryBuildFinishedAction', () => {
-  const config = {
+  const config: d.Config = {
     outputTargets: [],
     flags: {
       args: [],
     },
-  } as d.Config;
+  };
+
   const logger = mockLogger();
   const sys = createSystem();
 
@@ -27,7 +29,7 @@ describe('telemetryBuildFinishedAction', () => {
       duration: 100,
     } as d.CompilerBuildResults;
 
-    await telemetry.telemetryBuildFinishedAction(sys, config, logger, {}, results);
+    await telemetry.telemetryBuildFinishedAction(sys, config, logger, coreCompiler, results);
     expect(spyShouldTrack).toHaveBeenCalled();
 
     spyShouldTrack.mockRestore();
@@ -52,7 +54,7 @@ describe('telemetryAction', () => {
       })
     );
 
-    await telemetry.telemetryAction(sys, config, logger, {}, () => {});
+    await telemetry.telemetryAction(sys, config, logger, coreCompiler, () => {});
     expect(spyShouldTrack).toHaveBeenCalled();
 
     spyShouldTrack.mockRestore();
@@ -66,7 +68,7 @@ describe('telemetryAction', () => {
       })
     );
 
-    await telemetry.telemetryAction(sys, config, logger, {}, async () => {
+    await telemetry.telemetryAction(sys, config, logger, coreCompiler, async () => {
       new Promise((resolve) => {
         setTimeout(() => {
           resolve(true);
@@ -137,10 +139,10 @@ describe('prepareData', () => {
   const sys = createSystem();
 
   it('provides an object', async () => {
-    const data = await telemetry.prepareData({}, config, sys, 1000);
+    const data = await telemetry.prepareData(coreCompiler, config, sys, 1000);
     expect(data).toEqual({
       arguments: [],
-      build: 'unknown',
+      build: coreCompiler.buildId,
       component_count: undefined,
       cpu_model: '',
       duration_ms: 1000,
@@ -149,13 +151,13 @@ describe('prepareData', () => {
       os_version: '',
       packages: [],
       packages_no_versions: [],
-      rollup: 'unknown',
-      stencil: 'unknown',
+      rollup: coreCompiler.versions.rollup,
+      stencil: coreCompiler.versions.stencil,
       system: 'in-memory __VERSION:STENCIL__',
       system_major: 'in-memory __VERSION:STENCIL__',
       targets: [],
       task: undefined,
-      typescript: 'unknown',
+      typescript: coreCompiler.versions.typescript,
       yarn: false,
     });
   });
@@ -168,11 +170,11 @@ describe('prepareData', () => {
       outputTargets: [{ type: 'www', baseUrl: 'https://example.com', serviceWorker: { swDest: './tmp' } }],
     } as d.Config;
 
-    const data = await telemetry.prepareData({}, config, sys, 1000);
+    const data = await telemetry.prepareData(coreCompiler, config, sys, 1000);
 
     expect(data).toEqual({
       arguments: [],
-      build: 'unknown',
+      build: coreCompiler.buildId,
       component_count: undefined,
       cpu_model: '',
       duration_ms: 1000,
@@ -181,13 +183,13 @@ describe('prepareData', () => {
       os_version: '',
       packages: [],
       packages_no_versions: [],
-      rollup: 'unknown',
-      stencil: 'unknown',
+      rollup: coreCompiler.versions.rollup,
+      stencil: coreCompiler.versions.stencil,
       system: 'in-memory __VERSION:STENCIL__',
       system_major: 'in-memory __VERSION:STENCIL__',
       targets: ['www'],
       task: undefined,
-      typescript: 'unknown',
+      typescript: coreCompiler.versions.typescript,
       yarn: false,
     });
   });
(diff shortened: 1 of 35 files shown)


src/cli/telemetry/test/telemetry.spec.ts
```diff
@@ -137,10 +139,10 @@ describe('prepareData', () => {
   const sys = createSystem();
 
   it('provides an object', async () => {
-    const data = await telemetry.prepareData({}, config, sys, 1000);
+    const data = await telemetry.prepareData(coreCompiler, config, sys, 1000);
     expect(data).toEqual({
       arguments: [],
-      build: 'unknown',
+      build: coreCompiler.buildId,
```
alicewriteswrongs, line 145: these changes are just consequences of calling these functions with an actual `coreCompiler` instance instead of `{}`

src/cli/test/parse-flags.spec.ts
```diff
@@ -58,12 +58,13 @@ describe('parseFlags', () => {
...
     args.push('serve', '--address', '127.0.0.1', '--port', '8888', '4444');
 
-    sys.getEnvironmentVar = (key) => {
+    sys.getEnvironmentVar = (key: string): string => {
       if (key === 'npm_config_argv') {
         return JSON.stringify({
           original: ['run', 'serve', '--port', '4444'],
         });
       }
+      return '';
```
alicewriteswrongs, line 67: I believe this should be ok? No tests are failing soooo

src/compiler/build/test/build-stats.spec.ts
```diff
@@ -16,7 +16,7 @@ describe('generateBuildStats', () => {
   it('should return a structured json object', async () => {
     buildCtx.buildResults = generateBuildResults(config, compilerCtx, buildCtx);
 
-    const result: d.CompilerBuildStats = generateBuildStats(config, buildCtx);
+    const result = generateBuildStats(config, buildCtx) as d.CompilerBuildStats;
```
alicewriteswrongs, line 19: because of the two possible return types here I thought this was the easier way to get this passing (and fairly well-justified)

src/compiler/config/test/validate-config.spec.ts
```diff
@@ -256,10 +257,13 @@ describe('validation', () => {
...
-    const { config } = validateConfig(userConfig);
-    expect(config.outputTargets.some((o) => o.type === 'docs')).toBe(false);
-  });
+  it.each([DOCS_JSON, DOCS_CUSTOM, DOCS_README, DOCS_VSCODE])(
+    'should not add "%s" output target by default',
+    (targetType) => {
+      const { config } = validateConfig(userConfig);
+      expect(config.outputTargets.some((o) => o.type === targetType)).toBe(false);
+    }
+  );
```
alicewriteswrongs, line 266: there were two identical tests both testing what I believe to be an outdated / deprecated output target (`"docs"`), I replaced them with this one checking that none of the doc-related output targets get automatically added

src/compiler/output-targets/test/output-targets-dist.spec.ts
```diff
@@ -1,3 +1,4 @@
+// @ts-nocheck
```
alicewriteswrongs, line 1: this is debatable I think. I added `// @ts-nocheck` to the files in `src/compiler/output-targets/test` because they are all skipped currently and getting them passing again is a substantial amount of work (which we should do!).

I think at the same time that we get these passing we can remove this and get the typescript compiler passing here too.

alicewriteswrongs: There are a few other tests in other directories to which I did this too, but they are also files with `xdescribe`

src/runtime/vdom/test/event-listeners.spec.ts
```diff
@@ -21,7 +21,7 @@ describe('event listeners', () => {
 
     const vnode = h('div', { onClick: clicked }, h('a', null, 'Click my parent'));
 
-    patch(vnode0, vnode, document);
+    patch(vnode0, vnode);
```
alicewriteswrongs, line 24: I'm not sure why these were all receiving a third `document` argument still. It looks like an older version of the `patch` function took this argument but it was later removed, see https://github.com/ionic-team/stencil/commit/15c00c0f947f4976f3f0a465c147bd55c8382269

src/runtime/vdom/test/patch-svg.spec.ts
```diff
@@ -29,8 +29,8 @@ describe('renderer', () => {
       );
 
       expect(svgElm.namespaceURI).toEqual(SVG_NS);
-      expect(svgElm.firstChild.namespaceURI).toEqual(SVG_NS);
-      expect(svgElm.children[0].firstChild.namespaceURI).not.toEqual(SVG_NS);
+      expect((svgElm.firstChild as SVGSVGElement).namespaceURI).toEqual(SVG_NS);
```
alicewriteswrongs, line 32: 😕 I don't love this, but not sure of a good way to fix it

alicewriteswrongs (commented):

rwaskiewicz (changes requested):

src/compiler/build/test/build-stats.spec.ts
alicewriteswrongs: sure, that makes sense - I'm also not a big fan of sort of random `TypeA | TypeB` return types on functions. It would be one thing if we had something like `Result` or `Option` from Rust, so we could do `Result<TypeA, TypeB>` and have some more principled way for dealing with that, but we don't so :/

alicewriteswrongs: Created `STENCIL-461` for this

src/compiler/output-targets/test/output-targets-dist.spec.ts
alicewriteswrongs: yup agreed, I took a quick stab at getting one of these going (I think it was this file here, `output-target-dist.spec.ts`) and it isn't exactly trivial, especially since we don't really have good tests written right now, so it's not just a matter of getting what is there to pass.

src/compiler/service-worker/test/service-worker.spec.ts
```diff
@@ -1,3 +1,4 @@
+// @ts-nocheck
```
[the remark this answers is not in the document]
alicewriteswrongs: sure, will do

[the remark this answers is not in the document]
alicewriteswrongs: added `STENCIL-462`, `STENCIL-463`, and `STENCIL-464`

src/compiler/config/test/validate-service-worker.spec.ts
```diff
@@ -12,13 +13,28 @@ describe('validateServiceWorker', () => {
...
+   * `outputTarget.serviceWorker` is typed as
+   * `serviceWorker?: ServiceWorkerConfig | null | false;` we get type errors
+   * all over if we try to just access it directly. So instead, do a little
+   * check to see if it's falsy. If not, we return it, and if it is we fail the test.
+   */
+  function getServiceWorker(target: OutputTargetWww) {
+    if (target.serviceWorker) {
+      return target.serviceWorker;
+    } else {
+      fail("shouldn't get here");
```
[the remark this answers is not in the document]
alicewriteswrongs: yeah, that makes sense!

src/mock-doc/test/element.spec.ts
```diff
@@ -164,10 +164,11 @@ describe('element', () => {
     doc.head.appendChild(metaElm);
     expect(metaElm).toEqualHtml(`<meta content="value" id="test">`);
 
-    const elm = doc.getElementById('test');
+    const elm = doc.getElementById('test') as MockMetaElement;
     expect(elm).toEqualHtml(`<meta content="value" id="test">`);
 
     elm['content'] = 'updated';
+    elm.setAttribute('content', 'updated');
```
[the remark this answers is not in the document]
alicewriteswrongs: ayyy looks like we don't - I did a bit of this work a week ago or so and I think this line just snuck in from some experimenting I was doing

alicewriteswrongs:
@rwaskiewicz ok I think I addressed all of your comments!

rwaskiewicz (approved):

PR #3386, opened 2022-05-27
chore(lint): turn on the 'no-var' rule

alicewriteswrongs:
adding because it is turned on in the `@typescript-eslint` base eslint rule configuration file, here:

https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/configs/eslint-recommended.ts#L27

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

GitHub Issue Number: N/A

### DIFF
--- a/src/client/polyfills/css-shim/test/css-shim.spec.ts
+++ b/src/client/polyfills/css-shim/test/css-shim.spec.ts
@@ -366,8 +366,8 @@ describe('css-shim', () => {
     );
   });
 
-  var window: Window; // eslint-disable-line no-var
-  var document: Document; // eslint-disable-line no-var
+  var window: Window; // eslint-disable-line no-var -- shims will continue to use var while we support older browsers
+  var document: Document; // eslint-disable-line no-var -- shims will continue to use var while we support older browsers
 
   function style(text: string) {
     const elm = document.createElement('style');
--- a/src/client/polyfills/css-shim/test/css-shim.spec.ts
+++ b/src/client/polyfills/css-shim/test/css-shim.spec.ts
@@ -366,8 +366,8 @@ describe('css-shim', () => {
     );
   });
 
-  var window: Window;
-  var document: Document;
+  let window: Window;
+  let document: Document;
 
   function style(text: string) {
     const elm = document.createElement('style');
--- a/src/client/polyfills/css-shim/test/css-shim.spec.ts
+++ b/src/client/polyfills/css-shim/test/css-shim.spec.ts
@@ -366,8 +366,8 @@ describe('css-shim', () => {
     );
   });
 
-  let window: Window;
-  let document: Document;
+  var window: Window; // eslint-disable-line no-var
+  var document: Document; // eslint-disable-line no-var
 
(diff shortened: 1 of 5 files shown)


src/client/polyfills/css-shim/test/css-shim.spec.ts
```diff
@@ -366,8 +366,8 @@ describe('css-shim', () => {
     );
   });
 
-  var window: Window;
```
rwaskiewicz, line 369: I think for our shims, it's safer to leave `var` in place. They (the shims) are designed to run in older browsers (like Safari 10, IE 11). But I think these shims will go away in a future major release where we don't support those browsers. For now, can we ignore 'no-var' here?

src/dev-server/dev-server-client/polyfills.ts
```diff
@@ -24,12 +24,12 @@ const applyObjectAssign = () => {
   if (typeof Object.assign !== 'function') {
```
rwaskiewicz, line 24: I think for our shims, it's safer to leave `var` in place. They (the shims) are designed to run in older browsers (like Safari 10, IE 11). But I think these shims will go away in a future major release where we don't support those browsers. For now, can we ignore 'no-var' here?

rwaskiewicz (changes requested):

src/client/polyfills/css-shim/test/css-shim.spec.ts
alicewriteswrongs: sure, that makes sense, I'll make that edit now

alicewriteswrongs:
@rwaskiewicz added ignores for those `var` usages, and then also dealt with rebasing the branch over `main`

src/client/polyfills/css-shim/test/css-shim.spec.ts
```diff
@@ -366,8 +366,8 @@ describe('css-shim', () => {
...
-  var document: Document;
+  var window: Window; // eslint-disable-line no-var
+  var document: Document; // eslint-disable-line no-var
```
rwaskiewicz, line 370: One last thing - can we add the justification for disabling eslint here?
```suggestion
  var window: Window; // eslint-disable-line no-var -- shims will continue to use var while we support older browsers
  var document: Document; // eslint-disable-line no-var -- shims will continue to use var while we support older browsers
```

src/dev-server/dev-server-client/polyfills.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-var */
```
rwaskiewicz, line 1: One last thing - can we add the justification for disabling eslint here?
```suggestion
/* eslint-disable no-var -- shims will continue to use var while we support older browsers */
```

rwaskiewicz (changes requested):

alicewriteswrongs:
@rwaskiewicz just added eslint disable justification strings

PR #3396, opened 2022-05-31
chore(publish): remove explicit next tag publish 

rwaskiewicz:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes

Stencil publishes a 'next' tag for new releases, which we want more control over moving forward.

GitHub Issue Number: N/A

## What is the new behavior?

this commit removes an explicit push of the 'next' tag to npm as a part
of the publish process.

we're removing this from the publishing process for the following
reasons:
- a very small number of folks <5 are actively using the next tag
- we wish to explore moving to a less bespoke system in the future for
  publishing. popular libraries for doing so do not offer this ability
- we also wish to be able to use 'next' for upcoming alpha, beta, etc.
  releases
- it makes using the current publishing scripts difficult with one time
  passwords (otp), as the current flow can only accept one otp

as a consequence of this change, the 'next' tag will be frozen in time at
its current spot (until a team member updates it).

discussed in the team's internal ionic slack channel on 2022.04.11

- [ ] Yes
- [x] No

## Testing

I couldn't test this unfortunately, as there are no tests around this script :-(

Similarly, manually kicking it off requires a publish.  I did trace through it to ensure there weren't any one-time use fields that we were taking advantage of (that we could eliminate if there were), happy to take any suggestions.


alicewriteswrongs (approved):
I think this change makes sense! 👍

PR #3398, opened 2022-06-01
test(json-parsing): add unit tests 

rwaskiewicz:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

the `parseJson` utility was unnecessarily exported. while looking at the 
code, I took half an hour to clean it up a bit, add tests, etc.

GitHub Issue Number: N/A

## What is the new behavior?

this commit adds testing around the `parsePackageJson` function. it
calls an internal function `parseJson`, which was orginally
unnecessarily exported and removing the `export` keyword from this
function was the impetus for this work.

some types were tightened/changed:
- a concrete type, `ParsePackageJsonResult` was created to describe the
  return type of `parsePackageJson` and `parseJson`. this type
  explicitly marks the `diagnostic` and `data` fields as the union of
  their original type and `null`, allowing us to remove type assertions
  in `parseJson`.
- the second parameter of `parseJson` has been narrowed to no longer be
  optional, as no code path exists where a this function is called
  without a second argument.

this commit removes a check if the json that has been passed to
`parseJson` is a string or not. looking at the single caller of it's
wrapper function, we verify that this argument is of type string
already. i'm going to add a comment to discuss what we want to here
with respect to defensive programming

- [ ] Yes
- [x] No

## Testing

- Unit tests were added + passed

### DIFF
--- a/src/utils/util.ts
+++ b/src/utils/util.ts
@@ -117,20 +117,38 @@ export const readPackageJson = async (config: d.Config, compilerCtx: d.CompilerC
   }
 };
 
-export const parsePackageJson = (
-  pkgJsonStr: string,
-  pkgJsonFilePath: string
-): { diagnostic: d.Diagnostic; data: d.PackageJsonData; filePath: string } => {
+/**
+ * A type that describes the result of parsing a `package.json` file's contents
+ */
+export type ParsePackageJsonResult = {
+  diagnostic: d.Diagnostic | null;
+  data: any | null;
+  filePath: string;
+};
+
+/**
+ * Parse a string read from a `package.json` file
+ * @param pkgJsonStr the string read from a `package.json` file
+ * @param pkgJsonFilePath the path to the already read `package.json` file
+ * @returns the results of parsing the provided contents of the `package.json` file
+ */
+export const parsePackageJson = (pkgJsonStr: string, pkgJsonFilePath: string): ParsePackageJsonResult | null => {
   if (isString(pkgJsonFilePath)) {
     return parseJson(pkgJsonStr, pkgJsonFilePath);
   }
   return null;
 };
 
-export const parseJson = (jsonStr: string, filePath?: string) => {
-  const rtn = {
-    diagnostic: null as d.Diagnostic,
-    data: null as any,
+/**
+ * Parse a string read from a `package.json` file
+ * @param jsonStr the string read from a `package.json` file
+ * @param filePath the path to the already read `package.json` file
+ * @returns the results of parsing the provided contents of the `package.json` file
+ */
+const parseJson = (jsonStr: string, filePath: string): ParsePackageJsonResult => {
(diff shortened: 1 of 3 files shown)


src/utils/util.ts
```diff
@@ -117,39 +117,50 @@ export const readPackageJson = async (config: d.Config, compilerCtx: d.CompilerC
...
+ * @returns the results of parsing the provided contents of the `package.json` file
+ */
+const parseJson = (jsonStr: string, filePath: string): ParsePackageJsonResult => {
+  const rtn: ParsePackageJsonResult = {
+    diagnostic: null,
+    data: null,
     filePath,
   };
 
-  if (isString(jsonStr)) {
```
rwaskiewicz, line 137: I removed this as the only caller of `parsePackageJson` validates that the argument it passes in when calling `parsePackageJson` is a string.  There's something to be said about being defensive here in our programming, but I tend to lean removing this in favor of the following considerations:
- the contents of the string aren't coming directly from a user (although its the contents of a file from a user provided path)
- leaning on the type system

the latter isn't probably sufficient on its own (at least not at this time) for user provided values, but I _think_ this is OK given the first point.  Feel free to argue the alternative though

src/utils/util.ts
```diff
@@ -117,42 +117,40 @@ export const readPackageJson = async (config: d.Config, compilerCtx: d.CompilerC
...
-    diagnostic: null as d.Diagnostic,
-    data: null as any,
-    filePath,
+/**
+ * Parse a string read from a `package.json` file
+ * @param pkgJsonStr the string read from a `package.json` file
+ * @param pkgJsonFilePath the path to the already read `package.json` file
+ * @returns the results of parsing the provided contents of the `package.json` file
+ */
+export const parsePackageJson = (pkgJsonStr: string, pkgJsonFilePath: string): ParsePackageJsonResult => {
```
alicewriteswrongs, line 135: This is one of those functions where you'd really like to have `Result` or at the least `Option` hanging around...

src/utils/util.ts
alicewriteswrongs: Yeah it's a little tricky here because `JSON.parse` called with a non-string will sometimes throw (e.g. w/ `JSON.parse({})` or `JSON.parse([])` and sometimes not (e.g. `JSON.parse(true)` or `JSON.parse(3)`). So a non-string value at runtime isn't a guarantee that the parsing will throw and that we'll return errors in the diagnostic and whatnot.

But I think that this is probably ok anyhow, I don't think that just checking it's a string gives us that much more safety anyway. We could still pass all kinds of things through that are valid JSON but which don't conform to what we'd expect a `package.json` to look like, e.g. `"3"`, `"true"`, `"[]"`, etc.

alicewriteswrongs (approved):
I think this all LGTM! more tests 🎉

PR #3401, opened 2022-06-01
feat(telemetry): add stencil config to telemetry object

alicewriteswrongs:
This adds an anonymized copy of the Stencil configuration to the
telemetry object, so that various configuration options can be tracked
and observed over time.

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Other (please describe): Telemetry change!

GitHub Issue Number: N/A

## What is the new behavior?

This adds a function to anonymize the Stencil configuration and basically adds that anonymized object in to the object that we ship to the telemetry API endpoint.

The main thing to review here is whether the anonymization makes sense! I went through the output and removed everything I could think of that would have a filepath in it that might contain the current username, an internal URL, or something of that type.

- [ ] Yes
- [x] No

## Testing

I have added tests around the anonymization function and updated the other unit tests for the module to account for the changes, I'm not sure if we need / want to test anything else. I've also run a dev build based on these changes locally and confirmed that the data is shipping to where it's supposed to.

### DIFF
--- a/src/cli/telemetry/telemetry.ts
+++ b/src/cli/telemetry/telemetry.ts
@@ -3,7 +3,7 @@ import { shouldTrack } from './shouldTrack';
 import type * as d from '../../declarations';
 import { readConfig, updateConfig, writeConfig } from '../ionic-config';
 import { CoreCompiler } from '../load-compiler';
-import { WWW } from '../../compiler/output-targets/output-utils';
+import { isOutputTargetHydrate, WWW } from '../../compiler/output-targets/output-utils';
 
 /**
  * Used to within taskBuild to provide the component_count property.
@@ -160,7 +160,35 @@ type ConfigStringKeys = keyof {
   [K in keyof d.Config as Required<d.Config>[K] extends string ? K : never]: d.Config[K];
 };
 
-const OUTPUT_TARGET_KEYS_TO_KEEP = ['type', 'typesDir', 'baseUrl'];
+// props in output targets for which we retain their original values when
+// preparing a config for telemetry
+//
+// we omit the values of all other fields on output targets.
+const OUTPUT_TARGET_KEYS_TO_KEEP = [
+  'type',
+  'external',
+  'autoDefineCustomElements',
+  'generateTypeDeclarations',
+] as const;
+
+// top-level config props that we anonymize for telemetry
+const CONFIG_PROPS_TO_ANONYMIZE: ReadonlyArray<ConfigStringKeys> = [
+  'rootDir',
+  'fsNamespace',
+  'packageJsonFilePath',
+  'namespace',
+  'srcDir',
+  'srcIndexHtml',
+  'buildLogFilePath',
+  'cacheDir',
+  'configPath',
+  'tsconfig',
+];
+
+// Props we delete entirely from the config for telemetry
+//
+// TODO(STENCIL-469): Investigate improving anonymization for tsCompilerOptions and devServer
+const CONFIG_PROPS_TO_DELETE: ReadonlyArray<keyof d.Config> = ['sys', 'logger', 'tsCompilerOptions', 'devServer'];
 
 /**
  * Anonymize the config for telemetry, replacing potentially revealing config props
@@ -172,46 +200,44 @@ const OUTPUT_TARGET_KEYS_TO_KEEP = ['type', 'typesDir', 'baseUrl'];
  */
 export const anonymizeConfigForTelemetry = (config: d.Config): d.Config => {
   const anonymizedConfig = { ...config };
-  const propsToAnonymize: ConfigStringKeys[] = [
-    'rootDir',
-    'fsNamespace',
-    'packageJsonFilePath',
-    'namespace',
-    'srcDir',
-    'srcIndexHtml',
-    'buildLogFilePath',
-    'cacheDir',
-    'configPath',
-    'tsconfig',
-  ];
 
-  for (const prop of propsToAnonymize) {
+  for (const prop of CONFIG_PROPS_TO_ANONYMIZE) {
     if (anonymizedConfig[prop] !== undefined) {
       anonymizedConfig[prop] = 'omitted';
     }
   }
 
-  // Anonymize the outputTargets on our configuration, taking advantage of the
-  // optional 2nd argument to `JSON.stringify`. If anything is not a string or number
-  // we retain it so that any nested properties are handled, else we check
-  // whether it's in our 'keep' list to decide whether to keep it or replace it
-  // with `"omitted"`.
-  anonymizedConfig.outputTargets = JSON.parse(
-    JSON.stringify(config.outputTargets ?? [], (key, value) => {
-      if (!['string', 'number'].includes(typeof value)) {
-        return value;
-      }
-      if (OUTPUT_TARGET_KEYS_TO_KEEP.includes(key)) {
-        return value;
-      }
-      return 'omitted';
-    })
-  );
+  anonymizedConfig.outputTargets = (config.outputTargets ?? []).map((target) => {
+    // Anonymize the outputTargets on our configuration, taking advantage of the
+    // optional 2nd argument to `JSON.stringify`. If anything is not a string or number
+    // we retain it so that any nested properties are handled, else we check
+    // whether it's in our 'keep' list to decide whether to keep it or replace it
+    // with `"omitted"`.
+    const anonymizedOT = JSON.parse(
+      JSON.stringify(target, (key, value) => {
+        if (!['string', 'number'].includes(typeof value)) {
+          return value;
+        }
+        if (OUTPUT_TARGET_KEYS_TO_KEEP.includes(key)) {
+          return value;
+        }
+        return 'omitted';
+      })
+    );
+
+    // this prop has to be handled separately because it is an array
+    // so the replace function above will be called with all of its
+    // members, giving us `["omitted", "omitted", ...]`.
+    //
+    // Instead, we check for its presence and manually copy over.
+    if (isOutputTargetHydrate(target) && target.external) {
+      anonymizedOT['external'] = target.external.concat();
+    }
+    return anonymizedOT;
+  });
 
   // TODO(STENCIL-469): Investigate improving anonymization for tsCompilerOptions and devServer
-  const propsToDelete: Array<keyof d.Config> = ['sys', 'logger', 'tsCompilerOptions', 'devServer'];
-
-  for (const prop of propsToDelete) {
+  for (const prop of CONFIG_PROPS_TO_DELETE) {
     delete anonymizedConfig[prop];
   }
 
--- a/src/cli/telemetry/telemetry.ts
+++ b/src/cli/telemetry/telemetry.ts
@@ -195,6 +195,7 @@ export const anonymizeConfigForTelemetry = (config: d.Config): d.Config => {
     baseUrl: 'omitted',
   }));
 
+  // TODO(STENCIL-469): Investigate improving anonymization for tsCompilerOptions and devServer
   const propsToDelete: Array<keyof d.Config> = ['sys', 'logger', 'tsCompilerOptions', 'devServer'];
 
   for (const prop of propsToDelete) {
--- a/src/cli/telemetry/telemetry.ts
+++ b/src/cli/telemetry/telemetry.ts
@@ -151,8 +151,6 @@ export const prepareData = async (
   };
 };
 
-// Get keys in `d.Config` which map to a value of type `string`
-//
 // Setting a key type to `never` excludes it from a mapped type, so we
 // can get only keys which map to a string value by excluding all keys `K`
 // where `d.Config[K]` does not extend `string`.
@@ -189,8 +187,9 @@ export const anonymizeConfigForTelemtry = (config: d.Config): d.Config => {
 
   anonymizedConfig.outputTargets = config.outputTargets.map((target) => ({
     ...target,
-    dir: `actual value for dir omitted`,
+    dir: 'omitted',
     typesDir: 'omitted',
+    baseUrl: 'omitted',
   }));
 
   const propsToDelete: Array<keyof d.Config> = ['sys', 'logger'];
--- a/src/cli/telemetry/telemetry.ts
+++ b/src/cli/telemetry/telemetry.ts
@@ -127,7 +127,7 @@ export const prepareData = async (
   const cpu_model = sys.details.cpuModel;
   const build = coreCompiler.buildId || 'unknown';
   const has_app_pwa_config = hasAppTarget(config);
-  const anonymizedConfig = anonymizeConfigForTelemtry(config);
+  const anonymizedConfig = anonymizeConfigForTelemetry(config);
 
   return {
     yarn,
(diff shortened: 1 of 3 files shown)


src/cli/telemetry/telemetry.ts
```diff
@@ -137,9 +147,77 @@ export const prepareData = async (
...
+    'tsconfig',
+  ];
+
+  for (const prop of propsToAnonymize) {
+    if (anonymizedConfig[prop] !== undefined) {
+      anonymizedConfig[prop] = 'omitted';
+    }
+  }
+
+  anonymizedConfig.outputTargets = (config.outputTargets ?? []).map((target) => ({
```
rwaskiewicz, line 190: For anonymizing our output targets, the dev server, and the TypeScript compiler options, is there a way we could recursively apply something like `ConfigStringKeys` to anything that's an object literal? I'm a bit concerned about maintaining configurations where we would have to remember the anonymize certain items as we add/change these things.

Alternatively, we could always punt on that for now. ATM, we're really only concerned with the following fields for the purposes of determining which v3 flags to set (potentially) new defaults for:
- OutputTargetHydrate#external ([source](https://github.com/ionic-team/stencil/blob/04fb83090c968010a400cee96b129f067c5cb6f0/src/declarations/stencil-public-compiler.ts#L1867))
- OutputTargetDistCustomElements#autoDefineCustomElements ([source](https://github.com/ionic-team/stencil/blob/04fb83090c968010a400cee96b129f067c5cb6f0/src/declarations/stencil-public-compiler.ts#L1932))
- OutputTargetDistCustomElements#generateTypeDeclarations ([source](https://github.com/ionic-team/stencil/blob/04fb83090c968010a400cee96b129f067c5cb6f0/src/declarations/stencil-public-compiler.ts#L1936))

We _could_ temporarily hard code these as a temporary solution (if no recursive one is possible). In that scenario, I'd be OK with omitting the ts config and dev server config entirely for now.

rwaskiewicz (changes requested):

src/cli/telemetry/telemetry.ts
```diff
@@ -99,6 +99,15 @@ export async function getActiveTargets(config: d.Config): Promise<string[]> {
...
 
+/**
+ * Prepare data for telemetry
+ *
+ * @param coreCompiler the core compiler
+ * @param config the current Stencil config
+ * @param sys the compiler system instance in use
+ * @param duration_ms the duration of the action being tracked
+ * @param component_count the number of components being built (optional)
+ */
```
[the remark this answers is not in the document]
alicewriteswrongs: I always forget that! 🤦‍♀️

src/cli/telemetry/telemetry.ts
```diff
@@ -137,9 +147,77 @@ export const prepareData = async (
...
 
+// Setting a key type to `never` excludes it from a mapped type, so we
+// can get only keys which map to a string value by excluding all keys `K`
+// where `d.Config[K]` does not extend `string`.
+type ConfigStringKeys = keyof {
+  [K in keyof d.Config as Required<d.Config>[K] extends string ? K : never]: d.Config[K];
+};
+
+/**
+ * Anonymize the config for telemtry, replacing potentially revealing config props
```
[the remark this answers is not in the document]
alicewriteswrongs: 😬

src/cli/telemetry/telemetry.ts
alicewriteswrongs: Created `STENCIL-469` for this

alicewriteswrongs:
@rwaskiewicz ok I think comments should be addressed!

src/cli/telemetry/telemetry.ts
rwaskiewicz: I don't know if that covers all of our cases here 🤔 More immediately, we also need to consider output target fields since we're going to anonymize/serialize/deserialize those.

ATM, our output targets extend one of two base types, [`OutputTargetBase`](https://github.com/ionic-team/stencil/blob/c0401477c0074f280543bc19d7893a03ed1d5d91/src/declarations/stencil-public-compiler.ts#L2009) or [`OutputTargetBaseNext`
](https://github.com/ionic-team/stencil/blob/c0401477c0074f280543bc19d7893a03ed1d5d91/src/declarations/stencil-public-compiler.ts#L1971), and each output target can have multiple fields that are specific to the output target itself.  

Because of the number of fields, I think we may need to walk the entire object. Otherwise, I don't know if we can cover/anonymize every string field in the output targets.

I think this would look something like:
```typescript
const omitOutputTargetDetails = (outputTargetPiece: any): any => {
  if (!Array.isArray(outputTargetPiece) && typeof outputTargetPiece === 'object') {
    for (const [key, value] of Object.entries(outputTargetPiece)) {
      if (typeof value === 'string') {
        outputTargetPiece[key] = 'omitted';
     } else if (Array.isArray(value)) {
        outputTargetPiece[key] = 'omitted';
    } else if (typeof outputTargetPiece === 'object') {
        outputTargetPiece[key] = omitOutputTargetDetails(outputTargetPiece[key]);
    }
  }
}; 
``` 

and call it like so:

```typescript
anonymizedConfig.outputTargets = (config.outputTargets ?? []).map((outputTarget) => omitOutputTargetDetails(outputTarget));
```

or something like that (this was all back of the napkin prototyping).  But LMK if there's something I'm missing here (that's very possible)

alicewriteswrongs: what if instead of trying to omit all of the fields on the output targets that could possibly hold identifying information, we instead opt-in to the specific fields we're interested in looking at? Something like

```ts
  anonymizedConfig.outputTargets = (config.outputTargets ?? []).map((target) => {
    const { external, autoDefineCustomElements, generateTypeDeclarations } = target;
    return { external, autoDefineCustomElements, generateTypeDeclarations };
});
```

might have to do something a bit more complicated to deal with the types, but that would be the basic idea

alicewriteswrongs: maybe we'd want `target.type` in there too

rwaskiewicz: Agreed, we'd want to keep `target.type` in there for sure.  My vote here is for us to try to be an encompassing as possible for the output targets now. If we can find/use a generalized solution (that's not too rough on performance) we should pursue that route.  Ideally I'd like us to not have to add a new key to telemetry code when we modify the shape of output targets. The reason being that in addition to having to remember/decide if the field is important enough to serialize, I think we can run into some issues with typings where a field like `autoDefineCustomElements` doesn't exist on every output target.

That said, that solution isn't all too perfect either. If we go the other way and want to omit a field entirely (like the dev server), we then _do_ have to remember to strike it from the telemetry list

alicewriteswrongs: ok so a little investigation! I implemented two different ways to recursively omit any keys in the output target objects which aren't on a 'keep' list that we define:

```ts
const OUTPUT_TARGET_KEYS_TO_KEEP = ['type', 'typesDir', 'baseUrl'];
```

The first is based on what you suggested:

```ts
  const omitOutputTargetDetails = (outputTargetPiece: any): any => {
    if (!Array.isArray(outputTargetPiece) && typeof outputTargetPiece === 'object') {
      for (const [key, value] of Object.entries(outputTargetPiece)) {
        if (typeof value === 'string') {
          if (!OUTPUT_TARGET_KEYS_TO_KEEP.includes(key)) {
            outputTargetPiece[key] = 'omitted';
          }
        } else if (Array.isArray(value)) {
          outputTargetPiece[key] = 'omitted';
        } else if (typeof outputTargetPiece === 'object') {
          outputTargetPiece[key] = omitOutputTargetDetails(outputTargetPiece[key]);
        }
      }
    }
    return outputTargetPiece;
  };

  anonymizedConfig.outputTargets = (config.outputTargets ?? []).map(omitOutputTargetDetails);
```

the second uses `JSON.stringify`, passing the optional [2nd argument](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#the_replacer_parameter):


```ts
  anonymizedConfig.outputTargets = JSON.parse(
    JSON.stringify(config.outputTargets ?? [], (key, value) => {
      if (!['string', 'number'].includes(typeof value)) {
        return value;
      }
      if (OUTPUT_TARGET_KEYS_TO_KEEP.includes(key)) {
        return value;
      }
      return 'omitted';
    })
  );
```

I wrote a little hacky 'performance' test just to get a ballpark idea of how these perform:

```ts
  it.only('test performance of OT anonymizers', () => {
    const config = {
      flags: {
        args: [],
      },
      outputTargets: [
        { type: 'www', baseUrl: 'https://example.com', serviceWorker: { swDest: './tmp' } }
      ],
    } as d.Config;    

    let beforeJSON = new Date();
    for (let i = 0; i < 1000; i++) {
      anonymizeConfigForTelemetryJSON(config)
    }
    let afterJSON = new Date()
    // @ts-ignore
    let JSONDifference = afterJSON - beforeJSON;
    console.log(`JSON version took ${JSONDifference}`);

    let beforenonJSON = new Date();
    for (let i = 0; i < 10000; i++) {
      anonymizeConfigForTelemetry(config)
    }
    let afternonJSON = new Date()
    // @ts-ignore
    let nonJSONDifference = afternonJSON - beforenonJSON;
    console.log(`nonJSON version took ${nonJSONDifference}`);
  });
```

The results were kind of surprising, but possibly make sense? I thought that the `JSON.parse(JSON.stringify( ... ))` would be slower because it's basically doing some 'unnecessary' overhead work serializing and deserializing, but actually that version is quite a bit faster - anonymizing the above config 10000 times takes ~4-5ms with that version vs ~31-35ms for the recursive function call version. I suspect this is just because w/ `JSON.stringify` the descent through the object is handled by native code, rather than recursive function calls, `Object.entries`, etc.

So anywhoo - that's what I've found. I _personally_ find the `JSON.{stringify,parse}` version to be quite a bit more elegant, letting us just write a little function which decides what to do based on each `k, v` pair (sort of a context-free way to handle it, if you will) and we don't have to include any logic for iterating, recurring, etc.

What do you think?

alicewriteswrongs:
@rwaskiewicz ok so I just pushed the change to use `JSON.stringify` and `JSON.parse` to anonymize the output targets and added a few tests, should be ready for re-review now 🙂

src/cli/telemetry/telemetry.ts
```diff
@@ -159,8 +160,10 @@ type ConfigStringKeys = keyof {
   [K in keyof d.Config as Required<d.Config>[K] extends string ? K : never]: d.Config[K];
 };
 
+const OUTPUT_TARGET_KEYS_TO_KEEP = ['type', 'typesDir', 'baseUrl'];
+
 /**
- * Anonymize the config for telemetry, replacing potentially revealing config props
+ * Anonymize the config for telemetry, replacing potentially revealing configprops
```
rwaskiewicz, line 166: Minor, was this 'configprops' change intentional in the most recent commit? I don't see any other usages of this spelling anywhere

src/cli/telemetry/telemetry.ts
```diff
@@ -159,8 +160,10 @@ type ConfigStringKeys = keyof {
   [K in keyof d.Config as Required<d.Config>[K] extends string ? K : never]: d.Config[K];
 };
 
+const OUTPUT_TARGET_KEYS_TO_KEEP = ['type', 'typesDir', 'baseUrl'];
```
rwaskiewicz, line 163: Should we keep the 'typesDir'? Isn't it possible that folks will be providing an absolute path here? (It actually may get resolved to a full path by the compiler 🤔)

rwaskiewicz (commented):
Tested locally, everything looks good from an output standpoint 👍 

I had a few questions about our exclusions + a few tiny nits.

src/cli/telemetry/telemetry.ts
```diff
@@ -188,12 +191,22 @@ export const anonymizeConfigForTelemetry = (config: d.Config): d.Config => {
...
   }
 
-  anonymizedConfig.outputTargets = (config.outputTargets ?? []).map((target) => ({
-    ...target,
-    dir: 'omitted',
-    typesDir: 'omitted',
-    baseUrl: 'omitted',
-  }));
+  // Anonymize the outputTargets on our configuration, taking advantage of the
+  // optional 2nd argument to JSON.parse. If anything is not a string or number
```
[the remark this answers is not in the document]
alicewriteswrongs: 🤦‍♀️ yes it should be haha

src/cli/telemetry/telemetry.ts
```diff
@@ -188,12 +191,22 @@ export const anonymizeConfigForTelemetry = (config: d.Config): d.Config => {
...
+  // we retain it so that any nested properties are handled, else we check
+  // whether it's in our 'keep' list to decide whether to keep it or replace it
+  // with `"omitted"`.
+  anonymizedConfig.outputTargets = JSON.parse(
+    JSON.stringify(config.outputTargets ?? [], (key, value) => {
+      if (!['string', 'number'].includes(typeof value)) {
```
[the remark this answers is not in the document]
alicewriteswrongs: I didn't see anything in particular, and I do have a _bit_ of a hard time thinking of an example, but I just went with that to be on the safe side really. We could probably just check `!(typeof value === "string")` here if we wanted to keep it simple

src/cli/telemetry/telemetry.ts
alicewriteswrongs: yeah total brain fart on my end haha

I think the props we want to keep are:

```ts
const OUTPUT_TARGET_KEYS_TO_KEEP = [
  'type', 
  'external',
  'autoDefineCustomElements',
  'generateTypeDeclarations',
];
```

src/cli/telemetry/telemetry.ts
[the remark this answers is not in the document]
rwaskiewicz: My vote would be keeping it simple for now. Looking at the output target definitions, nothing jumps out at me as something that would be an issue here

src/cli/telemetry/telemetry.ts
rwaskiewicz: I think since the latter three are booleans (or, are supposed to be booleans) we can omit those as well

rwaskiewicz:
@alicewriteswrongs it looks like the build is failing on this PR - can you PTAL?

alicewriteswrongs:
@rwaskiewicz sorry for the noise, I pushed a few in-progress commits, I think it should be good now

src/cli/telemetry/telemetry.ts
```diff
@@ -137,13 +149,101 @@ export const prepareData = async (
...
+
+  for (const prop of CONFIG_PROPS_TO_ANONYMIZE) {
+    if (anonymizedConfig[prop] !== undefined) {
+      anonymizedConfig[prop] = 'omitted';
+    }
+  }
+
+  anonymizedConfig.outputTargets = (config.outputTargets ?? []).map((target) => {
+    // Anonymize the outputTargets on our configuration, taking advantage of the
+    // optional 2nd argument to `JSON.stringify`. If anything is not a string or number
```
rwaskiewicz, line 207:
```suggestion
    // optional 2nd argument to `JSON.stringify`. If anything is not a string
```

rwaskiewicz (approved):
LGTM! one non-blocking comment I think I can take care of in the GH UI

PR #3408, opened 2022-06-06
feat(mock-doc): add matrix and tspan props for svgelement

marlon-ionic:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Some methods and properties of SVGElement and its children elements are missing from `mock-doc`. 

## What is the new behavior?

Per request from ZenDesk ticket 41657:
- Added following members were added to `SVGElement`: `createSVGPoint()`, `getBBox()`, `getComputedTextLength()`, `getCTM()`, `getScreenCTM()`
- Added `focus()` and `blur()` to `MockElement`

- [ ] Yes
- [x] No

## Testing

Added test suites for the new svg members, one for the matrix-based values and another for getComputedTextLength() since that's related to SVG text elements. The thought was that if additional mockdoc updates are needed in the future, separating the suites would help with code organization.


src/mock-doc/element.ts
```diff
@@ -267,6 +267,27 @@ export class MockStyleElement extends MockHTMLElement {
   }
 }
 
+//Based on deprecated SVGMatrix - http://html5index.org/SVG%20-%20SVGMatrix.html
+export const defaultSVGMatrix = {
+  a: 1,
```
rwaskiewicz, line 272: For my own understanding, why do `a` and `d` get assigned a value of 1, but the other single letter properties of the matrix are assigned a value of 0?

src/mock-doc/element.ts
```diff
@@ -267,6 +267,27 @@ export class MockStyleElement extends MockHTMLElement {
   }
 }
 
+//Based on deprecated SVGMatrix - http://html5index.org/SVG%20-%20SVGMatrix.html
```
rwaskiewicz, line 270:
```suggestion
// Based on deprecated SVGMatrix - http://html5index.org/SVG%20-%20SVGMatrix.html
```

src/mock-doc/element.ts
```diff
@@ -297,6 +314,25 @@ export class MockSVGElement extends MockElement {
...
+  }
+  getBBox() {
+    return { x: 0, y: 0, width: 10, height: 10 };
+  }
+  getCTM() {
+    return defaultSVGMatrix;
+  }
+  getScreenCTM() {
+    return defaultSVGMatrix;
+  }
```
rwaskiewicz, line 335: These methods appear to not exist on `SVGElement`, but rather it's[ child class `SVGGraphicsElement`](https://developer.mozilla.org/en-US/docs/Web/API/SVGGraphicsElement).  Can we please move these to a (new) `MockSVGGraphicsElement` class that extends `MockSVGElement`? The reason being is that these methods shouldn't be available on `MockSVGElement` (and by extension, any of its child classes except `MockSVGGraphicsElement`)

src/mock-doc/element.ts
```diff
@@ -297,6 +314,25 @@ export class MockSVGElement extends MockElement {
...
+  createSVGPoint() {
+    return {
+      matrixTransform: defaultSVGMatrix,
+      x: 0,
+      y: 0,
+    };
+  }
+  getBBox() {
+    return { x: 0, y: 0, width: 10, height: 10 };
```
rwaskiewicz, line 328: Can we also return the [`top`, `right`, `bottom` and `left` properties](https://developer.mozilla.org/en-US/docs/Web/API/DOMRect#properties) on the `DOMRect` that's returned here to adhere to the interface a little better?

```suggestion
    return { x: 0, y: 0, width: 10, height: 10, left: 0, top: 0, right: 10, right: 10 };
```

src/mock-doc/element.ts
```diff
@@ -297,6 +314,25 @@ export class MockSVGElement extends MockElement {
   getTotalLength(): number {
     return 0;
   }
+  getComputedTextLength(): number {
+    return 0;
+  }
```
rwaskiewicz, line 319: This method appears to not exist on `SVGElement`, but [rather on `SVGTextContentElement`](https://developer.mozilla.org/en-US/docs/Web/API/SVGTextContentElement). Can we please move these to a (new) `SVGTextContentElement` class that extends (a new) `MockSVGGraphicsElement` that extends `MockSVGElement`? The reason being is that these methods shouldn't be available on MockSVGElement (and by extension, any of its child classes except `SVGTextContentElement`)

src/mock-doc/element.ts
rwaskiewicz, line 335: Can we add return types to the function signatures added here please, since they don't return `undefined`/`void`?

src/mock-doc/element.ts
```diff
@@ -297,6 +314,25 @@ export class MockSVGElement extends MockElement {
...
+  createSVGPoint() {
```
rwaskiewicz, line 320: This method appears to not exist on `SVGElement`, but [rather on `SVGPoint`](https://www.w3.org/TR/SVG11/coords.html#InterfaceSVGPoint). Can we please move these to a (new) `SVGTextContentElement` class that extends (a new) `MockSSVGPoint`?  The reason being is that these methods shouldn't be available on MockSVGElement (and by extension, any of its child classes)

src/mock-doc/element.ts
```diff
@@ -297,6 +314,25 @@ export class MockSVGElement extends MockElement {
...
+    return {
+      matrixTransform: defaultSVGMatrix,
```
rwaskiewicz, line 322: Looking at https://www.w3.org/TR/SVG11/coords.html#InterfaceSVGPoint, should this be a callable function rather than an attribute on the object?

```suggestion
      matrixTransform: () => defaultSVGMatrix,
```

src/mock-doc/element.ts
```diff
@@ -297,6 +314,25 @@ export class MockSVGElement extends MockElement {
...
+      x: 0,
+      y: 0,
+    };
+  }
+  getBBox() {
+    return { x: 0, y: 0, width: 10, height: 10 };
+  }
+  getCTM() {
+    return defaultSVGMatrix;
```
rwaskiewicz, line 331: I think the return value of `getCTM` needs to be a [DOMMatrix](https://developer.mozilla.org/en-US/docs/Web/API/DOMMatrix). Although it's similar to an `SVGMatrix`, it looks like there are different properties/methods on the two

src/mock-doc/element.ts
```diff
@@ -297,6 +314,25 @@ export class MockSVGElement extends MockElement {
...
+  }
+  getScreenCTM() {
+    return defaultSVGMatrix;
```
rwaskiewicz, line 334: I think the return value of `getScreenCTM` needs to be a [DOMMatrix](https://developer.mozilla.org/en-US/docs/Web/API/DOMMatrix). Although it's similar to an `SVGMatrix`, it looks like there are different properties/methods on the two

rwaskiewicz (changes requested):

src/mock-doc/node.ts
```diff
@@ -316,6 +320,10 @@ export class MockElement extends MockNode {
     return this.children[0] || null;
   }
 
+  focus() {
```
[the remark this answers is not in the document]
marlon-ionic:
```suggestion
  focus(_options?: { preventScroll?: boolean }) {
```
Since `options` is optional too right?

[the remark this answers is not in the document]
rwaskiewicz: Yep! Exactly

src/mock-doc/element.ts
marlon-ionic: Sure but can I use named types like `DOMMatrix`, or `DOMRect`? or should the return types be anonymous? I'm not sure what's available while running in the testing context.

rwaskiewicz: I think you can, there's prior art here in this file for using types like `DOMPoint` and `SVGSVGElement`

src/mock-doc/element.ts
marlon-ionic: It looks like `createSVGPoint()` exists in [SVGSVGElement](https://www.w3.org/TR/SVG/struct.html#InterfaceSVGSVGElement) so I created that instead

src/mock-doc/element.ts
```diff
@@ -237,6 +258,95 @@ patchPropAttributes(MockScriptElement.prototype, {
...
 });
 
+export class MockDOMMatrix {
+  static fromMatrix() {
+    return new MockDOMMatrix();
+  }
+  a: number = 1;
+  b: number = 0;
+  c: number = 0;
+  d: number = 1;
```
rwaskiewicz, line 268: I had this question in the [original implementation](https://github.com/ionic-team/stencil/pull/3408#discussion_r891194336) that I think still stands

> For my own understanding, why do `a` and `d` get assigned a value of 1, but the other single letter properties of the matrix are assigned a value of 0?

rwaskiewicz (commented):
Looks good! One outstanding question/comment

src/mock-doc/element.ts
marlon-ionic: Oh sorry, I missed responding to this one! I don't know. I simply used the default values returned when a DOMMatrix is instantiated (see screenshot).
Screen Shot 2022-06-13 at 11 18 30 AM

rwaskiewicz (changes requested):
Thanks @marlon-ionic! All in all this looks good to me.

The last thing to do is to is resolve a few `strictNullChecks` errors that this PR is introducing.  They're a bit hard to find at the moment, as it looks like our CI check for these isn't working correctly against forks.  I took the time to find/fix the violations that our tests are introducing, which you can see in the diff below. Can you apply  the changes to your fork please? (I've also included a patch file so that you can apply all of the changes all at once)

```diff
diff --git a/src/mock-doc/test/html-parse.spec.ts b/src/mock-doc/test/html-parse.spec.ts
index e015311d8..eb3286ae9 100644
--- a/src/mock-doc/test/html-parse.spec.ts
+++ b/src/mock-doc/test/html-parse.spec.ts
@@ -96,7 +96,8 @@ describe('parseHtml', () => {
         </svg>
       </svg>
     `);
-    const svgElem: MockSVGSVGElement = doc.body.firstElementChild.firstElementChild as MockSVGSVGElement;
+    const svgElem: MockSVGSVGElement = doc.body.firstElementChild?.firstElementChild as MockSVGSVGElement;
+    expect(svgElem).toBeDefined();
     expect(svgElem.getBBox()).toEqual(new MockSVGRect());
     expect(svgElem.createSVGPoint()).toEqual(new MockDOMPoint());
     expect(svgElem.getScreenCTM()).toEqual(new MockDOMMatrix());
@@ -112,9 +113,12 @@ describe('parseHtml', () => {
         </text>
       </svg>
     `);
-    const tspan: MockSVGTextContentElement = doc.body.firstElementChild.firstElementChild
-      .firstElementChild as MockSVGTextContentElement;
-    expect(doc.body.firstElementChild.firstElementChild.tagName).toEqual('text');
+    const text: MockSVGTextContentElement = doc.body.firstElementChild?.firstElementChild as MockSVGTextContentElement;
+    expect(text).toBeDefined();
+    expect(text.tagName).toEqual('text');
+
+    const tspan: MockSVGTextContentElement = text.firstElementChild as MockSVGTextContentElement;
+    expect(tspan).toBeDefined();
     expect(tspan.getComputedTextLength()).toEqual(0);
   });
 ```
[rm-null-errors.txt](https://github.com/ionic-team/stencil/files/8893763/rm-null-errors.txt)

rwaskiewicz (approved):

rwaskiewicz:
I'm merging this as an admin - the tech debt burndown (download error files and report) has a bug in it where it doesn't work against forks of the repo. All other CI checks are passing, and no new `strictNullChecks` violations nor unused exports are being introduced in this PR.

PR #3427, opened 2022-06-16
refactor(cli): refactor parse-flags module

alicewriteswrongs:
this refactors the parse-flags module, removing some code duplication
and changing how the `ConfigFlags` interface is declared.

Now the ConfigFlags interface is generated based on the content of a few
read-only arrays of strings. This means that config flags are listed in
one and only one placed and cannot drift / get out of sync.

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Currently we define a `ConfigFlags` object which defines CLI arguments that Stencil supports. We also have lists of supported CLI arguments in `src/cli/parse-flags.ts`. Although those lists of arguments (basically `ReadonlyArray<string>`) are actually used to construct the `ConfigFlags` object, we don't currently have a type-system-level guarantee of consistency between what is actually pulled out and stuck onto the `ConfigFlags` object and what is actually declared as a member of the `ConfigFlags` interface. This creates the possibility of subtle and confusing bugs when there is drift between the `ConfigFlags` object and our CLI arg parsing code in `src/cli/parse-flags.ts`.

I wrote more about this here: #3404 that PR fixed a bug that was introduced precisely because of a drift between these two things.

## What is the new behavior?

This PR changes the definition of the `ConfigFlags` object so that it is instead derived from our lists of supported arguments. This means that we used these arrays of strings at runtime to match arguments _and_ we push them into the type system, letting us ensure that things are consistent and that we declare supported arguments in one place.

- [ ] Yes
- [x] No

## Testing

I took some care to ensure that this refactor does not change any functionality. I don't believe I have changed anything - my confidence in that is helped by the fact that the test suite for the relevant module is passing with no modifications (I just added a few tests) and some smoke testing. But it should probably be tested locally more thoroughly!

### DIFF
--- a/scripts/bundles/cli.ts
+++ b/scripts/bundles/cli.ts
@@ -45,23 +45,6 @@ export async function cli(opts: BuildOptions): Promise<ReadonlyArray<RollupOptio
   dts = dts.replace('@stencil/core/internal', '../internal/index');
   await fs.writeFile(join(opts.output.cliDir, dtsFilename), dts);
 
-  // bundle config-flags.js
-  const esConfigFlagsOutput: OutputOptions = {
-    format: 'es',
-    file: join(outputDir, 'config-flags.js'),
-    preferConst: true,
-    sourcemap: true,
-    banner: getBanner(opts, `Stencil CLI Config Flags`, true),
-  };
-
-  const cjsConfigFlagsOutput: OutputOptions = {
-    format: 'cjs',
-    file: join(outputDir, 'config-flags.cjs'),
-    preferConst: true,
-    sourcemap: true,
-    banner: getBanner(opts, `Stencil CLI Config Flags (CommonJS)`, true),
-  };
-
   // copy config-flags.d.ts
   let configDts = await fs.readFile(join(inputDir, 'config-flags.d.ts'), 'utf8');
   configDts = configDts.replace('@stencil/core/declarations', '../internal/index');
@@ -78,7 +61,7 @@ export async function cli(opts: BuildOptions): Promise<ReadonlyArray<RollupOptio
 
   const cliBundle: RollupOptions = {
     input: join(inputDir, 'index.js'),
-    output: [esOutput, cjsOutput, esConfigFlagsOutput, cjsConfigFlagsOutput],
+    output: [esOutput, cjsOutput],
     external: ['path'],
     plugins: [
       relativePathPlugin('@stencil/core/testing', '../testing/index.js'),
--- a/scripts/bundles/cli.ts
+++ b/scripts/bundles/cli.ts
@@ -62,7 +62,6 @@ export async function cli(opts: BuildOptions): Promise<ReadonlyArray<RollupOptio
     banner: getBanner(opts, `Stencil CLI Config Flags (CommonJS)`, true),
   };
 
-
   // copy config-flags.d.ts
   let configDts = await fs.readFile(join(inputDir, 'config-flags.d.ts'), 'utf8');
   configDts = configDts.replace('@stencil/core/declarations', '../internal/index');
--- a/scripts/bundles/cli.ts
+++ b/scripts/bundles/cli.ts
@@ -48,15 +48,15 @@ export async function cli(opts: BuildOptions): Promise<ReadonlyArray<RollupOptio
   // copy config-flags.js
   const configFlags: OutputOptions = {
     format: 'es',
-    file: join(outputDir, "config-flags.js"),
+    file: join(outputDir, 'config-flags.js'),
     preferConst: true,
     sourcemap: true,
     banner: getBanner(opts, `Stencil CLI Config Flags`, true),
-  }
+  };
 
   // copy config-flags.d.ts
   let configDts = await fs.readFile(join(inputDir, 'config-flags.d.ts'), 'utf8');
-  configDts = configDts.replace('@stencil/core/internal', '../internal/index');
+  configDts = configDts.replace('@stencil/core/declarations', '../internal/index');
   await fs.writeFile(join(opts.output.cliDir, 'config-flags.d.ts'), configDts);
 
   // write @stencil/core/compiler/package.json
--- a/scripts/bundles/cli.ts
+++ b/scripts/bundles/cli.ts
@@ -45,6 +45,20 @@ export async function cli(opts: BuildOptions): Promise<ReadonlyArray<RollupOptio
   dts = dts.replace('@stencil/core/internal', '../internal/index');
   await fs.writeFile(join(opts.output.cliDir, dtsFilename), dts);
 
+  // copy config-flags.js
+  const configFlags: OutputOptions = {
+    format: 'es',
+    file: join(outputDir, "config-flags.js"),
+    preferConst: true,
+    sourcemap: true,
+    banner: getBanner(opts, `Stencil CLI Config Flags`, true),
+  }
+
+  // copy config-flags.d.ts
+  let configDts = await fs.readFile(join(inputDir, 'config-flags.d.ts'), 'utf8');
+  configDts = configDts.replace('@stencil/core/internal', '../internal/index');
+  await fs.writeFile(join(opts.output.cliDir, 'config-flags.d.ts'), configDts);
+
   // write @stencil/core/compiler/package.json
   writePkgJson(opts, opts.output.cliDir, {
     name: '@stencil/core/cli',
@@ -56,7 +70,7 @@ export async function cli(opts: BuildOptions): Promise<ReadonlyArray<RollupOptio
 
   const cliBundle: RollupOptions = {
     input: join(inputDir, 'index.js'),
-    output: [esOutput, cjsOutput],
+    output: [esOutput, cjsOutput, configFlags],
     external: ['path'],
     plugins: [
       relativePathPlugin('@stencil/core/testing', '../testing/index.js'),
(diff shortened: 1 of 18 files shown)


scripts/bundles/cli.ts
```diff
@@ -45,6 +45,28 @@ export async function cli(opts: BuildOptions): Promise<ReadonlyArray<RollupOptio
...
   await fs.writeFile(join(opts.output.cliDir, dtsFilename), dts);
 
+  // bundle config-flags.js
+  const esConfigFlagsOutput: OutputOptions = {
+    format: 'es',
+    file: join(outputDir, 'config-flags.js'),
+    preferConst: true,
+    sourcemap: true,
+    banner: getBanner(opts, `Stencil CLI Config Flags`, true),
+  };
```
alicewriteswrongs, line 55: this is my first time messing around in these bundle files, so I'm not sure this is the best approach, that we want to do this, etc.

There is also some slightly strange stuff going on here. For instance, for the `cli` module in general, we currently do not bundle the `index.d.ts` file which is generated by Typescript during compilation, but instead we read file `src/cli/public.d.ts` from disk, replace a string import with a different one to make the path to the stencil declarations resolve correctly, and then we write that to disk as `${outputDir}/cli/index.d.ts`. Why is this done in such a convoluted and strange way? It is very brittle and rather confusing I think.

Anyhow, for my change in order for the build to work I need to bundle and copy over `build/cli/config-flags.js` and `build/cli/config-flags.d.ts`. I made changes to do that, but just wanted to flag this since a) the overall way this aspect of building stencil is handled is strange to me and b) I'm not 100% that the way I did this is the best way to do so.

src/cli/config-flags.ts
```diff
@@ -0,0 +1,137 @@
+import type { LogLevel, TaskCommand } from '@stencil/core/declarations';
```
alicewriteswrongs, line 1: This PR creates a new module here which exports readonly arrays of all our supported CLI args and uses those to programmatically construct an interface (`ConfigFlags`) into which all of our CLI args get parsed. Because this is created on the basis of readonly arrays of strings, we have access to those keys both at runtime _and_ in the type system, so we can use them to write our actual arg parsing code and to create a type which describes the object we're going to parse them into. This lets us write a given flag down only in one place and have some consistency in the world.

src/cli/config-flags.ts
```diff
@@ -0,0 +1,137 @@
...
+export const STRING_CLI_ARGS = ['address', 'config', 'docsJson', 'emulate', 'root', 'screenshotConnector'] as const;
+
+/**
+ * All the LogLevel-type options supported by the Stencil CLI
+ *
+ * This is a bit silly since there's only one such argument atm,
+ * but this approach lets us make sure that we're handling all
+ * our arguments in a type-safe way.
+ */
+export const LOG_LEVEL_CLI_ARGS = ['logLevel'] as const;
```
alicewriteswrongs, line 57: this is obvs pretty similar to a string arg, but because we parse it into a `LogLevel` union type which is `"info" | "warn" | "debug" | "error"` we need to handle it separately

src/cli/parse-flags.ts
```diff
@@ -35,203 +54,197 @@ export const parseFlags = (args: string[], sys?: CompilerSystem): ConfigFlags =>
...
  */
-const parseArgs = (flags: any, args: string[], knownArgs: string[]) => {
-  BOOLEAN_ARG_OPTS.forEach((booleanName) => {
-    const alias = ARG_OPTS_ALIASES[booleanName];
-    const flagKey = configCase(booleanName);
+const parseArgs = (flags: ConfigFlags, args: string[]) => {
+  BOOLEAN_CLI_ARGS.forEach((argName) => parseBooleanArg(flags, args, argName));
+  STRING_CLI_ARGS.forEach((argName) => parseStringArg(flags, args, argName));
+  NUMBER_CLI_ARGS.forEach((argName) => parseNumberArg(flags, args, argName));
+  LOG_LEVEL_CLI_ARGS.forEach((argName) => parseLogLevelArg(flags, args, argName));
```
alicewriteswrongs, line 77: here I created a separate arg parsing function for each support arg type (boolean, string, etc). I was able to share some logic between some of them (see the `getValue`and `getEqualsValue` function below) and ensure that everything is typed correctly (no `any`)

alicewriteswrongs: the actual logic for how arguments are parsed should be the same as before the refactor, the test suite passes 100% without changing any tests (I added a few and added some more checks to one or two)

alicewriteswrongs (commented):
just calling out a few things

src/cli/parse-flags.ts
```diff
@@ -35,203 +54,197 @@ export const parseFlags = (args: string[], sys?: CompilerSystem): ConfigFlags =>
...
+    flags[configCaseName] = null;
+  }
 
-    for (let i = 0; i < args.length; i++) {
-      const cmdArg = args[i];
-
-      if (cmdArg.startsWith(`--${numberName}=`)) {
-        const values = cmdArg.split('=');
-        values.shift();
-        flags[flagKey] = parseInt(values.join(''), 10);
```
rwaskiewicz, line 140: We may be introducing a subtle bug here by not having a `.join()` call for the new `parseNumberArg` impl, but that seems pretty far-fetched to me.  I'm in favor of taking that risk here, especially if our tests are passing.

scripts/bundles/cli.ts
```diff
@@ -45,6 +45,28 @@ export async function cli(opts: BuildOptions): Promise<ReadonlyArray<RollupOptio
...
+
+  const cjsConfigFlagsOutput: OutputOptions = {
+    format: 'cjs',
+    file: join(outputDir, 'config-flags.cjs'),
+    preferConst: true,
+    sourcemap: true,
+    banner: getBanner(opts, `Stencil CLI Config Flags (CommonJS)`, true),
+  };
```
rwaskiewicz, line 63: I don't think something's right here - if I build this branch, I see the generated `config-flags` files, but it looks like they define/re-export functions like `run`, `runTask`.  Those functions are _also_ defined in `index.js/cjs`. Is that intentional?

rwaskiewicz (changes requested):

src/cli/config-flags.ts
```diff
@@ -0,0 +1,137 @@
...
+
+/**
+ * All the Number options supported by the Stencil CLI
+ */
+export const NUMBER_CLI_ARGS = ['maxWorkers', 'port'] as const;
+
+/**
+ * All the String options supported by the Stencil CLI
+ */
+export const STRING_CLI_ARGS = ['address', 'config', 'docsJson', 'emulate', 'root', 'screenshotConnector'] as const;
```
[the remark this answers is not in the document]
alicewriteswrongs: that was an oversight! I'll pull it over. I cannot find any references to it in the codebase however, so not sure if we do need it, but I will keep it in for now in the interest of not having any unintended consequences for this refactor

src/cli/config-flags.ts
```diff
@@ -0,0 +1,137 @@
...
+  [key in K[number]]?: T | null;
+};
+
+/**
+ * Type containing the possible Boolean configuration flags, to be included
+ * in ConfigFlags, below
+ */
+type BooleanConfigFlags = ObjectFromKeys<typeof BOOLEAN_CLI_ARGS, boolean>;
+/**
+ * Type containing the possible Boolean configuration flags, to be included
```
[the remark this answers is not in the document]
alicewriteswrongs: 🙃 haha

src/cli/parse-flags.ts
```diff
@@ -1,28 +1,47 @@
...
   if (sys && sys.name === 'node') {
     const envArgs = getNpmConfigEnvArgs(sys);
-    parseArgs(flags, envArgs, flags.knownArgs);
+    parseArgs(flags, envArgs);
 
     envArgs.forEach((envArg) => {
-      if (!flags.args.includes(envArg)) {
-        flags.args.push(envArg);
+      if (!flags.args!.includes(envArg)) {
+        flags.args!.push(envArg);
```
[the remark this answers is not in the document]
alicewriteswrongs: I did consider and try out making the type for `ConfigFlags` something like this:

```ts
export interface ConfigFlags extends BooleanConfigFlags, StringConfigFlags, NumberConfigFlags, LogLevelFlags {
  task: TaskCommand | null;
  args: string[];
  knownArgs: string[];
  unknownArgs: string[];
}
```

I really prefer that typing over what it is now:

```ts
export interface ConfigFlags extends BooleanConfigFlags, StringConfigFlags, NumberConfigFlags, LogLevelFlags {
  task?: TaskCommand | null;
  args?: string[];
  knownArgs?: string[];
  unknownArgs?: string[];
}
```

the reason I decided to leave it off of this changeset is just that I wanted to keep this as much an internal refactor w/in the `parse-flags` module as I could, and changing these types necessitates a bunch of changes in unrelated code, will most likely leave a bunch of scattered and no-longer-necessary nullchecks around, etc. But I agree we should make a change like this if we change the type of `Config` to make the `.flags` property required!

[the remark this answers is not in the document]
alicewriteswrongs: I'll make that `Array.isArray(args) ? ...` change, I think that makes sense

scripts/bundles/cli.ts
alicewriteswrongs: hmm that's not what I intended...

what I _wanted_ to do was just add this new `config-flags` module but not have to add a new file to the build, by just having it get bundled into the `index.js` output for the whole `cli` folder but doing an `export * from ...` pointing to the `config-flags` module lead to a bunch of other errors...I'll play around with it a little more and see what I can figure out.

PR #3430, opened 2022-06-17
chore(lint): add lint for @returns

rwaskiewicz:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

GitHub Issue Number: N/A

## What is the new behavior?

this commit enables multiple lint rules provided by the jsdoc plugin
regarding `@returns` in jsdocs.

the intent behind this commit is to ensure that jsdocs have a documented
`@returns` field and description where applicable, to catch missing ones
in ci rather than in code review.

there is one known limitation of these rules - when adding a jsdoc to a
function that also has a decorator applied:

```typescript
// This will error, since SomeDecorator must return a value, and 
// getLuckyNumber doesn't
/**
 * Prints a lucky number
 */
 @SomeDecorator()
 function getLuckyNumber(): void {
   console.log(7);
 }
```

the rule will error based on the return type of the decorator, rather
than the function.

there was only a single violation of this in all of stencil's code,
which was removed. i'm inclined to use this rule and use 'eslint-ignore'
directives if absolutely necessary, but would be open to arguments
against it

- [ ] Yes
- [x] No

## Testing

Lint passes 
## Other information

Prettier doesn't appear to be running against the `.eslintrc.js` file. I'll save that for another day

### DIFF
--- a/src/runtime/test/fixtures/cmp-a.tsx
+++ b/src/runtime/test/fixtures/cmp-a.tsx
@@ -73,9 +73,6 @@ export class CmpA {
   // * Method Definitions *
   // **********************
 
-  /**
-   * Method initialize
-   */
   @Method()
   init(): Promise<void> {
     return Promise.resolve(this._init());
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -38,6 +38,11 @@ module.exports = {
     // rely on TypeScript types to be the source of truth, minimize verbosity in comments
     "jsdoc/require-param-type": ["off"],
     "jsdoc/require-param-description": ["error"],
+    "jsdoc/require-returns-check": ["error"],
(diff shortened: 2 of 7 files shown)


src/runtime/test/fixtures/cmp-a.tsx
```diff
@@ -73,9 +73,6 @@ export class CmpA {
   // * Method Definitions *
   // **********************
 
-  /**
-   * Method initialize
-   */
```
rwaskiewicz, line 78: This is the removed JSDoc I mentioned in the PR summary

alicewriteswrongs (approved):
lgtm! :+1:

PR #3431, opened 2022-06-21
chore(tests): silence console.{log|warn|error} messages during tests

alicewriteswrongs:
this adds a little utility for replacing the `console` functions with
mocks to silence any error messages and get our jest output under
control.

This function calls `afterAll` internally, so it will clean up after
itself by restoring the original functions for `console.log`, `.error`,
etc.

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Other (please describe): testing quality-of-life improvement

GitHub Issue Number: N/A

## What is the new behavior?

Gaze upon this and feel at peace:

https://user-images.githubusercontent.com/6207644/174887868-ec9d7eb7-6862-471a-8cf0-91fec3ebdfa1.mov

- [ ] Yes
- [x] No

## Testing

This is just a testing change anyhow, so I think if tests are passing we are all good.

I did confirm that this setup does not inadvertently silence tests outside of the suites in which it's called.

### DIFF
--- a/src/testing/testing-utils.ts
+++ b/src/testing/testing-utils.ts
@@ -185,7 +185,7 @@ export async function withSilentWarn<T>(cb: SilentWarnFunc<T>): Promise<T> {
   const realWarn = console.warn;
   const warnMock = jest.fn();
   console.warn = warnMock;
-  const retval = await cb(warnMock);
+  const retVal = await cb(warnMock);
   console.warn = realWarn;
-  return retval;
+  return retVal;
 }
--- a/src/testing/testing-utils.ts
+++ b/src/testing/testing-utils.ts
@@ -95,3 +95,72 @@ function getAppUrl(config: d.Config, browserUrl: string, appFileName: string) {
 
   return browserUrl;
 }
+
+/**
+ * Utility for silencing `console` functions in tests.
+ *
+ * When this function is first called it grabs a reference to the `log`,
+ * `error`, and `warn` functions on `console` and then returns a per-test setup
+ * function which sets up a fresh set of mocks (via `jest.fn()`) and then
+ * assigns them to each of these functions. This setup function will return a
+ * reference to each of the three mock functions so tests can make assertions
+ * about their calls and so on.
+ *
+ * Because references to the original `.log`, `.error`, and `.warn` functions
+ * exist in closure within the function, it can use an `afterAll` call to clean
+ * up after itself and ensure that the original implementations are restored
+ * after the test suite finishes.
+ *
+ * An example of using this to silence log statements in a single test could look
+ * like this:
+ *
+ * ```ts
+ * describe("my-test-suite", () => {
+ *   const setupConsoleMocks = setupConsoleMocker()
+ *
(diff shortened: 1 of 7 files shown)


rwaskiewicz (commented):
The code/PR itself looks good to me.

I am a little concerned that we're suppressing error messages that we should be addressing in the test suites themselves though - 

For example, the log output from `src/compiler/style/test/optimize-css.spec.ts` is a consequence of how Autoprefixer is cached/installed. We have a task in the backlog to fix this (STENCIL-314). In fact, I managed to fix this once (STENCIL-131), but couldn't tell you for the life of me how it happened 😨 Unfortunately, not seeing that output is the best Acceptance Criteria that we have ATM to verify that we've fixed the caching of certain dependencies. As a result, I'm a little hesitant to suppress the output for this file.

Similarly, for the tests found in the `src/runtime/test` directory, I wonder how many of those tests can be fixed/reworked such that the original intent of the tests can be captured, but not require us to test in such a way that causes these errors to log.  Do we know if that's possible for those suites?

alicewriteswrongs:
thanks for that context, I ran across the code in `src/sys/node/bundles/autoprefixer.js` just today and was scratching my head as to what exactly is going on there.

I think at the minimum it might be nice to silence some of the 'prop changed' warnings which some tests spew and also use the log mocking stuff in files like `terminal-logger.spec.ts` where we actually need to make assertions about logs, and then leave addressing the autoprefixer stuff for another day. I'll give it a go for the `src/runtime/test` ones and see what I can come up with without spending too much time on it

alicewriteswrongs:
@rwaskiewicz ok just updated this to take a different strategy for the tests which are doing `console.warn`, basically I went through and for some I was able to figure out a small change to the test component to resolve the warning messages, and for those which I wasn't able to fix I added a little wrapper function called `withSilentWarn` which takes a callback, silences `console.warn`, `await`s the callback, unsilences `console.warn`, and returns whatever was returned by the callback. This is still silencing these messages, but in a more specific and targeted way.

I also removed the changes for the autoprefixer bit

src/testing/testing-utils.ts
```diff
@@ -95,3 +95,99 @@ function getAppUrl(config: d.Config, browserUrl: string, appFileName: string) {
...
+
+interface ConsoleMocker {
+  (): {
+    logMock: jest.Mock<typeof console.log>;
+    warnMock: jest.Mock<typeof console.warn>;
+    errorMock: jest.Mock<typeof console.error>;
+  };
+}
+
+// export function withSilentWarn<T extends () => Promise<any>>(
```
rwaskiewicz, line 168: Do we need this still? If not, can we remove it?

rwaskiewicz (changes requested):
LGTM, two tiny nits

src/testing/testing-utils.ts
alicewriteswrongs: oops 🤦‍♀️

rwaskiewicz (approved):

PR #3441, opened 2022-06-28
test(cli): add tests for run.ts

rwaskiewicz:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

`run.ts` is not under test.

GitHub Issue Number: N/A

## What is the new behavior?

this commit adds initial unit tests for `run.ts`. the goal of these
tests are to get the minimal amount of coverage under the `run` and
`runTask` functions so that we can (eventually) begin to change the
function signatures on the functions they invoke.

specifcially, the stencil team is looking to make the configuration
object that is used throughout the compiler stricter by making fields
no longer optional on the type. as a part of that effort, we are
reassessing what data fields functions need. we know that we will be
changing the signatures of the tasks for 'help' and 'telemetry'. by
adding tests around the invocation of these tasks, we can have greater
trust in the changes that we're making

- [ ] Yes
- [X] No

## Testing

Unit tests pass

## Other information

#3436 is the driving force behind this PR. I'd like to land this PR first, then I'll update #3436 to include the test suite, and update the tests accordingly

### DIFF
--- a/src/cli/test/run.spec.ts
+++ b/src/cli/test/run.spec.ts
@@ -1,0 +1,154 @@
+import type * as d from '../../declarations';
+import * as coreCompiler from '@stencil/core/compiler';
+import { mockCompilerSystem, mockConfig, mockLogger as createMockLogger } from '@stencil/core/testing';
+import * as ParseFlags from '../parse-flags';
+import { run, runTask } from '../run';
+import * as HelpTask from '../task-help';
+import * as TelemetryTask from '../task-telemetry';
+import { createTestingSystem } from '../../testing/testing-sys';
+
+describe('run', () => {
+  describe('run()', () => {
+    let cliInitOptions: d.CliInitOptions;
+    let mockLogger: d.Logger;
+    let mockSystem: d.CompilerSystem;
+
+    let parseFlagsSpy: jest.SpyInstance<
+      ReturnType<typeof ParseFlags.parseFlags>,
+      Parameters<typeof ParseFlags.parseFlags>
+    >;
+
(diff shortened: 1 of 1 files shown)


src/cli/test/run.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+      cliInitOptions = {
+        args: [],
+        logger: mockLogger,
+        sys: mockSystem,
+      };
+
+      parseFlagsSpy = jest.spyOn(ParseFlags, 'parseFlags');
+      parseFlagsSpy.mockReturnValue({
+        task: 'help',
+      });
```
alicewriteswrongs, line 34: should this `.mockReturnValue` bit here be nested under the `'help task'` describe block, below? j/w because it seems specific to testing the 'help' task, although I guess we could also leave it here as a default value for future tests of other tasks

alicewriteswrongs (changes requested):
good to get things like this under test! just had one question

src/cli/test/run.spec.ts
rwaskiewicz: > although I guess we could also leave it here as a default value for future tests of other tasks

that was my thought, since 'help' returns early, it seemed like a reasonable/safe default for other tests.  I'll add a comment here explaining that

rwaskiewicz: 👍 dc3055292

alicewriteswrongs (approved):
lgtm!

Issue #3276, opened 2022-03-10
bug: jest cli - no longer able to run individual unit tests


Kwooda:
**Stencil version:**

```
 @stencil/core@2.14.0
```

**Current behavior:**
CLI no longer allows running individual unit tests. We updated from v2.6.0 to v2.14.0 and have lost the ability to run individual unit tests from the command line.

**Expected behavior:**
Running ```npx stencil test --e2e --spec component``` -- where component is the name of an individual unit test, should run just that test. Instead, it runs the entire test suite.

**GitHub Reproduction Link:**

**Other information:**

rwaskiewicz:
Hey @Kwooda, this sound similar to #3275, but perhaps slightly different. Could you create a minimal reproduction repo so that I can make sure I understand the issue and add the correct test cases here?

Kwooda:
Not sure how that's related. We have a bunch of unit tests, and would like to run just one of them. We used to be able to do that by just running npx stencil test --e2e --spec {testName} and it would run just that test. But after upgrading Stencil, any attempt to run a unit test runs all the unit tests. I'll try to find some time to set up a repo, but it seems easy enough to test by just trying to run a single unit test.

Kwooda:
I have created a repo to demonstrate the issue. Details are in the readme.
https://github.com/Kwooda/StencilUnitTests

rwaskiewicz:
Thanks @Kwooda! I was able to use the reproduction case to determine this behavior started to occur in v2.13.0 - I'm going to label this issue so that it gets ingested into the team's internal backlog. From there we'll be able to prioritize it and look into it further. Thanks again!

alicewriteswrongs:
Hey @Kwooda, sorry for the wait on this issue! We've diagnosed the issue and are working on a fix. In the meantime, with the reproduction you kindly provided we were able to come up with a workaround that should let you run a single test file while we come up with a more robust fix for the problem.

If you change the `test` script in `package.json` to omit the `--coverage` flag, so that it just looks like this:

```json
  "test": "stencil test --spec --e2e",
```

then you should be able to do `npm test my-component` to run the specfile for that component. You could also try that out by doing `npx stencil test --spec my-component`.

The issue relates to how `stencil test` processes arguments before it passes them along to Jest. Basically, any flag-type argument (like `--coverage`) will currently be passed along to Jest, but arguments are being parsed incorrectly in the case that a test file pattern (like `my-component` in the example above) is passed as well.

[...]

PR #3444, opened 2022-06-28
fix(cli): add explicit support for Jest CLI arguments 

alicewriteswrongs:
This adds explicit support to our CLI argument parsing code for Jest's
CLI arguments by adding entries to the arrays we use to parse 'known'
arguments.

## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

GitHub Issue Number: N/A

## What is the new behavior?

This PR makes it so that we declare all the Jest CLI arguments (as pulled from [the file in which they're implemented](https://github.com/facebook/jest/blob/main/packages/jest-cli/src/cli/args.ts)) in `src/cli/config-flags.ts`, so all of these flags are now fully-supported 'first class' (if you will) CLI flags.

- [ ] Yes
- [x] No

## Testing

Tests are passing, which is encouraging!

I also confirmed that this fixes the problem shown by the reproduction case in #3276, the underlying cause of which is that we only officially supported some jest arguments and not all, but actually all CLI args were making their way through to jest in a convoluted way that produces somewhat inconsistent behavior.


src/cli/config-flags.ts
```diff
@@ -35,12 +35,65 @@ export const BOOLEAN_CLI_ARGS = [
   'verbose',
   'version',
   'watch',
+
+  // JEST CLI OPTIONS
+  'all',
+  'automock',
+  'bail',
+  // 'cache', Stencil already supports this argument
```
alicewriteswrongs, line 43: some of these options are dupes of options that Stencil already supports. I thought we might want to keep them in the list here just so that the list is, in the code at least, a complete list of the Jest CLI arguments. Open to just deleting these commented-out lines as well though!

src/cli/config-flags.ts
```diff
@@ -35,12 +35,65 @@ export const BOOLEAN_CLI_ARGS = [
   'verbose',
   'version',
   'watch',
+
+  // JEST CLI OPTIONS
```
alicewriteswrongs, line 39: I separated these will a little comment heading for ease of editing later if / when Jest's supported arguments change

src/cli/config-flags.ts
```diff
@@ -53,8 +106,71 @@ export const STRING_CLI_ARGS = [
...
+  'watchPathIgnorePatterns',
 ] as const;
 
+/**
+ * All the CLI arguments which may have string or number values
+ *
+ * `maxWorkers` is an argument which is used both by Stencil _and_ by Jest,
+ * which means that we need to support parsing both string and number values.
+ */
+export const STRING_NUMBER_CLI_ARGS = ['maxWorkers'] as const;
```
alicewriteswrongs, line 172: `maxWorkers` [can according to the Jest docs](https://jestjs.io/docs/27.x/cli#--maxworkersnumstring) be either a string or a number, so we need to introduce a new argument type to support that.

src/cli/parse-flags.ts
```diff
@@ -161,6 +164,47 @@ const parseNumberArg = (flags: ConfigFlags, args: string[], configCaseName: Numb
...
+ * the argument types for which we support both a string and a
+ * number value.
+ *
+ * The regex tests for the presence of at least one character which is
+ * _not_ a digit (`\d`) or a period (`\.`). Thus we'll match a string
+ * like `"50%"`, but not a string like `"50"` or `"5.0"`. If it matches
+ * a given string we conclude that the string should be parsed as a
+ * string literal, rather than using `parseInt` to convert it to a number.
+ */
+const CLI_ARG_STRING_REGEX = /[^\d\.]+/g;
```
alicewriteswrongs, line 206: I think this should be pretty solid, and I don't think we need to support negative numbers, separators like `_` or `,`, etc. But always good to be sure with regexes!

src/cli/test/parse-flags.spec.ts
```diff
@@ -160,12 +161,41 @@ describe('parseFlags', () => {
   it('should parse --ci', () => {
     args[0] = '--ci';
     const flags = parseFlags(args, sys);
+    expect(flags.knownArgs).toEqual(['--ci']);
     expect(flags.ci).toBe(true);
   });
 
+  describe.each(BOOLEAN_CLI_ARGS)('should parse boolean flag %s', (cliArg) => {
```
alicewriteswrongs, line 168: these comprehensive tests of all the supported args serve as regression tests against duplicating any of the arguments in the arrays - because of the way that the arg parsing algorithm works having a dupe will result in a value like `[true, true]` being set on `ConfigFlags`. We want to make sure that can't happen, so we have some simple tests here.

alicewriteswrongs (commented):

rwaskiewicz (changes requested):
All in all looks good, just a few minor comments

src/cli/config-flags.ts
```diff
@@ -35,12 +35,65 @@ export const BOOLEAN_CLI_ARGS = [
...
+  'runTestsByPath',
+  'showConfig',
+  'silent',
+  'skipFilter',
+  'testLocationInResults',
+  'updateSnapshot',
+  'useStderr',
+  // 'verbose', Stencil alrerady supports this argument
+  // 'version', Stencil alrerady supports this argument
+  // 'watch', Stencil alrerady supports this argument
```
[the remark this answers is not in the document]
alicewriteswrongs: omg `alrerady` 😅

src/cli/parse-flags.ts
alicewriteswrongs: ooo I forgot that JS had scientific notation support! will add that and a test case for it

rwaskiewicz (approved):

rwaskiewicz:
@alicewriteswrongs looks like Lint & Format is failing - can you PTAL?

src/cli/config-flags.ts
deleonio: please add `outputFile`

PR #3460, opened 2022-07-07
chore(types): remove optional types on ConfigFlags interface

alicewriteswrongs:
Removing the optional types on `ConfigFlags` lets us remove a few null
suppression errors here and there and removes some `strictNullChecks`
violations as well.

To make it easy to create a correctly-formed `ConfigFlags` object this
also introduces a `createConfigFlags` helper function which takes a
partial `ConfigFlags` object and returns one with default properties
set.

## Pull request checklist

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes

GitHub Issue Number: N/A

## What is the new behavior?

Basically this just cleans up the typing on the `ConfigFlags` object a little bit and introduced a helper function to make it easy to create correctly typed objects.

- [ ] Yes
- [x] No

## Testing

I just made sure everything is compiling and tests are passing.

### DIFF
--- a/src/cli/config-flags.ts
+++ b/src/cli/config-flags.ts
@@ -269,8 +269,27 @@ export interface ConfigFlags
     NumberConfigFlags,
     StringNumberConfigFlags,
     LogLevelFlags {
-  task?: TaskCommand | null;
-  args?: string[];
-  knownArgs?: string[];
-  unknownArgs?: string[];
+  task: TaskCommand | null;
+  args: string[];
+  knownArgs: string[];
+  unknownArgs: string[];
 }
+
+/**
+ * Helper function for initializing a `ConfigFlags` object. Provide any overrides
+ * for default values and off you go!
+ *
+ * @param init an object with any overrides for default values
+ * @returns a complete CLI flag object
+ */
+export const createConfigFlags = (init: Partial<ConfigFlags> = {}): ConfigFlags => {
+  const flags: ConfigFlags = {
+    task: null,
+    args: [],
+    knownArgs: [],
+    unknownArgs: [],
+    ...init,
+  };
+
+  return flags;
+};
--- a/src/cli/test/run.spec.ts
+++ b/src/cli/test/run.spec.ts
@@ -12,6 +12,7 @@ import * as ServeTask from '../task-serve';
 import * as TelemetryTask from '../task-telemetry';
 import * as TestTask from '../task-test';
 import { createTestingSystem } from '../../testing/testing-sys';
+import { createConfigFlags } from '../config-flags';
 
 describe('run', () => {
   describe('run()', () => {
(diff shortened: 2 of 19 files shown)


src/cli/config-flags.ts
```diff
@@ -269,8 +269,27 @@ export interface ConfigFlags
...
 }
+
+/**
+ * Helper function for initializing a `ConfigFlags` object. Provide any overrides
+ * for default values and off you go!
+ *
+ * @param init an object with any overrides for default values
+ * @returns a complete CLI flag object
+ */
+export const createConfigFlags = (init: Partial<ConfigFlags> = {}): ConfigFlags => {
```
rwaskiewicz, line 285: I wonder if it makes sense to have a separate one of these for creating the config flags for tests.  IDK if reasonable defaults for tests will always be the reasonable defaults we use in `src` 🤔 Thoughts?

rwaskiewicz (changes requested):
first pass looks good - I still need to go through the tests and verify calling the helper function with defaults doesn't drop any coverage (where adding new fields could theoretically change the codepath taken).  A few questions/comments in the mean time

src/cli/test/run.spec.ts
```diff
@@ -129,6 +138,9 @@ describe('run', () => {
       validatedConfig = mockConfig(sys);
       validatedConfig.outputTargets = [];
 
+      console.log('run.spec::beforeEach');
+      console.log(validatedConfig.flags);
+
```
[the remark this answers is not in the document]
alicewriteswrongs: sigh...I should read things more closely!

src/cli/config-flags.ts
alicewriteswrongs: we could have a separate one for tests if we wanted to, my intention here was to create an object which is the minimum which complies with the interface so I think we'd want _that_ in both test code and, uh, actual code, but if we wanted to have additional defaults for tests we could have a separate function for that, although I think at present it would look something like this:

```ts
const mockConfigFlags = (overrides: Partial<ConfigFlags> = {}): ConfigFlags => {
  const init = createConfigFlags();

  return { ...init, ...overrides };
}
```

src/compiler/compiler.ts
```diff
@@ -18,7 +18,7 @@ import ts from 'typescript';
  * @returns a new instance of a Stencil compiler
  * @public
  */
-export const createCompiler = async (userConfig: Config): Promise<Compiler> => {
+export const createCompiler = async (userConfig: ValidatedConfig): Promise<Compiler> => {
```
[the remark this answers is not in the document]
alicewriteswrongs: that makes sense - I think I was playing around with this and meant to leave a 'do we want to do this?' comment

src/cli/config-flags.ts
rwaskiewicz: Yeah - TBH I waffle on what the best thing to do is here. On one hand, it's nice to have a clear distinction between the two - one for testing, one for the impl. On the other, we'd probably end up writing exactly what you have there, which still uses the underlying version for tests.  I say let's let YAGNI ride and keep things as is for now

rwaskiewicz (approved):

PR #3466, opened 2022-07-13
chore(browserstack): add concurrency field

rwaskiewicz:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes

See 'new behavior' section below

GitHub Issue Number: N/A

## What is the new behavior?

this commit adds a concurrency string, 'run_browserstack_queue' to the
'run_browserstack' job in ci.

the name for the concurrency field was chosen to:
- make it easily identifiable as a part of running browswerstack
- not overlap with the name of the run_browserstack job completely to
   avoid any confusion when debugging in the future
    
this change prevents a condition in ci where two simultaneous browserstack 
runs are attempted at the same time. stencil's browserstack license only
allows for five parallel runs at a given time. as of the time of this
writing, stencil uses three browsers per run of browserstack. as a
result, two runs of the job will result in requesting more browsers than
our license permits. on the browserstack side, the extra run (number six
of the five) is queued to run once one of the five completes. however,
on the karma side things, we don't have a great way to check for these
queued jobs and poll until one is done. as a result, we experience
timeouts/disconnects with browserstack as a result of waiting for a
browser instance (that's queued) to start up. oftentimes, we hit a
jasmine/karma timeout of ~30_000 milliseconds.
    
with this change, we are able to continue to have the flexibility to
re-run just the browserstack tests, while blocking them from running
when we aren't sure that we can acquire all the necessary instances. it
should be noted that this does not handle cases where a developer is
tunneling into browserstack from their local machine

- [ ] Yes
- [x] No

## Testing

First, I changed `pull_request_target` to `pull_request` to test this in CI.

Next, I pushed two PRs up - this one and https://github.com/ionic-team/stencil/pull/3467. I made sure they didn't have the same commit history, but did have the same contents (I think GH is smart enough to know that it's testing the same SHA otherwise and will only do one run of the various workflows).  I let the workflows run all the way through - the reason being is that the Browserstack workflows for `pull_request_target` were still going to run, and cause Browserstack resource contention otherwise (which will not exist once this PR is merged).

Then, I kicked off a Run of Browserstack in both PRs, to ensure that one awaited the other:
Screen Shot 2022-07-13 at 4 11 39 PM

Once the first once completed, I observed the second one start. Once the second one had started, I kicked the first one back off to observe that one waiting. We did see the first one fail due to disconnects, but it's hard to get a good run in with other PRs not having the same blocking behavior 😆 I was able to get both to pass earlier in the day just fine, just the luck of the draw I suppose

## Other information

I have to revert the change of `pull_request_target` to `pull_request` before we land this

### DIFF
--- a/.github/workflows/test-browserstack.yml
+++ b/.github/workflows/test-browserstack.yml
@@ -41,6 +41,8 @@ jobs:
     name: Run Browserstack
     runs-on: ubuntu-latest
     needs: [ build_core ]
+
+    concurrency: run_browserstack_queue
     steps:
       - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # v2.4.0
         with:
--- a/.github/workflows/test-browserstack.yml
+++ b/.github/workflows/test-browserstack.yml
@@ -1,7 +1,7 @@
 name: BrowserStack Tests
 
 on:
-  pull_request:
+  pull_request_target:
     branches:
       - 'main'
       - 'v3.0.0-dev'
--- a/.github/workflows/test-browserstack.yml
+++ b/.github/workflows/test-browserstack.yml
@@ -1,7 +1,7 @@
 name: BrowserStack Tests
 
 on:
-  pull_request_target:
+  pull_request:
     branches:
       - 'main'
       - 'v3.0.0-dev'
--- a/.github/workflows/test-browserstack.yml
+++ b/.github/workflows/test-browserstack.yml
@@ -42,6 +42,13 @@ jobs:
     runs-on: ubuntu-latest
     needs: [ build_core ]
 
+    # The concurrency field allows us to block multiple invocations of this job across multiple workflow runs.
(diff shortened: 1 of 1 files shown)


.github/workflows/test-browserstack.yml
```diff
@@ -41,6 +41,14 @@ jobs:
     name: Run Browserstack
     runs-on: ubuntu-latest
     needs: [ build_core ]
+    # The concurrency field allows us to block multiple invocations of this job across multiple workflow runs.
+    # Stencil is only able to run 5 parallel tests in Browserstack, which is exceeded when more than one run of this
+    # workflow occurs - e.g. one or more pull requests submitted in proximity, a pull request and a merge to a branch
```
rwaskiewicz, line 46: note to self:
```suggestion
    # workflow occurs - e.g. more than one pull request is submitted in proximity, a pull request and a merge to a branch
```

do this in the editor, double check for exceeding 120 chars

alicewriteswrongs (approved):
looks good to me!

PR #3469, opened 2022-07-13
chore(compiler): add comments and update types for in-memory FS

alicewriteswrongs:
This updates the in-memory FS file with a bunch of comments explaining
what various methods on the `InMemoryFileSystem` object do.

It also refactors where the relevant types live. Instead of having a
separate interface declared outside of the file where the
implementation is, we instead derive the `InMemoryFileSystem` from the
return value of the function which creates the object. This lets us have
comments annotating the implementation code which will also show up as
type hints throughout the codebase.

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

## What is the new behavior?

This does not (or should not!) change how anything works, but is instead just a documentation / cleanup patch.

- [ ] Yes
- [x] No

## Testing

I confirmed that unit tests are passing and whatnot, and did a bit of smoke testing with some test components locally.

### DIFF
--- a/src/compiler/build/compiler-ctx.ts
+++ b/src/compiler/build/compiler-ctx.ts
@@ -2,6 +2,7 @@ import type * as d from '../../declarations';
 import { basename, dirname, extname, join } from 'path';
 import { buildEvents } from '../events';
 import { noop, normalizePath } from '@utils';
+import { InMemoryFileSystem } from '../sys/in-memory-fs';
 
 /**
  * The CompilerCtx is a persistent object that's reused throughout
@@ -26,7 +27,7 @@ export class CompilerContext implements d.CompilerCtx {
   collections: d.CollectionCompilerMeta[] = [];
   compilerOptions: any = null;
   events = buildEvents();
-  fs: d.InMemoryFileSystem;
+  fs: InMemoryFileSystem;
   hasSuccessfulBuild = false;
   isActivelyBuilding = false;
   lastBuildResults: d.CompilerBuildResults = null;
--- a/src/compiler/index.ts
+++ b/src/compiler/index.ts
@@ -3,6 +3,7 @@ import { createWorkerMessageHandler } from './worker/worker-thread';
 import { initWebWorkerThread } from './sys/worker/web-worker-thread';
 import { IS_WEB_WORKER_ENV } from './sys/environment';
 import ts from 'typescript';
+export { FsWriteResults } from './sys/in-memory-fs'
 
 if (IS_WEB_WORKER_ENV) {
   initWebWorkerThread(createWorkerMessageHandler(createSystem()));
--- a/src/compiler/index.ts
+++ b/src/compiler/index.ts
@@ -22,4 +22,3 @@ export { path } from './sys/modules/path';
 export { transpile, transpileSync } from './transpile';
 export { version, versions, vermoji, buildId } from '../version';
 export { ts };
-export { InMemoryFileSystem, FsWriteResults } from './sys/in-memory-fs';
--- a/src/compiler/index.ts
+++ b/src/compiler/index.ts
@@ -22,3 +22,4 @@ export { path } from './sys/modules/path';
 export { transpile, transpileSync } from './transpile';
 export { version, versions, vermoji, buildId } from '../version';
 export { ts };
+export { InMemoryFileSystem, FsWriteResults } from './sys/in-memory-fs';
--- a/src/compiler/index.ts
+++ b/src/compiler/index.ts
@@ -3,7 +3,7 @@ import { createWorkerMessageHandler } from './worker/worker-thread';
 import { initWebWorkerThread } from './sys/worker/web-worker-thread';
 import { IS_WEB_WORKER_ENV } from './sys/environment';
 import ts from 'typescript';
-export { FsWriteResults } from './sys/in-memory-fs'
+export { FsWriteResults } from './sys/in-memory-fs';
 
 if (IS_WEB_WORKER_ENV) {
   initWebWorkerThread(createWorkerMessageHandler(createSystem()));
--- a/src/compiler/sys/in-memory-fs.ts
+++ b/src/compiler/sys/in-memory-fs.ts
@@ -206,8 +206,6 @@ export const createInMemoryFs = (sys: d.CompilerSystem) => {
    *
    * @param src the path to the source file
    * @param dest the destination the source file should be copied to
-   * @returns a promise
-   *
    */
   const copyFile = async (src: string, dest: string): Promise<void> => {
     const item = getItem(src);
--- a/src/compiler/sys/in-memory-fs.ts
+++ b/src/compiler/sys/in-memory-fs.ts
@@ -115,6 +115,9 @@ interface FsReaddirOptions {
   excludeExtensions?: string[];
 }
 
+/**
+ * A result from a directory read operation
+ */
 interface FsReaddirItem {
   absPath: string;
   relPath: string;
@@ -262,7 +265,7 @@ export const createInMemoryFs = (sys: d.CompilerSystem) => {
    * Get the contents of a directory on the in-memory filesystem
    *
    * @param dirPath the path to the directory of interest
-   * @param opts an optional object containg configuration options
+   * @param opts an optional object containing configuration options
    * @returns a Promise wrapping a list of directory contents
    */
   const readdir = async (dirPath: string, opts: FsReaddirOptions = {}): Promise<FsReaddirItem[]> => {
@@ -839,7 +842,7 @@ export const createInMemoryFs = (sys: d.CompilerSystem) => {
   /**
    * Ensure that a series of directories are created.
    *
-   * If `inMemoryOnly` is true this wil not touch the disk but will only
+   * If `inMemoryOnly` is true this will not touch the disk but will only
    * modify the in-memory filesystem cache. Otherwise it will create directories
    * in the real FS.
    *
@@ -999,6 +1002,16 @@ export const createInMemoryFs = (sys: d.CompilerSystem) => {
     }
   };
 
+  /**
+   * Cancel pending delete operations on files cached in the in-memory FS.
+   * This will not reverse a delete operation if it has already been committed
+   * to disk, but will cancel any pending delete operations that have not yet
+   * been committed.
+   *
+   * Note that this will silently **not cancel delete operations on directories**!
+   *
+   * @param filePaths a list of filepaths which should not be deleted
+   */
   const cancelDeleteFilesFromDisk = (filePaths: string[]) => {
     for (const filePath of filePaths) {
       const item = getItem(filePath);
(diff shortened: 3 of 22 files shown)


src/compiler/build/compiler-ctx.ts
```diff
@@ -2,6 +2,7 @@ import type * as d from '../../declarations';
 import { basename, dirname, extname, join } from 'path';
 import { buildEvents } from '../events';
 import { noop, normalizePath } from '@utils';
+import { InMemoryFileSystem } from '../sys/in-memory-fs';
```
alicewriteswrongs, line 5: the `InMemoryFileSystem` type is now defined in the same file as `createInMemoryFileSystem` and is also exported from `src/compiler/index.ts`. I updated imports to use relative imports _within_ the `compiler` module (i.e. in any file below `src/compiler/`) and I imported from `@stencil/core/compiler` _outside_ of the module, for the sake of consistency and, well, modularity 😄

src/compiler/sys/in-memory-fs.ts
```diff
@@ -2,76 +2,225 @@ import type * as d from '../../declarations';
...
+ *
+ * // commit the results to disk
+ * const commitStats = await inMemoryFs.commit();
+ * ```
+ *
+ * In the above example the write operation and the delete operation (w/
+ * `.remove`) are both queue in the in-memory proxy but not committed to
+ * disk until the `.commit` method is called.
+ */
+export type InMemoryFileSystem = ReturnType<typeof createInMemoryFs>;
```
alicewriteswrongs, line 37: this type is now defined in this file, and is derived from the return type of `createInMemoryFs` instead of being declared separately as an interface. I think this just helps us keep things in one place.

src/compiler/sys/in-memory-fs.ts
```diff
@@ -2,76 +2,225 @@ import type * as d from '../../declarations';
...
+
+/**
+ * An node in the in-memory file system. This may represent a file or
+ * a directory, and pending copy, write, and delete operations may be stored
+ * on it.
+ */
+export interface FsItem {
```
alicewriteswrongs, line 44: some of these other interfaces are used outside of this file (although generally not too often). I considered moving them all to the bottom of the file below `createInMemoryFs` but didn't so yet. If that's a cleaner look I'm open to it!

src/compiler/sys/in-memory-fs.ts
```diff
@@ -2,76 +2,225 @@ import type * as d from '../../declarations';
...
+ * This allows the compiler to proceed freely as if it is modifying the
+ * filesystem, modifying the world in whatever way suits it, while deferring
+ * actual FS writes until the end of the compilation process, making actual
+ * changes to the filesystem on disk contingent on an error-free build or any
+ * other condition.
+ *
+ * @param sys a compiler system object
+ * @returns an in-memory filesystem interface
+ */
 export const createInMemoryFs = (sys: d.CompilerSystem) => {
```
alicewriteswrongs, line 154: As always, and perhaps especially with this file, these comments reflect what I learned from reading the code, looking for call sites across the codebase, and a little bit of github / slack spelunking here and there. So there very well may be inaccuracies!

src/compiler/sys/in-memory-fs.ts
```diff
@@ -2,76 +2,225 @@ import type * as d from '../../declarations';
...
-  const items: d.FsItems = new Map();
+  /**
+   * Map to hold the items in the in-memory cache which proxies the underlying
+   * OS filesystem.
+   */
+  const items: FsItems = new Map();
   const outputTargetTypes = new Map<string, string>();
 
-  const accessData = async (filePath: string) => {
+  /**
```
alicewriteswrongs, line 162: a note here - the `accessData` function wasn't called outside of this file, and was just sort of odd. I refactored this `access` function to work more like the `accessSync` function below and was able to delete `accessData`

src/compiler/sys/in-memory-fs.ts
```diff
@@ -719,13 +1072,26 @@ export const createInMemoryFs = (sys: d.CompilerSystem) => {
     return item;
   };
 
-  const clearCache = () => items.clear();
-
-  const keys = () => Array.from(items.keys()).sort();
```
alicewriteswrongs, line 724: I could find no references to these functions in the codebase, so I think they should be safe to delete

alicewriteswrongs (commented):

src/compiler/sys/in-memory-fs.ts
```diff
@@ -1,77 +1,226 @@
...
   };
 
-  const copyFile = async (src: string, dest: string) => {
+  /**
+   * Copy a file from `src` to `dest`. Note that this merely queues the file
+   * for copying, the copy isn't actually committed.
+   *
+   * @param src the path to the source file
+   * @param dest the destination the source file should be copied to
+   * @returns a promise
```
rwaskiewicz, line 209: No action required - IMO we can elide `@returns` for empty Promises from JSDoc. The linter is configured to permit that, but I don't believe we could configure it to enforce consistency (I.E. "warn us if we add `@returns` on a fn that returns `void` or `Promise<void>`), hence the no change needed here

src/compiler/sys/in-memory-fs.ts
```diff
@@ -289,7 +481,17 @@ export const createInMemoryFs = (sys: d.CompilerSystem) => {
     return fileText;
   };
 
-  const remove = async (itemPath: string) => {
+  /**
+   * Remove an item from the in-memory FS
+   *
+   * This is done by marking if for deletion. The item will remain in memory
```
rwaskiewicz, line 487:
```suggestion
   * This is done by marking it for deletion. The item will remain in memory
```

src/compiler/sys/in-memory-fs.ts
```diff
@@ -299,7 +501,13 @@ export const createInMemoryFs = (sys: d.CompilerSystem) => {
     }
   };
 
-  const removeDir = async (dirPath: string) => {
+  /**
+   * Remove an item form the in-memory FS by marking it to be deleted
```
rwaskiewicz, line 505:
```suggestion
   * Remove a directory from the in-memory FS by marking it to be deleted
```

src/compiler/sys/in-memory-fs.ts
```diff
@@ -323,14 +531,31 @@ export const createInMemoryFs = (sys: d.CompilerSystem) => {
     }
   };
 
-  const removeItem = async (filePath: string) => {
+  /**
+   * Remove an item form the in-memory FS by marking it to be deleted
```
rwaskiewicz, line 535:
```suggestion
   * Remove an item from the in-memory FS by marking it to be deleted
```

src/compiler/index.ts
```diff
@@ -22,3 +22,4 @@ export { path } from './sys/modules/path';
 export { transpile, transpileSync } from './transpile';
 export { version, versions, vermoji, buildId } from '../version';
 export { ts };
+export { InMemoryFileSystem, FsWriteResults } from './sys/in-memory-fs';
```
rwaskiewicz, line 25: Should we be exporting `InMemoryFileSystem`? Previously that was a private type. It's not that we can't export it, but it does expand the scope of things that are public facing that we maintain

rwaskiewicz (changes requested):

.github/workflows/test-browserstack.yml
```diff
@@ -97,6 +97,13 @@ jobs:
           # local-testing must be one of two values: 'start' or 'stop' to start/stop Browserstack
           local-testing: stop
 
+      - name: TEMP
+        # here we check that there are no changed / new files from our test
+        # run. we use `git status`, grep out the build zip (downloaded above),
+        # and check if there are more than 0 lines in the output.
+        run: git status --short | grep -c -v stencil-core-build.zip
+        shell: bash
+
```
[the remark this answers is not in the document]
alicewriteswrongs: it was failing before and I was trying to get it to print what changed haha

src/compiler/sys/in-memory-fs.ts
alicewriteswrongs: My thought was mainly just that it cuts down on verbosity and lets us have comments for these functions that live right next to their implementations, previously there were some comments in the actual implementation, some in the interface, etc. I did port over those 'Synchronous!! ...' comments to the new JSDocs for all those functions so we still have those around, not sure if that addresses what you're thinking about?

src/compiler/sys/in-memory-fs.ts
```diff
@@ -1,77 +1,226 @@
...
-    if (s) {
-      item.exists = s.exists;
-      item.isDirectory = s.isDirectory;
-      item.isFile = s.isFile;
-
-      data.exists = item.exists;
-      data.isDirectory = item.isDirectory;
-      data.isFile = item.isFile;
-    } else {
-      item.exists = false;
```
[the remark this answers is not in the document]
alicewriteswrongs: actually scratch that — it's been a while so I forgot, but the reason that I removed this code (`accessData`) was because it calls the `stat` function and, afaict, the assignments which modify the `item` returned by `getItem` are redundant with what is done there, see:

https://github.com/ionic-team/stencil/blob/2e773244f43cee77febb1715d1393bfa4df1309f/src/compiler/sys/in-memory-fs.ts#L558-L589

there we already have a series of checks which ensure that `.isFile`, `.isDirectory`, and so on are properly set

src/compiler/index.ts
alicewriteswrongs: yeah it's not necessary I think, tbh I don't remember why I added it here so I'll remove it

alicewriteswrongs:
@rwaskiewicz I think all comments should be addressed now and (I think...) your changes for that git context stuff should get the build passing now

src/compiler/sys/in-memory-fs.ts
rwaskiewicz: Yeah, I think that works!

src/compiler/index.ts
```diff
@@ -22,4 +22,3 @@ export { path } from './sys/modules/path';
 export { transpile, transpileSync } from './transpile';
 export { version, versions, vermoji, buildId } from '../version';
 export { ts };
-export { InMemoryFileSystem, FsWriteResults } from './sys/in-memory-fs';
```
rwaskiewicz: Is `FsWriteResults` re-exported anywhere else? The file that originally had that type definition `stencil-public-compiler` is a signal that this was a publicly exported type - I think we need to continue to re-export it to not break our API (unless I'm missing where it's being re-exported)

src/compiler/sys/in-memory-fs.ts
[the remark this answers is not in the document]
rwaskiewicz: Ah! Good catch!

rwaskiewicz (commented):
One last question/comment

src/compiler/index.ts
alicewriteswrongs: that makes sense, I will add it back 👍

rwaskiewicz (approved):

alicewriteswrongs:
@tanner-reits if you wouldn't mind taking a look — I think this is waiting for your sign-off

tanner-reits (approved):
LGTM! Appreciate you doing some archeology and adding these comments! I'm sure they'll come in handy later

PR #3477, opened 2022-07-19
chore(compiler): make outputTargets required on the ValidatedConfig

alicewriteswrongs:
This makes the `outputTargets` field required on the `ValidatedConfig`
type, allowing us to freely access `.outputTargets` on a validated
configuration object in the compiler code without having to use the `!`
operator, check `=== undefined`, etc.

This includes changes to the places where a `ValidatedConfig` object is
created to default `.outputTargets` to `[]` if no suitable value is
present.

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

`.outputTargets` is currently not a required field on our `ValidatedConfig` type, which means that we get a null check error every place we access this (very commonly accessed!) property, unless we do `.outputTargets ?? []` or something of the sort everywhere.

We also have a few places where, due to the callstack, a `ValidatedConfig` is _actually_ what's being passed around, but we haven't updated type signatures to reflect that yet.

## What is the new behavior?

`.outputTargets` is now a required field for `ValidatedConfig` and the type is changed from `d.Config` to `d.ValidatedConfig` in a few places. This does not eliminate every site where we could go from `d.Config` -> `d.ValidatedConfig` but just does to a few which are sort of relevant to the `.outputTargets` change.

- [ ] Yes
- [x] No

### DIFF
--- a/src/compiler/output-targets/dist-custom-elements/custom-elements-types.ts
+++ b/src/compiler/output-targets/dist-custom-elements/custom-elements-types.ts
@@ -12,12 +12,12 @@ import { normalizePath, dashToPascalCase } from '@utils';
  * @param typesDir the path to the directory where type declarations are saved
  */
 export const generateCustomElementsTypes = async (
-  config: d.Config,
+  config: d.ValidatedConfig,
   compilerCtx: d.CompilerCtx,
   buildCtx: d.BuildCtx,
   typesDir: string
 ): Promise<void> => {
-  const outputTargets = (config.outputTargets ?? []).filter(isOutputTargetDistCustomElements);
+  const outputTargets = config.outputTargets.filter(isOutputTargetDistCustomElements);
 
   await Promise.all(
     outputTargets.map((outputTarget) =>
@@ -36,7 +36,7 @@ export const generateCustomElementsTypes = async (
  * @param outputTarget the output target for which types are being currently generated
  */
 const generateCustomElementsTypesOutput = async (
-  config: d.Config,
+  config: d.ValidatedConfig,
   compilerCtx: d.CompilerCtx,
   buildCtx: d.BuildCtx,
   typesDir: string,
--- a/src/compiler/output-targets/dist-custom-elements/index.ts
+++ b/src/compiler/output-targets/dist-custom-elements/index.ts
@@ -97,9 +97,9 @@ export const getBundleOptions = (
  * Get bundle options for rollup, run the rollup build, optionally minify the
  * output, and write files to disk.
  *
- * @param the validated Stencil configuration we're using
- * @param buildCtx the current build context
+ * @param config the validated Stencil configuration we're using
  * @param compilerCtx the current compiler context
+ * @param buildCtx the current build context
  * @param outputTarget the outputTarget we're currently dealing with
  * @returns an empty promise
  */
--- a/src/compiler/output-targets/dist-custom-elements/index.ts
+++ b/src/compiler/output-targets/dist-custom-elements/index.ts
@@ -28,7 +28,7 @@ import ts from 'typescript';
  * which do actual work of generating the rollup configuration, creating an
  * entry chunk, running, the build, etc.
  *
- * @param config the user-supplied compiler configuration we're using
+ * @param config the validated compiler configuration we're using
  * @param compilerCtx the current compiler context
  * @param buildCtx the current build context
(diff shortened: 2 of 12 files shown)


src/compiler/output-targets/dist-custom-elements/custom-elements-types.ts
```diff
@@ -12,12 +12,12 @@ import { normalizePath, dashToPascalCase } from '@utils';
...
  */
 export const generateCustomElementsTypes = async (
-  config: d.Config,
+  config: d.ValidatedConfig,
   compilerCtx: d.CompilerCtx,
   buildCtx: d.BuildCtx,
   typesDir: string
 ): Promise<void> => {
-  const outputTargets = (config.outputTargets ?? []).filter(isOutputTargetDistCustomElements);
+  const outputTargets = config.outputTargets.filter(isOutputTargetDistCustomElements);
```
rwaskiewicz, line 20: Ahhhh - so good!

src/compiler/output-targets/dist-custom-elements/index.ts
```diff
@@ -34,15 +34,15 @@ import ts from 'typescript';
  * @returns an empty Promise which won't resolve until the work is done!
  */
 export const outputCustomElements = async (
-  config: d.Config,
+  config: d.ValidatedConfig,
```
rwaskiewicz, line 37: Minor, can we update the JSDoc description for this variable to better describe this is now a validated config? Something like:
```diff
- * @param config the user-supplied compiler configuration we're using
+ * @param config the validated compiler configuration we're using
```

src/compiler/output-targets/dist-custom-elements/index.ts
```diff
@@ -66,7 +66,7 @@ export const outputCustomElements = async (
  * @returns bundle options suitable for generating a rollup configuration
  */
 export const getBundleOptions = (
-  config: d.Config,
+  config: d.ValidatedConfig,
```
rwaskiewicz, line 69: Similar to my comment above, can we update the JSDoc `@param` description for `config` here as well?
```diff
- * @param config user-supplied Stencil configuration
+ * @param config the validated compiler configuration we're using
```

src/compiler/output-targets/dist-custom-elements/index.ts
```diff
@@ -104,7 +104,7 @@ export const getBundleOptions = (
  */
 
 export const bundleCustomElements = async (
-  config: d.Config,
+  config: d.ValidatedConfig,
```
rwaskiewicz, line 107: Similar to my comment above, can we update the JSDoc `@param` description for `config` here as well?
```diff
- * @param config user-supplied Stencil configuration
+ * @param config the validated compiler configuration we're using
```

It also looks like this JSDoc is 'floating' with a newline between it and the function. While you're in there, could you remove that pesky extra `\n` too please?

rwaskiewicz (changes requested):
looks great! nothing but nitpicks on my end to help differentiate a user-supplied config vs a validated one in JSDocs

src/compiler/output-targets/dist-custom-elements/custom-elements-types.ts
alicewriteswrongs: the joy of type checking 😌

rwaskiewicz (approved):
LGTM!

src/compiler/output-targets/dist-custom-elements/index.ts
```diff
@@ -96,15 +96,15 @@ export const getBundleOptions = (
 /**
  * Get bundle options for rollup, run the rollup build, optionally minify the
  * output, and write files to disk.
- * @param config user-supplied Stencil configuration
+ *
+ * @param the validated Stencil configuration we're using
```
rwaskiewicz, line 100: Can we fix this inline???
```suggestion
 * @param config the validated Stencil configuration we're using
```

rwaskiewicz (approved):

rwaskiewicz:
@alicewriteswrongs it looks like lint is failing. I tried to fix it to  no avail :-( Can you PTAL?

alicewriteswrongs:
@rwaskiewicz should be good now!

rwaskiewicz (approved):

Issue #3481, opened 2022-07-20
bug: cli args parsing fails


jasperbosch:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/.github/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

2.17.x

### Current Behavior

When executing `stencil test --spec --e2e --json --outputFile=output.json` the outputFile-argument is badly parsed. It will result in `jest --spec --e2e --json --outputFile=output.json output.json`, wich will result in not executing jest.

However executing `stencil test --spec --e2e --json --output-file output.json` will succeed.

Cause is in the `parse-flags.ts` file:
```typescript
args.forEach((arg, i) => {
    if (arg.startsWith(`--${dashCaseName}=`) || arg.startsWith(`--${configCaseName}=`)) {
      value = getEqualsValue(arg);
      matchingArg = arg;
    } else if (arg === `--${dashCaseName}` || arg === `--${configCaseName}`) {
      value = args[i + 1];
      matchingArg = arg;
    } else if (alias) {
      if (arg.startsWith(`-${alias}=`)) {
        value = getEqualsValue(arg);
        matchingArg = arg;
      } else if (arg === `-${alias}`) {
        value = args[i + 1];
        matchingArg = arg;
      }
    }
  });
  ```
  `matchingArg` (on line 4) should only contain `--outputFile=` but now contains `--outputFile=output.json`

### Expected Behavior

Expect to execute jest with: `jest --spec --e2e --json --outputFile=output.json`

### Steps to Reproduce

execute `stencil test --spec --e2e --json --outputFile=output.json`

### Code Reproduction URL

-

alicewriteswrongs:
hey @jasperbosch, thanks for filing this issue and for doing some code spelunking! You're right that this is an issue. I was just in this code recently so I'll fix it today.

alicewriteswrongs:
Thanks for reporting this! Just to update you, there's a bugfix PR up here: https://github.com/ionic-team/stencil/pull/3483 which just merged into `main`.

If you want to test it out to make sure it fixes your issue you could

- clone the repo locally
[...]

Issue #3471, opened 2022-07-14
bug: Since v1.7.1 our tests doesn't work anymore [fix with PR #3472]


deleonio:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/.github/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

1.7.1

### Current Behavior

`npm run test`

```bash
testMatch:  - 0 matches
```

### Expected Behavior

```bash
testMatch:  - >0 matches
```

### Steps to Reproduce

Try test exec with v1.7.1 with project should work in v.1.7.0

### Code Reproduction URL

-

alicewriteswrongs:
Hi deleonio, thanks for filing this issue! Would you mind providing a reproduction for the issue you're seeing?

alicewriteswrongs:
I think this is the same issue as #3481

alicewriteswrongs:
Thanks for reporting this! Just to update you, there's a bugfix PR up here: https://github.com/ionic-team/stencil/pull/3483 which just merged into `main`.

If you want to test it out to make sure it fixes your issue you could

- clone the repo locally
- check out [the branch](https://github.com/ionic-team/stencil/tree/ap/flag-parsing-bug)
- do `npm install`, `npm run build` and `npm pack`
- try installing the resulting `.tgz` file in our project by doing `npm install ./path/to/stencil.tgz`

If you'd rather wait for a pre-release or release build to try that out no worries :)

rwaskiewicz:
The fix for this issue has been officially released as a part of [Stencil v2.17.2](https://github.com/ionic-team/stencil/releases/tag/v2.17.2).

PR #3483, opened 2022-07-20
fix(cli): fix bug with parsing --fooBar=baz type CLI flags 

alicewriteswrongs:
This fixes a bug with how these flags were being parsed which was
messing with Jest. Basically, the issue related to how the `knownArgs`
array on the `ConfigFlags` object was being populated. For CLI key-value
args of the format `--argName=value` the whole string
`"--argName=value"` was being added _as well as_ the value string
`"value"`, which had the effect of passing duplicate args to Jest.

This refactors how such arguments are handled to always parse apart the
argument name and the value, adding only the argument name and then the
argument value, so that if you pass, for instance,
`--outputFile=output.json`, the args passed to Jest will look like

```ts
['--outputFile', 'output.json']
```

See https://github.com/ionic-team/stencil/issues/3471 and https://github.com/ionic-team/stencil/issues/3481 for more details

## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Previously CLI args of the format `--foo=bar` were not being passed to Jest correctly because of a bug in how they were being parsed.

## What is the new behavior?

Now they are handled correctly and should work well with jest.

- [ ] Yes
- [x] No

## Testing

Check out hte repro steps on the linked issues for some ideas of how to test this manually. Basically, in a simple stencil project (i.e. created w/ `npm init stencil`) if you try to do this w/ stencil installed from npm you should see an error:

```sh
npx stencil test --spec --e2e --json --outputFile=output.json
```

if you check out this branch, `npm run build`, `npm pack`, and then install the `.tgz` in that same project you should not see an error when running that same command.

### DIFF
--- a/src/cli/parse-flags.ts
+++ b/src/cli/parse-flags.ts
@@ -51,9 +51,13 @@ export const parseFlags = (args: string[], sys?: CompilerSystem): ConfigFlags =>
     }
   }
 
-  flags.unknownArgs = flags.args.filter((arg: string) => {
-    return !flags.knownArgs.includes(arg);
-  });
+  // to find unknown / unrecognized arguments we filter `args`, including only
+  // arguments whose normalized form is not found in `knownArgs`. `knownArgs`
+  // is populated during the call to `parseArgs` above. For arguments like
+  // `--foobar` the string `"--foobar"` will be added, while for more
+  // complicated arguments like `--bizBoz=bop` or `--bizBoz bop` just the
+  // string `"--bizBoz"` will be added.
+  flags.unknownArgs = flags.args.filter((arg: string) => !flags.knownArgs.includes(parseEqualsArg(arg)[0]));
 
   return flags;
 };
@@ -259,17 +263,18 @@ const getValue = (
 
   let value: string | undefined;
   let matchingArg: string | undefined;
+
   args.forEach((arg, i) => {
     if (arg.startsWith(`--${dashCaseName}=`) || arg.startsWith(`--${configCaseName}=`)) {
-      value = getEqualsValue(arg);
-      matchingArg = arg;
+      // our argument was passed at the command-line in the format --argName=arg-value
+      [matchingArg, value] = parseEqualsArg(arg);
     } else if (arg === `--${dashCaseName}` || arg === `--${configCaseName}`) {
+      // the next value in the array is assumed to be a value for this argument
       value = args[i + 1];
       matchingArg = arg;
     } else if (alias) {
       if (arg.startsWith(`-${alias}=`)) {
-        value = getEqualsValue(arg);
-        matchingArg = arg;
+        [matchingArg, value] = parseEqualsArg(arg);
       } else if (arg === `-${alias}`) {
         value = args[i + 1];
         matchingArg = arg;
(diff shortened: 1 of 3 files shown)


src/cli/parse-flags.ts
```diff
@@ -287,13 +292,43 @@ interface CLIArgValue {
...
+ * const [arg, value] = parseEqualsArg("--myArgument=myValue")
+ * ```
+ *
+ * Then `arg` will be `"--myArgument"` and `value` will be `"myValue"`.
+ *
+ * @param arg the arg in question
+ * @returns a tuple containing the arg name and the value (if present)
  */
-const getEqualsValue = (commandArgument: string) => commandArgument.split('=').slice(1).join('=');
+export const parseEqualsArg = (arg: string): [string, string] => {
```
alicewriteswrongs, line 327: I refactored this function to return a tuple of the argument name (parsed from the CLI input) and the value it is set to if present. The value will be present if an argument like `--foobar=bop` is passed, but not if arg like `--foobar` is passed.

The first value in the tuple will be `--foobar` for both of those examples, so even for non-'equals sign' type CLI args this function be be safely used to normalize the argument names, so that we can count on things like

```ts
parseEqualsArg("--foobar")[0] === parseEqualsArgs("--foobar=baz")[0]
```
being true

src/cli/parse-flags.ts
```diff
@@ -51,9 +51,13 @@ export const parseFlags = (args: string[], sys?: CompilerSystem): ConfigFlags =>
...
-  flags.unknownArgs = flags.args.filter((arg: string) => {
-    return !flags.knownArgs.includes(arg);
-  });
+  // to find unknown / unrecognized arguments we filter `args`, including only
+  // arguments whose normalized form is not found in `knownArgs`. `knownArgs`
+  // is populated during the call to `parseArgs` above. For arguments like
+  // `--foobar` the string `"--foobar"` will be added, while for more
+  // complicated arguments like `--bizBoz=bop` or `--bizBoz bop` just the
+  // string `"--bizBoz"` will be added.
+  flags.unknownArgs = flags.args.filter((arg: string) => !flags.knownArgs.includes(parseEqualsArg(arg)[0]));
```
alicewriteswrongs, line 60: I use `parseEqualsArg` to normalize arguments for building `unknownArgs`, here

alicewriteswrongs (commented):
just pointing one thing out

src/testing/jest/test/jest-config.spec.ts
```diff
@@ -19,6 +19,17 @@ describe('jest-config', () => {
     expect(jestArgv.maxWorkers).toBe(2);
   });
 
+  it('should pass --outputFile=path/to/file', () => {
+    const args = ['test', '--ci', '--outputFile=path/to/my-file'];
+    const config = mockValidatedConfig();
+    config.flags = parseFlags(args, config.sys);
+    expect(config.flags.args).toEqual(['--ci', '--outputFile=path/to/my-file']);
+    expect(config.flags.unknownArgs).toEqual([]);
+    config.testing = {};
```
rwaskiewicz, line 28: Minor, I think we can hoist this to be a part of the rest of the setup for this function:

```diff
  it('should pass --outputFile=path/to/file', () => {
    const args = ['test', '--ci', '--outputFile=path/to/my-file'];
    const config = mockValidatedConfig();
    config.flags = parseFlags(args, config.sys);
+    config.testing = {};
    expect(config.flags.args).toEqual(['--ci', '--outputFile=path/to/my-file']);
    expect(config.flags.unknownArgs).toEqual([]);
-    config.testing = {};
    const jestArgv = buildJestArgv(config);
    expect(jestArgv.outputFile).toBe('path/to/my-file');
  });
```

src/testing/jest/test/jest-config.spec.ts
```diff
@@ -19,6 +19,17 @@ describe('jest-config', () => {
     expect(jestArgv.maxWorkers).toBe(2);
   });
 
+  it('should pass --outputFile=path/to/file', () => {
```
rwaskiewicz, line 22: Minor, it's not immediately clear to me what this is testing from this description. Can we work the expected result into the test name here? Maybe something like:

```ts
  it("marks 'outputFile' as a jest argument", () => {
```

rwaskiewicz (changes requested):
👍 two minor asks

PR #3485, opened 2022-07-21
chore(mocks): update mockConfig, mockValidatedConfig to accept any property

rwaskiewicz:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Our mocks for configuration entities, both unvalidated and validated, are cumbersome
to override.

GitHub Issue Number: N/A

## What is the new behavior?

this commit adds a new parameter to `mockValidatedConfig`, `overrides`.
the argument is optional, and defaults to an empty object literal. it is
spread over the returned object to override and defaults put in place by
this method. the `sys` argument is also removed, as it now can be safely
derived from the `overrides` argument (and still falls back to a new
`TestingSystem` if `overrides.sys` is falsy

this commit adds an `overrides` parameter to `mockConfig`. `overrides`
is a partial instance of `UnvalidatedConfig`, which defaults to an empty
object literal. its contents are spread over the returned object,
overriding the values put in place by default. it also removes the `sys` arg.
' this value is now derived from the provided `overrides`, falling back to a
`TestingSystem` otherwise.

STENCIL-486: Update mockConfig, mockValidatedConfig to accept any property

- [ ] Yes
- [x] No

## Testing

Unit tests continue to pass

### DIFF
--- a/src/cli/telemetry/test/telemetry.spec.ts
+++ b/src/cli/telemetry/test/telemetry.spec.ts
@@ -14,7 +14,7 @@ describe('telemetryBuildFinishedAction', () => {
 
   beforeEach(() => {
     sys = createSystem();
-    config = mockValidatedConfig(sys);
+    config = mockValidatedConfig({ sys });
     config.outputTargets = [];
     config.flags.args = [];
   });
@@ -45,7 +45,7 @@ describe('telemetryAction', () => {
 
   beforeEach(() => {
     sys = createSystem();
-    config = mockValidatedConfig(sys);
+    config = mockValidatedConfig({ sys });
     config.outputTargets = [];
     config.flags.args = [];
   });
--- a/src/cli/telemetry/test/telemetry.spec.ts
+++ b/src/cli/telemetry/test/telemetry.spec.ts
@@ -14,9 +14,11 @@ describe('telemetryBuildFinishedAction', () => {
 
   beforeEach(() => {
     sys = createSystem();
-    config = mockValidatedConfig({ sys });
-    config.outputTargets = [];
-    config.flags.args = [];
+    config = mockValidatedConfig({
+      flags: { args: [], knownArgs: [], task: 'build', unknownArgs: [] },
+      outputTargets: [],
+      sys,
+    });
   });
 
   it('issues a network request when complete', async () => {
@@ -45,9 +47,11 @@ describe('telemetryAction', () => {
 
   beforeEach(() => {
     sys = createSystem();
-    config = mockValidatedConfig({ sys });
-    config.outputTargets = [];
-    config.flags.args = [];
+    config = mockValidatedConfig({
+      flags: { args: [], knownArgs: [], task: 'build', unknownArgs: [] },
+      outputTargets: [],
+      sys,
+    });
   });
 
   it('issues a network request when no async function is passed', async () => {
--- a/src/cli/telemetry/test/telemetry.spec.ts
+++ b/src/cli/telemetry/test/telemetry.spec.ts
@@ -15,7 +15,7 @@ describe('telemetryBuildFinishedAction', () => {
   beforeEach(() => {
     sys = createSystem();
     config = mockValidatedConfig({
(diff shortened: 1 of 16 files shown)


src/cli/telemetry/test/telemetry.spec.ts
```diff
@@ -14,9 +14,11 @@ describe('telemetryBuildFinishedAction', () => {
 
   beforeEach(() => {
     sys = createSystem();
-    config = mockValidatedConfig(sys);
-    config.outputTargets = [];
-    config.flags.args = [];
+    config = mockValidatedConfig({
+      flags: { args: [], knownArgs: [], task: 'build', unknownArgs: [] },
```
alicewriteswrongs, line 18: this could be a `createConfigFlags` call here if we wanted, just to clean things up a little bit

src/cli/test/task-generate.spec.ts
```diff
@@ -14,13 +14,15 @@ jest.mock('prompts', () => ({
 
 const setup = async () => {
   const sys = mockCompilerSystem();
-  const config: d.ValidatedConfig = mockValidatedConfig(sys);
-  config.configPath = '/testing-path';
-  config.srcDir = '/src';
+  const config: d.ValidatedConfig = mockValidatedConfig({
+    configPath: '/testing-path',
+    flags: { args: [], knownArgs: [], task: 'generate', unknownArgs: [] },
```
alicewriteswrongs, line 19: same comment about `createConfigFlags` here

src/compiler/output-targets/test/output-targets-dist-custom-elements.spec.ts
```diff
@@ -22,18 +22,21 @@ import { DIST_CUSTOM_ELEMENTS, DIST_CUSTOM_ELEMENTS_BUNDLE } from '../output-uti
...
-  const config: d.ValidatedConfig = mockValidatedConfig(sys);
+  const config: d.ValidatedConfig = mockValidatedConfig({
+    buildAppCore: true,
+    buildEs5: true,
+    configPath: '/testing-path',
+    namespace: 'TestApp',
+    outputTargets: [{ type: DIST_CUSTOM_ELEMENTS }],
+    srcDir: '/src',
+    sys,
+  });
```
alicewriteswrongs, line 33: this is just a much nicer pattern for putting things together like this 👍

src/testing/mocks.ts
```diff
@@ -21,71 +20,72 @@ import { noop } from '@utils';
...
  * Creates a mock instance of an internal, validated Stencil configuration object
- * @param sys an optional compiler system to associate with the config. If one is not provided, one will be created for
  * the caller
+ * @param overrides a partial implementation of `ValidatedConfig`. Any provided fields will override the defaults
+ * provided by this function.
  * @returns the mock Stencil configuration
  */
-export function mockValidatedConfig(sys?: CompilerSystem): ValidatedConfig {
-  const baseConfig = mockConfig(sys);
+export function mockValidatedConfig(overrides: Partial<ValidatedConfig> = {}): ValidatedConfig {
```
alicewriteswrongs, line 30: I think moving away from `sys` as a 'special' param here is a good example of making something more general and clear while at the same time preserving all of the functionality it had before (like if you want to supply a special `sys` param it's just a matter of `mockValidatedConfig({ sys })` 

it's a good pattern is what I'm saying 😄

alicewriteswrongs (changes requested):
this looks good to me! just a few comments about some small tweaks

src/cli/test/task-generate.spec.ts
rwaskiewicz: D'oh - how did I forget we had that? Good catch!

src/cli/telemetry/test/telemetry.spec.ts
rwaskiewicz: 👍 Done in 57d6a3bb8

alicewriteswrongs (approved):
lgtm! 👍

PR #3491, opened 2022-07-27
chore(config): make sys required on ValidatedConfig

rwaskiewicz:
## Pull request checklist

Please check if your PR fulfills the following requirements:
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [x] Build (`npm run build`) was run locally and any changes were pushed
- [x] Unit tests (`npm test`) were run locally and passed
- [x] E2E Tests (`npm run test.karma.prod`) were run locally and passed
- [x] Prettier (`npm run prettier`) was run locally and passed

## Pull request type

 

 

Please check the type of change your PR introduces:
- [ ] Bugfix
- [ ] Feature
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

## What is the current behavior?

`sys` is not a required property on `ValidatedConfig`, creating several
`strictNullCheck` violations in the codebase

GitHub Issue Number: N/A

## What is the new behavior?

this commit makes the `sys` property required on `ValidatedConfig`. this
work decreases the number of `strictNullCheck` violations in the
codebase. it does not fix every violation pertaining to `config.sys`
usage - rather, it seeks to be the minimal work necessary to

1. make the `sys` property required on `ValidatedConfig`
2. prevent introducing breaking compilation errors

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Testing

Unit tests, e2e tests continue to pass (no function signature changes were made).

I tried hard here to replicate the existing behavior (otherwise saying 'tests pass boss 👍' wouldn't _really_ have any ground to stand on).

PR #3079, opened 2021-09-23
refactor(telemetry) Refactor out the singleton

splitinfinities:
## Pull request checklist

Please check if your PR fulfills the following requirements:
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [x] Build (`npm run build`) was run locally and any changes were pushed
- [x] Unit tests (`npm test`) were run locally and passed
- [x] E2E Tests (`npm run test.karma.prod`) were run locally and passed
- [x] Prettier (`npm run prettier`) was run locally and passed

## Pull request type

 

 

Please check the type of change your PR introduces:
- [x] Bugfix
- [ ] Feature
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

## What is the current behavior?

Today, the introduction of the Singleton caused some customers monorepo's to fail. This PR fixes that by removing the singleton experiment. 

GitHub Issue Number: N/A

## What is the new behavior?

- No more singleton for telemetry

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Testing
Check this branch out, build and link. Go to a consuming app and ensure telemetry is enabled - `npx stencil telemetry on` - then run any command - notably `npx stencil build --debug --verbose`, and you will see the telemetry data passed back. you can also run `yarn && yarn run stencil build --debug --verbose` in the consuming app and you will receive

PR #3494, opened 2022-07-27
chore(run): remove unneeded CompilerSystem checks

rwaskiewicz:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

In 66d0476b (#3079), we added a truthy check for `CompilerSystem` entities.
With the recent move to making that entity required on a `ValidatedConfig` entity
(a6a9171 - #3491), these checks can be safely removed.

GitHub Issue Number: N/A

## What is the new behavior?

This PR removes cases throughout the code base where a truthy check is
required in order to call a function that requires it. In every case, we are able to
utilize the fact that the calling function has an instance of `ValidatedConfig`, which
now has a `sys` on it. 

The original belief was that this would need to be a breaking change (this was
months ago when the ticket was originally written). However, the foundation of this
PR  (a6a9171) does not require this to be a breaking change anymore. 

- [ ] Yes
- [x] No

## Testing

Unit tests updated

### DIFF
--- a/src/cli/task-build.ts
+++ b/src/cli/task-build.ts
@@ -6,7 +6,7 @@ import { startupCompilerLog } from './logs';
 import { taskWatch } from './task-watch';
 import { telemetryBuildFinishedAction } from './telemetry/telemetry';
 
-export const taskBuild = async (coreCompiler: CoreCompiler, config: d.ValidatedConfig, sys?: d.CompilerSystem) => {
+export const taskBuild = async (coreCompiler: CoreCompiler, config: d.ValidatedConfig) => {
   if (config.flags.watch) {
     // watch build
     await taskWatch(coreCompiler, config);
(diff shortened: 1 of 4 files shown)


src/cli/task-build.ts
```diff
@@ -24,10 +24,7 @@ export const taskBuild = async (coreCompiler: CoreCompiler, config: d.ValidatedC
     const compiler = await coreCompiler.createCompiler(config);
     const results = await compiler.build();
 
-    // TODO(STENCIL-148) make this parameter no longer optional, remove the surrounding if statement
-    if (sys) {
-      await telemetryBuildFinishedAction(sys, config, coreCompiler, results);
-    }
+    await telemetryBuildFinishedAction(config.sys, config, coreCompiler, results);
```
rwaskiewicz, line 27: There's an argument to be had here that since `sys` exists on `config`, we could update the function signature of `telemetryBuildFinishedAction` so that calling it looks something like:
```diff
- await telemetryBuildFinishedAction(config.sys, config, coreCompiler, results);
+ await telemetryBuildFinishedAction(config, coreCompiler, results);
```

I decided _not_ to go down this route for 2 reasons. 
1.  I don't think we have a good feel for what we want to do from a consistency standpoint (if anything) throughout the codebase
2. This would have grown in scope into updating additional functions in `telemetryBuildFinishedAction`, which I felt was out of scope here

alicewriteswrongs: makes sense — I've made changes along those lines in a few places and I think I generally think they're a good idea, but we don't need to always do them and I think making changesets more focused and small is basically always a good idea 👍

alicewriteswrongs (approved):
lgtm! 👍

tanner-reits (approved):
LGTM! 🧹

rwaskiewicz:
CI never got picked up for BS. Closing and opening again

PR #3495, opened 2022-07-27
 feat(ci): fail the browserstack tests if any files were changed or added 

alicewriteswrongs:
This adds a small check to the end of our end-to-end CI tests which just
tries to ensure that no files were changed or added while the tests were
running.

STENCIL-451: Add CI Checks for a Dirty Git Context

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [x] Build related changes

Right now we have no automatic visibility into whether tests running on CI change files that should be committed.

## What is the new behavior?

This adds a simple check (implemented using `git`, `grep`, `wc`, and `-ne`) to make sure there aren't any uncommitted changes.

Note that we download a build zip file, so we have a little `grep -v` to ignore that.

- [ ] Yes
- [x] No

## Testing

I confirmed that this will fail if there is a file changed or added by the tests.

### DIFF
--- a/.github/workflows/test-e2e.yml
+++ b/.github/workflows/test-e2e.yml
@@ -37,3 +37,10 @@ jobs:
       - name: End-to-End Tests
         run: npm run test.end-to-end -- --ci
         shell: bash
+
+      - name: Git status check
+        # here we check that there are no changed / new files from our test
+        # run. we use `git status`, grep out the build zip (downloaded above),
+        # and check if there are more than 0 lines in the output.
+        run: if [[ $(git status --short | grep -v stencil-core-build.zip | wc -l) -ne 0 ]]; then exit 1; fi
+        shell: bash
--- a/.github/workflows/test-e2e.yml
+++ b/.github/workflows/test-e2e.yml
@@ -37,10 +37,3 @@ jobs:
       - name: End-to-End Tests
         run: npm run test.end-to-end -- --ci
         shell: bash
-
-      - name: Git status check
-        # here we check that there are no changed / new files from our test
-        # run. we use `git status`, grep out the build zip (downloaded above),
-        # and check if there are more than 0 lines in the output.
-        run: if [[ $(git status --short | grep -v stencil-core-build.zip | wc -l) -ne 0 ]]; then exit 1; fi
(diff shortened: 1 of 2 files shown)


.github/workflows/test-e2e.yml
```diff
@@ -37,3 +37,10 @@ jobs:
       - name: End-to-End Tests
         run: npm run test.end-to-end -- --ci
         shell: bash
+
+      - name: Git status check
+        # here we check that there are no changed / new files from our test
+        # run. we use `git status`, grep out the build zip (downloaded above),
+        # and check if there are more than 0 lines in the output.
+        run: if [[ $(git status --short | grep -v stencil-core-build.zip | wc -l) -ne 0 ]]; then exit 1; fi
```
rwaskiewicz, line 45: No action required - we can shorten this with grep's `-c` flag just a smidge:
```suggestion
        run: if [[ $(git status --short | grep -c -v stencil-core-build.zip) -ne 0 ]]; then exit 1; fi
```

but that's purely stylistic according to https://www.shellcheck.net/wiki/SC2126.  Arguably (IMO) the way its written now is more readable, but I'm not super familiar with all of grep's flags.  Approving, LMK your thoughts - either way I think we'll be able to merge this tomorrow

rwaskiewicz (approved):

rwaskiewicz:
@alicewriteswrongs now that I think about this, I think we want this in `test-browserstack` instead of `test-e2e`, since that generates the component type decl files for `test-app` and the like

.github/workflows/test-e2e.yml
alicewriteswrongs: ohh managed not to know about `-c` for `grep`! yeah I think possibly using `wc` might make it more obvious what's going on, but on the other hand to figure out the `grep` flags requires like one minute w/ the terminal and `man` so....I think it's probably fine to use `-c -v`

alicewriteswrongs:
Right, that makes sense! We have STENCIL-451 about this and it didn't specify which test run and I thought for some reason that the e2e tests could write files 🤔 

anywhoo — I'll move it to the browserstack one rn

Issue #3512, opened 2022-08-05
bug: WebWorker doesn't work when stencil component library imported to nextjs app


jeski-bright:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/.github/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

2.13.0

### Current Behavior

When trying to use a component which implements a method delegated to a web worker I get a `Uncaught DOMException: Failed to execute 'importScripts' on 'WorkerGlobalScope': The URL '/_next/static/media/<file_name>.js' is invalid.` exception.

### Expected Behavior

The web worker starts properly & can communicate with the component.

### Steps to Reproduce

1. Clone the reproduction repo
2. Run `yarn install`
3. Run `yarn build`
4. Run `yarn dev`
5. Go to `localhost:3000` and check the console output.

### Code Reproduction URL

https://github.com/jeski-bright/stencil-next-example

### Additional Information

This is similar to #2731, however here it almost works, with webpack5 there is no bug with `import.meta.url` only the parsed url.href is incorrect.

alicewriteswrongs:
hi @jeski-bright, thanks for filing this issue (and for opening a related PR!). I was able to confirm that the error is happening in your reproduction - thanks for providing that with detailed instructions for use!

I'm going to mark this for ingestion into our backlog and the Stencil team will prioritize root-causing and dealing with the issue.

Thanks again for reporting!

Pfeil:
As the fix has been merged, I wonder to which degree this issue still exists? Did it not yet end up within any release?

PR #3513, opened 2022-08-05
fix(compiler): try to create web worker with the workerPath before falling back to blob

jeski-bright:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

GitHub Issue Number: #3512 

Currently the worker plugin creates the Web Worker by first creating a blob of a script which loads the actual script via `importScripts`. However the path to the script is calculated via `new URL(<file_name>, import.meta.url)`, which generates a relative path for some frameworks (as Nextjs mentioned in the issue). I would expect most SSR frameworks to do this to avoid mismatches between the server and client. The `importScripts` method fails on loading a relative path however.

## What is the new behavior?

Before loading the worker in the current way, try to create it with the resolved script path directly. In case of error, fall back to the current implementation.
This is in line with [src/compiler/sys/worker/web-worker-main.ts](https://github.com/ionic-team/stencil/blob/main/src/compiler/sys/worker/web-worker-main.ts#L56)

- [ ] Yes
- [x] No

## Testing

Clone the reproduction repo from the issue & apply this patch to stencil. The web worker should start now.


christian-bromann (approved):
@johnjenkins approved 😉

PR #3456, opened 2022-07-07
fix(compiler): Re-enable build cache

George-Payne:
- uncomment cache write to disk
- add `buildCacheDirName` config option to prevent clash with `ScreenshotConnector`

## Pull request checklist

Please check if your PR fulfills the following requirements:
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [x] Build (`npm run build`) was run locally and any changes were pushed
- [x] Unit tests (`npm test`) were run locally and passed
- [x] E2E Tests (`npm run test.karma.prod`) were run locally and passed
- [x] Prettier (`npm run prettier`) was run locally and passed

## Pull request type

 

 

Please check the type of change your PR introduces:
- [x] Bugfix
- [ ] Feature
- [ ] Refactoring (no functional changes, no api changes)
- [x] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

## What is the current behavior?

Issue 1:
Currently, the build cache is being populated, but not written to disk, so is not picked up in subsequent builds. 

The line of code here: 
https://github.com/ionic-team/stencil/blob/main/src/compiler/build/write-build.ts#L24

Was commented out in this commit: 
https://github.com/ionic-team/stencil/commit/63595bc540db9aa85295f5a8dc8738beb2a37611

Issue 2:

`ScreenshotConnector` also uses the cacheDir for it's cache, but a build clears out the cache here:
https://github.com/ionic-team/stencil/blob/c0aeac1176c474f4e0935e40f3944433f5c24860/src/compiler/cache.ts#L20

### Uncommenting the line that writes the cache to disk re-enables the cache.
This offers considerable speedup in large (~200 component) projects:

Before: ~71s (average of three runs [81.62s, 64.27s, 69.29s])
After: ~38s (average of three runs [39.37s, 38.11s, 37.98s]) (with cache populated)

### New config option `buildCacheDirName` with default `"build"` 

Places the cache in `\`${cacheDir}/${buildCacheDirName}`\` , matching the pattern of `ScreenshotConnector` (\`${cacheDir}/${screenshotCacheFileName}`\`. The means only `.stencil/build` is wiped when the cache is cleared, leaving the screenshot connector cache alone.

## Does this introduce a breaking change?

- [ ] Yes
- [ ] No
- [x] Don't think so

## Testing

See time measurements above

## Other information

[...]

PR #3521, opened 2022-08-12
chore(types): reduce strictNullChecks on file writes

rwaskiewicz:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

there are >2300 strict null checks violations in the code base

GitHub Issue Number: N/A

## What is the new behavior?

this commit moves the needle forward by ~10 less `strictNullChecks`
failures in the section of the codebase where files are written to disk.
this section was chosen as an 'easy Friday win' where function
signatures could be changed with little change to the rest of the
codebase

- [ ] Yes
- [x] No

## Testing

Type checker continues to pass


src/compiler/build/write-build.ts
```diff
@@ -21,11 +31,10 @@ export const writeBuild = async (config: d.Config, compilerCtx: d.CompilerCtx, b
 
     // successful write
     // kick off writing the cached file stuff
-    // await compilerCtx.cache.commit();
```
rwaskiewicz, line 24: just deleting commented out code here

src/compiler/build/write-build.ts
```diff
@@ -21,11 +31,10 @@ export const writeBuild = async (config: d.Config, compilerCtx: d.CompilerCtx, b
...
     buildCtx.debug(`in-memory-fs: ${compilerCtx.fs.getMemoryStats()}`);
-    // buildCtx.debug(`cache: ${compilerCtx.cache.getMemoryStats()}`);
 
-    await outputServiceWorkers(config, buildCtx), await validateBuildFiles(config, compilerCtx, buildCtx);
+    await outputServiceWorkers(config, buildCtx);
+    await validateBuildFiles(config, compilerCtx, buildCtx);
```
rwaskiewicz, line 37: The comma operator here wasn't doing anything...removed for clarity

src/compiler/build/write-build.ts
alicewriteswrongs: There's a PR open which uncomments these lines (and makes a few other changes): https://github.com/ionic-team/stencil/pull/3456

We may want to take a look at that at some point, I think it's fine to remove these lines for now though

alicewriteswrongs (approved):
this LGTM!

src/compiler/build/validate-files.ts
```diff
@@ -2,7 +2,19 @@ import type * as d from '../../declarations';
...
+ * @param config the Stencil configuration used for the build
+ * @param compilerCtx the compiler context associated with the build
+ * @param buildCtx the build context associated with the current build
+ * @returns an array containing empty-Promise results
+ */
+export const validateBuildFiles = (
+  config: d.ValidatedConfig,
+  compilerCtx: d.CompilerCtx,
+  buildCtx: d.BuildCtx
+): Promise<(void | void[])[]> => {
```
tanner-reits, line 17: Didn't know `Promise<(void | void[])[]>` and `Promise<[void, void[]]>` were the same type

src/compiler/build/write-build.ts
tanner-reits: Do these need to run synchronously? Or could we `Promise.all()` them?

src/compiler/service-worker/generate-sw.ts
```diff
@@ -4,11 +4,11 @@ import { buildWarn, catchError } from '@utils';
 import { isOutputTargetWww } from '../output-targets/output-utils';
 
 export const generateServiceWorker = async (
-  config: d.Config,
+  config: d.ValidatedConfig,
   buildCtx: d.BuildCtx,
   workbox: d.Workbox,
   outputTarget: d.OutputTargetWww
-) => {
+): Promise<Awaited<void[]> | void> => {
```
tanner-reits, line 11: I'm just now finding out about `Awaited`, but it looks like it's benefit is type inference from Promises. So, is it doing anything is this case?

tanner-reits (changes requested):
Few small questions, but overall looking good

src/compiler/build/write-build.ts
rwaskiewicz: I don't know if it makes sense to run them in parallel, as the latter there should be validating build output (ie the build should have been completed (including SW gen) before we try to verify anything)

src/compiler/build/validate-files.ts
rwaskiewicz: The `[]` distributes over the parens, so `Promise<(void | void[])[]>` is equivalent to `Promise<void[] | void[][]>`, rather than a tuple consisting of something assignable to `void` in the first position and something assignable to `void[]` in the second.

Surprisingly, TS lets you do something like: 😆 
```ts
type Voidey = [void, void[]];
const a: Voidey = [undefined, [undefined, undefined, undefined]];
```

src/compiler/service-worker/generate-sw.ts
rwaskiewicz: It's not actually, I can remove it I think. I don't recall any issues that required it to be there.

Rm'ed in ca28ef34a

tanner-reits (approved):

src/compiler/build/validate-files.ts
tanner-reits: Yeah I knew the `[]` distributed, but I woulda thought TS treated `void[] | void[][]` and `[void, void[]]` as different types 🤷‍♂️

rwaskiewicz: Hmmm I think we've got some disconnect here. They are different types - `[void, void[]]` is a tuple with two fields, whereas `void[] | void[][]` is either a `void` array or a `void` array of arrays (without a size bound).

Issue #2815, opened 2021-02-07
Output target collection does not transpile ts path config


patrickhaug:
```
 @stencil/core@2.0.4>
```

**I'm submitting a:**

[x] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**
When setting paths in tsconfig:
`"paths": {
      "@utils/*": ["./app/utils/*"]
    }`
it gets transpiled to this in output target collection. 
`import("@utils/xxx"))`

Other output targets work fine.

**Expected behavior:**
It should resolve to the correct file path from ts config and not the alias.

abdolian:
I have the same problem

liamdebeasi:
We are running into the same issue. I have provided a GitHub repo: https://github.com/liamdebeasi/stencil-alias

Steps to reproduce:

1. Clone repo.
2. Run `npm install`.
3. Run `npm run start`. Observe that there are no errors.
4. Quit the dev server.
5. Run `npm run build`.
6. Open `dist/collection/components/my-component/my-component.js`. Observe that the `format` is still being imported from `@utils/utils`;
7. Open `dist/components/my-component.js`. Observe that the `format` util is correctly inlined.

rwaskiewicz:
Thanks all! I spoke with @liamdebeasi about this issue today and confirmed its an issue. I've labeled it so that it gets ingested into our backlog

PR #3523, opened 2022-08-15
Fix/collection output transpile imports

tanner-reits:
## Pull request checklist

## Pull request type

 

 

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

The `dist-collection` output target currently does not transpile path aliases defined in a project's `tsconfig.json`. Instead, the import paths are left unchanged and thus reference modules that cannot be resolved.

GitHub Issue Number: https://github.com/ionic-team/stencil/issues/2815

## What is the new behavior?

- A transformation step has been added as a part of the `dist-collection` output target that will transpile path aliases and convert them in to relative paths based on the output files from the TS compilation

- [ ] Yes
- [ ] No

## Testing

Unit tests have been added for the new transformer, and existing unit tests continue to pass. (Further testing will be possible once infrastructure is in place related to STENCIL-223).

E2E tests continue to pass (as consistently as they did before)

Validated output on test repos (including Ionic framework)


src/compiler/transformers/test/map-imports-to-path-aliases.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+  });
+
+  /**
+   * This test fails
+   * The `transpileModule` helper method is designed to only transpile a single module's (file's)
+   * content. However, with the solution implemented to fix module imports not getting transformed respective of
+   * their path aliases in the `tsconfig.json` file for a project, this helper cannot resolve the defined imports so
+   * the text returned from the helper keeps the original import path intact.
+   */
+  it('should replace the path alias with the generated relative path', () => {
```
tanner-reits, line 20: Failing test. The testing helpers we have do not account for the need to test with multiple JS modules.

src/compiler/transformers/map-imports-to-path-aliases.ts
```diff
@@ -0,0 +1,58 @@
...
+        if (!importPath.startsWith('.')) {
+          /**
+           * When running unit tests, the `resolvedModule` property on the returned object is always undefined.
+           * In an actual build, the modules are resolved as expected.
+           *
+           * Not sure what is causing this to fail in tests. My guess is that the `transpileModule` helper method
+           * is somehow not giving context to paths for the options supplied on a `config.tsCompilerOptions`, but I am confused
+           * as to how this differs between tests and a build.
+           */
+          const module = ts.resolveModuleName(importPath, sourceFile, tsCompilerOptions, compilerHost);
```
tanner-reits, line 25: This is what is causing tests to fail for this transformer. In the testing context, TS is not able to resolve the import modules, so import paths are always left untouched.

src/compiler/transformers/test/map-imports-to-path-aliases.spec.ts
rwaskiewicz: So, this is tricky. We need to restore some functionality where we have a compiler instance along with a virtual filesystem of some kind, and a patched TS compiler (all for testing things in a more 'e2e' fashion).  In the mean time, for unit testing, does spying on this TS method make things go away? TAL at the patch I attached to this comment (GH doesn't support '.patch' extensions, go figure 🙄). But you should be able to apply it with `git apply` all the same.

[ts_spy.txt](https://github.com/ionic-team/stencil/files/9338991/ts_spy.txt)

I didn't take a super close look here just yet, but this should unblock you at the very least on the testing front

tanner-reits: @rwaskiewicz I can do something like that, but at that point I question if that test really adds any benefit since it's not actually performing the action we're relying on TS for 🤷‍♂️ The test doesn't blow up, it just fails cuz the text doesn't get replaced like it does in a build.

With that said, we won't get "correct" test coverage either way until STENCIL-223 is fixed. So, I'm curious how we wanna handle this. Feels like we'd either need a dev build to have this verified by users (and framework team) or some kind of config flag. I know we had considered this being a potential breaking change too

tanner-reits: Fair points! I'll get those tests and some general code comments added in another commit. I'll also add a "todo" to get the actual module resolution step tested once STENCIL-223 is done

src/compiler/output-targets/dist-collection/index.ts
```diff
@@ -30,7 +32,22 @@ export const outputCollection = async (
           outputTargets.map(async (o) => {
             const relPath = relative(config.srcDir, mod.jsFilePath);
             const filePath = join(o.collectionDir, relPath);
-            await compilerCtx.fs.writeFile(filePath, code, { outputTargetType: o.type });
+
+            // Transpile the already transpiled modules to apply
+            // a transformer to convert aliased import paths to relative paths
+            const { outputText } = ts.transpileModule(code, {
```
tanner-reits, line 38: Only downside to this approach is, for this target, we'll essentially end up transpiling each module twice. However, I ran a couple builds with this change and without and only seemed to affect Stencil build times within a project by a second or so (that testing was done with the framework code base)

tanner-reits:
Last thing we should need to decide is how to handle this release/deployment. Basically if we feel like this can be merged directly, considered a breaking change for V3, or something we would want to keep behind a config flag.

src/compiler/transformers/map-imports-to-path-aliases.ts
```diff
@@ -0,0 +1,60 @@
...
+ * with generated relative import paths during the transformation step of the TS compilation process.
+ * This action is taken to prevent issues with import paths not being transpiled at build time resulting in
+ * unknown imports in output code for some output targets (`dist-collection` for instance). Output targets that do not run through a bundler
+ * are unable to resolve imports using the aliased path names and TS intentionally does not replace resolved paths as a part of
+ * their compiler ({@link https://github.com/microsoft/TypeScript/issues/10866})
+ *
+ * @param config The Stencil configuration object.
+ * @returns A factory for creating a {@link ts.Transformer}.
+ */
+export const mapImportsToPathAliases = (config: d.Config): ts.TransformerFactory<ts.SourceFile> => {
```
rwaskiewicz, line 16: Since this is only called by `src/compiler/output-targets/dist-collection/index.ts#outputCollection`, and `outputCollection` passes a config of type `ValidatedConfig` to this fn, can we update this function signature accordingly?
```suggestion
export const mapImportsToPathAliases = (config: d.ValidatedConfig): ts.TransformerFactory<ts.SourceFile> => {
```

src/compiler/transformers/test/map-imports-to-path-aliases.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+    config = mockValidatedConfig({ tsconfig: './tsconfig.json', tsCompilerOptions: {} });
+
+    resolveModuleNameSpy = jest.spyOn(ts, 'resolveModuleName');
+  });
+
+  afterEach(() => {
+    resolveModuleNameSpy.mockReset();
+  });
+
+  it('should ignore relative imports', () => {
```
rwaskiewicz, line 25: Consider making these test descriptions assertive to minimize repetition of 'should':
```suggestion
  it('ignores relative imports', () => {
```

rwaskiewicz (changes requested):
> Last thing we should need to decide is how to handle this release/deployment. Basically if we feel like this can be merged directly, considered a breaking change for V3, or something we would want to keep behind a config flag.

I think we should add a configuration flag to the `collections` output target for folks to try this out and report any issues to us. My thoughts are we do the following:
- resolve comments from @alicewriteswrongs and I on this PR
- Create a separate PR that targets this one that adds the flag
  - We'll also want to update our internal doc for flags for V3 
- Create a sibling PR with the associated documentation

src/compiler/transformers/map-imports-to-path-aliases.ts
tanner-reits: Ah yep, good catch. Couldn't use the ValidatedConfig when this was running as a part of the main transpilation process

src/compiler/transformers/map-imports-to-path-aliases.ts
```diff
@@ -0,0 +1,60 @@
...
+  return (transformCtx) => {
+    let sourceFile: string;
```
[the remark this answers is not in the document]
tanner-reits: Yes this is needed. Using the approach you suggested errors out. `node.getSourceFile()` returns `undefined`

src/compiler/transformers/map-imports-to-path-aliases.ts
```diff
@@ -0,0 +1,60 @@
...
+        // imports from external modules/packages
+        if (!importPath.startsWith('.')) {
+          const module = ts.resolveModuleName(importPath, sourceFile, config.tsCompilerOptions, compilerHost);
+
+          if (
+            module.resolvedModule?.isExternalLibraryImport === false &&
+            module.resolvedModule?.resolvedFileName != null
+          ) {
+            importPath = normalize(
+              relative(sourceFile, module.resolvedModule.resolvedFileName).replace(module.resolvedModule.extension, '')
```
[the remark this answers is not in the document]
tanner-reits: Good catch. I changed this to replace all file extensions defined on `ts.Extensions`

src/compiler/transformers/map-imports-to-path-aliases.ts
```diff
@@ -28,12 +41,24 @@ export const mapImportsToPathAliases = (config: d.Config): ts.TransformerFactory
...
-            module.resolvedModule?.resolvedFileName != null
-          ) {
+          const hasResolvedFileName = module.resolvedModule?.resolvedFileName != null;
+          const isModuleFromNodeModules = module.resolvedModule?.isExternalLibraryImport === true;
+          const shouldTranspileImportPath = hasResolvedFileName && !isModuleFromNodeModules;
+
+          if (shouldTranspileImportPath) {
+            // Create a global extension matching regular expression
+            // This can be used to strip all file extensions off the generated import path string
+            const extensionRegex = new RegExp(Object.values(ts.Extension).join('|'), 'g');
```
rwaskiewicz, line 51: I think this is close, but still not quite right.  That's on me, I think my previous comment/concern wasn't worded well. In the initial implementation:

```ts
relative(sourceFile, module.resolvedModule.resolvedFileName).replace(module.resolvedModule.extension, '')
```

we would always replace the first instance of anything that matched the resolved module's extension. So if we had something like
```
foo.mjs.mjs
```
this would become
```
foo.mjs
```
which seemed correct, until we grouped the extensions:
```
foo[.mjs][.mjs]
// the first '.mjs' is what would get removed
foo[.mjs] // <- second extension stuck around
```


I think we still run into an issue here, as we might be removing more than we'd like for cases like
```
foo.mjs.js
```
we'd end up removing all the extensions, which I don't _think_ is what we want. Instead, we should rm the last entry only with something like:

```diff
- const extensionRegex = new RegExp(Object.values(ts.Extension).join('|'), 'g');
+ const extensionRegex = new RegExp(Object.values(ts.Extension).map((ext) => `${ext}$`).join('|'));
```

where:
- we append '$' to each extension to check that the filename ends with the extension
- we remove the 'g' flag, since we only want to remove one thing

Does that make sense? Or did I miss something here? 😆

src/compiler/transformers/map-imports-to-path-aliases.ts
```diff
@@ -11,15 +11,28 @@ import type * as d from '../../declarations';
...
  * @param config The Stencil configuration object.
+ * @param destinationFilePath The location on disk the file will be written to.
+ * @param destinationDirectory The destination on disk of the collection directory.
  * @returns A factory for creating a {@link ts.Transformer}.
  */
-export const mapImportsToPathAliases = (config: d.Config): ts.TransformerFactory<ts.SourceFile> => {
+export const mapImportsToPathAliases = (
+  config: d.ValidatedConfig,
+  destinationFilePath: string,
+  destinationDirectory: string
```
rwaskiewicz, line 21: Can we add a few test cases for these, where
- `destinationFilePath` is a parent of `destinationDirectory` and both are non-empty strings?
- `destinationDirectory` is a parent of `destinationFilePath` and both are non-empty strings?
- both are the same directory and both are non-empty strings?

this might get a little funky, since we'll probably have to rely on `normalize` in our tests for Win/Unix compatability

rwaskiewicz (changes requested):
Thanks! Just a few more comments on the revisions

src/compiler/transformers/map-imports-to-path-aliases.ts
tanner-reits: @rwaskiewicz This was kinda where I started, but then I tested a file with weird extensions like `foo.mts.ts`. This resulted in an import path like `../foo.mts` which throws an error because it's not a JS file import. Not sure how realistic files like that are, but it was causing issues

rwaskiewicz: Yeah, IDK of realistic it is, kind of a weird edge case for someone to have something like that working in Stencil 🤷

src/compiler/transformers/map-imports-to-path-aliases.ts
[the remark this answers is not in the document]
alicewriteswrongs: it looks to me like you have it up here because you can't get the right value until the function returned here is called:

https://github.com/ionic-team/stencil/blob/88a6974882c23e97fbd257e81cd8b1349235345e/src/compiler/transformers/map-imports-to-path-aliases.ts#L83-L79

and then you need to use it within the `visit` body. I think doing `let foo: string;` does always create a `strictNullChecks` error though because since it's not initialized with a string the variable is `undefined` (and therefore not a string).

One alternative approach could be to curry the `sourceFile` argument like this:

```ts
const visit = (sourceFile: string) => (node: ts.Node): ts.VisitResult<ts.Node> => {
```

then below you could do something like

```ts
    return (tsSourceFile) => {
      const sourceFile = tsSourceFile.fileName;

      return ts.visitEachChild(tsSourceFile, visit(sourceFile), transformCtx);
    };
```

which would I think fix the `strictNullChecks` error and is arguably "more correct" code since we don't declare a variable without initializing it and then access it later w/ the implicit assumption it's been assigned in the intervening time

alicewriteswrongs (changes requested):
two thoughts / questions

src/compiler/transformers/map-imports-to-path-aliases.ts
tanner-reits: @rwaskiewicz with that said, do we feel comfortable removing just the last extension, or want to play it safe and remove all?

rwaskiewicz: I say remove the last extension. If someone reports a bug, they can directly link this comment and blame me lol 😆

src/compiler/transformers/test/map-imports-to-path-aliases.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+describe('mapImportsToPathAliases', () => {
+  let module: ReturnType<typeof transpileModule>;
+  let config: ValidatedConfig;
+  let resolveModuleNameSpy: jest.SpyInstance<
+    ReturnType<typeof ts.resolveModuleName>,
+    Parameters<typeof ts.resolveModuleName>
+  >;
+
+  beforeEach(() => {
+    config = mockValidatedConfig({ tsconfig: './tsconfig.json', tsCompilerOptions: {} });
```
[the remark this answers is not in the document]
tanner-reits: This was a good catch. I didn't actually need to pull in that config anymore with how we handled these tests so I just stripped that entirely.

src/compiler/transformers/map-imports-to-path-aliases.ts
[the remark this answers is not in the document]
tanner-reits: This worked! Thanks for the suggestion!

src/compiler/transformers/test/map-imports-to-path-aliases.spec.ts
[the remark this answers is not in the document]
alicewriteswrongs: so right now basically the TS stuff is all being mocked out anyhow right? I didn't quite realize that when I first read through the tests

src/compiler/transformers/map-imports-to-path-aliases.ts
```diff
@@ -21,65 +21,63 @@ export const mapImportsToPathAliases = (
...
+            const shouldTranspileImportPath = hasResolvedFileName && !isModuleFromNodeModules;
 
-          if (shouldTranspileImportPath) {
-            // Create a global extension matching regular expression
-            // This can be used to strip all file extensions off the generated import path string
-            const extensionRegex = new RegExp(Object.values(ts.Extension).join('|'), 'g');
+            if (shouldTranspileImportPath) {
+              // Create a regular expression that will be used to remove the last file extension
+              // from the import path
+              const extensionRegex = new RegExp(Object.values(ts.Extension).join('$|'));
```
rwaskiewicz, line 51: Small bug - `join` will only be applied when there are >1 entries, and when it is applied, won't be applied to the last entry. So if `ts.Extension` was `['.js', '.ts', '.mjs']`, this only gives us:
```
new RegExp(Object.values(['.js', '.ts', '.mjs']).join('$|')) // Yields  /.js$|.ts$|.mjs/
```
I think we want something more like:
```
new RegExp(Object.values(['.js', '.ts', '.mjs']).map((ext) => `${ext}$`).join('|')) // Yields /.js$|.ts$|.mjs$/
```

Specifically:
```
new RegExp(Object.values(ts.Extension).map((ext) => `${ext}$`).join('|'))
```

rwaskiewicz (changes requested):

src/compiler/transformers/map-imports-to-path-aliases.ts
tanner-reits: Oof yeah you're right, silly mistake on my part. Thanks for the callout!

rwaskiewicz:
LGTM 👍 Gonna hold off on the official approval until we decide how we want to handle the child PR(s)

alicewriteswrongs (approved):
I think this looks good! 👍

rwaskiewicz (approved):

PR #3526, opened 2022-08-18
chore(tests): fix null check errors for jest preprocess

rwaskiewicz:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

There are ~9 `strictNullChecks` violations in our code related to jest preprocessing

GitHub Issue Number: N/A

## What is the new behavior?

this commit removes a handful of strict null check violations that can
be found in stencil's jest preprocessor. most changes are concerned with
making types agree - e.g. if a variable is declared to be of type `T`,
but is assigned by a function that returns `T | U`, the type of the
variable is expanded for agreement.

the two changes that are worth calling out in particular is introducting
a `throw` statement in the event that the `transformOptions` parameter
is falsey both when we process a file and attempt to retrieve a cache
key. from jest 23 thru 27, the function signatures of these two
functions changed a few times (documented in the code today). however,
a typing issue that was introduced in
e0640de27de7aa1b6ebcf96d4298a1c87d9954bb (#3171) was `transformOptions`
could be considered `undefined` by typescript. we know that for jest
[23,26], this value will never be falsey (as it is a part of the call
signature). for jest 27, we re-assign a value to this variable for
backwards compatability reasons.

there is a high level of confidence that this `throw` is safe to
introduce for the following reasons:
1. if `transformOptions` is undefined/null, stencil would already be
   erroring out when running the test task
2. the function signatures changes of these two functions have been
   handled in e0640de27de7aa1b6ebcf96d4298a1c87d9954bb
3. in jest [23,26], `transformOptions` _always_ has a `rootDir` field of
   type `string` (which is the field we're accessing here causing
   errors)

- [ ] Yes
- [x] No

## Testing

I spun up a small component library (`npm init stencil@latest component stencil-library`) and installed this branch on it.

I then proceeded to run `npm t -- --no-cache` with Jest 27 and then Jest 26

### DIFF
--- a/src/testing/jest/jest-preprocessor.ts
+++ b/src/testing/jest/jest-preprocessor.ts
@@ -14,7 +14,7 @@ type Jest27TransformOptions = { config: Jest26Config };
  * @returns `true` if the `obj` is a `Jest27TransformOptions`, false otherwise
  */
 const isJest27TransformOptions = (obj: unknown): obj is Jest27TransformOptions => {
-  return typeof obj === 'object' && obj.hasOwnProperty('config');
+  return obj != null && typeof obj === 'object' && obj.hasOwnProperty('config');
 };
 
 /**
@@ -59,7 +59,7 @@ export const jestPreprocessor = {
   ): string {
     // TODO(STENCIL-306): Drop support for versions of Jest <27
     /**
-     * As of Jest 27, `jestConfig` changes it's shape (as it's been moved into `transformOptions`). To preserve
+     * As of Jest 27, `jestConfig` changes its shape (as it's been moved into `transformOptions`). To preserve
      * backwards compatability, we allow Jest to pass 4 arguments and check the shape of the third and fourth arguments
      * to run Jest properly (in lieu of a global Jest version available to us). Support for this functionality will be
      * removed in a future major version of Stencil.
@@ -69,13 +69,22 @@ export const jestPreprocessor = {
       transformOptions = jestConfig.config;
     }
 
+    /**
+     * At this point, `transformOptions` should be a truthy value:
+     * - if we're running Jest 27, it should have been assigned in the previous conditional
+     * - if we're running Jest 26 and under, it should have been provided to us
+     */
+    if (!transformOptions) {
+      throw 'Unable to find Jest transformation options.';
+    }
+
     if (shouldTransform(sourcePath, sourceText)) {
       const opts: TranspileOptions = {
         file: sourcePath,
         currentDirectory: transformOptions.rootDir,
       };
 
-      const tsCompilerOptions: ts.CompilerOptions = getCompilerOptions(transformOptions.rootDir);
+      const tsCompilerOptions: ts.CompilerOptions | null = getCompilerOptions(transformOptions.rootDir);
       if (tsCompilerOptions) {
         if (tsCompilerOptions.baseUrl) {
           opts.baseUrl = tsCompilerOptions.baseUrl;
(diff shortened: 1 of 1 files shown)


src/testing/jest/jest-preprocessor.ts
```diff
@@ -14,7 +14,7 @@ type Jest27TransformOptions = { config: Jest26Config };
  * @returns `true` if the `obj` is a `Jest27TransformOptions`, false otherwise
  */
 const isJest27TransformOptions = (obj: unknown): obj is Jest27TransformOptions => {
-  return typeof obj === 'object' && obj.hasOwnProperty('config');
+  return obj != null && typeof obj === 'object' && obj.hasOwnProperty('config');
```
rwaskiewicz, line 17: Because `typeof obj === 'object'` is true when `obj` is `null` 🤦

src/testing/jest/jest-preprocessor.ts
```diff
@@ -136,6 +145,15 @@ export const jestPreprocessor = {
       transformOptions = jestConfigStr.config;
     }
 
+    /**
+     * At this point, `transformOptions` should be a truthy value:
+     * - if we're running Jest 27, it should have been assigned in the previous conditional
+     * - if we're running Jest 26 and under, it should have been provided to us
+     */
+    if (!transformOptions) {
+      throw 'Unable to find Jest transformation options.';
```
rwaskiewicz, line 154: This is _veryyyyy_ similar to https://github.com/ionic-team/stencil/pull/3526/files#diff-c877bce83f1eca295b644af29aac14781949edfa58775bc46b07006c8b18acabR78. I considered coalescing them into a function, but decided against it since the `transformOptions` reassignment conditional above each of them is ever so subtly different

tanner-reits (approved):
LGTM!

src/testing/jest/jest-preprocessor.ts
alicewriteswrongs: just a lovely little bit of JS lore 😂

src/testing/jest/jest-preprocessor.ts
```diff
@@ -220,7 +238,7 @@ function getCompilerOptions(rootDir: string): ts.CompilerOptions | null {
  * @returns `true` if the file should be transformed, `false` otherwise
  */
 export function shouldTransform(filePath: string, sourceText: string): boolean {
-  const ext = filePath.split('.').pop().toLowerCase().split('?')[0];
+  const ext = (filePath.split('.').pop() ?? '').toLowerCase().split('?')[0];
```
alicewriteswrongs, line 241: does this actually remove an error? in a node repl if I do

```js
"".split(".").pop()
```

I get `""`, so I'm not sure the `?? ""` gives us additional safety here. if I'm reading the [ECMA standard](https://tc39.es/ecma262/multipage/text-processing.html#sec-string.prototype.split) correctly if the substring to split on isn't found in the string to split then the return value should be a singleton array containing the original string (unless the original string is `""` _and_ the string / regexp on which to split matches that)

alicewriteswrongs (changes requested):
just had one question

src/testing/jest/jest-preprocessor.ts
rwaskiewicz: I think you're correct, the problem isn't ECMA but how TypeScript types `pop()` here:

```
   /**
     * Removes the last element from an array and returns it.
     * If the array is empty, undefined is returned and the array is not modified.
     */
    pop(): T | undefined;
```

Since TypeScript can't infer that the result of `split()` will always be a non-empty array

without the `?? ''`, we get:

```
src/testing/jest/jest-preprocessor.ts:241:15 - error TS2532: Object is possibly 'undefined'.

241   const ext = (filePath.split('.').pop()).toLowerCase().split('?')[0];
                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Sadly we add verbosity here because of TS 😢

alicewriteswrongs: typescript! \*shakes fist\*

alright sounds good :)

alicewriteswrongs (approved):
lgtm! 👍

PR #3528, opened 2022-08-18
fix up null checks in some mock-doc code

alicewriteswrongs:
This PR basically just entails fixing (some!) strict null checks errors in a few `mock-doc`-related files. I don't pretend to have fixed them _all_, but mainly tried to adjust the TS types so that they're more inline with how the code is actually used. For instance, we had a number of places where something was typed `string` but actually called with both strings and `null`. I tried to just smooth out some of that inconsistency where I could while changing as little actual code as possible.

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

GitHub Issue Number: N/A

## What is the new behavior?

I just tried to change the types in `src/mock-doc/node.ts` to reflect the values they actually are at runtime. I think I eliminated the null check errors from that file. There are a few ancillary changes to other files that sort of followed on from those changes.

- [ ] Yes
- [x] No

## Testing

Just made sure the tests are all passing.

### DIFF
--- a/src/mock-doc/node.ts
+++ b/src/mock-doc/node.ts
@@ -196,7 +196,7 @@ export class MockNode {
   }
 
   get textContent() {
-    return this._nodeValue;
+    return this._nodeValue ?? "";
   }
   set textContent(value: string) {
     this._nodeValue = String(value);
@@ -225,7 +225,7 @@ export class MockNodeList {
 
 export class MockElement extends MockNode {
   namespaceURI: string | null;
-  __attributeMap: MockAttributeMap | null;
+  __attributeMap: MockAttributeMap | null | undefined;
   __shadowRoot: ShadowRoot | null | undefined;
   __style: MockCSSStyleDeclaration | null | undefined;
 
@@ -266,8 +266,8 @@ export class MockElement extends MockNode {
     }
   }
 
-  get attributes() {
-    if (this.__attributeMap === null) {
+  get attributes(): MockAttributeMap {
+    if (this.__attributeMap == null) {
       const attrMap = createAttributeProxy(false)
       this.__attributeMap = attrMap
       return attrMap
@@ -495,7 +495,7 @@ export class MockElement extends MockNode {
     return this.getAttribute(attrName) !== null;
   }
 
-  hasAttributeNS(namespaceURI: string, name: string) {
+  hasAttributeNS(namespaceURI: string | null, name: string) {
     return this.getAttributeNS(namespaceURI, name) !== null;
   }
 
@@ -602,7 +602,7 @@ export class MockElement extends MockNode {
     }
   }
 
-  removeAttributeNS(namespaceURI: string, attrName: string) {
+  removeAttributeNS(namespaceURI: string | null, attrName: string) {
     const attr = this.attributes.getNamedItemNS(namespaceURI, attrName);
     if (attr != null) {
       this.attributes.removeNamedItemNS(attr);
@@ -1024,7 +1024,7 @@ function getElementsByTagName(elm: MockElement, tagName: string, foundElms: Mock
   const children = elm.children;
   for (let i = 0, ii = children.length; i < ii; i++) {
     const childElm = children[i];
-    if (tagName === '*' || childElm.nodeName.toLowerCase() === tagName) {
+    if (tagName === '*' || (childElm.nodeName ?? "").toLowerCase() === tagName) {
       foundElms.push(childElm);
     }
     getElementsByTagName(childElm, tagName, foundElms);
--- a/src/mock-doc/node.ts
+++ b/src/mock-doc/node.ts
@@ -17,14 +17,14 @@ import { NON_ESCAPABLE_CONTENT, SerializeNodeToHtmlOptions, serializeNodeToHtml 
 import { parseFragmentUtil } from './parse-util';
 
 export class MockNode {
-  private _nodeValue: string;
-  nodeName: string;
+  private _nodeValue: string | null;
+  nodeName: string | null;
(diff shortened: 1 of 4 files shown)


src/mock-doc/node.ts
```diff
@@ -261,9 +266,11 @@ export class MockElement extends MockNode {
...
-  get attributes() {
+  get attributes(): MockAttributeMap {
     if (this.__attributeMap == null) {
-      this.__attributeMap = createAttributeProxy(false);
+      const attrMap = createAttributeProxy(false);
+      this.__attributeMap = attrMap;
+      return attrMap;
     }
     return this.__attributeMap;
   }
```
alicewriteswrongs, line 276: this is a funny little bit here. I changed the type of `this.__attributeMap` to be `MockAttributeMap | null | undefined` which, afaict, reflects the types that the variable actually is set to by our code. This however causes the return value for this getter to be that type as well since the original implementation just returned `this.__attributeMap`:

https://github.com/ionic-team/stencil/blob/7f8e5c68a786cd82f8a60aaf4f0ee943ea6d5735/src/mock-doc/node.ts#L264-L269

Anywhoo - I wanted to make the type signature `() => MockAttributeMap` instead. When we check `this.__attributeMap == null` that narrows the type outside of the `if { }` block as long as we have a `return` inside of it, but unfortunately doing a reassignment like

```ts
this.__attributeMap = createAttributeProxy(false)
```

doesn't change the type of `this.__attributeMap`. So anyway, as a workaround I used a temporary variable `attrMap` (which will get the right type) and then just returned a reference to the same object that we stick on `this.__attributeMap` by returning that.

src/mock-doc/node.ts
```diff
@@ -1062,18 +1069,21 @@ export class MockHTMLElement extends MockElement {
...
     this.nodeName = value;
   }
 
-  override get attributes() {
+  override get attributes(): MockAttributeMap {
     if (this.__attributeMap == null) {
-      this.__attributeMap = createAttributeProxy(true);
+      const attrMap = createAttributeProxy(true);
+      this.__attributeMap = attrMap;
+      return attrMap;
```
alicewriteswrongs, line 1082: this change is for the same reason as `get attributes()` on `MockNode`, above

alicewriteswrongs (commented):
pointing out a few things

src/mock-doc/node.ts
```diff
@@ -261,9 +266,11 @@ export class MockElement extends MockNode {
     }
   }
 
-  get attributes() {
+  get attributes(): MockAttributeMap {
     if (this.__attributeMap == null) {
-      this.__attributeMap = createAttributeProxy(false);
+      const attrMap = createAttributeProxy(false);
+      this.__attributeMap = attrMap;
+      return attrMap;
```
rwaskiewicz, line 273: Non blocking - we could probably simplify this fn by not early returning:
```suggestion
      this.__attributeMap = createAttributeProxy(false);
```
by directly assigning to `__attributeMap` and letting the return statement outside of the `if` take care of the rest

src/mock-doc/node.ts
rwaskiewicz, line 1082: same comment as earlier, we may be able to simplify this a bit:
```suggestion
      this.__attributeMap = createAttributeProxy(true);
```

src/mock-doc/node.ts
```diff
@@ -297,7 +304,7 @@ export class MockElement extends MockNode {
 
   override cloneNode(_deep?: boolean): MockElement {
     // implemented on MockElement.prototype from within element.ts
-    return null;
+    return this;
```
rwaskiewicz, line 307: this is a little tricky
- we may signal to users that we're "cloning" a node when we're really not here (after all, it returned a value, why wouldn't they think it's a clone?)
- this may be considered a breaking change since we're changing the return value here

this may be one of the few cases where we want to return `null` (although IDK quite what the ramifications are there)

rwaskiewicz (changes requested):

src/mock-doc/node.ts
rwaskiewicz: Disregard, I missed the comments inline somehow 🤔

src/mock-doc/node.ts
rwaskiewicz: Disregard, I missed the comments inline somehow 🤔

src/mock-doc/node.ts
alicewriteswrongs: yeah I wasn't entirely sure how to proceed here, however I'm not sure that in normal use this method will ever get called - it's monkey-patched here:

https://github.com/ionic-team/stencil/blob/e6c8078b1d8c13505bc9691eb2d11b7fc1df7964/src/mock-doc/element.ts#L596-L616

which now makes me wonder why it's even here on `MockElement` 🤔🤔🤔

rwaskiewicz: Yeah, I'm not sure either. Beyond that comment on `element.ts#cloneNode` I don't have the full story myself.

Maybe the best thing to do here (which I really don't love) is to keep it as `null` and use a type assertion? 😨 

```suggestion
    // @ts-ignore - implemented on MockElement.prototype from within element.ts
    return null as unknown as MockElement;
```

alicewriteswrongs: sometimes

rwaskiewicz: Never before has Futurama been so apt 😆 😢

rwaskiewicz (approved):

tanner-reits (approved):
LGTM!

PR #3550, opened 2022-08-24
chore(ci): fix dirty git context for browserstack

rwaskiewicz:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes

ci fails when we modify the `scripts/` directory of the project

this is a symptom of a larger issue where the incorrect branch was built for browserstack tests,
we never noticed before because `scripts` contains both compiled and source code we archive,
and this is the first time we've modified it since the bug in CI was introduced

GitHub Issue Number: N/A

## What is the new behavior?

this commit modifies the `build_core` reusable workflow for stencil's ci
to take into account the event name when checking out the code to be
built. depending on the event name, the `HEAD` value used to check out
the code can differ. for event's with name 'pull_request_target', `HEAD`
is the head of the primary branch, in our case, `main`.

this lead to an issue in the stencil repo where two workflows would run
in order to build/run browserstack:

1. `build_core`, which would use `main@HEAD` when building the repo, and
   subesquently archiving the results for the next step
2. `Browserstack Tests`, which would checkout the code associated with
   the pull request, the unpack the saved dependencies from `main` over
   the contents of the pull request

because step 2 would take on files from `main` when the build archive was
unpacked, we would fail the git status check, because files from `main` would
clash with files on the checked out commit associated with the pr

this bug was originally introduced in 1cef74ff (#3454)

- [ ] Yes
- [x] No

## Testing

We use this same logic for checking code out when we run browserstack

Unfortunately, pull_request_target code needs to land to be tested 😢 so we're rolling the dice here 🎲

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,6 +12,11 @@ jobs:
     steps:
       - name: Checkout Code
         uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # v2.4.0
+        with:
+          # the pull_request_target event will consider the HEAD of `main` to be the SHA to use.
+          # attempt to use the SHA associated with a pull request and fallback to HEAD of `main`
+          ref: ${{ github.event_name == 'pull_request_target' && format('refs/pull/{0}/merge', github.event.number) || '' }}
+          persist-credentials: false
 
       - name: Get Core Dependencies
         uses: ./.github/workflows/actions/get-core-dependencies


tanner-reits (approved):
Little out of my realm of expertise, but think this makes sense based on your description and comments

alicewriteswrongs (approved):
🎲🤞😬 lets do it

Issue #3470, opened 2022-07-13
bug: rollup build cannot tree-shake properly when using dist-custom-elements + defineCustomElement + setAssetPath


jcfranco:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/.github/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

2.17.1

### Current Behavior

A rollup build of an app that uses `setAssetPath` and imports a subset of components (built w/ Stencil 2.17.1 + `defineCustomElement: true`), will not tree-shake properly and include code for **all** components. 

### Expected Behavior

The main bundle should only have code relevant to the imported component.

### Steps to Reproduce

See README from [repro case repo](https://github.com/jcfranco/stencil-use-case-where-build-is-not-tree-shaken/blob/master/README.md).

### Code Reproduction URL

https://github.com/jcfranco/stencil-use-case-where-build-is-not-tree-shaken

### Additional Information

This started after the single file export of all components got introduced by [Stencil v2.17.0](https://github.com/ionic-team/stencil/blob/main/CHANGELOG.md#-2170-2022-06-21).

A couple of suggestions to address this:

1. Add an option to `dist-custom-elements` to enable/disable exporting all components from `index.js`
2. Provide another way of importing `setAssetPath`

alicewriteswrongs:
Hi @jcfranco, thanks for filing this issue! I'm going to mark this for further discussion and refinement so we can figure out what to do exactly.

christian-bromann:
Hi @jcfranco 👋

We apologize for the delay in addressing this issue. The StencilJS team has been hard at work, and we appreciate your patience.

As StencilJS has undergone significant updates and improvements since this issue was originally posted, we believe that this specific concern may no longer be relevant with the latest version. Therefore, we are closing this issue.

If you are still experiencing this problem or have any new issues, please feel free to open a new issue with the latest details, and we will be happy to assist you.

Thank you for your understanding and support of StencilJS!

PR #3556, opened 2022-08-26
fix(compiler): revert component barrel export for `dist-custom-elements`

tanner-reits:
## Pull request checklist

## Pull request type

 

 

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

GitHub Issue Number: https://github.com/ionic-team/stencil/issues/3470

## What is the new behavior?

This commit reverts a change made to the `dist-custom-elements` output target that re-exported all component classes and `defineCustomElement` functions from a single `index.js` file in the output directory. This addresses the issue reported in STENCIL-500. Future improvements may be made as a result of STENCIL-457

- [X] Yes
- [ ] No

This will be released with Stencil V3. More changes related to this and the migration plan will come as a result of STENCIL-457

## Testing

Unit and E2E tests continue to pass locally. Validated `dist-custom-elements` output target output on an example repo.

### DIFF
--- a/src/compiler/output-targets/dist-custom-elements/index.ts
+++ b/src/compiler/output-targets/dist-custom-elements/index.ts
@@ -183,8 +183,6 @@ export const bundleCustomElements = async (
  */
 export const addCustomElementInputs = (buildCtx: d.BuildCtx, bundleOpts: BundleOptions): void => {
   const components = buildCtx.components;
-  // an array to store the imports of these modules that we're going to add to our entry chunk
-  const indexImports: string[] = [];
 
   components.forEach((cmp) => {
     const exp: string[] = [];
@@ -195,7 +193,6 @@ export const addCustomElementInputs = (buildCtx: d.BuildCtx, bundleOpts: BundleO
 
     if (cmp.isPlain) {
       exp.push(`export { ${importName} as ${exportName} } from '${cmp.sourceFilePath}';`);
-      indexImports.push(`export { {${exportName} } from '${coreKey}';`);
     } else {
       // the `importName` may collide with the `exportName`, alias it just in case it does with `importAs`
       exp.push(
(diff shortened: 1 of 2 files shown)


alicewriteswrongs:
one thing I see here is that we'll have to also undo the changes in https://github.com/ionic-team/stencil/pull/3468, which made a change to bring `dist/components/index.d.ts` inline with the then-recently-added component re-export change.

Which raises a question - does it make sense to delete this code now, or go ahead and work on the `OutputTargetExportBehavior` change we talked about?

rwaskiewicz:
You bring up a good point - we added the code, "should" delete it (we are here), only to re-add it 🤣 

Perhaps the better route to take here is

1. add `CustomElementsExportBehavior` (or whatever we call the enum we discussed on 2022.08.25) _with_ `ExportAll` (again, I'm not married to the name)
2.  remove the boolean `autoDefineCustomElements` and add `AutoDefine` as an option to `CustomElementsExportBehavior` as a follow up

The implication there being we don't revert to the `v2.16.0` behavior as I had originally thought we would, but rather would effectively put the current behavior behind a flag (`CustomElementsExportBehavior#ExportAll`), then do the refactoring for `autoDefineCustomElements`.  

Thoughts?

alicewriteswrongs:
I think what you proposed makes sense, this is all happening in v3 land anyhow!

tanner-reits:
I think that's all good. For this PR, I can add that new enum config option but for now only have the option of `exportAll` (or obviously `undefined` where nothing will happen). Then we can do follow-up PRs to replace `autoDefineCustomElements` and/or investigate the `defineCustomElements` helper function. Sound good to y'all?

rwaskiewicz:
@tanner-reits 👍

tanner-reits:
Gonna actually close this PR out and make a new one with the updated approach discussed above.

PR #3559, opened 2022-08-27
fix(compiler): add config flags for transpiling path aliases

tanner-reits:
## Pull request checklist

## Pull request type

 

 

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

Currently, aliased import paths are not transpiled at all. A previous PR to add new transpilation behavior would run by default. This PR adds functionality to guard that behavior behind a config flag.

GitHub Issue Number: https://github.com/ionic-team/stencil/issues/2815

## What is the new behavior?

This commit adds config flags to enable the transpilation of import path aliases as a part of the `dist-collection` output target. Part of the work for STENCIL-437

- [ ] Yes
- [X] No

## Testing

Unit tests were added to verify the new behavior only runs when explicitly enabled

### DIFF
--- a/src/compiler/config/outputs/validate-dist.ts
+++ b/src/compiler/config/outputs/validate-dist.ts
@@ -67,7 +67,7 @@ export const validateDist = (config: d.ValidatedConfig, userOutputs: d.OutputTar
           dir: distOutputTarget.dir,
           collectionDir: distOutputTarget.collectionDir,
           empty: distOutputTarget.empty,
-          transpileAliasedImportPaths: distOutputTarget.transpileAliasedImportPathsInCollection,
+          transformAliasedImportPaths: distOutputTarget.transformAliasedImportPathsInCollection,
         });
         outputs.push({
           type: COPY,
@@ -135,7 +135,7 @@ const validateOutputTargetDist = (config: d.ValidatedConfig, o: d.OutputTargetDi
     copy: validateCopy(o.copy ?? [], []),
     polyfills: isBoolean(o.polyfills) ? o.polyfills : undefined,
     empty: isBoolean(o.empty) ? o.empty : true,
-    transpileAliasedImportPathsInCollection: o.transpileAliasedImportPathsInCollection ?? false,
+    transformAliasedImportPathsInCollection: o.transformAliasedImportPathsInCollection ?? false,
   };
 
   if (!isAbsolute(outputTarget.buildDir)) {
--- a/src/compiler/config/outputs/validate-dist.ts
+++ b/src/compiler/config/outputs/validate-dist.ts
@@ -67,6 +67,7 @@ export const validateDist = (config: d.ValidatedConfig, userOutputs: d.OutputTar
           dir: distOutputTarget.dir,
           collectionDir: distOutputTarget.collectionDir,
           empty: distOutputTarget.empty,
+          transpileAliasedImportPaths: distOutputTarget.transpileAliasedImportPathsInCollection,
         });
         outputs.push({
           type: COPY,
@@ -134,6 +135,7 @@ const validateOutputTargetDist = (config: d.ValidatedConfig, o: d.OutputTargetDi
     copy: validateCopy(o.copy ?? [], []),
     polyfills: isBoolean(o.polyfills) ? o.polyfills : undefined,
     empty: isBoolean(o.empty) ? o.empty : true,
+    transpileAliasedImportPathsInCollection: o.transpileAliasedImportPathsInCollection ?? false,
   };
 
   if (!isAbsolute(outputTarget.buildDir)) {
--- a/src/compiler/config/test/validate-output-dist-collection.spec.ts
+++ b/src/compiler/config/test/validate-output-dist-collection.spec.ts
@@ -1,0 +1,103 @@
+import type * as d from '@stencil/core/declarations';
+import { validateConfig } from '../validate-config';
+import { mockConfig, mockLoadConfigInit } from '@stencil/core/testing';
+
+describe('validateDistCollectionOutputTarget', () => {
+  let config: d.Config;
+
+  beforeEach(() => {
+    config = mockConfig();
+  });
+
+  it('sets correct default values', () => {
+    const target: d.OutputTargetDistCollection = {
+      type: 'dist-collection',
+      empty: false,
+      dir: null,
+      collectionDir: null,
+    };
+    config.outputTargets = [target];
+
+    const { config: validatedConfig } = validateConfig(config, mockLoadConfigInit());
+
+    expect(validatedConfig.outputTargets).toEqual([
+      {
+        type: 'dist-collection',
+        empty: false,
+        dir: '/dist/collection',
+        collectionDir: null,
+        transformAliasedImportPaths: false,
+      },
+    ]);
+  });
+
+  it('sets specified directory', () => {
+    const target: d.OutputTargetDistCollection = {
+      type: 'dist-collection',
+      empty: false,
+      dir: '/my-dist',
+      collectionDir: null,
+    };
+    config.outputTargets = [target];
+
+    const { config: validatedConfig } = validateConfig(config, mockLoadConfigInit());
+
+    expect(validatedConfig.outputTargets).toEqual([
+      {
+        type: 'dist-collection',
+        empty: false,
+        dir: '/my-dist',
+        collectionDir: null,
+        transformAliasedImportPaths: false,
+      },
+    ]);
+  });
+
+  describe('transformAliasedImportPaths', () => {
+    it('sets option false when explicitly false in config', () => {
+      const target: d.OutputTargetDistCollection = {
+        type: 'dist-collection',
+        empty: false,
+        dir: null,
+        collectionDir: null,
+        transformAliasedImportPaths: false,
+      };
+      config.outputTargets = [target];
+
+      const { config: validatedConfig } = validateConfig(config, mockLoadConfigInit());
+
+      expect(validatedConfig.outputTargets).toEqual([
+        {
+          type: 'dist-collection',
+          empty: false,
+          dir: '/dist/collection',
+          collectionDir: null,
+          transformAliasedImportPaths: false,
+        },
+      ]);
+    });
+
+    it('sets option true when explicity true in config', () => {
+      const target: d.OutputTargetDistCollection = {
+        type: 'dist-collection',
+        empty: false,
+        dir: null,
+        collectionDir: null,
+        transformAliasedImportPaths: true,
+      };
+      config.outputTargets = [target];
+
+      const { config: validatedConfig } = validateConfig(config, mockLoadConfigInit());
+
+      expect(validatedConfig.outputTargets).toEqual([
+        {
+          type: 'dist-collection',
+          empty: false,
+          dir: '/dist/collection',
+          collectionDir: null,
+          transformAliasedImportPaths: true,
+        },
+      ]);
+    });
+  });
+});
--- a/src/compiler/config/test/validate-output-dist-collection.spec.ts
+++ b/src/compiler/config/test/validate-output-dist-collection.spec.ts
@@ -54,50 +54,30 @@ describe('validateDistCollectionOutputTarget', () => {
   });
 
   describe('transformAliasedImportPaths', () => {
-    it('sets option false when explicitly false in config', () => {
-      const target: d.OutputTargetDistCollection = {
-        type: 'dist-collection',
-        empty: false,
-        dir: null,
-        collectionDir: null,
-        transformAliasedImportPaths: false,
-      };
-      config.outputTargets = [target];
-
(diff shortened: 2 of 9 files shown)


src/declarations/stencil-public-compiler.ts
```diff
@@ -1845,6 +1845,27 @@ export interface OutputTargetDist extends OutputTargetBase {
...
+   *     "@utils/*": ['/src/utils/*']
+   *   }
+   * }
+   *
+   * // Source file
+   * import * as dateUtils from '@utils/date-utils';
+   * // Output file
+   * import { dateUtils } from '../utils/date-utils';
+   */
+  transpileAliasedImportPathsInCollection?: boolean | null;
```
tanner-reits, line 1867: I'm not sure how I feel about the approach needed here to pass this config flag on to the `dist-collection` target. The `dist` output target creates a `dist-collection` output automatically. I'm still confused and concerned about how these targets play together (and appear to not be able to be opted out of)

src/compiler/config/outputs/validate-dist.ts
```diff
@@ -67,6 +67,7 @@ export const validateDist = (config: d.ValidatedConfig, userOutputs: d.OutputTar
           dir: distOutputTarget.dir,
           collectionDir: distOutputTarget.collectionDir,
           empty: distOutputTarget.empty,
+          transpileAliasedImportPaths: distOutputTarget.transpileAliasedImportPathsInCollection,
```
alicewriteswrongs, line 70: is `transpile` the right verb? maybe we're transforming them, or dereferencing them? I'm not sure, maybe transpile is right

alicewriteswrongs (changes requested):
a few questions and thoughts

src/compiler/config/outputs/validate-dist.ts
tanner-reits: Wasn't totally sure about this either 😅 I just carried it over from the initial issue this change was based on. "Transform" might be the better option.

Why does terminology have to be the hardest part 😂

src/declarations/stencil-public-compiler.ts
```diff
@@ -1845,6 +1845,27 @@ export interface OutputTargetDist extends OutputTargetBase {
...
+   * {
+   *   paths: {
+   *     "@utils/*": ['/src/utils/*']
+   *   }
+   * }
+   *
+   * // Source file
+   * import * as dateUtils from '@utils/date-utils';
+   * // Output file
+   * import { dateUtils } from '../utils/date-utils';
```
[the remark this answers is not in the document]
tanner-reits: Oops yeah thanks for catching this

src/compiler/config/outputs/validate-dist.ts
rwaskiewicz: Before I read this thread, I was going to say "transform" might be the better option here 😆

src/declarations/stencil-public-compiler.ts
```diff
@@ -1845,6 +1845,27 @@ export interface OutputTargetDist extends OutputTargetBase {
   dir?: string;
 
   collectionDir?: string | null;
+  /**
+   * When `true` this flag will transpile aliased import paths defined in
```
rwaskiewicz, line 1849: Just in case we do 'transpile' -> 'transform':
```suggestion
   * When `true` this flag will transform aliased import paths defined in
```

src/declarations/stencil-public-compiler.ts
rwaskiewicz: Yeah - IMO for now we need to put this on both `OutputTargetDistCollection` and  `OutputTargetDist`. Otherwise the only way to configure the former without explicitly declaring the latter.

As a consquence of that, we'd have to add validation logic to `validate-collection.ts` as well

rwaskiewicz (changes requested):

src/compiler/config/outputs/validate-dist.ts
alicewriteswrongs: I think "transform" sounds good!

src/declarations/stencil-public-compiler.ts
```diff
@@ -1845,6 +1845,27 @@ export interface OutputTargetDist extends OutputTargetBase {
...
   collectionDir?: string | null;
+  /**
+   * When `true` this flag will transform aliased import paths defined in
+   * a project's `tsconfig.json` to relative import paths in the compiled output's
+   * `dist-collection` bundle if it is generated (i.e. `collectionDir` is set).
+   *
+   * Paths will be left in aliased format if `false` or `undefined`.
+   *
+   * // tsconfig.json
+   * {
```
rwaskiewicz, line 1856: Can we wrap this in a code block to make sure that formatting is adhered to in renderers for JSDoc?

Sometimes, editors like VSCode will not respect newlines like so:
Screen Shot 2022-08-30 at 4 02 01 PM

src/declarations/stencil-public-compiler.ts
```diff
@@ -1845,6 +1845,27 @@ export interface OutputTargetDist extends OutputTargetBase {
...
+   *   paths: {
+   *     "@utils/*": ['/src/utils/*']
+   *   }
+   * }
+   *
+   * // Source file
```
rwaskiewicz, line 1862: Can we wrap this in a code block to make sure that formatting is adhered to in renderers for JSDoc?

src/declarations/stencil-public-compiler.ts
```diff
@@ -1858,6 +1879,25 @@ export interface OutputTargetDistCollection extends OutputTargetBase {
...
```
rwaskiewicz, line 1895: Can we wrap this in a code block to make sure that formatting is adhered to in renderers for JSDoc?

src/declarations/stencil-public-compiler.ts
```diff
@@ -1858,6 +1879,25 @@ export interface OutputTargetDistCollection extends OutputTargetBase {
   empty?: boolean;
   dir: string;
   collectionDir: string;
+  /**
+   * When `true` this flag will transform aliased import paths defined in
+   * a project's `tsconfig.json` to relative import paths in the compiled output.
+   *
+   * Paths will be left in aliased format if `false` or `undefined`.
+   *
+   * // tsconfig.json
```
rwaskiewicz, line 1888: Can we wrap this in a code block to make sure that formatting is adhered to in renderers for JSDoc?

rwaskiewicz (commented):
Minor comments on the code itself. I'm gonna start putting some comments on the feature branch as a whole in separate threads

src/declarations/stencil-public-compiler.ts
tanner-reits: Leverage the JSDoc `@example` tag. That preserves standard code formatting and syntax highlighting. Up in [646bc5c](https://github.com/ionic-team/stencil/pull/3559/commits/646bc5c5906e74d45ad5d69ec137f05624764deb)

rwaskiewicz:
@tanner-reits 

I believe I've found a few bugs in the implementation when testing this end to end.

To make the reproductions easier to explain, I've create a repo with 2 branches, one for each bug:

- [issue1/missing-imports](https://github.com/rwaskiewicz/collection-ts-path-transform/tree/issue1/missing-imports) demonstrates some wonkiness in imports
- [issue2/potential-race-condition](https://github.com/rwaskiewicz/collection-ts-path-transform/tree/issue2/potential-race-condition) demonstrates a potential race condition

each branch has a step-by-step git commit history, as well as a readme that gives an overview of the issue. LMK if you have any questions on these!

tanner-reits:
> * issue2/potential-race-condition

Hmm okay I can take a look at these. Super weird this is happening even without any `paths` defined in the `tsconfig.json` files

tanner-reits:
> @tanner-reits
> 
> I believe I've found a few bugs in the implementation when testing this end to end.
> 
> To make the reproductions easier to explain, I've create a repo with 2 branches, one for each bug:
> 
> * [issue1/missing-imports](https://github.com/rwaskiewicz/collection-ts-path-transform/tree/issue1/missing-imports) demonstrates some wonkiness in imports
> * [issue2/potential-race-condition](https://github.com/rwaskiewicz/collection-ts-path-transform/tree/issue2/potential-race-condition) demonstrates a potential race condition
> 
> each branch has a step-by-step git commit history, as well as a readme that gives an overview of the issue. LMK if you have any questions on these!



> @tanner-reits
> 
> I believe I've found a few bugs in the implementation when testing this end to end.
> 
> To make the reproductions easier to explain, I've create a repo with 2 branches, one for each bug:
> 
> * [issue1/missing-imports](https://github.com/rwaskiewicz/collection-ts-path-transform/tree/issue1/missing-imports) demonstrates some wonkiness in imports
> * [issue2/potential-race-condition](https://github.com/rwaskiewicz/collection-ts-path-transform/tree/issue2/potential-race-condition) demonstrates a potential race condition
> 
> each branch has a step-by-step git commit history, as well as a readme that gives an overview of the issue. LMK if you have any questions on these!

Hmm okay I can take a look at these. Super weird this is happening even without any paths defined in the tsconfig.json files

rwaskiewicz (approved):
Approving, we'll tackle those issues I found in the root PR after we merge this

src/compiler/config/test/validate-output-dist-collection.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+          type: 'dist-collection',
+          empty: false,
+          dir: '/dist/collection',
+          collectionDir: null,
+          transformAliasedImportPaths: false,
+        },
+      ]);
+    });
+
+    it('sets option true when explicity true in config', () => {
```
alicewriteswrongs, line 80: might be a good idea to parametrize this test and the one immediately above with an `it.each`, just to DRY things up a lil bit

src/compiler/config/test/validate-output-dist.spec.ts
```diff
@@ -107,4 +109,90 @@ describe('validateDistOutputTarget', () => {
...
+      {
+        dir: path.join(rootDir, 'my-dist'),
+        type: 'dist-types',
+        typesDir: path.join(rootDir, 'my-dist', 'types'),
+      },
+      {
+        collectionDir: path.join(rootDir, 'my-dist', 'collection'),
+        dir: path.join(rootDir, '/my-dist'),
+        empty: false,
+        transformAliasedImportPaths: true,
```
alicewriteswrongs, line 168: here we're basically testing that this is passed through from the user-supplied `'dist'` output target, right? just wondering if it makes sense to narrow this assertion down to just test that, something like

```ts
let distCollectionOT = config.outputTargets.find(ot => ot.type === "dist-collection");
expect(distCollectionOT.transformAliasedImportPaths).toBe(true);
```

idk, might just be a bit more compact :)

alicewriteswrongs (commented):
I think this all looks good, just had a few small suggestions on tests!

src/compiler/config/test/validate-output-dist-collection.spec.ts
tanner-reits: Never used this before but it's cool! Thanks for showing me that!

alicewriteswrongs (approved):
lgtm! 👍

src/compiler/config/test/validate-output-dist.spec.ts
rwaskiewicz: IDK if I agree with this - the reason for which being is that tests like this, while admittedly wide in their assertions, are our primary way of checking if a change we've made affects other output targets as well.  Since we call `validateConfig` instead of `validateSpecifcOutputTarget`, we get all the side effects.  I would agree that what I'm describing isn't what is often taught as the "ideal" way of unit testing, it's our best way of catching regressions of this type today.  

Taking a step back, I'm also a bit squirmy about adding filtering logic to our tests - in my experience it makes it a little too easy for tests to creep in. e.g. if we change `isOutputTargetDist`, we've just affected the outcome of our tests too.

My vote is that we revert the changes to `src/compiler/config/test/validate-output-dist.spec.ts` in https://github.com/ionic-team/stencil/pull/3559/commits/c94c48836a1b467f96f6a438cf8e739b7c7cef23. Thoughts?

tanner-reits: I think I can go either way. I agree maybe the filter logic should be changed to just check the target `type` if we keep it. As for filtering the targets themselves, maybe we do revert at least for now. I think you make a good point that some of these targets are so coupled and tangled together that we might be safer checking the entirety of them. If/when we start to decouple them, then we could probably feel better about only checking certain ones

alicewriteswrongs: mmmmm @rwaskiewicz I think you are sharing wisdom here and probably right :)

rwaskiewicz (changes requested):
@tanner-reits can you fix the broken paths in Windows tests please? After that it should be good to go

rwaskiewicz (approved):

alicewriteswrongs (approved):
lgtm! 👍

PR #3502, opened 2022-08-01
feat(cli): update flag defaults for V3

tanner-reits:
## Pull request checklist

Please check if your PR fulfills the following requirements:
- [X] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [X] Build (`npm run build`) was run locally and any changes were pushed
- [X] Unit tests (`npm test`) were run locally and passed
- [X] E2E Tests (`npm run test.karma.prod`) were run locally and passed
- [X] Prettier (`npm run prettier`) was run locally and passed

## Pull request type

 

 

Please check the type of change your PR introduces:
- [ ] Bugfix
- [ ] Feature
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [X] Other (please describe): Updates CLI flag defaults for `sourceMaps` on the Stencil config and `generateTypeDeclarations` on the `dist-custom-elements` output target

## What is the current behavior?

- Stencil config `sourceMaps` flag defaults to `false`.
- `dist-custom-elements` config `generateTypeDeclarations` flag defaults to `false`

GitHub Issue Number: N/A

## What is the new behavior?

**NOTE:** Only affecting Stencil V3
- Stencil config `sourceMaps` flag now defaults to `true`.
- `dist-custom-elements` config `generateTypeDeclarations` flag now defaults to `true`

## Does this introduce a breaking change?

- [X] Yes
- [ ] No

Will be covered in V3 migration guide

## Testing

Updated unit tests for custom-element output target, validating the stencil config, and validating the stencil config sourceMaps flag.

PR #3562, opened 2022-08-29
feat(compiler): add `CustomElementExportBehavior` to custom elements …

tanner-reits:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

Currently, treeshaking for `dist-custom-elements` does not work when `autoDefineCustomElements` is enabled. This is due to a change in [v2.17](https://github.com/ionic-team/stencil/commit/ff0e8cc54e5e68631cd83302d59b19f2626d43cb) that adding re-exporting of components from the `index.js` file of the specified output location.

GitHub Issue Number: https://github.com/ionic-team/stencil/issues/3470

## What is the new behavior?

This commit adds a config option to the `dist-custom-elements` output target that will control the behavior for re-exporting class definitions and (eventually) defining them as custom elements. Essentially, when this option is not set, this output target will revert to the previous export behavior from v2.17. Start of addressing STENCIL-500

- [X] Yes
- [ ] No

This will be released as a part of Stencil v3. 

## Testing

Existing unit tests were updated to run against the new config option and additional tests were added to cover behavior when the new option is not specified. Generated output was correct on test repos


src/declarations/stencil-public-compiler.ts
```diff
@@ -1975,6 +1975,18 @@ export interface OutputTargetBaseNext {
...
+/**
+ * This enum defines all behavioral options available for
+ * exporting elements from the `dist-custom-elements` output target.
+ */
+export enum CustomElementsExportBehavior {
+  /**
+   * All components will be re-exported from the specified directories
+   * root `index.js` barrel file.
+   */
+  BARREL = 'barrel',
```
tanner-reits, line 1987: Went with "barrel" here since "exportAll" felt a little ambiguous. All the components get exported either way, all this behavior does is additionally export them from a single file. Not married to the name, so open to changing this if we feel there is something better

tanner-reits:
Called out one change. Also had an error pop up compiling using the build via `npm link` on a test project. I think it was just because the output imports from `@stencil/core` which would've been referencing a different version that did not have the export for the `CustomElementsExportBehavior` enum

src/declarations/stencil-public-compiler.ts
alicewriteswrongs: I hadn't heard `"barrel"` before you used it the other day, is your sense that's the most used terminology for this? If there's something generally used in the rollup world or in the wider JS world I'm down for whatever is the standard.

For whatever reason it makes me think of pirates, so maybe we can go with a whole nautical theme 🛢🏴‍☠️🦜⚓️⛵️

tanner-reits: AFAIK, "barrel" is common term. There are even VS code extensions for generating them like "auto barrel" or "... barrel generator". But, again, I'm totally open to using another term if we have concerns with this being too general or an unfamiliar term 🤷

rwaskiewicz: FWIW, I had never heard of "barrel" either - a cursory glance at the Google makes it seem like it is indeed used in the JS/TS world, I've just managed to miss it for ~7 years 😆. I'd personally be in favor of renaming it, but I'm struggling to think of what to call it otherwise (I'd probably call it "export file" or "entrypoint" to align with https://nodejs.org/api/packages.html#package-entry-points, personally)

tanner-reits: @alicewriteswrongs @rwaskiewicz How do we feel about something like "single entrypoint"? That seems to align with Node's practices that Ryan called out while also giving a little more context/explicitness to the fact that this will export everything from a single file

rwaskiewicz: 👍

alicewriteswrongs: I think something along the lines of 'single entrypoint' makes sense...not sure precisely how to word it, maybe something like 

```ts
enum CustomElementsExportBehavior {
  DEFAULT = "default",
  SINGLE_MODULE = "single-module",
}
```

I was also wondering if in this PR we should add some code into the output target validation to ensure that the property is always set on `DIST_CUSTOM_ELEMENTS` output targets w/ the default if the user doesn't set it. Ideally we'd push that up into the type system too. But if we always knew it would be defined then I think in general it would be good to rewrite the section of the code response for actually writing files to disk to look (at a super high level) something like

```ts
switch (outputTarget.customElementsExportBehavior) {
  case (CustomElementsExportBehavior.DEFAULT):
    // do everything required for the default case
  case (CustomElementsExportBehavior.SINGLE_MODULE):
    // do everything required for this
};
```

tanner-reits: I was just gonna go with `SINGLE_ENTRY_POINT`, but `SINGLE_EXPORT_MODULE` could also work.

I can add some logic to the target validation to enforce this is always defined after that runs, but there might be a decent work effort to refactor the code into a switch like you're proposing since the different values we're introducing will be leveraged in different functions along the call stack. Not saying we shouldn't do it, but maybe that level of work is best broken off into its own task

tanner-reits: Pushed up the changes for the option rename and adding the validation logic for the custom element output target

rwaskiewicz (changes requested):
I took a first pass at this, but wasn't able to get this to run on a local starter component. After running `npm init stencil component test-tanner`, I installed this branch on the component and made the following change:
```diff
diff --git a/stencil.config.ts b/stencil.config.ts
index dfa32ee..0514eef 100644
--- a/stencil.config.ts
+++ b/stencil.config.ts
@@ -1,4 +1,4 @@
-import { Config } from '@stencil/core';
+import { Config, CustomElementsExportBehavior } from '@stencil/core';
 
 export const config: Config = {
   namespace: 'test-tanner',
@@ -9,6 +9,7 @@ export const config: Config = {
     },
     {
       type: 'dist-custom-elements',
+      customElementsExportBehavior: CustomElementsExportBehavior.SINGLE_EXPORT_MODULE
     },
     {
       type: 'docs-readme',
```
which resulted in

```
> test-tanner@0.0.1 build
> stencil build --docs

[54:39.4]  @stencil/core
[54:39.5]  [LOCAL DEV] 🐧

[ ERROR ]  TypeError: Cannot read properties of undefined (reading 'SINGLE_EXPORT_MODULE') at Object.<anonymous>
           (/private/tmp/test-tanner/stencil.config.ts:14:73) at Module._compile
           (node:internal/modules/cjs/loader:1105:14) at Object.require.extensions..ts
           (/private/tmp/test-tanner/node_modules/@stencil/core/compiler/stencil.js:9770:18) at Module.load
           (node:internal/modules/cjs/loader:981:32) at Function.Module._load (node:internal/modules/cjs/loader:822:12)
           at Module.require (node:internal/modules/cjs/loader:1005:19) at require
           (node:internal/modules/cjs/helpers:102:18) at nodeRequire
           (/private/tmp/test-tanner/node_modules/@stencil/core/compiler/stencil.js:9777:24) at evaluateConfigFile
           (/private/tmp/test-tanner/node_modules/@stencil/core/compiler/stencil.js:67040:23) at loadConfigFile
           (/private/tmp/test-tanner/node_modules/@stencil/core/compiler/stencil.js:67012:34)
```

Is there something I'm not doing right here? I followed https://github.com/ionic-team/stencil-site/pull/909/files#diff-67136d57fb649bb16216c3eddc9028b9b6b0b8342c1b3d632ca9b750e9a2e19cR49, but maybe I missed something

src/declarations/stencil-public-compiler.ts
```diff
@@ -1975,6 +1975,22 @@ export interface OutputTargetBaseNext {
...
+ * exporting elements from the `dist-custom-elements` output target.
+ */
+export enum CustomElementsExportBehavior {
+  /**
+   * No additional export or definition behavior will happen.
+   */
+  DEFAULT = 'default',
+  /**
+   * All components will be re-exported from the specified directory's
+   * root `index.js` file.
```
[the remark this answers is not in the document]
tanner-reits: That's not entirely true tho since treeshaking does work as long as `autoDefineCustomElements` isn't enabled. Which, by the time V3 is released, those two behaviors will not be able to be used together.

src/internal/stencil-core/index.d.ts
```diff
@@ -50,3 +50,5 @@ export {
 } from '../stencil-public-runtime';
 
 export type { StencilConfig as Config, PrerenderConfig } from '../stencil-public-compiler';
+
+export { CustomElementsExportBehavior } from '../stencil-public-compiler';
```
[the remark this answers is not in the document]
tanner-reits: You can't use enums as a value if they're imported as a `type`, so, rather than changing the existing import from a `type` import I added this for the values we need as standard imports.

src/declarations/stencil-public-compiler.ts
```diff
@@ -1993,6 +2009,12 @@ export interface OutputTargetDistCustomElements extends OutputTargetBaseNext {
    * Enables the generation of type definition files for the output target.
    */
   generateTypeDeclarations?: boolean;
+  /**
+   * Define the export/definition behavior for the output target's generated output.
+   * This controls if/how custom elements will be defined or where components will be exported from.
+   * If omitted, no auto-definition behavior or re-exporting will happen.
+   */
+  customElementsExportBehavior?: CustomElementsExportBehavior | null;
```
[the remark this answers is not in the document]
tanner-reits: Yeah I think that can be removed since I made a change to add validation logic to the target so that after validation runs there is always a value assigned and it defaults to `DEFAULT`

src/compiler/output-targets/test/output-targets-dist-custom-elements.spec.ts
```diff
@@ -130,55 +130,99 @@ describe('Custom Elements output target', () => {
...
+        buildCtx.components = [componentOne, componentTwo];
+
+        const bundleOptions = getBundleOptions(
+          config,
+          buildCtx,
+          compilerCtx,
+          config.outputTargets[0] as OutputTargetDistCustomElements
+        );
+        addCustomElementInputs(buildCtx, bundleOptions, config.outputTargets[0] as OutputTargetDistCustomElements);
+        expect(bundleOptions.loader['\0core'].replace(/\s/g, '')).toEqual(
```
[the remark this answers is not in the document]
tanner-reits: It was cuz tests were complaining about spacing. Turns out it was just new-lines so I'll remove these calls.

src/compiler/output-targets/dist-custom-elements/custom-elements-types.ts
```diff
@@ -54,16 +59,20 @@ const generateCustomElementsTypesOutput = async (
...
-      //
-      // To construct this path we:
-      //
-      // - get the relative path to the component's source file from the source directory
-      // - join that relative path to the relative path from the `index.d.ts` file to the
-      //   directory where typedefs are saved
-      const componentSourceRelPath = relative(config.srcDir, component.sourceFilePath).replace('.tsx', '');
-      const componentDTSPath = join(componentsTypeDirectoryRelPath, componentSourceRelPath);
+
+      let componentDTSPath = join(componentsTypeDirectoryRelPath, component.tagName, component.tagName);
```
rwaskiewicz, line 63: ATM, `isBarrelExport` is falsey IFF (if & only if) `outputTarget.customElementsExportBehavior === 'default'`. In the scenario for 'default', where we don't generate any additional types, and use the `componentDTSPath` calculated here. However, I don't believe this path is correct - by:
1. Spinning up a new project with `npm init stencil component`
2. Installing this branch in the starter project 
3. Adding the following to `stencil.config.ts`:
```diff
diff --git a/stencil.config.ts b/stencil.config.ts
index b8fb12f..76697ff 100644
--- a/stencil.config.ts
+++ b/stencil.config.ts
@@ -9,7 +9,8 @@ export const config: Config = {
     },
     {
       type: 'dist-custom-elements',
+      customElementsExportBehavior: 'default',
+      generateTypeDeclarations: true,
     },
     {
```
4. Renaming the default `MyComponent` to `MyComponentClazz` and the tag to `my-component-diff-tag`
4. Running `npm run build` yields a file with an import path like so:
```ts
// dist/components/index.d.ts
export { MyComponentClazz as MyComponentDiffTag } from '../types/components/my-component-diff-tag/my-component-diff-tag';
```
However, that path in the export statement does not exist on the filesystem.

I think this is occurring because in v2.X, where the original changes from https://github.com/ionic-team/stencil/pull/3468/files occurred, we don't create types be default. However in v3.X, we do generate those types https://github.com/ionic-team/stencil/pull/3502/files#diff-4a8e48905c6379413a82ea0524130976379f57f8976e2abdfdb44082c0fc0499R40.  

I think we need to do a little more here to ensure type generation is taken into account - can you PTAL?

src/declarations/stencil-public-compiler.ts
```diff
@@ -1975,6 +1975,25 @@ export interface OutputTargetBaseNext {
   dir?: string;
 }
 
+/**
+ * The collection of valid export behaviors.
+ * Used to generate a type for typed configs as well as output target validation
+ * for the `dist-custom-elements` output target.
+ *
+ * Adding an value to this const array will automatically add it as a valid option on the
```
rwaskiewicz, line 1983:
```suggestion
 * Adding a value to this const array will automatically add it as a valid option on the
```

rwaskiewicz (changes requested):

src/compiler/output-targets/dist-custom-elements/custom-elements-types.ts
tanner-reits: Think I got this squared away. Now, these paths will only be generated if the `single-export-module` behavior is set and uses the correct logic to generate the paths (I initially reverted behavior _too_ far which is why you were seeing paths that didn't exist). Tested the output with combinations of type generation and `customElementsExportBehavior` we currently have.

rwaskiewicz (changes requested):

rwaskiewicz (approved):

alicewriteswrongs (approved):
all looks good!

rwaskiewicz:
Holding off on the merge until https://github.com/ionic-team/stencil-site/pull/909 is ready to go

Issue #3253, opened 2022-02-23
bug: ref is set to null after rerender due to changing prop value


amandaejohnston:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/.github/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

2.13.0

### Current Behavior

In the Ionic Framework, on the `display` test of `ion-datetime`, changing the `presentation` prop of the `datetime` component sometimes causes the `ref` in `renderCalendarBody` to fire twice and/or get set to `null`. (See steps to repro for details)

### Expected Behavior

The `ref` should not be set to `null`, because the element in question is present in the DOM.

### Steps to Reproduce

1. Clone the `ionic-framework` repo.
2. Open `core/src/components/datetime/datetime.tsx`. In the `renderCalendarBody` function, add a `console.log` to the `ref` callback like so: `<div class="calendar-body ion-focusable" ref={el => { console.log('ref:', el); this.calendarBodyRef = el} } tabindex="0">`
3. Run `npm run start` in the `/core` directory. Once the build is finished, go here in your browser: http://localhost:3333/src/components/datetime/test/display
4. Change the "presentation" dropdown to "time-date." Note in the console that the `ref` callback fires twice, receiving `null` first, then the proper element.
5. Change the "presentation" back to "date-time." Note that the callback fires twice again, but in the opposite order, such that `this.calendarBodyRef` finishes at `null`.

### Code Reproduction URL

http://localhost:3333/src/components/datetime/test/display

### Additional Information

- The source for the `display` test can be found at `/core/src/components/datetime/test/display/index.html`.
- Changing "presentation" to "date" only fires the `ref` once.
- Changing it to "time" sets the `ref` to null, which in this case is correct behavior, since there is no calendar body to render in this mode.
- `renderCalendarBody` is called in `renderCalendar`, which is called in `renderDatetime`, which returns an array that gets passed directly to the `render` function.
- Both `date-time` and `time-date` presentations lead to the same render functions being called in `renderDatetime`, just in a different order.

alicewriteswrongs:
[...]

PR #3572, opened 2022-08-31
test(runtime): add regression tests for `updateChildren` algorithm

alicewriteswrongs:
This adds tests which ensure that children are being correctly
reconciled between re-renders and, in particular, that the handling of
`ref` attrs set on those children (or _their_ children) are handled
correctly.

This is in part adding a regression test for the fix for the issue
described in #3253.

## Pull request checklist

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Other (please describe): adding test coverage

GitHub Issue Number: N/A

## What is the new behavior?

Just adding a few tests

- [ ] Yes
- [x] No


src/runtime/test/render-vdom.spec.tsx
```diff
@@ -1097,5 +1097,93 @@ describe('render-vdom', () => {
       await waitForChanges();
       expect(rootInstance.counter).toEqual(2);
     });
+
+    it('should not call ref cb w/ null when children are reordered', async () => {
+      @Component({ tag: 'cmp-a' })
+      class CmpA {
+        counter = 0;
```
rwaskiewicz, line 1104: Does this field ever get used? Can we remove it?

src/runtime/test/render-vdom.spec.tsx
```diff
@@ -1097,5 +1097,93 @@ describe('render-vdom', () => {
...
+      // We've changed the state and forced a re-render. This tests one of the
+      // ways in which children can be re-ordered that the `updateChildren` algo
+      // can handle without having `key` attrs set.
+      expect(rootInstance.divRef).toEqual(root.querySelector('.a'));
+    });
+
+    it('should not call ref cb w/ null when children w/ keys are reordered', async () => {
+      @Component({ tag: 'cmp-a' })
+      class CmpA {
+        counter = 0;
```
rwaskiewicz, line 1143: Does this field ever get used? Can we remove it?

src/runtime/test/render-vdom.spec.tsx
```diff
@@ -1097,5 +1097,93 @@ describe('render-vdom', () => {
...
+            ? [this.renderB(), this.renderA(), this.renderB(), this.renderB()]
+            : [this.renderA(), this.renderB(), this.renderB()];
+        }
+      }
+
+      const { root, rootInstance, waitForChanges } = await newSpecPage({
+        components: [CmpA],
+        html: `<cmp-a></cmp-a>`,
+      });
+
```
rwaskiewicz, line 1170: Does it make sense for us to put an assertion before we force the re-render that `divRef` is `undefined`?
```suggestion
expect(rootInstance.divRef).toBeUndefined();
```

That way, we know that the re-render definitely happened, as opposed to a scenario where something funky happens and the contents of `renderA()` were being rendered the whole time

rwaskiewicz (changes requested):

src/runtime/test/render-vdom.spec.tsx
alicewriteswrongs: ahh indeed it does not 🤦‍♀️

src/runtime/test/render-vdom.spec.tsx
```diff
@@ -1097,5 +1097,93 @@ describe('render-vdom', () => {
...
+            ? [this.renderB(), <div>middle</div>, this.renderA()]
+            : [this.renderA(), <div>middle</div>, this.renderB()];
+        }
+      }
+
+      const { root, rootInstance, waitForChanges } = await newSpecPage({
+        components: [CmpA],
+        html: `<cmp-a></cmp-a>`,
+      });
+
```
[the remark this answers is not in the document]
alicewriteswrongs: yeah I think it's a good idea to keep a check there - in this case though we'll actually be checking that it's set to the right element after the first render - what we're testing here is basically that the ref is called the right way on the first render and then is _not_ overwritten with `null`

src/runtime/test/render-vdom.spec.tsx
```diff
@@ -1097,5 +1097,94 @@ describe('render-vdom', () => {
...
+        html: `<cmp-a></cmp-a>`,
+      });
+      // ref should be set correctly after the first render
+      expect(rootInstance.divRef).toEqual(root.querySelector('.a'));
+      root.state = false;
+      await waitForChanges();
+      // We've changed the state and forced a re-render. This tests one of the
+      // ways in which children can be re-ordered that the `updateChildren` algo
+      // can handle without having `key` attrs set.
+      expect(rootInstance.divRef).toEqual(root.querySelector('.a'));
```
rwaskiewicz, line 1137: Hmmm I don't think I understood this the first time around..

Is the idea here that after the re-render, `divRef` hasn't changed?

alicewriteswrongs: yeah, basically the tests are checking that in these situations the issue which showed up in Framework doesn't happen, in the one because we're doing a relatively simple reorder and the other because we're setting key attributes (on that one if you remove the `key` attr then the test should fail)

rwaskiewicz: Gotcha, I think where my understanding falls apart is that there's only one place we assign `this.divRef`, https://github.com/ionic-team/stencil/pull/3572/files/5549f41e2500b0a7104cf6ae7c1cfa5ca821443f#diff-ec92564e32a1aaa4fbee8ec77fa6269cdfb57581ca28f2d4764c276360fa60abR1109.

Should that also be on the `div` returned by `renderB`, or am I not know what I don't know? 😆

alicewriteswrongs: right so that `ref` function on there will be called when the related DOM node is created and when it is destroyed. So when the component first renders we call the `renderA` method and the `ref` will fire w/ the new DOM node as an argument. Then later on we re-render, and basically what we're testing is that when we re-render _with the children in a different order_ the `updateChildren` algorithm for detecting when child nodes in the new and old children are the same works correctly.

Basically if that algorithm is not working correctly then what will happen is that on render no. 2 we'll trigger that `ref` race condition. In that scenario, Stencil's vdom doesn't recognize that the output of `renderA` is present both in new children and old children, and so it will remove the DOM nodes associated with `renderA` in the old children and create new ones for the new children. If this happens in a particular order due to the order of the child nodes in the old and new children then the removal of the old nodes will happen after the addition of the new ones, causing the ref callback to be called 3 times in total

1. on the first re-render
2. when the new child is added for render no. 2
3. and then again with `null` as the child from render no. 1 is removed

So! Basically the test with the keys is testing out a scenario where the arrangement of the children will trigger this issue if the keys aren't present.

does that all make sense? possibly I should add some more comments to the tests describing what we're trying to test here

rwaskiewicz: Yep! That all makes sense - I do like the idea of adding a few more comments here for future folks reading these, would you please add those?

alicewriteswrongs: will do!

alicewriteswrongs: @rwaskiewicz ok just pushed a slight update to add some comments explaining what's up

rwaskiewicz (approved):

PR #3579, opened 2022-09-02
chore(output-targets): soft-remove the dist-custom-elements-bundle…

alicewriteswrongs:
… output target

This soft-removes the dist-custom-elements-bundle output target by
removing it from our list of valid output targets and fixing the few
issues that causes w/ tests, type checking, etc.

This also adds TODO(STENCIL-561) comments that (hopefully) point out the
main places we'll need to delete code later one when we tear the whole
thing out all the way.

STENCIL-260: Soft Removal of dist-custom-elements-bundle

## Pull request checklist

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Other (please describe): soft-removing the dist-custom-elements-bundle

GitHub Issue Number: N/A

## What is the new behavior?

This basically removes the OT from a list of valid output targets that a user could supply in a config but keeps the code around for now, we'll circle back later to fully remove it.

- [x] Yes
- [ ] No

Previously this was soft-deprecated with a little deprecation message. This instead makes it so that configurations trying to use the `dist-custom-elements-bundle` won't work anymore, so users currently relying on the now-deprecated behavior will have to change how they build their Stencil projects.


tanner-reits (approved):
Just so I'm clear on our process here, we're gonna remove `dist-custom-elements-bundle` as an exposed output target in V3 (this PR) and then follow-up with a future PR somewhere after V3 where all the logic for generating the output is removed, correct?

alicewriteswrongs:
that's what I gathered from the relevant ticket, I think the approach generally makes sense and it's going to be a lot more work to actually tear all this stuff out so let's leave that as a job for future stencil team to do :)

src/compiler/output-targets/dist-custom-elements-bundle/custom-elements-build-conditionals.ts
```diff
@@ -1,3 +1,4 @@
+// TODO(STENCIL-561): fully delete dist-custom-elements-bundle code
```
rwaskiewicz, line 1: Nit/fun fact - this file is also used in `src/compiler/output-targets/dist-custom-elements/index.ts` - can we update the TODO to move it to `dist-custom-elements/` and rename it?

test/end-to-end/test-end-to-end-dist.js
```diff
@@ -10,9 +10,6 @@ fs.accessSync(path.join(distDir, 'index.cjs.js'));
 fs.accessSync(path.join(distDir, 'index.js'));
 
 const customElementsDir = path.join(distDir, 'custom-elements');
```
rwaskiewicz, line 12: Does this variable for `/dist/custom-elements` still get used? If not, can we rm it?

rwaskiewicz (changes requested):

src/compiler/output-targets/dist-custom-elements-bundle/custom-elements-build-conditionals.ts
alicewriteswrongs: that is indeed "fun"

test/todo-app/src/components.d.ts
```diff
@@ -15,6 +15,14 @@ export namespace Components {
         "text": string;
     }
 }
+export interface TodoInputCustomEvent<T> extends CustomEvent<T> {
```
rwaskiewicz, line 18: Did this change get triggered by other things in the PR? Or is this a case of us (me) merging `main` into this branch and not running a build?

alicewriteswrongs: oh right, I meant to point this out / ask

I had some test failures I had to work through with the `test.analysis` npm script, and running it locally led to these changes. Should we commit them?

rwaskiewicz: Ah, I see what happened here. These should have been committed a while ago 😨 

My vote is we split them out into a separate PR. Can you do me a favor and create a branch off of `v3.0.0-dev` and run `npm run build` in the `test` dir? That should be enough to replicate these changes. While we're there, it may make sense to propagate our check for a dirty git context over to the main CI workflow (the irony being, IIRC, that's where we originally had it in the first implementation PR before I asked for it only to be in Browserstack 🤦 )

alicewriteswrongs: haha ok sure, will do, and I'll delete the file here 👍

alicewriteswrongs: https://github.com/ionic-team/stencil/pull/3582

rwaskiewicz (approved):

PR #3589, opened 2022-09-08
chore(types): fix logs.ts strictNullChecks errors

rwaskiewicz:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Stencil has a few `strictNullChecks` violations in `logs.ts`

GitHub Issue Number: N/A

## What is the new behavior?

this commit resolves the `strictNullChecks` errors in `logs.ts` by:

1. creating a default `CompilerSystem` object for `loadedCompilerConfig` in the event that the fields are not defined - making the fields read by this function were considered, but would be a breaking change to the public api and didn't seem worth it (at this time)
2. replacing `Config` with `ValidatedConfig` where applicable/possible

this commit also adds JSDoc to the file

- [ ] Yes
- [x] No

## Testing

Type checker continues to pass


src/cli/logs.ts
```diff
@@ -28,13 +49,29 @@ export const startupLogVersion = (logger: Logger, task: TaskCommand, coreCompile
...
-) => {
-  const sysDetails = sys.details;
+): void => {
+  const sysDetails: SystemDetails = sys.details ?? {
+    platform: '',
+    cpuModel: 'UNKNOWN',
+    freemem: () => 0,
+    release: 'UNKNOWN',
+    totalmem: 0,
+  };
```
rwaskiewicz, line 74: Technically this is a change in functionality. However, prior to this change, `sysDetails` could be `undefined`, and users would run into "trying to access [PROPERTY_NAME] of undefined" errors on `run()`

tanner-reits: Does it make sense for these logs to print if there are no `sysDetails`? Should another message be printed instead?

Also, for my knowledge, in what case would the details not exist?

rwaskiewicz: > Does it make sense for these logs to print if there are no sysDetails? Should another message be printed instead?

Fair, we could alternatively log something like "Unable to determine System Details"

> Also, for my knowledge, in what case would the details not exist?

Someone could invoke `run()` manually via the Node API, where they handcraft the `sys` entity :-(

tanner-reits: > Fair, we could alternatively log something like "Unable to determine System Details"

I'd be in favor of something like this instead. Feels clearer than us generating "fake" information.

rwaskiewicz: 👍 SGTM - I'll wait for @alicewriteswrongs to weigh in and make those changes once we have consensus

src/cli/logs.ts
```diff
@@ -1,16 +1,37 @@
-import type { Config, Logger, CompilerSystem, TaskCommand } from '../declarations';
+import type { Logger, CompilerSystem, SystemDetails, TaskCommand, ValidatedConfig } from '../declarations';
 import type { ConfigFlags } from './config-flags';
 import type { CoreCompiler } from './load-compiler';
 
-export const startupLog = (logger: Logger, task: TaskCommand) => {
+/**
+ * Log the name of this package (`@stencil/core`) to an output stream
+ *
+ * The output stream is determined by the {@link Logger} instances that is provided as an argument to this function
```
alicewriteswrongs, line 8:
```suggestion
 * The output stream is determined by the {@link Logger} instance that is provided as an argument to this function
```

src/cli/logs.ts
```diff
@@ -1,16 +1,37 @@
...
   }
 
   logger.info(logger.cyan(`@stencil/core`));
 };
 
-export const startupLogVersion = (logger: Logger, task: TaskCommand, coreCompiler: CoreCompiler) => {
+/**
+ * Log this package's version to an output stream
+ *
+ * The output stream is determined by the {@link Logger} instances that is provided as an argument to this function
```
alicewriteswrongs, line 26:
```suggestion
 * The output stream is determined by the {@link Logger} instance that is provided as an argument to this function
```

alicewriteswrongs (changes requested):
a few comments / suggestions

src/cli/logs.ts
rwaskiewicz: thanks all - added https://github.com/ionic-team/stencil/pull/3589/commits/bf69f45b9f46f7a516e21978ab1a9c8f3a3b14c8 in order to keep similar behavior for cases where the logger is set to debug or `flags.CI` is truthy, without introducing a separate conditional. LMK what yinz think

tanner-reits (approved):

alicewriteswrongs (approved):
lgtm!

PR #3604, opened 2022-09-13
chore(compiler): add JSDocs and some types related to @State decorator

alicewriteswrongs:
This updates the docstrings and typings for a few functions which have to do with the `@State` decorator and, in particular, the code which is used to convert class field declarations to getters.

This also switches over to use `ts.factory` methods in a few places where deprecation messages were popping up.

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

GitHub Issue Number: N/A

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/src/compiler/transformers/decorators-to-static/state-decorator.ts
+++ b/src/compiler/transformers/decorators-to-static/state-decorator.ts
@@ -3,6 +3,18 @@ import ts from 'typescript';
 import { createStaticGetter } from '../transform-utils';
 import { isDecoratorNamed } from './decorator-utils';
 
+/**
+ * Convert class fields decorated with `@State` to static getters
+ *
+ * This function takes a list of decorated properties pulled off of a class
+ * declaration AST Node and builds up equivalent static getter AST nodes
+ * with which they can be replaced.
+ *
+ * @param decoratedProps TypeScript AST nodes representing class members
+ * @param watchable set of names of fields which should be watched for changes
+ * @param newMembers an out param containing new class members
+ *
+ */
 export const stateDecoratorsToStatic = (
   decoratedProps: ts.ClassElement[],
   watchable: Set<string>,
@@ -14,11 +26,24 @@ export const stateDecoratorsToStatic = (
     .filter((state) => !!state);
 
   if (states.length > 0) {
-    newMembers.push(createStaticGetter('states', ts.createObjectLiteral(states, true)));
+    newMembers.push(createStaticGetter('states', ts.factory.createObjectLiteralExpression(states, true)));
   }
 };
 
-const stateDecoratorToStatic = (prop: ts.PropertyDeclaration, watchable: Set<string>) => {
+/**
+ * Convert a property declaration decorated with `@State` to a property
+ * assignment AST node which maps the name of the state property to an empty
+ * object.
+ *
+ * Note that this function will return null if the property declaration is
+ * decorated with other decorators.
+ *
+ * @param prop A TypeScript AST node representing a class property declaration
+ * @param watchable set of names of fields which should be watched for changes
+ * @returns a property assignment AST Node which maps the name of the state
+ * prop to an empty object
+ */
+const stateDecoratorToStatic = (prop: ts.PropertyDeclaration, watchable: Set<string>): ts.PropertyAssignment | null => {
   const stateDecorator = prop.decorators.find(isDecoratorNamed('State'));
   if (stateDecorator == null) {
     return null;
@@ -26,5 +51,8 @@ const stateDecoratorToStatic = (prop: ts.PropertyDeclaration, watchable: Set<str
 
   const stateName = prop.name.getText();
   watchable.add(stateName);
-  return ts.createPropertyAssignment(ts.createLiteral(stateName), ts.createObjectLiteral([], true));
+  return ts.factory.createPropertyAssignment(
+    ts.createLiteral(stateName),
+    ts.factory.createObjectLiteralExpression([], true)
+  );
 };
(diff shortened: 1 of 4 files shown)


src/compiler/transformers/decorators-to-static/state-decorator.ts
```diff
@@ -3,6 +3,18 @@ import ts from 'typescript';
...
+ * Convert class fields decorated with `@State` to static getters
+ *
+ * This function takes a list of decorated properties pulled off of a class
+ * declaration AST Node and builds up equivalent static getter AST nodes
+ * with which they can be replaced.
+ *
+ * @param decoratedProps TypeScript AST nodes representing class members
+ * @param watchable set of names of fields which should be watched for changes
+ * @param newMembers an out param containing new class members
+ *
```
rwaskiewicz, line 16: Nit - do we need this extra newline?

src/compiler/transformers/decorators-to-static/state-decorator.ts
```diff
@@ -14,17 +26,33 @@ export const stateDecoratorsToStatic = (
...
   const stateDecorator = prop.decorators.find(isDecoratorNamed('State'));
   if (stateDecorator == null) {
     return null;
   }
 
   const stateName = prop.name.getText();
   watchable.add(stateName);
-  return ts.createPropertyAssignment(ts.createLiteral(stateName), ts.createObjectLiteral([], true));
+  return ts.factory.createPropertyAssignment(
+    ts.createLiteral(stateName),
```
rwaskiewicz, line 55: Could this be:
```suggestion
    ts.factory.createStringLiteral(stateName),
```
? IDK if you played around with that or not (and if the former if something awful happened as a result 😆)

rwaskiewicz (changes requested):
2 minor questions

src/compiler/transformers/decorators-to-static/state-decorator.ts
alicewriteswrongs: no we do not!

src/compiler/transformers/decorators-to-static/state-decorator.ts
alicewriteswrongs: I think we want to do `ts.factory.createIdentifier` - tried that out and it seems fine?

rwaskiewicz: Hmmm, I'm not sure I follow. Why `createIdentifier`?

Looking at the [current implementation of `createLiteral`](https://github.com/microsoft/TypeScript/blob/b23f1d6b59beabde1072d14b7fb65781ce209c15/src/deprecatedCompat/4.0/nodeFactoryTopLevelExports.ts#L963-L984), it defaults to `createStringLiteral` for arguments of type `string`:
```ts
    export const createLiteral = Debug.deprecate(function createLiteral(value: string | number | PseudoBigInt | boolean | StringLiteral | NoSubstitutionTemplateLiteral | NumericLiteral | Identifier): PrimaryExpression {
        // ...
        if (typeof value === "string") {
            return factory.createStringLiteral(value, /*isSingleQuote*/ undefined);
        }
        return factory.createStringLiteralFromNode(value);
    } // ...
```
but I don't think the same can be said for `createIdentifier` - https://github.com/microsoft/TypeScript/blob/a4cabe725b413f154f738b48c1fe2f053cff7d26/src/compiler/factory/nodeFactory.ts#L877

Unrelated, looks like non-factory based calls will start to error in TS 5.0 - https://github.com/microsoft/TypeScript/blob/b23f1d6b59beabde1072d14b7fb65781ce209c15/src/deprecatedCompat/4.0/nodeFactoryTopLevelExports.ts#L1-L5 😢  Gonna put something in the backlog to convert these over

alicewriteswrongs: tbh I just found an example in the Angular codebase using `ts.factory.createIdentifier` as the first argument to `ts.factory.createPropertyAssignment` so I went with that...but yeah, `createStringLiteral` might be the better option here. I'll swap it out and make sure it builds and whatnot

alicewriteswrongs: yeah that seems to work, let's go with `createStringLiteral`

alicewriteswrongs: I think an `Identifier` node does work in a lot of situations though and I don't see a way it would create problems, if you put the following snippet [into a TypeScript AST viewer](https://ts-ast-viewer.com/#code/DYUwLgBAZg9jEF4IG8BQEPTgLggIgCEBDAJz1QF8g):

```ts
let foo = {
  foo: "Bar"
}
```

then it parses the line `foo: "Bar"` into

```
PropertyAssignment
  Identifier
  StringLiteral
```

generated with these factory calls:

```ts
factory.createPropertyAssignment(
  factory.createIdentifier("foo"),
  factory.createStringLiteral("Bar")
)
```

However one reason to prefer a string in a `PropertyAssignment` node is if there's a possibility of other characters like `-` being present in the name which can't be present in an `Identifier` - I don't think we'd run into this with `@State` fields though since the name is the name of the field on the class and has to be an identifier itself.

Anywhoo - happy to stay with the string because it doesn't seem like that would cause any problem, but just a little brain dump from me

rwaskiewicz: Love it - thanks for that!

rwaskiewicz (approved):

tanner-reits (approved):

PR #3290, opened 2022-03-18
docs(declarations): typos in public compiler

sean-perkins:
## Pull request checklist

Please check if your PR fulfills the following requirements:
- [ ] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [ ] Build (`npm run build`) was run locally and any changes were pushed
- [ ] Unit tests (`npm test`) were run locally and passed
- [ ] E2E Tests (`npm run test.karma.prod`) were run locally and passed
- [x] Prettier (`npm run prettier`) was run locally and passed

## Pull request type

 

 

Please check the type of change your PR introduces:
- [ ] Bugfix
- [ ] Feature
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

## What is the current behavior?

GitHub Issue Number: N/A

## What is the new behavior?

Fixes several typo/documentation issues in the Stencil public compiler interfaces.

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

PR #3612, opened 2022-09-14
Fix/custom event type exports

tanner-reits:
## Pull request checklist

## Pull request type

 

 

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

Currently, types for custom events are not exported from the generated `components.d.ts` file

GitHub Issue Number: N/A

## What is the new behavior?

Types for custom events are now re-exported from the `components.d.ts` file. This allows those types to be easily accessed from the root of the type distribution (so long as the `index.ts` file is set to export all definitions from the `components.d.ts` file):
`import { MyCustomType } from '@my-lib/types';`

Or, when using `dist-custom-elements`, these types can now be accessed from the custom element output:
`import { MyCustomType } from '@my-custom-elements-output';`

- [ ] Yes
- [X] No

## Testing

Unit and e2e tests continue to pass. New tests added for altered function. Tested output in starter projects.


src/declarations/stencil-public-compiler.ts
```diff
@@ -917,7 +917,7 @@ export interface CompilerSystem {
   /**
    * Add a callback which will be ran when destroy() is called.
    */
-  addDestory(cb: () => void): void;
```
rwaskiewicz, line 920: FYI - although I think this is a good rename, it's technically a breaking change since this is a public facing API 😢 

The good news is that it's fixed in the `v3.0.0-dev` branch via https://github.com/ionic-team/stencil/pull/3290

tanner-reits:
Think this should be good for review. Not sure why the e2e tests are failing (don't know if it's just flakiness or if something is actually wrong, they run totally fine locally)

src/compiler/types/generate-app-types.ts
```diff
@@ -19,7 +19,7 @@ import { updateReferenceTypeImports } from './update-import-refs';
  * @returns `true` if the type declaration file written to disk has changed, `false` otherwise
  */
 export const generateAppTypes = async (
-  config: d.Config,
+  config: d.ValidatedConfig,
```
rwaskiewicz, line 22: Verified all the callers of this fn are calling it with a `ValidatedConfig` instance. 👍 👍

test/karma/test-app/components.d.ts
```diff
@@ -7,6 +7,8 @@
 import { HTMLStencilElement, JSXBase } from "@stencil/core/internal";
 import { SomeTypes } from "./util";
 import { TestEventDetail } from "./event-custom-type/cmp";
+export { SomeTypes } from "./util";
```
rwaskiewicz, line 10: Can you do me a favor and run `npm run build` from the `end-to-end` directory? I think some changes there haven't been generated

src/compiler/types/generate-app-types.ts
```diff
@@ -92,29 +92,31 @@ const generateComponentTypesFile = (config: d.Config, buildCtx: d.BuildCtx, areT
...
+  // Map event type metadata to partial expressions (omitting import/export keywords)
+  // e.g. { TestEvent } from '../path/to/event/test-event.interface';
+  const expressions = Object.keys(typeImportData).map((filePath) => {
+    const typeData = typeImportData[filePath];
+    let importFilePath: string;
+    if (isAbsolute(filePath)) {
+      importFilePath = normalizePath('./' + relative(config.srcDir, filePath)).replace(/\.(tsx|ts)$/, '');
+    } else {
+      importFilePath = filePath;
+    }
```
rwaskiewicz, line 104: Minor, we can probably simplify this a bit by doing the initial assignment to `importFilePath` in lieu of an `else` statement:
```suggestion
    let importFilePath = filePath;
    if (isAbsolute(importFilePath)) {
      importFilePath = normalizePath('./' + relative(config.srcDir, filePath)).replace(/\.(tsx|ts)$/, '');
    } 
```

src/compiler/types/generate-app-types.ts
```diff
@@ -92,29 +92,31 @@ const generateComponentTypesFile = (config: d.Config, buildCtx: d.BuildCtx, areT
...
+          return `${td.importName}`;
+        } else {
+          return `${td.localName} as ${td.importName}`;
+        }
+      })
+      .join(`, `)} } from "${importFilePath}";`;
+  });
+
+  // Write all import and export statements for event types
+  c.push(...expressions.map((ref) => `import ${ref}`), ...expressions.map((ref) => `export ${ref}`));
```
rwaskiewicz, line 119: With this change, we don't export only custom event types. We also end up exporting types that are declared on props as well. Take for instance a component spun up using `npm init stencil component` then modified as such:
```diff
+ export type Foo = string;
+ export type Bar = string;

@Component({
  tag: 'my-component',
  styleUrl: 'my-component.css',
  shadow: true,
})
export class MyComponent {
  /**
   * The first name
   */
-  @Prop() first: string;
+  @Prop() first: Foo;

  /**
   * The middle name
   */
-  @Prop() middle: string;
+  @Prop() middle: Bar;
```

running `npm run build` with this branch installed in the project results in the following (note the exports added):
```diff

diff --git a/src/components.d.ts b/src/components.d.ts
index cf9e773..f93c8fc 100644
--- a/src/components.d.ts
+++ b/src/components.d.ts
@@ -5,12 +5,14 @@
  * It contains typing information for all components that exist in this project.
  */
 import { HTMLStencilElement, JSXBase } from "@stencil/core/internal";
+import { Bar, Foo } from "./components/my-component/my-component";
+export { Bar, Foo } from "./components/my-component/my-component";
 export namespace Components {
     interface MyComponent {
         /**
           * The first name
          */
-        "first": string;
+        "first": Foo;
         /**
           * The last name
          */
@@ -18,7 +20,7 @@ export namespace Components {
         /**
           * The middle name
          */
-        "middle": string;
+        "middle": Bar;
     }
 }```

is this intentional?

File: src/compiler/types/tests/generate-app-types.spec.ts [hunk 6]
```diff
@@ -0,0 +1,153 @@
...
+      '/components.d.ts',
+      `/* eslint-disable */
+/* tslint:disable */
+/**
+ * This is an autogenerated file created by the Stencil compiler.
+ * It contains typing information for all components that exist in this project.
+ */
+import { HTMLStencilElement, JSXBase } from "@stencil/core/internal";
+import { UserImplementedEventType } from "./some/stubbed/path/resources";
+export { UserImplementedEventType } from "./some/stubbed/path/resources";
```
rwaskiewicz, line 111: Can we add a few additional test scenarios here?

1. Exporting multiple types from a single location to exercise the `join()` on multiple types
2. A type that is defined/used in multiple components (ideally with a different hierarchy). Consider the example below where `common-types` has a type that is used in both `level-1-component` and `level-2-component`. This should exercise path resolution of types we export
```
src
├── components
│   └── level-1
│       ├── level-1-component.tsx
│       └── level-2
│            └── level-2-component.tsx
└── utils
    └── common-types.ts
```

3. Naming collisions - If we have two components that each define and use a custom event type of the same name (but a different type), what gets exported? e.g.
```ts
// src/component-a.tsx

export type MyEvent = string;
```
```ts
// src/component-b.tsx

export type MyEvent = number;
```

rwaskiewicz (changes requested):

src/compiler/types/generate-app-types.ts
tanner-reits: I saw this while working on it as well. I was a little on the fence, but thinking about I don't think there is any risk to these types being exported. This could be helpful for users as well if their properties are expecting complex types they can use these types/interfaces to ensure their type safety without needing to duplicate a type/interface in their consuming app.

test/karma/test-app/components.d.ts
rwaskiewicz: https://github.com/ionic-team/stencil/pull/3625 created to catch these programmatically

src/compiler/types/tests/generate-app-types.spec.ts
tanner-reits: Managed to get the name collisions for local imports resolved. Added those test cases and a few more for prop types based on the comments in [this thread](https://github.com/ionic-team/stencil/pull/3612#discussion_r975424761)

src/compiler/sys/stencil-sys.ts
```diff
@@ -289,12 +289,12 @@ export const createSystem = (c?: { logger?: Logger }) => {
...
     };
 
-    remoreDirSyncRecursive(p, opts, results);
+    removeDirSyncRecursive(p, opts, results);
 
     return results;
   };
 
-  const remoreDirSyncRecursive = (
+  const removeDirSyncRecursive = (
```
alicewriteswrongs, line 297: `remore`, good stuff haha

src/compiler/transpile/run-program.ts
```diff
@@ -60,7 +60,34 @@ export const runTsProgram = async (
...
+  // name collisions prevention logic in the type generation code can
+  // correctly identify when to use an aliased type name.
+  //
+  // We can not add this as a part of the actual TS transpilation
+  // process because that process analyzes the AST to generate static
+  // Stencil event metadata (i.e. this is not something we control).
+  //
+  // These mutations happen on the original component object references
+  // to prevent breaking logic down the pipeline that relies on these
+  // references to the original objects staying in tact.
```
alicewriteswrongs, line 77:
```suggestion
  // references to the original objects staying intact.
```

src/compiler/transpile/run-program.ts
```diff
@@ -60,7 +60,34 @@ export const runTsProgram = async (
...
   buildCtx.components = getComponentsFromModules(buildCtx.moduleFiles);
+  // To fix an issue with custom event type name collisions
+  // when the types are defined in the same file as components,
+  // we need to "hack" adding the file path to the custom type reference so that our
+  // name collisions prevention logic in the type generation code can
+  // correctly identify when to use an aliased type name.
+  //
+  // We can not add this as a part of the actual TS transpilation
+  // process because that process analyzes the AST to generate static
+  // Stencil event metadata (i.e. this is not something we control).
```
alicewriteswrongs, line 73: I don't understand what this means exactly, the types are created through the compilation process? what is it that we don't control exactly?

I think I grok in general that we need add the filepath to the events included in a given component in order to ensure that there is something to differentiate events with a possible name collision when they are re-exported later on, but I'm not clear on this note about compilation here

alicewriteswrongs (changes requested):
just had a question about possibly clarifying a comment

src/compiler/transpile/run-program.ts
tanner-reits: Thanks for calling me out here because I was actually able to find a better location for this change to happen when trying to add clarity to this comment lol

Moved this change and updated comments for it [here](https://github.com/ionic-team/stencil/pull/3612/commits/476c3aca5f1a28a0de85c4e339e6676af04bb416)

rwaskiewicz (approved):

alicewriteswrongs (approved):
lgtm!

PR #3614, opened 2022-09-15
update handling of decorators to support emitting ES2018+

alicewriteswrongs:
There is a lot of detail here! Check the diff for an essay in a comment on the `handleClassFields` function which explains much of what's going on here.

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

GitHub Issue Number: N/A

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No


src/compiler/transformers/decorators-to-static/convert-decorators.ts
```diff
@@ -105,8 +120,9 @@ const removeStencilDecorators = (classMembers: ts.ClassElement[]) => {
           m.body
         );
       } else if (ts.isPropertyDeclaration(m)) {
-        return ts.updateProperty(m, newDecorators, m.modifiers, m.name, m.questionToken, m.type, m.initializer);
+        return m;
```
rwaskiewicz, line 123: IIUC, we're not removing Stencil-decorators on the class member anymore if it's a property declaration and deferring that to later - is that correct? If that's the case, it may be worth adding a comment stating why we are singling out property declarations to only return them

rwaskiewicz (commented):
I wasn't terribly thorough here, so my inline comments can be ignored for the time being.

The one bigger question I have is, is this the right path we should be taking? We had originally slated this a v3 feature with the intent that we may cause breaking changes to the runtime here.  However, I haven't spent as much time with this as you have, so I suppose my actual question is "Is there a path forward where we align with the ECMAScript standard instead of having to patch the ctor?"

alicewriteswrongs:
I think that we will be able to do a much better implementation of all the decorator-based functionality in Stencil once [the ES decorators proposal](https://github.com/tc39/proposal-decorators) is accepted. In particular, I think we could add the `accessor` keyword at compile-time and then we'd be able to supply a decorator function with the right name that's specific to whatever needs we have at runtime which would do the equivalent of what we set up with `Object.defineProperty` in `proxy-component.ts`:

https://github.com/ionic-team/stencil/blob/a3f9af26e8cf1756e686ec845e3d03e0b6318641/src/runtime/proxy-component.ts#L39-L68

That would overall be a much more elegant solution and something where we'd be outputting nice, standards-compliant ES classes and not hacking things up 😅

My guess is that proposal will get accepted fairly soon and then will need a bit of time to get full support everywhere. What would you think about merging something along these lines in the meantime, so that we can support `ESNext`, and filing a ticket to update this code once decorators are supported in browsers?

rwaskiewicz:
> In particular, I think we could add the accessor keyword at compile-time and then we'd be able to supply a decorator function with the right name that's specific to whatever needs we have at runtime which would do the equivalent of what we set up with Object.defineProperty in proxy-component.ts

LMK if I'm understanding this correctly - at Stencil's compile time (before we invoke the TypeScript compiler), we add `accessor` to the class field decorated with `@State` and still convert that field static, then metadata? Or did you have something else in mind?

> Assuming that the proposal is accepted more or less as it's written currently, that would overall be a much more elegant solution and something where we'd be outputting nice, standards-compliant ES classes and not hacking things up 😅

FWIW I'd consider it "stable" at Stage 3/with the TS team is looking at implementing it for 4.9 - it's both terrifying and exciting 😆 Although 4.9 is in beta, it doesn't appear the Decorators proposal has landed yet. The RC is slated for Nov 1 at the time of this writing, with the final build and official releases on Nov 11th and 15th, respectively. But either way, we unfortunately don't have an implementation to prototype against :-/

src/compiler/transformers/decorators-to-static/convert-decorators.ts
```diff
@@ -143,3 +159,198 @@ const filterDecorators = (
...
+ * **Note**: this function will modify a constructor if one is already present on
+ * the class or define a new one otherwise.
+ *
+ * @param classNode a TypeScript AST node for a Stencil component class
+ * @param classMembers the class members that we need to update
+ * @returns a list of updated class elements which can be inserted into the class
+ */
+function handleClassFields(classNode: ts.ClassDeclaration, classMembers: ts.ClassElement[]): ts.ClassElement[] {
+  const statements = [];
+  const updatedClassMembers = [];
```
rwaskiewicz, line 259: Minor, can we add type declarations to these two?
```suggestion  
  const statements: ts.ExpressionStatement[] = [];
  const updatedClassMembers: ts.ClassElement[] = [];
```

src/compiler/transformers/decorators-to-static/convert-decorators.ts
```diff
@@ -143,3 +159,198 @@ const filterDecorators = (
...
+    if (currentDecorators !== newDecorators && ts.isPropertyDeclaration(member)) {
+      // this is a class field that we'll need to handle, so lets push a statement for
+      // initializing the value onto our statements list
+      statements.push(
+        ts.factory.createExpressionStatement(
+          ts.factory.createBinaryExpression(
+            ts.factory.createPropertyAccessExpression(
+              ts.factory.createThis(),
+              // @ts-ignore
+              ts.factory.createIdentifier(member.name.escapedText)
```
rwaskiewicz, line 273: Can we determine if there's another property that doesn't require us to use `@ts-ignore`? I've been bit by that one being undefined before (or maybe it was `text`?), IDK if that's what we want to move forward with or not

src/compiler/transformers/decorators-to-static/convert-decorators.ts
```diff
@@ -143,3 +159,198 @@ const filterDecorators = (
...
+
+  // create or update a constructor which contains the initializing statements
+  // we created above
+  return updateConstructor(classNode, updatedClassMembers, statements);
+}
+
+/**
+ * Helper util for updating the constructor on a class declaration AST node.
+ *
+ * @param classNode the class node whose constructor we're going to be updated
```
rwaskiewicz, line 295: this reads a little funky
```suggestion
 * @param classNode the class node whose constructor that's going to be updated
```

src/compiler/transformers/decorators-to-static/convert-decorators.ts
```diff
@@ -143,3 +159,198 @@ const filterDecorators = (
...
+  const constructorMethod = classMembers[constructorIndex];
+
+  if (constructorIndex >= 0 && ts.isConstructorDeclaration(constructorMethod)) {
+    const hasSuper = constructorMethod.body.statements.some((s) => s.kind === ts.SyntaxKind.SuperKeyword);
+
+    if (!hasSuper && needsSuper(classNode)) {
+      statements = [createConstructorBodyWithSuper(), ...statements];
+    }
+
+    classMembers[constructorIndex] = ts.updateConstructor(
```
rwaskiewicz, line 316: Can we use the `factory` method here? I _think_ it's 
```suggestion
    classMembers[constructorIndex] = ts.factory.updateConstructorDeclaration(
```

src/compiler/transformers/decorators-to-static/convert-decorators.ts
```diff
@@ -143,3 +159,198 @@ const filterDecorators = (
...
+      constructorMethod,
+      constructorMethod.decorators,
+      constructorMethod.modifiers,
+      constructorMethod.parameters,
+      ts.updateBlock(constructorMethod.body, statements)
```
rwaskiewicz, line 321: Can we use the `factory` method here? I _think_ it's 
```suggestion
      ts.factory.updateBlock(constructorMethod.body, statements)

```

src/compiler/transformers/decorators-to-static/convert-decorators.ts
```diff
@@ -143,3 +159,198 @@ const filterDecorators = (
...
+ */
+const needsSuper = (classDeclaration: ts.ClassDeclaration): boolean =>
+  classDeclaration.heritageClauses && classDeclaration.heritageClauses.length > 0;
+
+/**
+ * Create a statement with a call to `super()` suitable for including in the body of a constructor.
+ * @returns a {@link ts.ExpressionStatement} node equivalent to `super()`
+ */
+const createConstructorBodyWithSuper = (): ts.ExpressionStatement => {
+  return ts.createExpressionStatement(ts.createCall(ts.createIdentifier('super'), undefined, undefined));
```
rwaskiewicz, line 355: Should we handle cases like https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-6.html#allowing-code-in-constructors-before-super?

rwaskiewicz (changes requested):

src/compiler/transformers/decorators-to-static/convert-decorators.ts
```diff
@@ -73,6 +73,8 @@ const visitClassDeclaration = (
     listenDecoratorsToStatic(diagnostics, decoratedMembers, newMembers);
   }
 
+  const newerMembers = handleClassFields(classNode, newMembers);
```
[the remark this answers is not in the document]
alicewriteswrongs: makes sense! `newerMembers` is also not the best name, I'll come up with something better

src/compiler/transformers/decorators-to-static/convert-decorators.ts
alicewriteswrongs: lol indeed it does 😅

src/compiler/transformers/decorators-to-static/convert-decorators.ts
alicewriteswrongs: Came up with this, it's a bit wordy but no type assertions or `@ts-ignore` necessary:

```ts
      const declarationName: ts.DeclarationName = ts.getNameOfDeclaration(member);

      const memberName =
        declarationName.kind === ts.SyntaxKind.ComputedPropertyName
          ? declarationName.expression.getText()
          : declarationName.getText();
```

alicewriteswrongs: `memberName` is then a string that we can just pass in to `ts.factory.createIdentifier`

src/compiler/transformers/decorators-to-static/convert-decorators.ts
alicewriteswrongs: what would you suggest we do? right now this will basically add a `super()` call if one isn't present in the existing constructor or if we're creating a new one if the class inherits from another class. Right now it always gets added as the first statement in the constructor

rwaskiewicz: Hmmm - are you saying it's always added there today? If so, maybe I jumped the gun and we should punt on this for now.  If not, I wonder if we can find the index of the `super()` call in an existing ctor (if it exists) and use that? And if not we just put it at the start?

alicewriteswrongs:
@rwaskiewicz just an quick update on this PR. I made two improvements:

- only write a constructor on a class if we actually have some statements to put into the constructor (silly, I know)
- the initial changeset wasn't handling `@Event` fields correctly, and was rewriting _them_ to be initialized in the class' constructor rather than remaining class fields. to fix this I narrowed the range of which decorators we rewrite to just be `@State` and `@Prop`, which should fix that issue

I think things should now be ready for another look!

src/compiler/transformers/test/convert-decorators.spec.ts
```diff
@@ -0,0 +1,172 @@
+import { transpileModule } from './transpile';
```
rwaskiewicz, line 1: Can we add a few tests around `super()` please?

rwaskiewicz (changes requested):

src/compiler/transformers/decorators-to-static/convert-decorators.ts
```diff
@@ -82,35 +91,75 @@ const visitClassDeclaration = (
...
+          // update the property to remove decorators
+          return ts.updateProperty(
+            member,
+            newDecorators,
+            member.modifiers,
+            member.name,
+            member.questionToken,
+            member.type,
+            member.initializer
+          );
```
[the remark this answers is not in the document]
alicewriteswrongs: yes that's right - basically we don't want to do this edit (from class field to constructor-initialized field, idk what to call that exactly) to _all_ fields, but just to ones decorated with `@State` and `@Prop`. So this is the behavior that was previously there for all fields.

src/compiler/transformers/decorators-to-static/convert-decorators.ts
```diff
@@ -82,35 +91,75 @@ const visitClassDeclaration = (
...
 
-const removeStencilDecorators = (classMembers: ts.ClassElement[]) => {
-  return classMembers.map((m) => {
-    const currentDecorators = m.decorators;
-    const newDecorators = filterDecorators(m, MEMBER_DECORATORS_TO_REMOVE);
+/**
+ * Take a list of `ClassElement` AST nodes and remove any decorators from
+ * method elements which are Stencil-specific decorators. We implement the
+ * intended behavior for these Stencil-specific decorators (things like
+ * `@Watch`, `@PropDidChange`, etc) through a combination of compile- and
```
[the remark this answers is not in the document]
alicewriteswrongs: hum, should we remove them while we're in here?

src/compiler/transformers/decorators-to-static/convert-decorators.ts
```diff
@@ -143,3 +192,234 @@ const filterDecorators = (
...
+ */
+function handleClassFields(classNode: ts.ClassDeclaration, classMembers: ts.ClassElement[]): ts.ClassElement[] {
+  const statements: ts.ExpressionStatement[] = [];
+  const updatedClassMembers: ts.ClassElement[] = [];
+
+  for (const member of classMembers) {
+    if (shouldInitializeInConstructor(member) && ts.isPropertyDeclaration(member)) {
+      const declarationName: ts.DeclarationName = ts.getNameOfDeclaration(member);
+
+      const memberName =
```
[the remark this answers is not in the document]
alicewriteswrongs: Yeah I'll add a comment explaining - I'm not sure this ever _could_ be a `ComputedPropertyName` in practice, but the nodes are typed that way so I put the checks to keep typescript happy

src/compiler/transformers/decorators-to-static/convert-decorators.ts
```diff
@@ -143,3 +192,234 @@ const filterDecorators = (
...
+      // this is a class field that we'll need to handle, so lets push a statement for
+      // initializing the value onto our statements list
+      statements.push(
+        ts.factory.createExpressionStatement(
+          ts.factory.createBinaryExpression(
+            ts.factory.createPropertyAccessExpression(ts.factory.createThis(), ts.factory.createIdentifier(memberName)),
+            ts.factory.createToken(ts.SyntaxKind.EqualsToken),
+            // if the member has no initializer we should default to setting it to
+            // just 'undefined'
+            member.initializer ?? ts.factory.createIdentifier('undefined')
```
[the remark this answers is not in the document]
alicewriteswrongs: hum I don't think so, according to my editor `member.initializer` will be `ts.Expression` (i.e. an AST node representing an expression) or `undefined`, so I'm basically ensure that if it _is_ undefined we hoist that value into AST-land instead of JS-runtime-value-land, if that makes sense? 🤔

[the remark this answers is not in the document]
alicewriteswrongs: another way to say that is that yeah, I think if the expression evaluates to `null` that won't come into play until later on in the backend of the TS compiler when it goes to emit JS and has to turn such AST nodes into concrete values in the emitted code

src/compiler/transformers/test/convert-decorators.spec.ts
alicewriteswrongs: oh yeah that's a good idea, will do

src/compiler/transformers/decorators-to-static/convert-decorators.ts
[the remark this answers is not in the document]
alicewriteswrongs: the TypeScript AST viewer [does allow me to use a computed name for a declaration](https://ts-ast-viewer.com/#code/MYewdgzgLgBARgQwE4wLwwESKRgUL4AGwQghgDEQQYBvXGBmAbWwF01NEAvDAblwC+QA), like so:

```ts
const bar = "bar"

class Foo {
    [bar] = "baz";
}
```

it parses it to

Screen Shot 2022-10-04 at 4 03 46 PM

I didn't realize you could do this! But it does seem like that is valid ECMAScript.

MDN has a brief note about it in this section: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Public_class_fields#public_instance_fields

I read some of the ECMAScript standard and it seems to be in accordance with that: https://tc39.es/ecma262/multipage/ecmascript-language-expressions.html#prod-PropertyName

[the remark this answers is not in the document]
alicewriteswrongs: I'll add a little note with some examples to clarify

alicewriteswrongs:
@rwaskiewicz made a few more updates, should be ready for another look

src/compiler/transformers/decorators-to-static/convert-decorators.ts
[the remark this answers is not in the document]
rwaskiewicz: I haven't looked at your changes you've made since this comment. If you haven't already, it's probably best we remove them in an explicit commit/PR (e.g. "Remove Deprecated Decorator Names")

src/compiler/transformers/decorators-to-static/convert-decorators.ts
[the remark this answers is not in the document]
rwaskiewicz: > the TypeScript AST viewer [does allow me to use a computed name for a declaration](https://ts-ast-viewer.com/#code/MYewdgzgLgBARgQwE4wLwwESKRgUL4AGwQghgDEQQYBvXGBmAbWwF01NEAvDAblwC+QA), like so:

thanks JavaScript 😏

rwaskiewicz (approved):

alicewriteswrongs:
@rwaskiewicz just pushed a comment-only commit to add TODOs for STENCIL-591

rwaskiewicz (approved):

tanner-reits (approved):
Read through the discussions and code and lgtm! Nice work!

PR #3618, opened 2022-09-16
chore(compiler): convert a Set to a readonly array

alicewriteswrongs:
This converts a Set of decorators that we remove during compilation from `Set<string>` to `ReadonlyArray<string>` because by doing so we get much better type hint popups.

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

GitHub Issue Number: N/A

## What is the new behavior?

We have two `Set` objects which basically should really be constants, since they are never modified at runtime. This just changes them to be `ReadonlyArray<string>` in order to push that information into the type system (and get some better type hints when you run across these things in a file).

- [ ] Yes
- [x] No

### DIFF
--- a/src/compiler/transformers/transform-utils.ts
+++ b/src/compiler/transformers/transform-utils.ts
@@ -1,4 +1,4 @@
-import { augmentDiagnosticWithNode, buildError, normalizePath } from '@utils';
+import { augmentDiagnosticWithNode, buildError, normalizePath, readOnlyArrayHasStringMember } from '@utils';
 import ts from 'typescript';
 
 import type * as d from '../../declarations';
@@ -415,7 +415,7 @@ export const validateReferences = (
 ) => {
   Object.keys(references).forEach((refName) => {
     const ref = references[refName];
-    if (ref.path === '@stencil/core' && MEMBER_DECORATORS_TO_REMOVE.includes(refName as any)) {
+    if (ref.path === '@stencil/core' && readOnlyArrayHasStringMember(MEMBER_DECORATORS_TO_REMOVE, refName)) {
       const err = buildError(diagnostics);
       augmentDiagnosticWithNode(err, node);
     }
--- a/src/compiler/transformers/transform-utils.ts
+++ b/src/compiler/transformers/transform-utils.ts
@@ -415,7 +415,7 @@ export const validateReferences = (
 ) => {
   Object.keys(references).forEach((refName) => {
     const ref = references[refName];
-    if (ref.path === '@stencil/core' && MEMBER_DECORATORS_TO_REMOVE.has(refName)) {
+    if (ref.path === '@stencil/core' && MEMBER_DECORATORS_TO_REMOVE.includes(refName as any)) {
       const err = buildError(diagnostics);
       augmentDiagnosticWithNode(err, node);
     }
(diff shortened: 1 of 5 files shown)


src/compiler/transformers/transform-utils.ts
```diff
@@ -415,7 +415,7 @@ export const validateReferences = (
 ) => {
   Object.keys(references).forEach((refName) => {
     const ref = references[refName];
-    if (ref.path === '@stencil/core' && MEMBER_DECORATORS_TO_REMOVE.has(refName)) {
+    if (ref.path === '@stencil/core' && MEMBER_DECORATORS_TO_REMOVE.includes(refName as any)) {
```
tanner-reits, line 418: Does `refName` need to be cast as `any` here?

tanner-reits (changes requested):

src/compiler/transformers/transform-utils.ts
alicewriteswrongs: yeah unfortunately, basically if you have a `ReadonlyArray` with strings like `"foo" | "bar"` as it's members then all the array methods which are generic across the type of the elements of the array (like `Array.prototype.map`, `Array.prototype.filter`, etc) inherit that type. so normally you want like `Array<T>.map` to have a type like `<F>((el: T) => F) => F[]`, but in this case it basically means that if you have a `ReadonlyArray` like

```ts
const readOnly = [ "foo", "bar" ] as const;
```

then the type of methods like `includes` will look like

```ts
(el: "foo" | "bar") => boolean
```

rather than what we'd sort of want here, which is

```ts
(el: string) => boolean
```

or maybe

```ts
(el: string) => el is "foo" | "bar"
```

so then basically you get a type error if you pass something of type `string` to the `.includes` on such an array, because `"foo" | "bar"` is a narrower type than `string`.

I put a little note in one other spot about this:

https://github.com/ionic-team/stencil/blob/aed400b1815c7328afd325bdfb5e5f63d1e07e04/src/cli/parse-flags.ts#L331-L337

I could add the same here, or maybe create a little helper to use for these things, like 

```ts
function readOnlyArrayHasStringMember<T, F extends T>(xs: ReadonlyArray<T>, member: F | string): boolean {
  return xs.includes(member as any)
}
```

tanner-reits: Hmm okay, I thought `includes` kept a more generic signature. Feels like it adopting a different signature defeats the purpose of the method 🤷‍♂️

tanner-reits (approved):

src/compiler/transformers/transform-utils.ts
rwaskiewicz: I think the utility function here that @alicewriteswrongs suggested would be a good idea, 'cause I think I asked the same question @tanner-reits did here on `parse-flags.ts`.

We could probably narrow that type assertion a smidge by asserting that `member` is `typeof xs[number]`, although TBH I'm not sure how much _actually_ helps in practice
```
function readOnlyArrayHasStringMember<T, F extends T>(xs: ReadonlyArray<T>, member: F | string): boolean {
  return xs.includes(member as typeof xs[number]);
}
```

rwaskiewicz (commented):
LGTM - thoughts on adding the suggested helper fn?

src/compiler/transformers/transform-utils.ts
alicewriteswrongs: just pushed a change to add the little helper, lmk what y'all think

rwaskiewicz:
I dig it 👍 Thanks!

rwaskiewicz (approved):

alicewriteswrongs:
imports! \*shakes fist\* should be good now

tanner-reits (approved):
LGTM! I like the helper function! 💯

PR #3619, opened 2022-09-19
feat(compiler): add `defineCustomElements` method & signature typedef

tanner-reits:
## Pull request checklist

- [x] Docs have been reviewed and added / updated if needed (for bug fixes / features) - https://github.com/ionic-team/stencil-site/pull/921

## Pull request type

 

 

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

`dist-custom-elements` currently does not have a `defineCustomElements` function to quickly register all components on the custom element registry.

GitHub Issue Number: [#3136](https://github.com/ionic-team/stencil/issues/3136#issuecomment-1130718870)

## What is the new behavior?

`bundle` is added as a `customElementsExportBehavior` option. This behavior will mirror that of `dist-custom-elements-bundle` in exporting a `defineCustomElements` function. This is intended to be used as a quick migration for users currently building with `dist-custom-elements-bundle`.

- [ ] Yes
- [ ] No

## Testing

Steps taken for testing:
1. Create a Stencil component starter
2. Build and pack/link this branch and use within the starter created above
3. In the starter, make sure `dist-custom-elements` is an output target and set `customElementsExportBehavior: 'bundle'`
4. Also set the appropriate targets in the stater's `package.json` according to the [Stencil documentation](https://stenciljs.com/docs/custom-elements#distributing-custom-elements) for the custom elements target.
5. Build and pack the starter app output
6. Add the packaged output as a dep to a starter framework project (I used Angular)
7. Import the `defineCustomElements()` function and call it on app bootstrap
8. Verify that any components in the component lib can be used in HTML

Unit tests have been added to the output target spec file as well as the type generation spec file.


tanner-reits:
I'll create a PR for the stencil-site docs once this is good to go

src/compiler/output-targets/dist-custom-elements/custom-elements-types.ts
```diff
@@ -97,6 +98,22 @@ const generateCustomElementsTypesOutput = async (
...
     `export declare const setPlatformOptions: (opts: SetPlatformOptions) => void;`,
+    ...(isBundleExport
+      ? [
+          ``,
+          `/**`,
+          ` * Utility to define all custom elements within this package using the tag name provided in the component's source.`,
+          ` * When defining each custom element, it will also check it's safe to define by:`,
+          ` *`,
+          ` * 1. Ensuring the "customElements" registry is available in the global context (window).`,
+          ` * 2. The component tag name is not already defined.`,
```
alicewriteswrongs, line 109: small wording suggestion:

```suggestion
          ` * When defining each custom element, it will also check it's safe to define by:`,
          ` *`,
          ` * 1. Ensuring the "customElements" registry is available in the global context (window).`,
          ` * 2. Ensuring that the component tag name is not already defined.`,
```

src/compiler/output-targets/dist-custom-elements/index.ts
```diff
@@ -191,6 +191,8 @@ export const addCustomElementInputs = (
   const components = buildCtx.components;
   // an array to store the imports of these modules that we're going to add to our entry chunk
   const indexImports: string[] = [];
+  const indexExports: string[] = [];
```
alicewriteswrongs, line 194: might be helpful to have a short comment above this line (and above the line below) explaining what these arrays are for?

src/compiler/output-targets/dist-custom-elements/index.ts
```diff
@@ -216,18 +218,40 @@ export const addCustomElementInputs = (
...
+    bundleOpts.loader!['\0core'] += indexImports.join('\n');
+  }
+
   // Only re-export component definitions if the barrel export behavior is set
   if (outputTarget.customElementsExportBehavior === 'single-export-module') {
-    bundleOpts.loader!['\0core'] += indexImports.join('\n');
+    bundleOpts.loader!['\0core'] += indexExports.join('\n');
+  }
+
+  if (outputTarget.customElementsExportBehavior === 'bundle') {
```
alicewriteswrongs, line 242: is there any reason why we wouldn't want to merge this into the other block checking the value of `outputTarget` on line 233? just thinking it would be good to keep the behavior specific to each `customElementsExportBehavior` bundled together as much as we can

alicewriteswrongs: might be good also to hang just a brief explanatory comment above line 233, along the olines of line 237

alicewriteswrongs (changes requested):
a few comments and questions

src/compiler/output-targets/dist-custom-elements/index.ts
tanner-reits: @alicewriteswrongs We probably can. I have them separate for now to try and keep all the import/export declarations at the top of the file and everything else in the "body". I'll see if there is a way to restructure how the file gets written to make this easier.

tanner-reits: @rwaskiewicz @alicewriteswrongs I pushed a commit that moves some of this stuff around a bit and changes how the entry-point gets generated. Changes the order of the output slightly, but, functionally, all the same. It just makes it so the output file has separate imports, exports, and body segments so that can stay more consistent as the output target behavior inevitably changes in the future!

rwaskiewicz:
Doing a final review/test now. @tanner-reits can you get started on the docs for this?

rwaskiewicz:
@tanner-reits do you have a project template you were using to test this out that @alicewriteswrongs & I could leverage?

tanner-reits:
> @tanner-reits do you have a project template you were using to test this out that @alicewriteswrongs & I could leverage?

I was testing against [this](https://github.com/jcfranco/stencil-use-case-where-build-is-not-tree-shaken) repo (it was from a previous issue, but works well for these types of output target changes). If you alter `src/main.js` to use the `defineCustomElements` function, then, after building, the `public/main.js` file will pull in all the components

rwaskiewicz:
@tanner-reits can you provide a step-by-step here? Ideally, we should be able to reproduce your testing steps from the "Testing" section of the PR summary. Right now it's not clear how we verified the output in the browser (and rather than come up with my own here, I'd like to make sure we're all on the same page). That way if there's a bug, we can come back to this commit and re-run the manual testing procedure if need be

src/compiler/output-targets/dist-custom-elements/index.ts
```diff
@@ -233,49 +231,76 @@ export const addCustomElementInputs = (
...
 
-  imp.push(
+  // Exports that are always present
+  exports.push(
     `export { setAssetPath, setPlatformOptions } from '${STENCIL_INTERNAL_CLIENT_ID}';`,
     `export * from '${USER_INDEX_ENTRY_ID}';`
   );
 
+  // Content related to global scripts
   if (outputTarget.includeGlobalScripts !== false) {
```
alicewriteswrongs, line 263: is it our intention that the user could set `outputTarget.includeGlobalScripts` to `null` and have this condition evaluate to true? Seems off to me...

(just because `null !== false` evaluates to `true`)

we don't have to change that in this PR but just calling it out because I noticed it

src/compiler/output-targets/dist-custom-elements/index.ts
```diff
@@ -233,49 +231,76 @@ export const addCustomElementInputs = (
...
+  let content = '';
+
+  // Add imports to file content
+  content += imports.length ? imports.join('\n') + '\n' : '';
+  // Add exports to file content
+  content += exports.length ? exports.join('\n') + '\n' : '';
+  // Add body to file content
+  content += body.length ? body.join('\n') + '\n' : '';
+
+  return content;
```
alicewriteswrongs, line 303: I like this setup a lot more 👍 things are cleanly factored and co-located

src/compiler/output-targets/dist-custom-elements/index.ts
```diff
@@ -233,49 +231,76 @@ export const addCustomElementInputs = (
...
+  // defined above. This keeps the file structure consistent as
+  // new export behaviors may be added
+  let content = '';
+
+  // Add imports to file content
+  content += imports.length ? imports.join('\n') + '\n' : '';
+  // Add exports to file content
+  content += exports.length ? exports.join('\n') + '\n' : '';
+  // Add body to file content
+  content += body.length ? body.join('\n') + '\n' : '';
```
alicewriteswrongs, line 301: wondering if we should put an extra newline before the body content? just so we get something like

```ts
import { globalScripts } from '${STENCIL_APP_GLOBALS_ID}';
export { setAssetPath, setPlatformOptions } from '${STENCIL_INTERNAL_CLIENT_ID}';
export * from '${USER_INDEX_ENTRY_ID}';

globalScripts();
```

just for a little readability improvement

alicewriteswrongs (changes requested):
few small comments questions, which I think could all probably be ignored if desired 😄

src/compiler/output-targets/dist-custom-elements/index.ts
tanner-reits: That does seem off. I'm hesitant to change it just because that was the logic already in place, but I can make note of that and take a look outside the context of this PR (this is going into v3 anyway)

tanner-reits:
> @tanner-reits can you provide a step-by-step here? Ideally, we should be able to reproduce your testing steps from the "Testing" section of the PR summary. Right now it's not clear how we verified the output in the browser (and rather than come up with my own here, I'd like to make sure we're all on the same page). That way if there's a bug, we can come back to this commit and re-run the manual testing procedure if need be

@rwaskiewicz I updated the PR summary with testing steps.

rwaskiewicz:
@tanner-reits I gave this a run through with the steps provided, and couldn't get this to work quite right. I'm sure it's probably I'm doing on my end - to try to chase down the error, I created https://github.com/rwaskiewicz/stencil-customelementsbehavior-bundle-test/commits/main, which has commits that roughly translate to the steps you've outlined above:

> Steps taken for testing:
> 1. Create a Stencil component starter
- [run: npm init stencil component my-lib](https://github.com/rwaskiewicz/stencil-customelementsbehavior-bundle-test/commit/6d1e236f741ea05ff70ac12974e1d8bc28abb47d)
- [generate package-lock with "npm i"](https://github.com/rwaskiewicz/stencil-customelementsbehavior-bundle-test/commit/8e80992b02b04f8a681ba69760df77c7f046a15b)
> 2. Build and pack/link this branch and use within the starter created above
- [copy over branch's tarball, install it](https://github.com/rwaskiewicz/stencil-customelementsbehavior-bundle-test/commit/05535ebce42f8edbfb168b7ebe159b80d083a179)
> 3. In the starter, make sure dist-custom-elements is an output target and set customElementsExportBehavior: 'bundle'
- [add customElementsExportBehavior: bundle to stencil.config.ts](https://github.com/rwaskiewicz/stencil-customelementsbehavior-bundle-test/commit/fae0a026dc04d99cbcb25696790a2887eae9ab19)
> 4. Build and pack the starter app output
- [commit tarball for my-lib, after running "npm run build && npm pack"](https://github.com/rwaskiewicz/stencil-customelementsbehavior-bundle-test/commit/026adfb99c7bc69ce55b88979693c031472f8e6f)
> 5. Add the packaged output as a dep to a starter framework project (I used Angular)
- [add screenshot of responses to ionic start cmd](https://github.com/rwaskiewicz/stencil-customelementsbehavior-bundle-test/commit/03e4bdd746b6c023afe95580bfdb44f8b622c916)
- [add output of "ionic start" command](https://github.com/rwaskiewicz/stencil-customelementsbehavior-bundle-test/commit/778426bc95cd9de6e996633634f93930199e37ff)
- [install my-lib tarball in my-app](https://github.com/rwaskiewicz/stencil-customelementsbehavior-bundle-test/commit/356ac7b608b6c9cf8335a4fe83d85fe705360b13)
> 6. Import the defineCustomElements() function and call it on app bootstrap
- [attempt to use my-component in html](https://github.com/rwaskiewicz/stencil-customelementsbehavior-bundle-test/commit/54f0362dd724f9127fee8eb2776f7da56ac11483)
> 7. Verify that any components in the component lib can be used in HTML
- i ran `npm start` here, and couldn't see anything here - I suspect the issue is with the import I'm doing on bootstrap, LMK what you think

tanner-reits:
@rwaskiewicz Took a look at your test repo and looks like the issue is where you're importing `defineCustomElements` from. The current import statement in `my-app/src/main.ts` doesn't actually point to output that has the function based on what your Stencil lib's `package.json` has defined for paths. I likely should've included the steps mentioned on the Stencil site [here](https://stenciljs.com/docs/custom-elements#distributing-custom-elements) in my test reproduction. 

If you change your import statement to be `import { defineCustomElements } from 'my-lib/dist/components';`, all should be well!

I can update this PR description with those additional steps.

rwaskiewicz:
Ah I knew it was something silly 🤦  Thanks!

rwaskiewicz (approved):

alicewriteswrongs (approved):
lgtm!

alicewriteswrongs:
looks like there are a few failing unit tests though

PR #3628, opened 2022-09-21
chore(ci): fix build out-of-memory errors

rwaskiewicz:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes

we're seeing sporadic build errors in ci, which makes us sad

GitHub Issue Number: N/A

## What is the new behavior?

this commit introduces an `.npmrc` file to the project that specifies a new value for `--max-old-space-size`. by default, node allocates 2GB of memory to a process. it is possible for stencil builds to near this limit, causing an out of memory failure. we have seen this more often in continuous integration (github actions) than anywhere else.

an `.npmrc` file was chosen here as an operating system, shell, and build-agnostic solution. that is, this solution is intended to work:
- on linux, macos, and windows (although we only use linux for building, and the team develops solely on macos)
- on any shell (no hacks to the run command either in `package.json` or in the ci job for building the project)
- we use the same value in ci and locally (machine configuraton parity)

the value of '4096' was chosen for this configuration field by doubling the existing default of '2048'

- [ ] Yes
- [x] No

## Testing

I've run the build step 18 times (9 regular builds, 9 for browserstack), which is probably enough to instill some confidence. Worst case scenario, we revert it

I then pushed [09c5955](https://github.com/ionic-team/stencil/pull/3628/commits/09c5955ece3c7eb4e62f0480e10e7e92fd81cc02), which drops the memory ceiling to 1GB to verify that the setting was being picked up by CI (I did the same thing locally first) ([failing log example 1](https://github.com/ionic-team/stencil/actions/runs/3100550170/jobs/5020955511), [failing log example 2](https://github.com/ionic-team/stencil/actions/runs/3100550091/jobs/5020955137)), then reverted it. That revert gives us 2 extra builds, so that's a total of 20 times we'd build with this working.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,6 @@
+# By default, Node allocates 2 GB for a process to run.
+# When building Stencil, it may reach that point before the garbage collector is invoked, causing an out-of-memory
+# related failure.
+# If this value is changed, please ensure that it works both locally and in a continuous integration environment in
+# a repeatable manner (i.e. it can run many times, one after the other, without failing due to out-of-memory errors).
+node_options=--max-old-space-size=4096
--- a/.npmrc
+++ b/.npmrc
@@ -3,4 +3,4 @@
 # related failure.
 # If this value is changed, please ensure that it works both locally and in a continuous integration environment in
 # a repeatable manner (i.e. it can run many times, one after the other, without failing due to out-of-memory errors).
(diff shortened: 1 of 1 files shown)


tanner-reits (approved):
LGTM. I've had to do this on other projects in the past

PR #3682, opened 2022-10-05
chore(test): upgrade karma infrastructure

rwaskiewicz:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes

Our Karma infrastructure is a little out of date.
I wonder if upgrading could potentially help us with the flakiness we see in Browserstack.
It's a long shot, but a cheap thing to test out (and if it work, 🎉)
If it doesn't work, well, at least we've updated some deps.
 
GitHub Issue Number: N/A

## What is the new behavior?

this commit upgrades our karma infrastructure, bumping the versions of
all `karma-*` packages. this is being done not necessarily to keep the
packages up to date (although it is a benefit), but to test out if new packages
could have the positive side effect of helping with the stability of tests in our
continuous integration environment

this commit removes a test hack from 'delegates-focus' where a test
would return early if the userAgent wasn't 'Chrome'. this test has been
silently passing in Firefox before now, but with the new infrastructure
updates, it now fails.

- [ ] Yes
- [x] No

## Testing

Karma tests were run locally prior to pushing to GitHub
CI runs as expected


test/karma/test-app/delegates-focus/karma.spec.ts
```diff
@@ -1,12 +1,6 @@
 import { setupDomTests, waitForChanges } from '../util';
 
 describe('delegates-focus', function () {
-  if (navigator.userAgent.indexOf('Chrome') === -1 || navigator.userAgent.indexOf('Edge') > -1) {
```
rwaskiewicz, line 4: this is the only instance of `navigator.userAgent.indexOf` I see in the test code. For firefox (which I re-enabled sometime in the last year), this would always early return (because that left hand side of the conditional would always be true). In the new infra, Jasmine says "Hey, I found a describe with no children, that's a problem!" when this code is present/running on Firefox, hence its removal

`delegatesFocus` _is_ supported by Edge these days https://caniuse.com/?search=delegatesfocus

alicewriteswrongs (approved):
let's try it! 🤞

tanner-reits (approved):
What's the worst that can happen?

PR #3684, opened 2022-10-05
fix(compiler): fix determination of when to `super()` in convert-deco…

alicewriteswrongs:
…rators.ts

This fixes an issue where we were inadvertently adding a `super()` call to 'synthetic' constructors created in the `updateConstructor` function.

In particular, this could happen when a class declaration had an `implements` keyword on it, like

```ts
class Foo implements Bar {
    foo = "bar";
}
```

In that case we _don't_ want to add a `super()` call to any constructor we create, because the class has no superclass. So we need to tell that situation apart from

```ts
class Foo extends Bar {
    foo = "bar";
}
```

in which we _do_ want to add a `super()` call.

## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

GitHub Issue Number: N/A

## What is the new behavior?

-
-
-

- [ ] Yes
- [x] No

### DIFF
--- a/src/compiler/transformers/decorators-to-static/convert-decorators.ts
+++ b/src/compiler/transformers/decorators-to-static/convert-decorators.ts
@@ -414,8 +414,20 @@ export const updateConstructor = (
  * @param classDeclaration a class declaration AST node
  * @returns whether this class has parents or not
  */
-const needsSuper = (classDeclaration: ts.ClassDeclaration): boolean =>
-  classDeclaration.heritageClauses && classDeclaration.heritageClauses.length > 0;
+const needsSuper = (classDeclaration: ts.ClassDeclaration): boolean => {
+  const hasHeritageClauses = classDeclaration.heritageClauses && classDeclaration.heritageClauses.length > 0;
+
+  if (hasHeritageClauses) {
+    // A {@link ts.SyntaxKind.HeritageClause} node may be for extending a
+    // superclass _or_ for implementing an interface. We only want to add a
+    // `super()` call to our synthetic constructor here in the case that there
+    // is a superclass, so we can check for that situation by checking for the
+    // presence of a heritage clause with the `.token` property set to
+    // `ts.SyntaxKind.ExtendsKeyword`.
+    return classDeclaration.heritageClauses.some((clause) => clause.token === ts.SyntaxKind.ExtendsKeyword);
+  }
+  return false;
+};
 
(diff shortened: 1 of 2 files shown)


src/compiler/transformers/decorators-to-static/convert-decorators.ts
```diff
@@ -414,8 +414,20 @@ export const updateConstructor = (
  * @param classDeclaration a class declaration AST node
  * @returns whether this class has parents or not
  */
-const needsSuper = (classDeclaration: ts.ClassDeclaration): boolean =>
-  classDeclaration.heritageClauses && classDeclaration.heritageClauses.length > 0;
+const needsSuper = (classDeclaration: ts.ClassDeclaration): boolean => {
+  let hasHeritageClauses = classDeclaration.heritageClauses && classDeclaration.heritageClauses.length > 0;
```
rwaskiewicz, line 418:
```suggestion
  const hasHeritageClauses = classDeclaration.heritageClauses?.length > 0;
```

src/compiler/transformers/decorators-to-static/convert-decorators.ts
```diff
@@ -414,8 +414,20 @@ export const updateConstructor = (
...
+
+  if (hasHeritageClauses) {
+    // A {@link ts.SyntaxKind.HeritageClause} node may be for extend a
+    // superclass _or_ for implementing an interface. We only want to add a
+    // `super()` call to our synthetic constructor here in the case that there
+    // is a superclass, so we can check for that situation by checking for the
+    // present of a heritage clause with the `.token` property set to
+    // `ts.SyntaxKind.ExtendsKeyword`.
```
rwaskiewicz, line 426: TIL

rwaskiewicz (changes requested):

src/compiler/transformers/decorators-to-static/convert-decorators.ts
alicewriteswrongs: me too 🤦‍♀️

tanner-reits: Kinda dumb imo. They should offer a better way of checking this 🙄

tanner-reits (approved):

rwaskiewicz (approved):

PR #3686, opened 2022-10-05
chore(compiler): convert to `ts.factory` methods in `add-static-style…

alicewriteswrongs:
….ts`

This just updates called to the deprecated methods like `ts.createIdentifier` to be calls to their `ts.factory` equivalents (like `ts.ts.factory.createIdentifier`, etc).

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

GitHub Issue Number: N/A

## What is the new behavior?

-
-
-

- [ ] Yes
- [x] No

### DIFF
--- a/src/compiler/transformers/add-static-style.ts
+++ b/src/compiler/transformers/add-static-style.ts
@@ -31,18 +31,23 @@ export const addStaticStyleGetterWithinClass = (
 
 /**
  * Adds static "style" property to the class variable.
+ *
  * ```typescript
  * const MyComponent = class {}
  * MyComponent.style = "styles";
  * ```
+ *
  * @param styleStatements a list of statements containing style assignments to a class
  * @param cmp the metadata associated with the component being evaluated
  */
 export const addStaticStylePropertyToClass = (styleStatements: ts.Statement[], cmp: d.ComponentCompilerMeta): void => {
   const styleLiteral = getStyleLiteral(cmp, false);
   if (styleLiteral) {
-    const statement = ts.createStatement(
-      ts.createAssignment(ts.createPropertyAccess(ts.createIdentifier(cmp.componentClassName), 'style'), styleLiteral)
+    const statement = ts.factory.createExpressionStatement(
+      ts.factory.createAssignment(
+        ts.factory.createPropertyAccessExpression(ts.factory.createIdentifier(cmp.componentClassName), 'style'),
+        styleLiteral
+      )
     );
     styleStatements.push(statement);
   }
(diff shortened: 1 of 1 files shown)


src/compiler/transformers/add-static-style.ts
```diff
@@ -92,11 +97,11 @@ const getMultipleModeStyle = (
...
 
-  return ts.createObjectLiteral(styleModes, true);
+  return ts.factory.createObjectLiteralExpression(styleModes, true);
 };
 
 const createPropertyAssignment = (mode: string, initializer: ts.Expression) => {
-  const node = ts.createPropertyAssignment(mode, initializer);
+  const node = ts.factory.createPropertyAssignment(mode, initializer);
   return node;
 };
```
alicewriteswrongs, line 106: this function strikes me as quite pointless, thoughts on removing it?

alicewriteswrongs (commented):
one thing

src/compiler/transformers/add-static-style.ts
tanner-reits: I'd condense it

src/compiler/transformers/add-static-style.ts
```diff
@@ -31,18 +31,23 @@ export const addStaticStyleGetterWithinClass = (
...
+ *
  * @param styleStatements a list of statements containing style assignments to a class
  * @param cmp the metadata associated with the component being evaluated
  */
 export const addStaticStylePropertyToClass = (styleStatements: ts.Statement[], cmp: d.ComponentCompilerMeta): void => {
   const styleLiteral = getStyleLiteral(cmp, false);
   if (styleLiteral) {
-    const statement = ts.createStatement(
-      ts.createAssignment(ts.createPropertyAccess(ts.createIdentifier(cmp.componentClassName), 'style'), styleLiteral)
+    const statement = ts.factory.createExpressionStatement(
```
tanner-reits, line 46: Just wanna verify that `createExpressionStatement` is the equivalent of `createStatement`?

tanner-reits (approved):
Looks good. Just left one question for verification

src/compiler/transformers/add-static-style.ts
alicewriteswrongs: yes, one way you can check is if you can find a `.createStatement` call somewhere in the codebase there should be a type hint with a deprecation message which gives the equivalent `ts.factory` function

src/compiler/transformers/add-static-style.ts
rwaskiewicz: On one hand, I agree.

On the other, I see how the level of abstraction _could_ help here. With wrappers like these, in theory we'd only ever need to update in once place should the TS API change again (rather than hunt for `/ts\.[^factory]/` like we're gonna do). 

Butttttt I don't think we're in a good place to take on _that_ work - so I say let's just rm it

PR #3711, opened 2022-10-11
chore(types): fix strict null checks violations for copy validation

rwaskiewicz:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

GitHub Issue Number: N/A

## What is the new behavior?

this commit adds unit tests for validate-copy#validateCopy to get a
better understanding/conveyance of what the function does in pursuit of
trying to drop the number of strict null checks related to the function

with that in mind, i was then able to:

1. fix one strict null check for the `dist-custom-element`
output target,
2. adds JSDoc to the function performing the copy
validation
3. fix a single snc error related to calling `validateCopy` for www
4.  fix a typing error on the `reduce()` call by narrowing the types
   of output targets we generate as a result of this function call

- [ ] Yes
- [x] No

## Testing

Tests continue to pass (and were added)


src/compiler/config/outputs/validate-custom-element.ts
```diff
@@ -48,7 +48,7 @@ export const validateCustomElement = (
       });
     }
 
-    outputTarget.copy = validateCopy(outputTarget.copy, []);
+    outputTarget.copy = validateCopy(outputTarget.copy ?? false, []);
```
rwaskiewicz, line 51: Since the default list is an empty array ,`false` is fine here

src/compiler/config/outputs/validate-www.ts
```diff
@@ -32,47 +32,53 @@ export const validateWww = (config: d.ValidatedConfig, diagnostics: d.Diagnostic
...
+        type: COPY,
+        dir: buildDir,
+        copyAssets: 'dist',
+      });
+
+      // Copy for www
+      outputs.push({
+        type: COPY,
+        dir: outputTarget.appDir,
+        copy: validateCopy(outputTarget.copy ?? [], [
```
rwaskiewicz, line 66: Since the default list is not an empty array, provide an empty array to copy the defaults into

src/compiler/config/outputs/validate-www.ts
```diff
@@ -32,47 +32,53 @@ export const validateWww = (config: d.ValidatedConfig, diagnostics: d.Diagnostic
...
-
-    // Generate global style with original name
-    outputs.push({
-      type: DIST_GLOBAL_STYLES,
-      file: join(buildDir, `${config.fsNamespace}.css`),
-    });
-
-    return outputs;
-  }, []);
+  return userWwwOutputs.reduce(
```
rwaskiewicz, line 35: Because I narrowed `outputs`, that union bumped us down to a newline. I suggest looking at this with whitespace hidden (12 LoC changed)

src/compiler/config/validate-copy.ts
```diff
@@ -2,7 +2,19 @@ import { unique } from '@utils';
 
 import type * as d from '../../declarations';
 
-export const validateCopy = (copy: d.CopyTask[] | boolean, defaultCopy: d.CopyTask[] = []): d.CopyTask[] => {
+/**
+ * Validate a series {@link d.CopyTask}s
```
alicewriteswrongs, line 6:
```suggestion
 * Validate a series of {@link d.CopyTask}s
```

src/compiler/config/validate-copy.ts
```diff
@@ -2,7 +2,19 @@ import { unique } from '@utils';
...
+ * @param copy the copy tasks to validate, or a boolean to specify if copy tasks are enabled
+ * @param defaultCopy default copy tasks to add to the returned validated list if not present in the first argument
+ * @returns the validated copy tasks. this function _may_ return an empty list if the first argument is:
```
alicewriteswrongs, line 9: it looks like it's also possible if `copy` is `[]` and `defaultCopy` is either not passed or `[]` is passed, should we note that too?

alicewriteswrongs (changes requested):
hooray more tests! just had like one question

src/compiler/config/validate-copy.ts
rwaskiewicz: I went back and forth here. I didn't want to add _too_ much detail in the event that if the implementation changed, I was concerned we'd get out of sync with the JSDoc. Instead, [I decided to remove these details ](https://github.com/ionic-team/stencil/pull/3711/commits/a026828a70db8ce0802fbb00dd0e6d9d9720b491)and let the tests provide the documenation for us

alicewriteswrongs: I think return type of `Foo[]` implies the possibility of `[]` as the returned value anyhow, so that's a 👍 from me

alicewriteswrongs (approved):
lgtm!

tanner-reits (approved):

Issue #3724, opened 2022-10-17
bug: --json flag no longer parsed properly for tests


rwaskiewicz:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/.github/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

2.18.1

### Current Behavior

When running Stencil tests with the `--json` flag, previously, JSON would be output on a successful test run. With Stencil v2.18.1, it appears that this flag no longer works as expected - no output is created when tests are run with this flag.

### Expected Behavior

JSON is output when `stencil test --json` is run with for a project.

This works on a project using v2.16:
```

npx stencil test --spec --json -- my-component.spec.ts
[36:14.5]  @stencil/core
[36:14.6]  v2.16.1 🎻
[36:14.7]  testing spec files
[36:14.7]  jest args: --json -- my-component.spec.ts --spec --max-workers=8
 PASS  src/components/my-component/my-component.spec.ts
  my-component
    ✓ renders (9 ms)
    ✓ renders with values (2 ms)

Test Suites: 1 passed, 1 total
Tests:       2 passed, 2 total
Snapshots:   0 total
Time:        0.24 s, estimated 1 s
Ran all test suites matching /my-component.spec.ts|--spec|--max-workers=8/i.
[...]

PR #3730, opened 2022-10-18
fix(jest): adjust conversion of CLI args to Jest args

alicewriteswrongs:
This adjusts the conversion of CLI args (parsed by our `cli/parse-flags` module) to arguments for Jest to fix a problematic situation when passing an argument which is not recognized by our `parse-flags` module.

A common pattern in Jest is to use a filename filter to narrow which tests are matched and run by the test runner, like so:

```
jest -- some-file-name.test.ts
```

Our CLI arg parsing module recognizes all Jest CLI args (like `--json`, `--watch`, and more) as 'first class' args and parses them into a `ConfigFlags` object. For the arguments shown above (`'--'`, `'some-file-name.test.ts'`), however, it will _not_ recognize them as first-class arguments and will instead preserve them in the `unknownArgs` array on the `ConfigFlags` object.

The flags (known and unknown) parsed by the `parse-flags` module are then converted into settings for Jest using the `buildJestArgv` function. Prior to this commit we took the known and unknown args off of the `ConfigFlags` object like so:

```ts
const args = [...config.flags.unknownArgs.slice(), ...config.flags.knownArgs.slice()];
```

If a filename match pattern is used alone (as in the example above) this will result in an array that looks like this:

```ts
const args = ['--', 'some-file-name.test.ts'];
```

That will translate correctly to what we want to communicate to Jest, and it will only run test files which match the pattern.

The problem comes in when a filename match pattern is used in conjunction with a boolean flag, like so:

```
jest --json -- some-file-name.test.ts
```

Or, in a Stencil context, doing something like this:

```
npx stencil test --spec --json -- my-component.spec.ts
```

which would result in the following `args` array being passed to Jest:

```ts
const args = [
  '--',
  'some-file-name.test.ts',
  '--spec',
  '--json',
];
```

Jest expects that `'--'` is _only_ followed by filename patterns and that all boolean flags like `--json` are before it in the argument list, so it will interpret this array as trying to match filenames which contain the string `'--json'`, rather than setting the `.json` flag on the Jest config.

The solution to this is to switch the order of the arguments passed to Jest, passing the known arguments _first_, followed by the unknown arguments.

Note: this used to work properly before #3444 because a flag like `--json` would have been stuck in the `unknownArgs` array, so the argument array passed to Jest would have looked like:

```ts
const args = [
  '--json',
  '--',
  'some-file-name.test.ts'
];
```

When we added explicit support for all Jest args this behavior was broken.

This commit adds a regression test which fails without the fix (changing the argument order) and adds the fix as well.

## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

GitHub Issue Number: #3724

## What is the new behavior?

This fixes how we pass known and unknown CLI args to Jest to fix an issue where the `--json` flag was being interpreted as a filename path to match against.

- [ ] Yes
- [x] No

## Testing

I followed the reproduction steps in the issue and confirmed that this change fixes the issue. I don't believe it will have any other consequences.

### DIFF
--- a/src/testing/jest/jest-config.ts
+++ b/src/testing/jest/jest-config.ts
@@ -46,7 +46,7 @@ function getLegacyJestOptions(): Record<string, boolean | number | string> {
 export function buildJestArgv(config: d.ValidatedConfig): Config.Argv {
   const yargs = require('yargs');
 
-  const args = [...config.flags.unknownArgs.slice(), ...config.flags.knownArgs.slice()];
+  const args = [...config.flags.knownArgs.slice(), ...config.flags.unknownArgs.slice()];
 
   if (!args.some((a) => a.startsWith('--max-workers') || a.startsWith('--maxWorkers'))) {
     args.push(`--max-workers=${config.maxConcurrentWorkers}`);
--- a/src/testing/jest/test/jest-config.spec.ts
+++ b/src/testing/jest/test/jest-config.spec.ts
@@ -189,4 +189,16 @@ describe('jest-config', () => {
     expect(jestArgv.spec).toBe(true);
     expect(jestArgv.passWithNoTests).toBe(true);
   });
+
+  it('should parse a setup with a filepath constraint', () => {
+    const args = ['test', '--spec', '--json', '--', 'my-component.spec.ts'];
+    const config = mockValidatedConfig();
+    config.flags = parseFlags(args);
+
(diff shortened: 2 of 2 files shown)


rwaskiewicz (approved):

tanner-reits (approved):
Nice quick fix and awesome explanation!

PR #3736, opened 2022-10-20
chore(github): create auto-assign action

tanner-reits:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [X] Other (please describe): Github action

GitHub Issue Number: N/A

## What is the new behavior?

Adds a Github action that runs when a PR is opened to auto-assign Stencil team reviewers when the PR was opened by an internal member of the Ionic organization

- [ ] Yes
- [X] No


.github/workflows/actions/auto-request-reviews/main.js
```diff
@@ -0,0 +1,45 @@
+const core = require('@actions/core');
```
tanner-reits, line 1: This file is just the compiled output from the action source code.

.github/workflows/pr-assign.yml
```diff
@@ -0,0 +1,19 @@
+name: Auto Request Review
+
+on:
+  pull_request:
+    # TODO remove synchronize event
```
tanner-reits, line 5: We won't need "synchronize" once this is g2g. Only here for testing so it runs each time the PR has code pushed to it

package.json
```diff
@@ -54,7 +54,8 @@
     "tsc": "tsc --incremental",
     "tsc.prod": "tsc",
     "tsc.scripts": "tsc -p scripts/tsconfig.json",
-    "tsc.watch": "tsc --incremental --watch"
+    "tsc.watch": "tsc --incremental --watch",
+    "build.action": "ncc build ./.github/workflows/actions/auto-request-reviews/main.js -o .github/workflows/actions/auto-request-reviews"
```
tanner-reits, line 58: This will go away in final code. Not sure if we need to have some script to do this. Alternatively, we can publish the action outside of the Stencil repo so we wouldn't need to keep any of the action-specific code around

package.json
```diff
@@ -151,5 +152,10 @@
   "volta": {
     "node": "16.13.0",
     "npm": "8.1.1"
+  },
+  "dependencies": {
+    "@actions/core": "^1.9.1",
+    "@actions/github": "^5.0.3",
+    "@vercel/ncc": "^0.34.0"
```
tanner-reits, line 159: I think these should be able to move to dev-deps, since we only need them when compiling the action code.

tanner-reits (commented):
Left some comments for remaining cleanup work just as FYIs

.github/workflows/actions/auto-request-reviews/action.yml
```diff
@@ -0,0 +1,16 @@
+name: Auto Request Review
+
+inputs:
+  token:
+    description: 'GitHub token'
```
rwaskiewicz, line 5: For each of these descriptions, can we add additional detail? Especially for GitHub nomenclature like 'organization' and 'team slug', a little more detail + an explanation can go a long way for someone looking at this for the first time/needing to maintain it

.github/workflows/actions/auto-request-reviews/main.js
rwaskiewicz, line 1: Is there a way to write these in TypeScript? I'm admittedly very inexperienced in writing actions, but from a maintainability standpoint, I'm hesitant for us to keep this as JS if we can do it in TS

.github/workflows/actions/auto-request-reviews/main.js
```diff
@@ -0,0 +1,45 @@
...
+const github = require('@actions/github');
+
+async function run() {
+  const token = core.getInput('token');
+  const octokit = github.getOctokit(token);
+
+  const organization = core.getInput('org');
+  const teamSlug = core.getInput('team');
```
rwaskiewicz, line 9: Do these lookups fail at runtime if they're not provided?

.github/workflows/actions/auto-request-reviews/main.js
```diff
@@ -0,0 +1,45 @@
...
+      pull_number: github.context.payload.number,
+      reviewers,
+    });
+  }
+}
+
+try {
+  run();
+} catch (error) {
+  core.setFailed(error.message);
```
rwaskiewicz, line 44: It looks like there's at least 3 places where we `await` something that could fail in `run` - does this handle all of them? Do we know that `message` will always be a property on the error & be populated?

.github/workflows/pr-assign.yml
```diff
@@ -0,0 +1,19 @@
...
+    types: [opened, ready_for_review, reopened, synchronize]
+
+jobs:
+  auto-request-review:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579
```
rwaskiewicz, line 12: Minor, can we add a version comment that states what semver version this commit is tied to?
```suggestion
      - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # 2.4.0
```

package.json
rwaskiewicz: Is there anything stopping us from putting these in a `package.json` file in `.github/workflows/actions/auto-request-reviews`? Is there a requirement for them to be in the root level `package.json`? I wonder if it'd be better to keep these separate from the rest of the dependencies 🤔

rwaskiewicz (changes requested):

.github/workflows/actions/auto-request-reviews/package.json
```diff
@@ -0,0 +1,16 @@
...
+    "@actions/core": "^1.9.1",
+    "@actions/github": "^5.0.3",
+    "@vercel/ncc": "^0.34.0",
+    "typescript": "4.7.4"
+  },
+  "volta": {
+    "node": "12.22.7",
+    "npm": "6.14.15"
+  }
+}
```
rwaskiewicz, line 16: Can you do me a favor and add this manifest to `.github/dependabot.yml`?

I think we'd want something pretty similar to the root level `package.json`:
```
version: 2
updates:
  # existing config for root level package.json
  - package-ecosystem: 'npm' # See documentation for possible values
    directory: './.github/workflows/actions/auto-request-reviews'
    schedule:
      interval: 'daily'
    open-pull-requests-limit: 50
    # Disable rebasing for pull requests, as having several open pull requests all get simultaneously rebased gets
    # noisy from a notification standpoint
    rebase-strategy: 'disabled'
```

I don't think we need the `ignore` field (AFAIK), since this action will only be running on GHA. We'll just need to rebuild the mini project when we update things like typescript.

rwaskiewicz (changes requested):
@tanner-reits Thanks! I had one last minor ask. There are few general questions from my original round of reviews - can you PTAL and answer those as well?

package.json
tanner-reits: Good call, was able to make that change

.github/workflows/actions/auto-request-reviews/main.js
tanner-reits: Turns out they don't fail (which seemed weird to me 🤷‍♂️ ). So I added some manually checks that will throw errors in those cases

package.json
tanner-reits: Moved this to the action's dedicated `package.json` so we have a quick way of compiling the action if it changes (or deps change)

.github/workflows/actions/auto-request-reviews/main.js
tanner-reits: I updated this to appropriately handle the async execution (i.e. `Promise.catch()`). The message will be populated if any of the Github API calls fail, and I populated messages for the manually thrown errors if any of the expected params aren't provided.

.github/workflows/actions/auto-request-reviews/package.json
tanner-reits: @rwaskiewicz I think I added this correctly. Obviously let me know if I did it wrong. Not sure how to validate it...

.github/workflows/actions/auto-request-reviews/main.js
alicewriteswrongs: is the idea that we want to cut down on startup time by not having to do a compilation step? if so, would it be too heavy to try to do some sort of content-addressed hash of the current state of the `main.ts` file? Just might be nice to avoid having to do a PR which just commits the new file if something changes.

Maybe just running the script with `ts-node` ?

.github/workflows/actions/auto-request-reviews/main.ts
```diff
@@ -0,0 +1,55 @@
+import * as core from '@actions/core';
+import * as github from '@actions/github';
```
alicewriteswrongs, line 2: small suggestion: possibly name this import `githubActions` or similar

.github/workflows/actions/auto-request-reviews/tsconfig.json
```diff
@@ -0,0 +1,17 @@
+{
+  "compilerOptions": {
+    "alwaysStrict": true,
```
alicewriteswrongs, line 3: can we set `"strict": true` in here?

alicewriteswrongs (changes requested):
just had a few questions, mostly looks good to me!

.github/workflows/actions/auto-request-reviews/main.js
tanner-reits: This is my first delve into custom Github actions, so what I say here might be completely wrong lol

AFAICT, there's no way to execute the uncompiled TS code with `ts-node` as a part of the workflow (which is a bummer cuz that would be handy). Also, most of the resources I have found walkthrough manually compiling the code and committing that to the repo. So, I feel like that's pretty standard.

Couple thoughts on the need to update this when the action changes:
1. I realistically don't see the action changing too much. Maybe early on as we're feeling things out, but in the long run it's not likely
2. In the cases where we would change the action code, we'd likely want to compile it locally anyway to make sure it builds successfully. So it's just a matter of committing it at that point. Also, if you want to use `act` to test it locally, you need to compiled code anyway.

rwaskiewicz (commented):
This LGTM. Once @alicewriteswrongs 👍's it, we'll get started with the integration work (tokens and whatnot)

.github/workflows/actions/auto-request-reviews/main.js
alicewriteswrongs: cool, that all makes sense - just figured I would ask because the big JS blob made me wonder a little bit haha

alicewriteswrongs (approved):
lgtm!

rwaskiewicz (commented):
Sweet! I noticed that I wasn't assigned when this step passed. Is that expected?

I have two documentation/logging based asks now that this is in a working state:
1. Can you document this action's configuration in our Notion page that documents the Stencil repo setup? Specifically, can you add a few sentences about the different secrets used/why we need 2 tokens? LMK if you need confirmation which page I'm talking about
2. Can we add some logging to the success-path? If I were to debug this, I'm not sure how to best do it. In cases like that, I'd look at logs of the Action's run. But I think we're only logging errors ATM. I can see a case where it'd be helpful to see something like:
```
tanner-reits requested a review
retrieving stencil team members....
found alicewriteswrongs, rwaskiewicz
assigning alicewriteswrongs, rwaskiewicz
success!
```
(total mockup)

rwaskiewicz:
@tanner-reits With https://github.com/ionic-team/stencil/pull/3736/commits/8e3fd02e45fc74a2cfa7bdf6e7cbc3096c5b0b65, I didn't see us recompiling the project in that commit. Do we need to do that before we land this?

tanner-reits:
> @tanner-reits With [8e3fd02](https://github.com/ionic-team/stencil/commit/8e3fd02e45fc74a2cfa7bdf6e7cbc3096c5b0b65), I didn't see us recompiling the project in that commit. Do we need to do that before we land this?

Yes, thank you! Forgot to push that up

rwaskiewicz:
@tanner-reits 👍 Can we temporarily revert https://github.com/ionic-team/stencil/pull/3736/commits/374a6df79a72fe966c2f86e39a85dd19ffbeef56 to see this running with the re-compiled output?

tanner-reits:
@rwaskiewicz https://github.com/ionic-team/stencil/actions/runs/3640145814/jobs/6144428678 logs from latest run

rwaskiewicz:
Hmmmm I think I missed something here. 

I thought this would be setting the _assignees_, rather than the reviewers
Screenshot 2022-12-07 at 10 10 44 AM

Maybe I missed something? Some of the the 'assign', 'review' nomenclature is a bit foggy (GH's documentation doesn't help here at all IMO). Do we want this to set the _reviewers_ (instead of the Stencil team group), or the assignees?

rwaskiewicz:
And FWIW should we go the 'assignment' route, IIUC, all pull request entities in GitHub are issue entities (but not all issue entities are pull requests). In theory, https://octokit.github.io/rest.js/v19#issues-add-assignees could be used to assign folks

tanner-reits:
> Hmmmm I think I missed something here.
> 
> I thought this would be setting the _assignees_, rather than the reviewers Screenshot 2022-12-07 at 10 10 44 AM
> 
> Maybe I missed something? Some of the the 'assign', 'review' nomenclature is a bit foggy (GH's documentation doesn't help here at all IMO). Do we want this to set the _reviewers_ (instead of the Stencil team group), or the assignees?

@rwaskiewicz I was intending this to set individual reviewers (not assignees) because then you can leverage "scheduled reminder" Slack notifications with the integration. We could, have it set both reviewers and assignees

rwaskiewicz:
Today, the reason that we assign `@stencil/team` under reviewers ([originally added here](https://github.com/ionic-team/stencil/pull/2977)) as opposed to individual folks was so that as the team took shape, we didn't have to change `.github/CODEOWNERS` every time the team changed. I think this PR achieves the same thing in that regard, so I don't have any objection there. This achieves the same effect today.

Looking towards growing the team in 2023, I believe we're going to be splitting `CODEOWNERS` up more granularly, similar to [how the Ionic Framework did recently](https://github.com/ionic-team/ionic-framework/pull/26442/files), where specific individuals will take more ownership of different parts of the codebase. I'd like to continue to use `CODEOWNERS` in this regard (assuming we split ownership on a directory/file path basis), since that logic is managed by GitHub (I.E. we don't have to maintain it). In this scenario, say if you were the owner of `src/compiler/optimize`, GItHub would perform the action requesting you to review it if someone else wanted to make a change to it. Similarly, as a PR grows in scope and includes additional directories/files that requires additional owners not included in the original PR, GitHub manages that for us as well.

IMO using Assignment is a little more expressive of the current state of a PR, but that's my personal opinion 😆 I _do_ think that using Assignment aligns with our current working agreement about reviewing code, but maybe it's something we revisit. Curious to know what you and @alicewriteswrongs think what we should do moving forward.

tanner-reits:
@rwaskiewicz Couple thoughts:
1. I think we should still request review from the Stencil team. I guess using the API to assign reviewers overrides the automation that was setting the team as a reviewer, so I'll TAL to make sure that still happens if individual reviewers get set. 
2. I like the idea of using code owners more. Obviously not baked in this action now, but I know there are auto-assign actions that leverage code owners, so we could for sure modify this when/if we go down that path

alicewriteswrongs:
Using `CODEOWNERS` to auto-request reviews from particular people sounds like a good workflow to me. I'm not particularly opinionated about how it's achieved, but the end goal I'd like to see is an easy way to filter in the github UI on PRs I should read and then respond to - whether we do that with assignment or requesting reviewers, I feel a bit agnostic on haha.

rwaskiewicz (approved):
Discussed today during our Stencil Alignment Series (see Notion for doc) - we're going to go with "reviewers" over "assignees" and will re-work as needed when `CODEOWNERS` is split up further

tanner-reits:
@alicewriteswrongs @rwaskiewicz When the branch rebased that triggered the action to run again because we hadn't removed the "synchronize" trigger yet. I pushed a commit to remove that, so this just needs a final stamp of approval

rwaskiewicz (approved):

alicewriteswrongs (approved):
lgtm!

PR #3741, opened 2022-10-21
proposal: remove merge-source-map in favor of depending on source-map

alicewriteswrongs:
This removes `merge-source-map` in favor of depending on and using the types from the `source-map` package, which 1) has types 2) is maintained and 3) `merge-source-map` depends on already.

This adds a function based on `merge-source-map` which gives us the same functionality.

I'd like to propose making this because (as I'll note with some code comments) I ran into some issues when trying to make a different change relating to the module declaration for source map. That led me to notice how old the `merge-source-map` package was, and looking at how little code is in there I thought, why not use `source-map` directly so we can benefit from the type declarations available there, export one less interface from our giant type declaration blob, and maybe decouple some code a little bit?

Anywhoo - that's the proposal. I have tested running this change in framework without issues.

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

GitHub Issue Number: N/A

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No


src/declarations/stencil-private.ts
```diff
@@ -32,16 +34,6 @@ import type {
   VNodeData,
 } from './stencil-public-runtime';
 
-export interface SourceMap {
```
alicewriteswrongs, line 35: since we're depending on the `RawSourceMap` interface from `source-map` now we no longer need our own interface

src/compiler/optimize/optimize-module.ts
```diff
@@ -216,3 +216,64 @@ export const prepareModule = async (
...
+ * If old or new source map value is falsy, return the other as-is.
+ *
+ * Based on the implementation of `merge-source-map`, see:
+ * https://github.com/keik/merge-source-map/blob/master/index.js
+ *
+ * @param oldMap old source map object
+ * @param newMap new source map object
+ * @returns a source map with the new merged into the old
+ */
+async function mergeSourceMaps(oldMap?: RawSourceMap, newMap?: RawSourceMap): Promise<RawSourceMap> {
```
alicewriteswrongs, line 232: this function is basically [copy-pasted from `merge-source-map`](https://github.com/keik/merge-source-map/blob/master/index.js)

alicewriteswrongs: I just made a few changes to add type annotations and so on

types/merge-source-map.d.ts
```diff
@@ -1,28 +0,0 @@
...
-    oldMap: import('../src/declarations').SourceMap,
-    newMap: import('../src/declarations').SourceMap
-  ): import('../src/declarations').SourceMap | undefined;
-
-  /**
-   * mark the `merge` function as the default export of the module, so that the JSDoc for `merge` is properly picked up
-   * by IntelliSense
-   */
-  export = merge;
-}
```
alicewriteswrongs, line 28: when trying to make a different change (related to trying to lessen the coupling between modules) I ran into some issues with this declaration file and the rollup build, so I decided to take a look at removing this package.

alicewriteswrongs (commented):
noting a few things

rwaskiewicz:
@tanner-reits this might be of interest to you with regard to the Terser upgrade (#3687)

tanner-reits (approved):
I'm cool with going this direction and think everything here looks good. It'll have some conflicts with #3687, but I can take care of those since the work to be compatible with the new version of Terser will still be needed even with this change 👍

rwaskiewicz:
@alicewriteswrongs can you post the testing instructions & fill out the remainder of the GH summary? I tried to build this, install it in a local component library and wasn't able to build with `sourceMap: true` set in my `stencil.config.ts`. May very well be PEBKAC, but just wanted to be sure 😆

alicewriteswrongs:
@rwaskiewicz yup there is indeed a build issue, looking into it now. I'll ping when I figure out what's up!

PR #3747, opened 2022-10-21
proposal: use project references to carve off strictly-checked sections of the project

alicewriteswrongs:
This updates our TypeScript build setup to use Project References to split off sub-projects within the codebase. This has two advantages for us:

1. TypeScript's `--build` subcommand will avoid rebuilding sub-projects if they haven't changed, improving the performance of our TypeScript build, and
2. Each sub-project has it's own `tsconfig.json`, which allows _each one to set compiler options_. This will let us turn on `strict` for subsets of the codebase without requiring it everywhere.

You can read a good deal about how this works here: https://www.typescriptlang.org/docs/handbook/project-references.html

The basic gist is:

- create a single tsconfig file for all others to inherit from
- add `tsconfig.json` files to directories that you want to build as separate projects
- define which files should be included therein, as well as which other projects that project should depend on

This PR really just puts the configuration changes in place and carves off a few basic sub-projects (see changes to the main `tsconfig.json` for details). Ideally, eventually we'd have each directory under `src/` building as a separate project, each with a whole bunch of sub-projects, and so on.

The difficult part is that typescript does not allow circular relationships between the sub-projects. So in order to do this we need to be able to define a DAG of projects, something that is made fairly difficult by some of how types flow through the project (especially the `src/declarations` stuff).

But even so, I think this is a pretty great option!

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

GitHub Issue Number: N/A

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No


tsconfig-base.json
```diff
@@ -0,0 +1,51 @@
+{
```
alicewriteswrongs, line 1: this is now the base TypeScript config which all of the other ones inherit from

alicewriteswrongs: so we can change options here and they'll be reflected in all the sub project

tsconfig.json
```diff
@@ -1,77 +1,36 @@
...
     "src/internal/index.ts",
     "src/mock-doc/index.ts",
-    "src/sys/node/index.ts",
-    "src/sys/node/worker.ts",
-    "src/sys/node/public.ts",
     "src/screenshot/index.ts",
     "src/screenshot/pixel-match.ts",
+    "src/sys/node/index.ts",
+    "src/sys/node/public.ts",
+    "src/sys/node/worker.ts",
```
alicewriteswrongs, line 33: I just alphabetized this list

tsconfig.json
```diff
@@ -1,77 +1,36 @@
 {
```
alicewriteswrongs, line 1: This is the new root `tsconfig.json` for the whole project. You can see that there isn't too much going on here - here we basically:

- inherit from our base config
- use `files` to include a bunch of files
- list sub project which this main project depends on in `references`

src/compiler/diagnostic/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{ 
+  "extends": "../../../tsconfig-base.json",
+  "compilerOptions": {
+    "composite": true,
+    "strict": true
+  },
+  "include": [
+    "./index.ts"
+  ]
+}
```
alicewriteswrongs, line 10: this is the simplest sub-project I created. It just includes a single file (`src/compiler/diagnostic/index.ts`) and doesn't depend on any other projects.

Note that if you try to import from outside of the current project into a file that is not included in it TypeScript will give you an error, requiring you do explicitly define and delineate the relationships among the subprojects. I think this is a good thing!

src/compiler/sys/logger/tsconfig.json
```diff
@@ -0,0 +1,17 @@
...
+  "include": [
+    "./index.ts",
+    "./logger.ts",
+    "./console-logger.ts",
+    "./terminal-logger.ts"
+  ],
+  "references": [
+    { "path": "../environment" },
+    { "path": "../../diagnostic" }
+  ]
```
alicewriteswrongs, line 16: this project is a little more substantial. It includes four files and depends on (`'references'`) the `environment` and `diagnostic` projects.

Note also that we have `"strict": true` here, so the files under here are now somewhat permanently strictly-checked.

package.json
```diff
@@ -52,9 +52,9 @@
     "test.watch": "jest --watch",
     "test.watch-all": "jest --watchAll --coverage",
     "tsc": "tsc --incremental",
-    "tsc.prod": "tsc",
+    "tsc.prod": "tsc --build --force",
     "tsc.scripts": "tsc -p scripts/tsconfig.json",
-    "tsc.watch": "tsc --incremental --watch"
+    "tsc.watch": "tsc --incremental --watch --build"
```
alicewriteswrongs, line 57: In order to build with project references we need to use the `--build` flag. If you want to check out what it's doing you can also pass `--verbose` and it will show you want its building / rebuilding and why.

src/compiler/diagnostic/index.ts
```diff
@@ -0,0 +1,20 @@
+export interface Diagnostic {
```
alicewriteswrongs, line 1: I created this new sub-project just for the diagnostic file in order to basically make it a 'leaf' node within a future dependency graph - if we actually want to do this more widely throughout the project we'll need to make as many things leaf nodes as possible, otherwise we'll have a really hard time dealing with circular dependency issues (TypeScript does not allow _any_ circular dependencies among project references, unlike within a single project).

src/compiler/sys/logger/logger.ts
```diff
@@ -0,0 +1,75 @@
...
+ * same level in the hierarchy than `currentLevel`, like so:
+ *
+ * ```ts
+ * LOG_LEVELS.indexOf(msgLevel) >= LOG_LEVELS.indexOf(currentLevel)
+ * ```
+ *
+ * NOTE: for the reasons described above, do not change the order of the entries
+ * in this array without good reason!
+ */
+export const LOG_LEVELS = ['debug', 'info', 'warn', 'error'] as const;
```
alicewriteswrongs, line 24: I moved the stuff for logger into this module so that the sub-project this module is part of won't have to depend on the declaration files when they are eventually built as part of a separate project later on.

I tried (both here and with the `Diagnostic` change) to do two things:

- retain the same exports from `src/declarations` by re-exporting from their new location
- update all imports across the project to point at the new location rather than importing from `src/declarations`

alicewriteswrongs (commented):
pointing out a few things

alicewriteswrongs (commented):
one other thing to note

package.json
```diff
@@ -52,9 +52,9 @@
     "test.watch": "jest --watch",
     "test.watch-all": "jest --watchAll --coverage",
     "tsc": "tsc --incremental",
-    "tsc.prod": "tsc",
+    "tsc.prod": "tsc --build --force",
```
alicewriteswrongs, line 55: the `--force` flag rebuilds all projects even if `tsc` already has an up-to-date version hanging around

alicewriteswrongs:
not going to do this

PR #3743, opened 2022-10-21
feat(typescript): support typescript 4.8

rwaskiewicz:
## Note

- [X] Please review/land https://github.com/ionic-team/stencil/pull/3835 first
## Pull request checklist

Please check if your PR fulfills the following requirements:
- [x] Tests for the changes have been added (for bug fixes / features)
- [X] Docs have been reviewed and added / updated if needed (for bug fixes / features)
    - See https://github.com/ionic-team/stencil-site/pull/934
- [x] Build (`npm run build`) was run locally and any changes were pushed
- [x] Unit tests (`npm test`) were run locally and passed
- [x] E2E Tests (`npm run test.karma.prod`) were run locally and passed
- [x] Prettier (`npm run prettier`) was run locally and passed

## Pull request type

 

 

Please check the type of change your PR introduces:
- [ ] Bugfix
- [x] Feature
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

## What is the current behavior?

Stencil uses up to and through TypeScript v4.7.4

GitHub Issue Number: N/A

## What is the new behavior?

this pr updates typescript for stencil to v4.8.4. it leverages the work done
in [#3748](https://github.com/ionic-team/stencil/pull/3748) (a043e5d) in order to facilitate the
creation of nodes in the syntax tree that contain `ModifierLike` entities.
it also leverages the work from [#3835](https://github.com/ionic-team/stencil/pull/3835) in order
to provide testing to a more fickle aspect of the codebase.

starting with typescript 4.8, decorators are not longer directly accessible on a
syntax tree node. this is a result of the decorators proposal reaching stage 3, and
soon to be implemented by the typescript team. in order to prepare for the
implementation, where decorators are stored has changed. more information on
this change can be found in https://devblogs.microsoft.com/typescript/announcing-typescript-4-8/#decorators-are-placed-on-modifiers-on-typescripts-syntax-trees.

in order for decorators to be retrieved, typescript provides a series of helper
functions to retrieve both decorators and modifiers off of a node. this pr
introduces helper methods of our own that abstract over typescript's own. this
is done to reduce some of the verbosity in attempting to retrieve decorators/modifiers
off of a node. these functions return either the found decorators/modifiers on a node,
`undefined` if the node can have a decorator/modifier but does not have any, or
`undefined` if the node cannot have a decorator/modifier on it. this is an intentional
design decision, as opposed to returning an empty array in place of `undefined` for the
[...]

PR #3748, opened 2022-10-21
refactor(compiler): create nodes using factory methods

rwaskiewicz:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

We use several deprecated TypeScript APIs in our codebase. This is a blocker for the TypeScript 4.8 upgrade due to [breaking changes in how decorators are treated](https://devblogs.microsoft.com/typescript/announcing-typescript-4-8/#decorators-are-placed-on-modifiers-on-typescripts-syntax-trees) and Stencil's use of TS APIs.
(When I upgraded the codebase to TS 4.8, there was a lot of logging spam warnings about using these APIs too, but that's not the primary thing here, we can live with some build/test spam worst case)

Specifically, we often use a layer of indirection provided to us by the TS team when creating nodes in the tree:
```
Stencil code -> ts.createSomeTypeOfNode() 
```
where `ts.createSomeTypeOfNode` pipes to something like this in TypeScript v4.7:
```ts
/** @deprecated Use `factory.createSomeTypeOfNode` or the factory supplied by your transformation context instead. */
ts.createSomeTypeOfNode = ts.Debug.deprecate(ts.factory.createSomeTypeOfNode, factoryDeprecation);
```
in cases like the example provided above, this is just a straight pass-thru from the deprecated function to the new factory version. These types of conversions are often the easiest, since we can just add `.factory` to the callsite and be done.

In other cases however, there's more complicated logic provided to the first argument of `ts.Debug.deprecate` (again TS v4.7):
```ts
    /** @deprecated Use `factory.updateClassExpression` or the factory supplied by your transformation context instead. */
    ts.updateClassExpression = ts.Debug.deprecate(function updateClassExpression(node, modifiers, name, typeParameters, heritageClauses, members) {
        return ts.factory.updateClassExpression(node, /*decorators*/ undefined, modifiers, name, typeParameters, heritageClauses, members);
    }, factoryDeprecation);
```

In upgrading us to TS 4.8, I found that Node creation wouldn't work properly, due to decorator usage in Stencil and changes in the API. As I upgraded more and more of these deprecated calls, the number of type errors went down. I got us to a point where the code _compiled_, but wouldn't run at all. In upgrading _all_ of the deprecated APIs to use supported ones, it appears we are always properly created Nodes for the CST.

GitHub Issue Number: N/A

## What is the new behavior?

this commit updates stencil to create nodes in the typescript syntax tree using `ts.factory.*` functions, rather than the (create|update|etc.) functions that were originally exported directly on `ts`. said functions were deprecated in v4.0 of typescript, and are slated to be removed in v5.0.

this commit has two motivations.
1. these functions will be removed in typescript 5.0. this is a piece of technical debt the team was going to be tackling in preparation for said upgrade.
2. when attempting to upgrade to typescript 4.8, we ran into two issues:
   - several warnings regarding the use of these deprecated methods began to spring up, which wasn't a pleasant dev experience
   - it broke stencil - static members were not being properly transplated from the original source code to our intermediate representations. code that was originally written as:
```tsx
@Component({ ... }) export class Foo {
  @Prop bar: boolean;
}
```
would be transformed to:
```tsx
export class {}
```

the changes in this PR help with the 4.8 upgrade, in that moving to functions that are no longer deprecated allow for the code to be transpile properly. in TS 4.8, decorators are no longer a separate field on `Node` derived classes, but are added to the `modifiers` field. this change does _not_ take those changes directly into account. it only moves to the non-deprecated factory functions. the upgrade to TS 4.8 will modify some of these functions calls to manage `Node` typing changes

- [ ] Yes
- [x] No

### General Call Checks

The general strategy for this was to see each deprecated call, and checking two things:
- The type signature of the deprecated function in `typescript.d.ts`, which suggests a replacement 
- Pop open `typescript.js` and look for the assignment of the old/deprecated function. See what that actually calls through to

### Basic
CI continues to pass - this should be a 'refactor' where we convert from the deprecated fn to the proper, non-deprecated, factory-based one

### Ionic
We can also test this in Ionic Framework by:
- building a tarball of this branch (`npm ci && npm run clean && npm run build && npm tar`)
- opening `ionic-framework/core` (assuming you've cloned the repo) and cleaning up (`git pull && npm clean -fd && npm ci`)
- installing our tarball (`npm i [PATH_TO_TAR]`)
- running `npm start` locally. this will start a little Stencil dev server.
- navigate into various component directories. in each's `test` we  can play with various components to see things still appear to work

## Other information

I made this a parent PR to https://github.com/ionic-team/stencil/pull/3743


src/compiler/transformers/update-component-class.ts
```diff
@@ -50,17 +50,25 @@ const createConstClass = (
...
     constModifiers,
-    ts.createVariableDeclarationList(
+    ts.factory.createVariableDeclarationList(
       [
-        ts.createVariableDeclaration(
+        ts.factory.createVariableDeclaration(
           className,
           undefined,
-          ts.createClassExpression(classModifiers, undefined, classNode.typeParameters, heritageClauses, members)
+          undefined,
```
rwaskiewicz, line 63: We add `undefined` here because the fn sig changes when we move to the `factory` version
```diff
- createClassExpression: (modifiers: readonly Modifier[] | undefined, name: string | Identifier | undefined, typeParameters: readonly TypeParameterDeclaration[] | undefined, heritageClauses: readonly HeritageClause[] | undefined, members: readonly ClassElement[]) => ClassExpression;
+ createClassExpression: (decorators: readonly Decorator[] | undefined, modifiers: readonly Modifier[] | undefined, name: string | Identifier | undefined, typeParameters: readonly TypeParameterDeclaration[] | undefined, heritageClauses: readonly HeritageClause[] | undefined, members: readonly ClassElement[]): ClassExpression;
```

src/compiler/transformers/transform-utils.ts
```diff
@@ -727,18 +733,23 @@ export const createRequireStatement = (importFnNames: string[], importPath: stri
...
     undefined,
-    ts.createVariableDeclarationList(
+    ts.factory.createVariableDeclarationList(
       [
-        ts.createVariableDeclaration(
+        ts.factory.createVariableDeclaration(
           importBinding,
           undefined,
-          ts.factory.createCallExpression(ts.factory.createIdentifier('require'), [], [ts.createLiteral(importPath)])
+          undefined,
```
rwaskiewicz, line 747: We add `undefined` here because the fn sig changes when we move to the `factory` version
```diff
- createClassExpression: (modifiers: readonly Modifier[] | undefined, name: string | Identifier | undefined, typeParameters: readonly TypeParameterDeclaration[] | undefined, heritageClauses: readonly HeritageClause[] | undefined, members: readonly ClassElement[]) => ClassExpression;
+ createClassExpression: (decorators: readonly Decorator[] | undefined, modifiers: readonly Modifier[] | undefined, name: string | Identifier | undefined, typeParameters: readonly TypeParameterDeclaration[] | undefined, heritageClauses: readonly HeritageClause[] | undefined, members: readonly ClassElement[]): ClassExpression;
```

src/compiler/transformers/style-imports.ts
```diff
@@ -95,11 +95,11 @@ const createEsmStyleImport = (
   const importName = ts.factory.createIdentifier(style.styleIdentifier);
   const importPath = getStyleImportPath(transformOpts, tsSourceFile, cmp, style, style.externalStyles[0].absolutePath);
 
-  return ts.createImportDeclaration(
+  return ts.factory.createImportDeclaration(
     undefined,
     undefined,
-    ts.createImportClause(importName, undefined),
-    ts.createLiteral(importPath)
+    ts.factory.createImportClause(false, importName, undefined),
```
rwaskiewicz, line 101: We add `false` here because the function signature has changed such that `isTypeOnly` is no longer optional, is typed as `boolean`, and is now the first parameter to the function:
```diff
- createImportClause(isTypeOnly: boolean, name: Identifier | undefined, namedBindings: NamedImportBindings | undefined): ImportClause;
+ createImportClause(name: Identifier | undefined, namedBindings: NamedImportBindings | undefined, isTypeOnly?: any): ImportClause;
```

src/compiler/transformers/style-imports.ts
```diff
@@ -135,14 +135,19 @@ const createCjsStyleRequire = (
...
     undefined,
-    ts.createVariableDeclarationList(
+    ts.factory.createVariableDeclarationList(
       [
-        ts.createVariableDeclaration(
+        ts.factory.createVariableDeclaration(
           importName,
           undefined,
-          ts.factory.createCallExpression(ts.factory.createIdentifier('require'), [], [ts.createLiteral(importPath)])
+          undefined,
```
rwaskiewicz, line 145: the `type: TypeNode | undefined` argument must be explicitly set now:
```diff
- createVariableDeclaration(name: string | BindingName, exclamationToken?: ExclamationToken, type?: TypeNode, initializer?: Expression): VariableDeclaration;
+ createVariableDeclaration(name: string | BindingName, exclamationToken: ExclamationToken | undefined, type: TypeNode | undefined, initializer: Expression | undefined): VariableDeclaration;
```

src/compiler/transformers/transform-utils.ts
```diff
@@ -664,24 +688,6 @@ export const isMethod = (member: ts.ClassElement, methodName: string): member is
   return ts.isMethodDeclaration(member) && member.name && (member.name as any).escapedText === methodName;
 };
 
-export const isAsyncFn = (typeChecker: ts.TypeChecker, methodDeclaration: ts.MethodDeclaration) => {
```
rwaskiewicz, line 667: Dead code. Be gone! 👻

src/compiler/transformers/transform-utils.ts
```diff
@@ -705,19 +711,19 @@ export const createImportStatement = (importFnNames: string[], importPath: strin
     );
   });
 
-  return ts.createImportDeclaration(
+  return ts.factory.createImportDeclaration(
     undefined,
     undefined,
-    ts.createImportClause(undefined, ts.createNamedImports(importSpecifiers)),
-    ts.createLiteral(importPath)
+    ts.factory.createImportClause(false, undefined, ts.factory.createNamedImports(importSpecifiers)),
```
rwaskiewicz, line 717: We add `false` here because the function signature has changed such that `isTypeOnly` is no longer optional, is typed as `boolean`, and is now the first parameter to the function:
```diff
- createImportClause(isTypeOnly: boolean, name: Identifier | undefined, namedBindings: NamedImportBindings | undefined): ImportClause;
+ createImportClause(name: Identifier | undefined, namedBindings: NamedImportBindings | undefined, isTypeOnly?: any): ImportClause;
```

tanner-reits (approved):
lgtm!

alicewriteswrongs (approved):
I think this all LGTM!

PR #3789, opened 2022-11-02
fix(compiler): move `inlineDynamicImports` to Rollup output

tanner-reits:
This commit updates the `inlineDyamicImports` Rollup config option to be moved to the `output` section of the config as the root-level config is deprecated. This also resolves an issue that was causing build errors when the flag was set on `dist-custom-elements` due to multiple inputs

## Pull request checklist

## Pull request type

 

 

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

Currently, setting `inlineDynamicImports` on the `dist-custom-elements` output target will result in a build error

GitHub Issue Number: N/A

## What is the new behavior?

- Updates our mapping of `inlineDynamicImports` in our `BundleOptions` interface to a Rollup config so that the option is set as a part of the `output` config object. This is the new approach as the root option was marked as deprecated (https://rollupjs.org/guide/en/#outputinlinedynamicimports)
- Also adds logic so the option can only be set when the number of inputs to the bundler is 1 or fewer (this is the error that was being thrown during the build process before)

- [ ] Yes
- [X] No

## Testing

- Verified the build was successful
- Added unit tests for the changes and conditional logic

## Other information

I looked at Amplitude to see if the option was actual used on `dist-custom-elements` and looked like there were some instances of it recently (not sure how since builds likely would've been failing). But, for now I think we should keep that option around

### DIFF
--- a/src/compiler/bundle/bundle-output.ts
+++ b/src/compiler/bundle/bundle-output.ts
@@ -96,6 +96,10 @@ export const getRollupOptions = (
   const afterPlugins = config.rollupPlugins.after || [];
   const rollupOptions: RollupOptions = {
     input: bundleOpts.inputs,
+    output: {
+      // This option cannot be set if multiple inputs are provided
+      inlineDynamicImports: bundleOpts.inlineDynamicImports ? Object.keys(bundleOpts.inputs).length <= 1 : false,
+    },
 
     plugins: [
       coreResolvePlugin(config, compilerCtx, bundleOpts.platform, bundleOpts.externalRuntime),
@@ -129,7 +133,6 @@ export const getRollupOptions = (
     ],
 
     treeshake: getTreeshakeOption(config, bundleOpts),
-    inlineDynamicImports: bundleOpts.inlineDynamicImports,
     preserveEntrySignatures: bundleOpts.preserveEntrySignatures ?? 'strict',
 
     onwarn: createOnWarnFn(buildCtx.diagnostics),
--- a/src/compiler/bundle/test/bundle-output.spec.ts
+++ b/src/compiler/bundle/test/bundle-output.spec.ts
@@ -1,0 +1,52 @@
+import { mockBuildCtx, mockCompilerCtx, mockValidatedConfig } from '@stencil/core/testing';
+import { OutputOptions } from 'rollup';
+
+import { BuildCtx, CompilerCtx, ValidatedConfig } from '../../../declarations';
+import { BundleOptions } from '../bundle-interface';
+import { getRollupOptions } from '../bundle-output';
+
+describe('bundle output', () => {
+  describe('getRollupOptions', () => {
+    let config: ValidatedConfig;
+    let compilerCtx: CompilerCtx;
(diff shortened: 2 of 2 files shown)


src/compiler/bundle/bundle-output.ts
```diff
@@ -96,6 +96,10 @@ export const getRollupOptions = (
   const afterPlugins = config.rollupPlugins.after || [];
   const rollupOptions: RollupOptions = {
     input: bundleOpts.inputs,
+    output: {
+      // This option cannot be set if multiple inputs are provided
+      inlineDynamicImports: bundleOpts.inlineDynamicImports ? Object.keys(bundleOpts.inputs).length <= 1 : false,
```
rwaskiewicz, line 101: For my own edification, in what scenarios is `bundleOpts.inputs` equal to zero? This might be something worth adding a comment to

src/compiler/bundle/test/bundle-output.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+        inputs: {
+          index: 'core',
+        },
+      };
+    });
+
+    it('should default `inlineDynamicImports` to `false` in the output section', () => {
+      const options = getRollupOptions(config, compilerCtx, buildCtx, bundleOptions);
+
+      expect((options.output as OutputOptions).inlineDynamicImports).toBe(false);
```
rwaskiewicz, line 31: I really wish there was a way for `expect` assertions to narrow a type for us. `options.output` could feasibly be `undefined`, `OutputOptions`, or `OutputOptions[]`. Could we add a few assertions to make using this assertion a little safer for us?
```
      expect(options.output).toBeDefined();
      expect(Array.isArray(options.output)).toBe(false);
```

rwaskiewicz (changes requested):
A few additional asks:

1. Can we update the testing section as to how we tested this for `lazy`, `hydrate` and `custom-elements`? I see this field being used in a few places not explicitly called out, and want to make sure we have it covered:
Screen Shot 2022-11-08 at 9 26 12 AM
2. Can we add JSDoc to the fields (`src/compiler/bundle/bundle-interface.ts#BundleOptions.inlineDynamicImports`) and the `inlineDynamicImports` fields for the output targets in the public compiler type interface?
3. Can we update the docs with this field?

tanner-reits:
@rwaskiewicz I took another look at this, and, honestly, we might want to just remove the `inlineDynamicImports` option from `dist-custom-elements` (and `dist-custom-elements-bundle`, but obviously that's going away anyway).

My reasoning is, the option is only valid if the number of inputs is one or less. But, for those targets, there will always be at least two "inputs" because we have the "core" input and then any components. Which, nobody is going to have a Stencil lib without any components in it 😂 

We'll still keep the option around for the `BundleOptions` since we hard-set that `true` for the hydrate app bundling.

rwaskiewicz:
@tanner-reits I think I'm OK with that 👍

alicewriteswrongs:
me too! :+1:

PR #3813, opened 2022-11-10
chore(compiler): add JSDoc to validate-service-worker.ts

alicewriteswrongs:
This adds JSDoc comments to validate-service-worker.ts, and does a (very light) refactor just moving around a bit of logic.

This service worker configuration is using the Workbox library for generating a caching service worker automatically, but at present we do not depend on the published types for Workbox. Since there is an compatibility issue using web worker typings alongside the DOM typings we're already using (by setting the `lib` property in `tsconfig.json`) we can't depend on those published typings directly until we can separately set compilation options in the code (using project references, for instance). As a workaround the current typings for the options interface for Workbox are copied into the code.

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

No JSDocs in the validation code for service worker configuration!

## What is the new behavior?

Now we have JSDocs in this module!

- [ ] Yes
- [x] No

### DIFF
--- a/src/compiler/config/validate-service-worker.ts
+++ b/src/compiler/config/validate-service-worker.ts
@@ -12,11 +12,11 @@ import type * as d from '../../declarations';
  * assets on the client. More here: https://developer.chrome.com/docs/workbox/
  *
  * This function first checks that the service worker config set on the
- * supplied `OutputTarget` is not any of it's possible empty values and that we
- * are not currently in development mode. In those cases it will early return.
+ * supplied `OutputTarget` is not empty and that we are not currently in
+ * development mode. In those cases it will early return.
  *
- * If we do find a service worker configuration then we do some validation to
- * ensure that things are set up correctly.
+ * If we do find a service worker configuration we do some validation to ensure
+ * that things are set up correctly.
  *
  * @param config the current, validated configuration
  * @param outputTarget the `www` outputTarget whose service worker
--- a/src/compiler/config/validate-service-worker.ts
+++ b/src/compiler/config/validate-service-worker.ts
@@ -42,9 +42,10 @@ export const validateServiceWorker = (config: d.ValidatedConfig, outputTarget: d
     return;
   }
 
-  const globDirectory = typeof outputTarget.serviceWorker?.globDirectory === 'string'
-    ? outputTarget.serviceWorker.globDirectory
-    : outputTarget.appDir;
+  const globDirectory =
+    typeof outputTarget.serviceWorker?.globDirectory === 'string'
+      ? outputTarget.serviceWorker.globDirectory
+      : outputTarget.appDir;
 
   outputTarget.serviceWorker = {
     ...outputTarget.serviceWorker,
--- a/src/compiler/config/validate-service-worker.ts
+++ b/src/compiler/config/validate-service-worker.ts
@@ -42,11 +42,15 @@ export const validateServiceWorker = (config: d.ValidatedConfig, outputTarget: d
     return;
   }
 
+  const globDirectory = typeof outputTarget.serviceWorker?.globDirectory === 'string'
+    ? outputTarget.serviceWorker.globDirectory
+    : outputTarget.appDir;
+
   outputTarget.serviceWorker = {
     ...outputTarget.serviceWorker,
-    globDirectory: outputTarget.serviceWorker?.globDirectory ?? outputTarget.appDir,
+    globDirectory,
     swDest: isString(outputTarget.serviceWorker?.swDest)
-      ? outputTarget.serviceWorker?.swDest
+      ? outputTarget.serviceWorker.swDest
       : join(outputTarget.appDir ?? '', DEFAULT_FILENAME),
   };
 
--- a/src/compiler/config/validate-service-worker.ts
+++ b/src/compiler/config/validate-service-worker.ts
@@ -3,7 +3,27 @@ import { isAbsolute, join } from 'path';
 
 import type * as d from '../../declarations';
 
-export const validateServiceWorker = (config: d.ValidatedConfig, outputTarget: d.OutputTargetWww) => {
+/**
+ * Validate that a service worker configuration is valid, if it is present and
+ * accounted for.
+ *
+ * Note that our service worker configuration / support is based on
+ * Workbox, a package for automatically generating Service Workers to cache
+ * assets on the client. More here: https://developer.chrome.com/docs/workbox/
+ *
+ * This function first checks that the service worker config set on the
+ * supplied `OutputTarget` is not any of it's possible empty values and that we
+ * are not currently in development mode. In those cases it will early return.
+ *
+ * If we do find a service worker configuration then we do some validation to
+ * ensure that things are set up correctly.
+ *
+ * @param config the current, validated configuration
+ * @param outputTarget the `www` outputTarget whose service worker
+ * configuration we want to validate. **Note**: the `.serviceWorker` object
+ * _will be mutated_ if it is present.
+ */
+export const validateServiceWorker = (config: d.ValidatedConfig, outputTarget: d.OutputTargetWww): void => {
   if (outputTarget.serviceWorker === false) {
     return;
   }
@@ -22,11 +42,11 @@ export const validateServiceWorker = (config: d.ValidatedConfig, outputTarget: d
     return;
   }
 
-  if (typeof outputTarget.serviceWorker !== 'object') {
-    // what was passed in could have been a boolean
-    // in that case let's just turn it into an empty obj so Object.assign doesn't crash
-    outputTarget.serviceWorker = {};
-  }
+  outputTarget.serviceWorker = {
+    ...outputTarget.serviceWorker,
+    globDirectory: outputTarget.serviceWorker?.globDirectory ?? outputTarget.appDir,
+    swDest: outputTarget.serviceWorker?.swDest ?? join(outputTarget.appDir, DEFAULT_FILENAME),
+  };
 
   if (!Array.isArray(outputTarget.serviceWorker.globPatterns)) {
     if (typeof outputTarget.serviceWorker.globPatterns === 'string') {
@@ -34,10 +54,6 @@ export const validateServiceWorker = (config: d.ValidatedConfig, outputTarget: d
     } else if (typeof outputTarget.serviceWorker.globPatterns !== 'string') {
       outputTarget.serviceWorker.globPatterns = DEFAULT_GLOB_PATTERNS.slice();
     }
-  }
-
-  if (typeof outputTarget.serviceWorker.globDirectory !== 'string') {
-    outputTarget.serviceWorker.globDirectory = outputTarget.appDir;
   }
 
   if (typeof outputTarget.serviceWorker.globIgnores === 'string') {
@@ -54,16 +70,19 @@ export const validateServiceWorker = (config: d.ValidatedConfig, outputTarget: d
     outputTarget.serviceWorker.swSrc = join(config.rootDir, outputTarget.serviceWorker.swSrc);
   }
 
-  if (!isString(outputTarget.serviceWorker.swDest)) {
-    outputTarget.serviceWorker.swDest = join(outputTarget.appDir, DEFAULT_FILENAME);
-  }
-
   if (!isAbsolute(outputTarget.serviceWorker.swDest)) {
     outputTarget.serviceWorker.swDest = join(outputTarget.appDir, outputTarget.serviceWorker.swDest);
   }
 };
 
-const addGlobIgnores = (config: d.Config, globIgnores: string[]) => {
+/**
+ * Add file glob patterns to the `globIgnores` for files we don't want to cache
+ * with the service worker.
+ *
+ * @param config the current, validated configuration
+ * @param globIgnores list of file ignore patterns. **Note**: will be mutated.
+ */
+const addGlobIgnores = (config: d.ValidatedConfig, globIgnores: string[]) => {
   globIgnores.push(
     `**/host.config.json`, // the filename of the host configuration
     `**/*.system.entry.js`,
--- a/src/compiler/config/validate-service-worker.ts
+++ b/src/compiler/config/validate-service-worker.ts
@@ -46,7 +46,7 @@ export const validateServiceWorker = (config: d.ValidatedConfig, outputTarget: d
     ...outputTarget.serviceWorker,
     globDirectory: outputTarget.serviceWorker?.globDirectory ?? outputTarget.appDir,
     swDest: isString(outputTarget.serviceWorker?.swDest)
-      ? outputTarget.serviceWorker?.swDest 
+      ? outputTarget.serviceWorker?.swDest
       : join(outputTarget.appDir ?? '', DEFAULT_FILENAME),
   };
 
--- a/src/compiler/config/validate-service-worker.ts
+++ b/src/compiler/config/validate-service-worker.ts
@@ -45,7 +45,9 @@ export const validateServiceWorker = (config: d.ValidatedConfig, outputTarget: d
   outputTarget.serviceWorker = {
     ...outputTarget.serviceWorker,
     globDirectory: outputTarget.serviceWorker?.globDirectory ?? outputTarget.appDir,
-    swDest: outputTarget.serviceWorker?.swDest ?? join(outputTarget.appDir ?? '', DEFAULT_FILENAME),
+    swDest: isString(outputTarget.serviceWorker?.swDest)
+      ? outputTarget.serviceWorker?.swDest 
+      : join(outputTarget.appDir ?? '', DEFAULT_FILENAME),
   };
 
   if (!Array.isArray(outputTarget.serviceWorker.globPatterns)) {
--- a/src/compiler/config/validate-service-worker.ts
+++ b/src/compiler/config/validate-service-worker.ts
@@ -45,7 +45,7 @@ export const validateServiceWorker = (config: d.ValidatedConfig, outputTarget: d
   outputTarget.serviceWorker = {
     ...outputTarget.serviceWorker,
     globDirectory: outputTarget.serviceWorker?.globDirectory ?? outputTarget.appDir,
-    swDest: outputTarget.serviceWorker?.swDest ?? join(outputTarget.appDir, DEFAULT_FILENAME),
+    swDest: outputTarget.serviceWorker?.swDest ?? join(outputTarget.appDir ?? '', DEFAULT_FILENAME),
   };
 
   if (!Array.isArray(outputTarget.serviceWorker.globPatterns)) {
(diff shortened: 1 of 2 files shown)


src/declarations/stencil-public-compiler.ts
```diff
@@ -2192,30 +2192,268 @@ export type OutputTarget =
...
 
+/**
+ * Our custom configuration interface for generated caching Service Workers
+ * using the Workbox library (see https://developer.chrome.com/docs/workbox/).
+ *
+ * Although we are using Workbox we are unfortunately unable to depend on the
+ * published types for the library because they must be compiled using the
+ * `webworker` lib for TypeScript, which cannot be used at the same time as
+ * the `dom` lib. As a workaround the properties are copy-pasted here from
+ * https://github.com/DefinitelyTyped/DefinitelyTyped/blob/c7b4dadae5b320ad1311a8f82242b8f2f41b7b8c/types/workbox-build/generate-sw.d.ts#L3
```
alicewriteswrongs, line 2203: as I note here we can't at present depend on these types because you need to have the `webworker` lib enabled (and that conflicts with the `dom` lib so both can't be set at the same time 😢).

If we have project references enabled at some time in the future, allowing parts of the codebase to be compiled as separate units, I think we'd be able to carve off the service-worker related code into it's own module and compile it against that lib instead, but for now we'll have to copy-paste this :/

alicewriteswrongs (commented):
just noting one thing

src/compiler/config/validate-service-worker.ts
```diff
@@ -17,18 +37,16 @@ export const validateServiceWorker = (config: d.ValidatedConfig, outputTarget: d
...
 
-  if (typeof outputTarget.serviceWorker !== 'object') {
-    // what was passed in could have been a boolean
-    // in that case let's just turn it into an empty obj so Object.assign doesn't crash
-    outputTarget.serviceWorker = {};
-  }
+  outputTarget.serviceWorker = {
+    ...outputTarget.serviceWorker,
+    globDirectory: outputTarget.serviceWorker?.globDirectory ?? outputTarget.appDir,
+    swDest: outputTarget.serviceWorker?.swDest ?? join(outputTarget.appDir ?? '', DEFAULT_FILENAME),
```
tanner-reits, line 48: Do we need to validate that this is a string?
```suggestion
    swDest: isString(outputTarget.serviceWorker?.swDest) ? outputTarget.serviceWorker.swDest : join(outputTarget.appDir ?? '', DEFAULT_FILENAME),
```

src/compiler/config/validate-service-worker.ts
```diff
@@ -56,16 +70,19 @@ export const validateServiceWorker = (config: d.ValidatedConfig, outputTarget: d
     outputTarget.serviceWorker.swSrc = join(config.rootDir, outputTarget.serviceWorker.swSrc);
   }
 
-  if (!isString(outputTarget.serviceWorker.swDest)) {
-    outputTarget.serviceWorker.swDest = join(outputTarget.appDir, DEFAULT_FILENAME);
-  }
-
-  if (!isAbsolute(outputTarget.serviceWorker.swDest)) {
-    outputTarget.serviceWorker.swDest = join(outputTarget.appDir, outputTarget.serviceWorker.swDest);
+  if (isString(outputTarget.serviceWorker.swDest) && !isAbsolute(outputTarget.serviceWorker.swDest)) {
```
tanner-reits, line 73: If we change the condition above (previous comment), then this `isString` check can go away
```suggestion
  if (!isAbsolute(outputTarget.serviceWorker.swDest)) {
```

src/declarations/stencil-public-compiler.ts
```diff
@@ -2192,30 +2192,268 @@ export type OutputTarget =
...
+
+  /**
+   * An optional ID to be prepended to cache names. This is primarily useful for
+   * local development where multiple sites may be served from the same
+   * `http://localhost:port` origin.
+   */
+  cacheId?: string | undefined;
+
+  /**
+   * Whether or not Workbox should attempt to identify an delete any precaches
```
tanner-reits, line 2242:
```suggestion
   * Whether or not Workbox should attempt to identify and delete any precaches
```

tanner-reits (changes requested):
Couple small things

src/compiler/config/validate-service-worker.ts
alicewriteswrongs: do you mean the ternary? I don't think that's the case because this line isn't inside of an `if` block which will cause the `.swDest` type to be narrowed.

Also, and this is a bit funky, but the narrowing with `isString` takes it from `swDest?: string | undefined` to `swDest?: string` (i.e. it's still effectively `string | undefined`) so we don't make too much headway there :/

src/compiler/config/validate-service-worker.ts
```diff
@@ -17,18 +37,18 @@ export const validateServiceWorker = (config: d.ValidatedConfig, outputTarget: d
...
   }
 
-  if (typeof outputTarget.serviceWorker !== 'object') {
-    // what was passed in could have been a boolean
-    // in that case let's just turn it into an empty obj so Object.assign doesn't crash
-    outputTarget.serviceWorker = {};
-  }
+  outputTarget.serviceWorker = {
+    ...outputTarget.serviceWorker,
+    globDirectory: outputTarget.serviceWorker?.globDirectory ?? outputTarget.appDir,
```
rwaskiewicz, line 47: Since we're validating user input here, it's probably best we keep the `typeof` check on `globDirectory` (which we could refactor to `isString()`) that [we originally had](https://github.com/ionic-team/stencil/pull/3813/files#diff-c2b7ab30b2bdcc180082ae16ba7b8aa1da392a53ea5054cea9bfb506b98d3fe0L41). That way we don't allow something like a numeric literal or object literal to be used for the value of `globDirectory`

src/declarations/stencil-public-compiler.ts
```diff
@@ -2192,30 +2192,268 @@ export type OutputTarget =
...
+ */
 export interface ServiceWorkerConfig {
-  // https://developers.google.com/web/tools/workbox/modules/workbox-build#full_generatesw_config
+  swSrc?: string;
   unregister?: boolean;
 
+  /**
+   * The local directory you wish to match `globPatterns` against. The path is relative to the current directory.
+   */
+  globDirectory?: string;
```
rwaskiewicz, line 2212: It looks like we [updated this type from `string | string[]` ](https://github.com/ionic-team/stencil/pull/3813/files#diff-c396c2ecfcbdf4562ae518122990719c12b5ed9bd64908ef2142620f6d1440dbL2202)to `string` - this should probably be considered a breaking change since we're changing the public interface of this type.

src/declarations/stencil-public-compiler.ts
```diff
@@ -2192,30 +2192,268 @@ export type OutputTarget =
...
   swDest?: string;
-  swSrc?: string;
-  globPatterns?: string[];
-  globDirectory?: string | string[];
+
+  /**
+   * A list of entries to be precached, in addition to any entries that are
+   * generated as part of the build configuration.
+   */
+  additionalManifestEntries?: any[] | undefined;
```
rwaskiewicz, line 2224: I don't see this in the diff prior to this commit, nor mention of adding new fields in the PR summary. 

In [this comment](https://github.com/ionic-team/stencil/pull/3813#discussion_r1019400967), you mentioned:

> as I note here we can't at present depend on these types because you need to have the `webworker` lib enabled (and that conflicts with the `dom` lib so both can't be set at the same time 😢).

> If we have project references enabled at some time in the future, allowing parts of the codebase to be compiled as separate units, I think we'd be able to carve off the service-worker related code into it's own project and compile it against that lib instead, but for now we'll have to copy-paste this :/

But I think I'm missing a piece of the puzzle here. What necessitates updating these types? Do they come from the latest version of Workbox? If not, where does this field and these new fields come from?

- babelPresetEnvTargets
- cleanupOutdatedCaches
- globFollow
- globStrict
- importScripts
- inlineWorkboxRuntime
- mode
- navigationPreload
- offlineGoogleAnalytics
- sourcemap

src/declarations/stencil-public-compiler.ts
```diff
@@ -2192,30 +2192,268 @@ export type OutputTarget =
...
+
+  /**
+   * Any search parameter names that match against one of the RegExp in this array
+   * will be removed before looking for a precache match. This is useful if your
+   * users might request URLs that contain, for example, URL parameters used to
+   * track the source of the traffic.
+   *
+   * @default [/^utm_/]
+   */
+  ignoreURLParametersMatching?: RegExp[] | undefined;
```
rwaskiewicz, line 2322: Similar to my comment above, it looks like we narrow this type from `any[]` to `RegExp[]`, which we might want to consider a breaking change as well

src/declarations/stencil-public-compiler.ts
```diff
@@ -2192,30 +2192,268 @@ export type OutputTarget =
...
+   * @default false
+   */
+  inlineWorkboxRuntime?: boolean | undefined;
+
+  /**
+   * One or more functions which will be applied sequentially against the
+   * generated manifest. If `modifyURLPrefix` or `dontCacheBustURLsMatching` are
+   * also specified, their corresponding transformations will be applied first.
+   */
+  manifestTransforms?: any[] | undefined;
```
rwaskiewicz, line 2347: Similar to my comment above, it looks like we narrow this type from `any` to `any[]`, which we might want to consider a breaking change as well

src/declarations/stencil-public-compiler.ts
```diff
@@ -2192,30 +2192,268 @@ export type OutputTarget =
...
+
+  /**
+   * A mapping of prefixes that, if present in an entry in the precache manifest,
+   * will be replaced with the corresponding value. This can be used to, for example,
+   * remove or add a path prefix from a manifest entry if your web hosting setup
+   * doesn't match your local filesystem setup. As an alternative with more flexibility,
+   * you can use the `manifestTransforms` option and provide a function that modifies
+   * the entries in the manifest using whatever logic you provide.
+   */
+  modifyURLPrefix?:
```
rwaskiewicz, line 2373: Similar to my comment above, it looks like we narrow this type from `any` to `{ [key: string]: string; }`, which we might want to consider a breaking change as well

src/declarations/stencil-public-compiler.ts
```diff
@@ -2192,30 +2192,268 @@ export type OutputTarget =
...
+
+  /**
+   * An optional array of regular expressions that restricts which URLs the configured
+   * `navigateFallback` behavior applies to. This is useful if only a subset of
+   * your site's URLs should be treated as being part of a
+   * [Single Page App](https://en.wikipedia.org/wiki/Single-page_application). If
+   * both `navigateFallbackDenylist` and `navigateFallbackAllowlist` are
+   * configured, the denylist takes precedent.
+   */
+  navigateFallbackDenylist?: RegExp[] | undefined;
```
rwaskiewicz, line 2396: I _think_ this was previously `navigateFallbackBlacklist`, is that the case?

src/declarations/stencil-public-compiler.ts
```diff
@@ -2192,30 +2192,268 @@ export type OutputTarget =
...
+
+  /**
+   * An optional array of regular expressions that restricts which URLs the configured
+   * `navigateFallback` behavior applies to. This is useful if only a subset of
+   * your site's URLs should be treated as being part of a
+   * [Single Page App](https://en.wikipedia.org/wiki/Single-page_application). If
+   * both `navigateFallbackDenylist` and `navigateFallbackAllowlist` are
+   * configured, the denylist takes precedent.
+   */
+  navigateFallbackAllowlist?: RegExp[] | undefined;
```
rwaskiewicz, line 2406: I _think_ this was previously `navigateFallbackWhitelist`, is that the case?

src/declarations/stencil-public-compiler.ts
```diff
@@ -2192,30 +2192,268 @@ export type OutputTarget =
...
-  navigateFallback?: string;
-  navigateFallbackWhitelist?: RegExp[];
-  navigateFallbackBlacklist?: RegExp[];
-  cacheId?: string;
-  skipWaiting?: boolean;
-  clientsClaim?: boolean;
-  directoryIndex?: string;
-  runtimeCaching?: any[];
-  ignoreUrlParametersMatching?: any[];
-  handleFetch?: boolean;
```
rwaskiewicz, line 2218: This property isn't included in the updated interface. Do we still need it?

rwaskiewicz (changes requested):
Most of my comments/questions revolve around the typings. I think I'm missing a piece of the puzzle here, which I explained a bit more in a comment on the review (I can't link a pending comment 😬)

src/declarations/stencil-public-compiler.ts
alicewriteswrongs: yeah the type is just copy-pasted out of workbox. Unfortunately we can't depend on their type declarations directly. Since they are dealing with service worker stuff they depend on having the service worker lib enabled in TypeScript and they therefore won't compiled without our current project setup, since you can't have the 'dom' and service worker libs enabled at the same time (they fight over the typings for a few global names). So I figured the best course is to at least have a more updated interface!

I'll go through and just keep whatever we were 'custom typing' things as before when there's a conflict.

alicewriteswrongs: Ok! So I looked at this a bit again and decided to just back out this part of the change.

For a little context, what I tried to do was change our setup to use the types for workbox which are published on DefinitelyTyped, I wasn't able to do that because they depend on the `webworker` lib being set and so they won't compile in our project, _but_ in trying to do that I noticed that our config was different than what's published, so I thought we'd want to pull that interface in somehow. I think for now it will be easier to just keep using this interface we maintain though, and we can revisit sometime in the future.

src/compiler/config/validate-service-worker.ts
```diff
@@ -17,18 +37,18 @@ export const validateServiceWorker = (config: d.ValidatedConfig, outputTarget: d
...
+    swDest: isString(outputTarget.serviceWorker?.swDest)
+      ? outputTarget.serviceWorker?.swDest
```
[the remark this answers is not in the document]
alicewriteswrongs: looks like we don't need it! Just removed it

alicewriteswrongs:
should be ready for another look!

src/compiler/config/validate-service-worker.ts
tanner-reits: Hmm yeah that's a bit confusing. I was thinking the type of `swDest` was narrowed to explicitly a string in lines 53-55 so we would only need to check if that path was absolute or not.

src/compiler/config/validate-service-worker.ts
```diff
@@ -54,16 +77,19 @@ export const validateServiceWorker = (config: d.ValidatedConfig, outputTarget: d
     outputTarget.serviceWorker.swSrc = join(config.rootDir, outputTarget.serviceWorker.swSrc);
   }
 
-  if (!isString(outputTarget.serviceWorker.swDest)) {
-    outputTarget.serviceWorker.swDest = join(outputTarget.appDir, DEFAULT_FILENAME);
```
rwaskiewicz, line 58: Is there a reason we're able to remove the `DEFAULT_FILENAME` assignment? I'm not sure I understand how/why we're able to combine these two `if` statements into one

alicewriteswrongs: The [change on lines 53-55](https://github.com/ionic-team/stencil/pull/3813/files#diff-c2b7ab30b2bdcc180082ae16ba7b8aa1da392a53ea5054cea9bfb506b98d3fe0R53-R55) covers part of what we were doing here:

```ts
    swDest: isString(outputTarget.serviceWorker?.swDest)
      ? outputTarget.serviceWorker.swDest
      : join(outputTarget.appDir ?? '', DEFAULT_FILENAME),
```

I wanted to move it up in the file so that earlier in the function we were sort of declaring what a good default value for this looks like. As I see it, there's two things we want to do w/r/t the `swDest`:

1. ensure that if no `swDest` is set we provide a known-good default value (`join(outputTarget.appDir ?? '', DEFAULT_FILENAME`)
2. ensure that if a value is set for `swDest` that it is an absolute path, and if not adjust it to be so

it was a while since I opened the PR so tbh I don't remember exactly my chain of reasoning but I think I wanted to separate no. 1 from no. 2, putting no. 1 higher in the function body in one spot where we set default values if there is no input provided (we do that for `globDirectory` too).


Also the diff is presented slightly wrong here (on github) - what I actually did was delete this code:

https://github.com/ionic-team/stencil/blob/65444226303bf2ccb678576b4fb57b4df35e0a18/src/compiler/config/validate-service-worker.ts#L57-L59

and move it up top and then add an `isString` check to this bit:

https://github.com/ionic-team/stencil/blob/65444226303bf2ccb678576b4fb57b4df35e0a18/src/compiler/config/validate-service-worker.ts#L61-L63

which was just to fix a `strictNullChecks` error since the type of `.swDest` is `string | undefined`, so I wanted to add the `isString` typeguard because `isAbsolute` takes `string`.

tanner-reits (approved):
lgtm based on some of the conversation

src/compiler/config/validate-service-worker.ts
```diff
@@ -22,11 +42,18 @@ export const validateServiceWorker = (config: d.ValidatedConfig, outputTarget: d
     return;
   }
 
-  if (typeof outputTarget.serviceWorker !== 'object') {
-    // what was passed in could have been a boolean
-    // in that case let's just turn it into an empty obj so Object.assign doesn't crash
-    outputTarget.serviceWorker = {};
-  }
+  const globDirectory =
+    typeof outputTarget.serviceWorker?.globDirectory === 'string'
```
rwaskiewicz, line 46: Do we have a guarantee here that `outputTarget.serviceWorker` is an object here? I don't believe it's part of `ValidatedConfig`, and (TBH) I'm not sure if the comment in the removed `if` statement holds true or not

rwaskiewicz (approved):
One non-blocking comment. Otherwise, LGTM. If we need to make a code change based on my comment, I'm OK with merging w/o me taking another look

src/compiler/config/validate-service-worker.ts
alicewriteswrongs: we don't have such a guarantee, but I think the optional chaining here takes care of that - if `.serviceWorker` returns undefined then iirc it will short-circuit the `?.globDirectory` and the whole expression (`outputTarget.serviceWorker?.globDirectory`) will evaluate to `undefined`

alicewriteswrongs: also this test will test this case, where `.serviceWorker` is `undefined` and `globDirectory` should then be set to `appDir`:

https://github.com/ionic-team/stencil/blob/cac647b67cb655477646fda072c5693c9df7f159/src/compiler/config/test/validate-service-worker.spec.ts#L81-L88

PR #3818, opened 2022-11-11
chore(compiler): move JSDoc to function definitions in transpile.ts

alicewriteswrongs:
This moves the JSDocs for the `transpile` and `transpileSync` functions from `public.ts` to the file where the function definitions actually are. Those functions are then re-exported from the `public.ts` file to make sure anything importing from that file will continue to work correctly.

## Pull request checklist

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

The JSDoc comments describing how these functions work aren't collocated with the function declaration.

## What is the new behavior?

Basically just moving some code from one place to another, there shouldn't really be any change in how things work.

- [ ] Yes
- [x] No

### DIFF
--- a/src/compiler/transpile.ts
+++ b/src/compiler/transpile.ts
@@ -1,7 +1,7 @@
 import { rollupPluginUtils } from '@compiler-deps';
 import { catchError, getInlineSourceMappingUrlLinker, isString } from '@utils';
 
-import { Config, TransformCssToEsmInput, TransformOptions, TranspileOptions, TranspileResults } from '../declarations';
+import { Config, TransformCssToEsmInput, TransformOptions, TranspileOptions, TranspileResults } from '@stencil/core/internal'
 import { getTranspileConfig, getTranspileCssConfig, getTranspileResults } from './config/transpile-options';
 import { transformCssToEsm, transformCssToEsmSync } from './style/css-to-esm';
 import { patchTypescript } from './sys/typescript/typescript-sys';
--- a/src/compiler/transpile.ts
+++ b/src/compiler/transpile.ts
@@ -1,7 +1,13 @@
 import { rollupPluginUtils } from '@compiler-deps';
+import type {
+  Config,
+  TransformCssToEsmInput,
+  TransformOptions,
+  TranspileOptions,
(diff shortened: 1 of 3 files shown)


src/compiler/transpile.ts
```diff
@@ -1,14 +1,40 @@
 import { rollupPluginUtils } from '@compiler-deps';
+import {
+  Config,
+  TransformCssToEsmInput,
+  TransformOptions,
+  TranspileOptions,
+  TranspileResults,
+} from '@stencil/core/internal';
```
tanner-reits, line 8: Is changing this import what is causing the the need to update `scripts/bundles/compiler.ts`?

alicewriteswrongs: not exactly, the change is necessary because now the generated `d.ts` file for `src/compiler/index.ts` depends on a type exported from `src/compiler/transpile.ts` (so in terms of build output `build/compiler/index.d.ts` depends on `build/compiler/transpile.d.ts`)

alicewriteswrongs: the underlying issue is that typescript doesn't provide any functionality for bundling together `.d.ts` files and I haven't found a good solution for doing that automatically yet, so until and if we find a good way to do that when we add a type export from a module like this we'll have to make a change in the build scripts to copy the `.d.ts` over as well.

src/compiler/transpile.ts
```diff
@@ -1,14 +1,40 @@
 import { rollupPluginUtils } from '@compiler-deps';
+import {
```
rwaskiewicz, line 2: Would this also work?
```suggestion
import type {
```

rwaskiewicz (commented):
LGTM, one small comment/question?

src/compiler/transpile.ts
alicewriteswrongs: I think so!

alicewriteswrongs: made the change, but I accidentally squashed it into the most recent commit on the branch

rwaskiewicz (approved):
All good 👍

Issue #3788, opened 2022-11-02
bug: generated types does not transform import alias paths to relative paths


AaronDDM:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/.github/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

2.19.2

### Current Behavior

Running the `build` command and generating `Types` for components does not replace `alias import` paths with relative paths.

Given:
```json
{
  "compilerOptions": {
    "baseUrl": "./",
    "paths": {
      "@interfaces/*": ["./src/interfaces/*"]
    }
  },
}

```

And
```ts
// components/my-component/my-component.tsx
import { MyInterface } from "@interfaces/my-interfaces";
// ...
```

And
```ts
// stencil.config.ts
import { Config } from '@stencil/core';

// https://stenciljs.com/docs/config

export const config: Config = {
  namespace: 'my-components',
  globalStyle: 'src/global/app.css',
  globalScript: 'src/global/app.ts',
  taskQueue: 'async',
  outputTargets: [
    {
      type: 'www',
      // comment the following line to disable service workers in production
      serviceWorker: null,
      baseUrl: 'https://myapp.local/',
    },
    {
      type: "dist",
    }
  ],
};
```

Then output for `dist/types/components/my-component/my-component.d.ts` is:
```ts
// dist/types/components/my-component/my-component.d.ts
import { MyInterface } from '@interfaces/my-interface';
export declare class MyComponent {
  myProperty: MyInterface;
  render(): any;
}
```

### Expected Behavior

The generated `dist/types/*` files should transform aliased import paths automatically to relative paths, otherwise the generates types are invalid.

Given:
```json
{
  "compilerOptions": {
    "baseUrl": "./",
    "paths": {
      "@interfaces/*": ["./src/interfaces/*"]
    }
  },
}

```

And
```ts
// components/my-component/my-component.tsx
import { MyInterface } from "@interfaces/my-interfaces";
// ...
```

And
```ts
// stencil.config.ts
import { Config } from '@stencil/core';

// https://stenciljs.com/docs/config

export const config: Config = {
  namespace: 'my-components',
  globalStyle: 'src/global/app.css',
[...]

PR #3819, opened 2022-11-11
refactor(build): use 'paths' set in TS config in alias plugin

alicewriteswrongs:
This refactors our 'alias' rollup plugin to determine what paths to resolve things to based on the settings we have in `tsconfig.json`.

This is relevant for the mappings that we have set in `paths`, where we map things like `@utils` to `src/utils/index.ts`. TypeScript, because reasons, does not provide an option to automatically resolve those paths to normal relative imports when compiling.

Our solution to this at present is to maintain a list of modules which should be rewritten in this `alias` plugin. However, the information about which module names map to which files is already set in our `tsconfig.json`! So it would be nicer if we didn't have to maintain that information in two places.

This change modifies the `alias` plugin so that it looks up module names in the `paths` object in `tsconfig` and then uses TypeScript's module resolution helper function to ensure it has the right path for the module, and then re-writes the path to point to the right compiled file in the built output.

Note that (unfortunately) this change cannot do the same thing for our `.d.ts` files :( We'll need to figure out another solution to hopefully copy those automatically and re-write import paths within them to point to the right places.

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

GitHub Issue Number: N/A

## What is the new behavior?

-
-
-

- [ ] Yes
- [x] No

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -30,6 +30,7 @@
       "@app-globals": ["src/app-globals/index.ts"],
       "@compiler-deps": ["src/compiler/sys/modules/compiler-deps.ts"],
       "@dev-server-process": ["src/dev-server/server-process.ts"],
+      "@environment": ["src/compiler/sys/environment.ts"],
       "@hydrate-factory": ["src/hydrate/runner/hydrate-factory.ts"],
       "@platform": ["src/client/index.ts"],
       "@runtime": ["src/runtime/index.ts"],
--- a/scripts/bundles/plugins/alias-plugin.ts
+++ b/scripts/bundles/plugins/alias-plugin.ts
@@ -1,5 +1,6 @@
 import { join } from 'path';
 import type { PartialResolvedId, Plugin } from 'rollup';
+import { CompilerOptions, nodeModuleNameResolver, parseConfigFileTextToJson, sys } from 'typescript';
 
 import type { BuildOptions } from '../../utils/options';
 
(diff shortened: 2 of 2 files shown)


tsconfig.json
```diff
@@ -30,6 +30,7 @@
       "@app-globals": ["src/app-globals/index.ts"],
       "@compiler-deps": ["src/compiler/sys/modules/compiler-deps.ts"],
       "@dev-server-process": ["src/dev-server/server-process.ts"],
+      "@environment": ["src/compiler/sys/environment.ts"],
```
alicewriteswrongs, line 33: I believe that we programmatically insert an import to this module into TypeScript during the rollup build (so an import from `@environment` doesn't appear anywhere in our code) but if we want it to resolve correctly the easiest / best thing to do is just add it here to our `paths`. We could also then start using this alias throughout our code if we want to.

alicewriteswrongs (commented):
just noting one funny bit

tanner-reits (approved):
This makes sense to me. Some of the existing code for `empty` and `alias` is a bit strange, but if it's been working 🤷‍♂️ These path alias type things keep coming up. I wonder if we'll be able to abstract some of the code between this, the change for `dist-collection` a while back and, #3788

alicewriteswrongs:
yeah, possibly this plugin could be used both to build stencil and by stencil?

rwaskiewicz:
@alicewriteswrongs is this ready to be reviewed/assigned?

alicewriteswrongs:
yes, sorry for not assigning!

alicewriteswrongs:
Just an update on this, I do want to do _something_ along these lines to clear up the `paths`-based module identifiers in the Stencil build, and while I still think this is a better way to do it than what we have in `main` right now I don't think it is \~the way\~, so while I would be fine with merging this for now if the team agrees we could also just stay tuned for further developments for now

PR #3820, opened 2022-11-14
chore(github): add environment details to issue template

PengBoUESTC:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Other (please describe):

GitHub Issue Number: N/A

## What is the new behavior?

add systeminfo input textarea to issus template
-
-
-

- [ ] Yes
- [x] No


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -31,6 +31,12 @@ body:
       description: A clear description of what you expected to happen.
     validations:
       required: true
+  - type: textarea
+    attributes:
+      label: System Info
+      description: Output of `npx envinfo --system --npmPackages @stencil/core --binaries --browsers`
+      render: shell
```
rwaskiewicz, line 38: Instead of using `envinfo`, can we ask folks to just provide what they think is the most important info?
```suggestion
      description: Operating system & version, npm, node, and browser(s) & version(s) (if applicable)
```

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -31,6 +31,12 @@ body:
...
+      placeholder: System, Binaries, Browsers
```
rwaskiewicz, line 39: "Binaries" might be a little too specific. Maybe "Environment"?
```suggestion
      placeholder: System, Environment, Browsers
```

rwaskiewicz (changes requested):
Thanks @PengBoUESTC!

I like this idea, and think there's a good way for us to incorporate this work into the Issues Template. 

I'm hesitant to ask folks to run a command that invokes a script that the team doesn't have control of. I don't have any guarantees the package works as intended on every machine, nor do we have any safeguards in the event that `envinfo` is compromised from a security standpoint.

Similarly, I think this might pull in a little more data than we may actually want when trying to debug an issue. For instance, a bug may occur in one specific browser, but this tool might report all browsers that are installed. In cases like this, it's best that bug reporters are as specific as possible.

Take a look at my feedback, let me know what you think. Thanks!

PengBoUESTC:
> Thanks @PengBoUESTC!
> 
> I like this idea, and think there's a good way for us to incorporate this work into the Issues Template.
> 
> I'm hesitant to ask folks to run a command that invokes a script that the team doesn't have control of. I don't have any guarantees the package works as intended on every machine, nor do we have any safeguards in the event that `envinfo` is compromised from a security standpoint.
> 
> Similarly, I think this might pull in a little more data than we may actually want when trying to debug an issue. For instance, a bug may occur in one specific browser, but this tool might report all browsers that are installed. In cases like this, it's best that bug reporters are as specific as possible.
> 
> Take a look at my feedback, let me know what you think. Thanks!

this command is used to remind coder that he/she can get the system info by running this command, and user can select the key information directly to fill in.

and you are right not all of the machines work well  with this command  (https://github.com/tabrindle/envinfo/issues/232), but as a suggest maybe it's not bad

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -31,6 +31,14 @@ body:
       description: A clear description of what you expected to happen.
     validations:
       required: true
+  - type: textarea
+    attributes:
+      label: System Info
+      description: |
+        Operating system & version, npm, node, and browser(s) & version(s) (if applicable), 
+        Maybe you can try to run `npx envinfo --system --npmPackages @stencil/core --binaries --browsers` to get the necessary information
```
rwaskiewicz, line 39: Can we remove this second line please? Although it's only a suggestion, I'm not comfortable asking folks to run something like this that may or may not work on their machine.
```suggestion
        Operating system & version, npm, node, and browser(s) & version(s) (if applicable)
```

rwaskiewicz (changes requested):

.github/ISSUE_TEMPLATE/bug_report.yml
PengBoUESTC: OK

rwaskiewicz (approved):
Thanks! Looks good to me!

Issue #3825, opened 2022-11-16
bug: CLI args of test names not propagated to Jest


duhem-s:
**Stencil version:**
```
 @stencil/core@2.19.x
```

**Current behavior:**
I used to do `npm run test:unit -- mg-form` to only test my `mg-form` component but since the version 2.19.x I'm not able to do so, all my unit tests are launched.

My `test:unit` script run `stencil test --spec`.

**Expected behavior:**
Be able to run a specific test file.

**GitHub Reproduction Link:**
https://github.com/duhem-s/stencil-cli

**Other information:**

With the command: `npm run test -- second` on the GH reproduction :

With stencil@2.18.1
![image](https://user-images.githubusercontent.com/113420581/202201370-84f637b0-cb4d-4ed8-9259-055130791ed4.png)

With stencil@2.19.0
![image](https://user-images.githubusercontent.com/113420581/202202127-1646821d-f3bf-4028-97d8-c1cd1502a9c6.png)

alicewriteswrongs:
Is it possible this is a duplicate of #3785?

alicewriteswrongs:
I just confirmed that this is the same issue as #3785, I have a fix which fixes the problem in both the reproduction for that issue and this one. I'm going to open a PR for that shortly, but I'm going to close this issue for bookkeeping purposes.

duhem-s:
Hi @alicewriteswrongs 👋

I opened a new one https://github.com/ionic-team/stencil/issues/3828 do you think your PR will also fix this ?

alicewriteswrongs:
Unfortunately that does look like a separate issue - at present we have a manually-defined set of short aliases like that which we support, and `-u` is, sadly, not among them at present 😢

alicewriteswrongs:
Ok so I just realized, belatedly, that #3785 is _not_ the same issue as this one, so I'm going to re-open this issue so we can keep track of making sure that our next release fixes it.

Sorry for the noise!

rwaskiewicz:
@duhem-s 👋 

I was able to verify that as of [Stencil v2.20.0, ](https://github.com/ionic-team/stencil/releases/tag/v2.20.0) this issue has been fixed using your reproduction case:
![Screenshot 2022-12-05 at 1 02 03 PM](https://user-images.githubusercontent.com/1930213/205709252-1f0e46f3-7d33-4846-b30f-7e885c7880f1.png)

[...]

Issue #3785, opened 2022-11-01
bug: Unable to run Jest spec files from subdirectory or absolute path


sean-perkins:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/.github/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

v2.19.2

### Current Behavior

When running `stencil test --spec` in Ionic Framework (`main`), we are only able to run tests for the entire `src/utils` directory with the following command:

```
stencil test --spec --max-workers=2 src/utils/
```

If we provide an additional sub directory or absolute path to our spec file, we receive the following error:

```
  ● Test suite failed to run

    TypeError: virtualMocks.get is not a function

      at Resolver._getVirtualMockPath (node_modules/@jest/core/node_modules/jest-resolve/build/resolver.js:468:25)
```

This occurs with any of the following:

```
stencil test --spec --max-workers=2 src/utils/keyboard/

stencil test --spec --max-workers=2 src/utils/keyboard/test/keyboard.spec.ts
```

The test runs and passes correctly if we run from `src/utils`. 

### Expected Behavior

Developers should be able to run the `stencil test --spec` command with any valid directory or absolute path and the spec files should execute consistently; regardless of if ran exclusively or as part of a larger directory.

### Steps to Reproduce

1. Clone Ionic Framework
2. Checkout `main` branch
3. Run `npm run test.spec src/utils`
4. Observe: Tests run successfully
5. Run `npm run test.spec src/utils/keyboard`
6. Observe: Tests fail with the error mentioned above

### Code Reproduction URL

https://github.com/ionic-team/ionic-framework

rwaskiewicz:
`npx stencil test --spec --max-workers=2 src/utils/keyboard` works, but `npm run test.spec -- src/utils/keyboard` and `npm run test.spec src/utils/keyboard` do not 🤔

alicewriteswrongs:
found a few issues which might be related: https://github.com/playwright-community/jest-playwright/issues/665 and https://github.com/facebook/jest/pull/11267#issuecomment-834101929

Unfortunately upgrading Jest in framework doesn't fix it :/

alicewriteswrongs:
[...]

PR #3827, opened 2022-11-16
fix(cli): ensure that argument order is correct for Jest

alicewriteswrongs:
This ensures that CLI arguments are ordered correctly when we pass them to Jest. In particular, we want to ensure that all flags of the type `--maxWorkers=0` and the like are _before_ any flags which we don't know about, like `my-spec-file.ts` (the latter being actually a match pattern for which spec files to run).

In other words, if the user passes arguments like this:

```
--e2e foo.spec.ts
```

or

```
--e2e -- foo.spec.ts
```

We need to ensure we use something like

```
["--e2e", "foo.spec.ts"]
```

to generate the Jest config.

The wrinkle is that after we've parsed our known CLI arguments (in the `cli` module) we do some addition modification of the arguments, based on values set in the `Config`, in the `buildJestArgv` function. In particular, we'll look to see if a `maxWorkers` flag is already passed and, if not, we add one to the args before they're used to build the Jest configuration.

Before this change that would result in an array like this being passed to Jest:

```
["--e2e", "foo.spec.ts", "--maxWorkers=0"]
```

because we simply stuck the new `--maxWorkers` arg right on the end of the already-combined array (combined because it's basically `knownArgs + unknownArgs`). No good!

Instead, we modify a copy of `knownArgs` with flags like this and then produce a new, combined array to pass to Jest when we're all done, so it looks like this instead:

```
["--e2e", "--maxWorkers=0", "foo.spec.ts"]
```

Much better!

## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

In certain cases passing a filename pattern to `stencil test` will not cause the right test files to run. Boo!

## What is the new behavior?

The bug is fixed by re-ordering arguments before they're using to generate a Jest configuration.

- [ ] Yes
- [x] No

## Testing

There are reproductions linked in both #3785 and #3825. I confirmed that the fix addressed both of those issues, and also added regression tests which fail without the fix.

### DIFF
--- a/src/testing/jest/jest-config.ts
+++ b/src/testing/jest/jest-config.ts
@@ -46,15 +46,27 @@ function getLegacyJestOptions(): Record<string, boolean | number | string> {
 export function buildJestArgv(config: d.ValidatedConfig): Config.Argv {
   const yargs = require('yargs');
 
-  const args = [...config.flags.knownArgs.slice(), ...config.flags.unknownArgs.slice()];
+  const knownArgs = config.flags.knownArgs.slice();
 
-  if (!args.some((a) => a.startsWith('--max-workers') || a.startsWith('--maxWorkers'))) {
-    args.push(`--max-workers=${config.maxConcurrentWorkers}`);
+  if (!knownArgs.some((a) => a.startsWith('--max-workers') || a.startsWith('--maxWorkers'))) {
+    knownArgs.push(`--max-workers=${config.maxConcurrentWorkers}`);
   }
 
   if (config.flags.devtools) {
-    args.push('--runInBand');
+    knownArgs.push('--runInBand');
   }
+
+  // we combine the modified args and the unknown args here and declare the
+  // result read only, providing some typesystem-level assurance that we won't
+  // mutate it after this point.
+  //
+  // We want that assurance because Jest likes to have any filepath match
+  // patterns at the end of the args it receives. Those args are going to be
+  // found in our `unknownArgs`, so while we want to do some stuff in this
+  // function that adds to `knownArgs` we need a guarantee that all of the
+  // `unknownArgs` are _after_ all the `knownArgs` in the array we end up
+  // generating the Jest configuration from.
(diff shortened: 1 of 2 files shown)


rwaskiewicz (approved):

src/testing/jest/jest-config.ts
```diff
@@ -46,15 +46,27 @@ function getLegacyJestOptions(): Record<string, boolean | number | string> {
...
+  // result read only, providing some typesystem-level assurance that we won't
+  // mutate it after this point.
+  //
+  // We want that assurance because Jest likes to have any filepath match
+  // patterns at the end of the args it receives. Those args are going to be
+  // found in our `unknownArgs`, so while we want to do some stuff in this
+  // function that adds to `knownArgs` we need a guarantee that all of the
+  // `unknownArgs` are _after_ all the `knownArgs` in the array we end up
+  // generating the Jest configuration from.
+  const args: ReadonlyArray<string> = [...knownArgs, ...config.flags.unknownArgs];
```
tanner-reits, line 69: UnknownArgs doesn't get mutated anywhere, correct? I just noticed we created a copy before with `slice` but not doing that here. Probably okay, just erring on the side of caution

tanner-reits (changes requested):
One small thing to confirm. If it's not a problem, then this looks good!

src/testing/jest/jest-config.ts
tanner-reits: Nope you're good. Just making sure it was an intentional change!

tanner-reits (approved):

bamsargeant:
# Bug

## Details
Unable to run single e2e test after upgrading from _v2.16.1_ to _v2.19.3_

Command used: `npx stencil test --e2e dialog.e2e.ts`

## Work around
I am able to get the same functionality by adding `--` as an extra arg

Command used: `npx stencil test --e2e -- dialog.e2e.ts`

## Screenshots

### v2.16.1
image

### v2.19.3
image

### v2.19.3 with workaround
image

PR #3597, opened 2022-09-09
refactor(decorators): fix `strictNullCheck` violation

rwaskiewicz:
## Pull request checklist

Please check if your PR fulfills the following requirements:
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [x] Build (`npm run build`) was run locally and any changes were pushed
- [x] Unit tests (`npm test`) were run locally and passed
- [x] E2E Tests (`npm run test.karma.prod`) were run locally and passed
- [x] Prettier (`npm run prettier`) was run locally and passed

## Pull request type

 

 

Please check the type of change your PR introduces:
- [ ] Bugfix
- [ ] Feature
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

## What is the current behavior?

there's a single `strictNullCheck` violation in the file in this PR, which coincidentally has its code
duplicated elsewhere
GitHub Issue Number: N/A

## What is the new behavior?

this commit fixes a single `strictNullChecks` violation where the result of an expression could set a local variable to be of type `string | false`, where `false` was not an accepted value on `Set.prototype.has`. this change ought to be backwards compatible, as no `Set` provided to this function includes `false` (or any value that is coerced to `false`).

the function signature was updated to include a return type, and the function name was updated to provide a better description of what the function does.

the deprecated call to `createNodeArray` has been moved to the `ts.factory` for general code health.

during this exercise, it was discovered there was an unused instance of `removeDecorators` in the project's utilities. it has been removed as a part of this commit

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Testing

Smoke tested locally by building this branch (`npm clean && npm run build && npm pack`) into a locally spun up component lib (`npm init stencil component test-lib`). Build works as expected with the following diff applied:

```diff
--- a/src/components/my-component/my-component.tsx
+++ b/src/components/my-component/my-component.tsx
@@ -1,5 +1,9 @@
 import { Component, Prop, h } from '@stencil/core';
 import { format } from '../../utils/utils';
+const printMemberName = (_target: any, memberName: string) => {
+  console.log(memberName);
+};
+
 
 @Component({
   tag: 'my-component',
@@ -10,8 +14,8 @@ export class MyComponent {
   /**
    * The first name
    */
-  @Prop() first: string;
+  @Prop() @printMemberName first: string;
```

## Other information

[...]

PR #3835, opened 2022-11-21
refactor(filterDecorators): narrow first parameter 

rwaskiewicz:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

The `filterDecorators` fn is untested

GitHub Issue Number: N/A

## What is the new behavior?

this commit narrows the first parameter type of `filterDecorators` to
accept a list of decorator nodes, rather than a single node itself and
subsequently pulling the decorators off the node as the first argument.

this change has two connected motivating factors:
1. this function can return either a new array of decorators, or the
   existing one. for cases where referential equality is important (i.e.
   comparing if the array provided to this function is the same as the
   one returned), it will be beneficial to have a reference to that
   array before calling this function because:
2. in typescript 4.8, decorators will no longer have direct accessors on
   a node. instead, helper functions provided by the language will be
   used to retrieve decorators. these helper methods allocate a new
   array when returning an array of decorators, making referential
   equality checks impossible as they are today (calling the
   `decorators` getter directly today will return a ref to the same
   array)

this commit adds unit tests to the `filterDecorators` utility fn. this
is done in an effort to make this utility a more maintainable function,
by taking the changes to remove strict null checks violations in https://github.com/ionic-team/stencil/commit/611756be93e4cd7a945fba0068c8502a35a1a90a
(https://github.com/ionic-team/stencil/pull/3597) and improve its testing.

the function is made public as a result of this commit. although it
breaks encapsulation of the module, it allows us to easily get this
function under test for the typescript 4.8 upgrade + necessary
refactors. we accept that in favor of better testing/higher confidence
in the refactor.

- [ ] Yes
- [x] No

## Testing

The unit tests were written under the current implementation (prior to applying this patch). For their development, I stepped through the tests with the debugger to verify they were hitting the code areas that I intended to exercise. Then I refactored the code, and stepped through the tests again.

I also rolled this functionality in #3743, in order to use it to fix a bug that was introduced during the initial development of the feature.

## Other information

I intend to land this before #3743 

There are new strictNullChecks errors, those will be removed in #3743


tanner-reits (approved):
lgtm!

alicewriteswrongs (approved):
changes look good (more clarity) and good to add some tests specifically for this function 👍

PR #3863, opened 2022-12-01
feat(typescript): add support for typescript 4.9

rwaskiewicz:
## Pull request checklist

  - https://github.com/ionic-team/stencil-site/pull/939 

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

We support TS 4.7 (the 4.8 work has landed, just hasn't been released as of this writing)
GitHub Issue Number: N/A

## What is the new behavior?

this commit increments typescript to v4.9 for the project, and updates the dependabot configuration to not create a pull request for a version of typescript newer than 4.9.

the `@types/node` package for the karma test suite is updated as a part of this commit, as the currently installed version has the following error when being type checked by v4.9:

```
[ ERROR ]  TypeScript: ../node_modules/@types/node/globals.d.ts:114:13
           Subsequent variable declarations must have the same type. Variable
           'AbortSignal' must be of type '{ new (): AbortSignal; prototype:
           AbortSignal; abort(reason?: any): AbortSignal; timeout(milliseconds:
           number): AbortSignal; }', but here has type '{ new (): AbortSignal;
           prototype: AbortSignal; }'.
```

similarly `@types/node` were previously a child dependency for our bundler tests. it has been explictly
installed to prevent that very error.

the remainder of the changes in this pr are for fixing type errors where there is no overlap between two types being compared. this occurs in two scenarios:

  1. parsing css. in the css parser, there were three cases where an overlap did not occur in a comparison. in reading through the code, i cannot find a scenario where such a value could be returned such that the comparison is valid. as a result, such comparisons are removed.

     prior to removing these comparisons, an alternative (else block) was written with a logging statement in it. stencil's unit/karma tests were run, as well as the ionic framework's tests with this branch installed. the logging statement was never triggered. that in addition to the rich testing available for css, makes us confident that this change is safe to make.

  2. validation of a configuration flag. this occurred for the dev server, hydration config, and service worker config. these fields _should_ be properly set via type checking. if users are providing invalid values, we may want to reconsider some of the "silent failures" occurring and defaults we're setting in cases like this.

- [ ] Yes
- [x] No
 
Updates to TypeScript always introduce the possibility of upgrade pains. This may land and be released in a minor version of Stencil, in adherence to our [Versioning Policy](https://stenciljs.com/docs/versioning)

> A minor release will be published when a new feature is added or API changes that are non-breaking are introduced. We will heavily test any changes so that we are confident with the release, but with new code comes the potential for new issues. ...
> * This statement applies to the Stencil team upgrading its version of TypeScript as well. For more information, please see the team's [support policy regarding TypeScript](https://stenciljs.com/docs/support-policy#typescript-support)

and our [support policy regarding TypeScript](https://stenciljs.com/docs/support-policy#typescript-support)
> The Stencil team acknowledges that TypeScript minor version releases may contain breaking changes. The Stencil team will do everything in its power to avoid propagating breaking changes to its user base.

## Testing

This commit is best tested in a real project, such as the Ionic Framework. To do so, first build this branch:
```sh
npm run clean && \
npm ci && \
npm run build && \
npm pack
```
which ought to result in a tarball (which will be the last line of the output)

Next, pull down the Framework repository, install this branch in the core package, build and test:
```sh
cd core && \
npm i [TARBALL] --force && \
npm ci --force && \
npm run build && \
npm t
```

Next, we can look at components on an individual basis, by running a dev server from `core/`:
```sh
npm start
```

Navigate to `src/components` in the browser window that pops up. For each component dir (most of them) there's a `test` sub directory that can be entered to smoke test components

## Other information

For folks performing the upgrade and are seeing an error like:
```
[ ERROR ]  TypeScript: ../node_modules/@types/node/globals.d.ts:114:13
           Subsequent variable declarations must have the same type. Variable
           'AbortSignal' must be of type '{ new (): AbortSignal; prototype:
           AbortSignal; abort(reason?: any): AbortSignal; timeout(milliseconds:
           number): AbortSignal; }', but here has type '{ new (): AbortSignal;
           prototype: AbortSignal; }'.
```
Please try updating your `@types/node` version to the latest minor version for the version of Node you're developing against. If `@types/node` is not explicitly installed (e.g. isn't in the `dev-dependencies` section of `package.json`), you may have it as a sub-dependency. You can check this with `npm ls @types/node` to determine where the package is being pulled in. In this scenario, please install `@types/node` as an explicit (dev) dependency

### DIFF
--- a/src/compiler/config/validate-dev-server.ts
+++ b/src/compiler/config/validate-dev-server.ts
@@ -100,7 +100,7 @@ export const validateDevServer = (
     devServer.protocol = devServer.https ? 'https' : addressProtocol ? addressProtocol : 'http';
   }
 
-  if (devServer.historyApiFallback !== null && devServer.historyApiFallback !== false) {
+  if (devServer.historyApiFallback !== null) {
     devServer.historyApiFallback = devServer.historyApiFallback || {};
 
     if (!isString(devServer.historyApiFallback.index)) {
--- a/src/compiler/config/validate-dev-server.ts
+++ b/src/compiler/config/validate-dev-server.ts
@@ -101,7 +101,9 @@ export const validateDevServer = (
   }
 
   if (devServer.historyApiFallback !== null) {
-    devServer.historyApiFallback = devServer.historyApiFallback || {};
+    if (Array.isArray(devServer.historyApiFallback) || typeof devServer.historyApiFallback !== 'object') {
+      devServer.historyApiFallback = {};
+    }
 
     if (!isString(devServer.historyApiFallback.index)) {
       devServer.historyApiFallback.index = 'index.html';
--- a/src/compiler/config/validate-hydrated.ts
+++ b/src/compiler/config/validate-hydrated.ts
@@ -12,11 +12,12 @@ export const validateHydrated = (config: UnvalidatedConfig): HydratedFlag | unde
   /**
    * If `config.hydratedFlag` is set to `null` that is an explicit signal that we
    * should _not_ create a default configuration when validating and should instead
-   * just return `undefined`.
+   * just return `undefined`. It may also have been set to `false`; this is an invalid
+   * value as far as the type system is concerned, but users may ignore its.
    *
    * See {@link HydratedFlag} for more details.
    */
-  if (config.hydratedFlag === null || Array.isArray(config.hydratedFlag) || typeof config.hydratedFlag !== 'object') {
+  if (config.hydratedFlag === null || (config.hydratedFlag as unknown as boolean) === false) {
     return undefined;
   }
 
--- a/src/compiler/config/validate-hydrated.ts
+++ b/src/compiler/config/validate-hydrated.ts
@@ -16,7 +16,7 @@ export const validateHydrated = (config: UnvalidatedConfig): HydratedFlag | unde
    *
    * See {@link HydratedFlag} for more details.
    */
-  if (config.hydratedFlag === null || config.hydratedFlag === false) {
+  if (config.hydratedFlag === null) {
     return undefined;
   }
 
--- a/src/compiler/config/validate-hydrated.ts
+++ b/src/compiler/config/validate-hydrated.ts
@@ -13,7 +13,7 @@ export const validateHydrated = (config: UnvalidatedConfig): HydratedFlag | unde
    * If `config.hydratedFlag` is set to `null` that is an explicit signal that we
    * should _not_ create a default configuration when validating and should instead
    * just return `undefined`. It may also have been set to `false`; this is an invalid
-   * value as far as the type system is concerned, but users may ignore its.
+   * value as far as the type system is concerned, but users may ignore this.
    *
    * See {@link HydratedFlag} for more details.
    */
--- a/src/compiler/config/validate-hydrated.ts
+++ b/src/compiler/config/validate-hydrated.ts
@@ -16,7 +16,7 @@ export const validateHydrated = (config: UnvalidatedConfig): HydratedFlag | unde
    *
    * See {@link HydratedFlag} for more details.
    */
-  if (config.hydratedFlag === null) {
+  if (config.hydratedFlag === null || Array.isArray(config.hydratedFlag) || typeof config.hydratedFlag !== 'object') {
     return undefined;
   }
 
--- a/src/compiler/config/validate-service-worker.ts
+++ b/src/compiler/config/validate-service-worker.ts
@@ -17,9 +17,7 @@ export const validateServiceWorker = (config: d.ValidatedConfig, outputTarget: d
     return;
   }
 
-  if (outputTarget.serviceWorker === true) {
-    outputTarget.serviceWorker = {};
-  } else if (!outputTarget.serviceWorker && config.devMode) {
+  if (!outputTarget.serviceWorker && config.devMode) {
     outputTarget.serviceWorker = null;
     return;
   }
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -15,7 +15,7 @@ updates:
       - dependency-name: '@types/node'
         versions: ['17', '18']
       - dependency-name: 'typescript'
-        versions: ['>=4.9']
+        versions: ['>=4.10', '>=5.0']
       # disable Jest updates until the new testing architecture is in place
       - dependency-name: '@types/jest'
         versions: ['>=28']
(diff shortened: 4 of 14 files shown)


src/compiler/config/validate-dev-server.ts
```diff
@@ -100,7 +100,7 @@ export const validateDevServer = (
     devServer.protocol = devServer.https ? 'https' : addressProtocol ? addressProtocol : 'http';
   }
 
-  if (devServer.historyApiFallback !== null && devServer.historyApiFallback !== false) {
```
rwaskiewicz, line 103: Feel free to push back on this.

src/compiler/config/validate-hydrated.ts
```diff
@@ -16,7 +16,7 @@ export const validateHydrated = (config: UnvalidatedConfig): HydratedFlag | unde
    *
    * See {@link HydratedFlag} for more details.
    */
-  if (config.hydratedFlag === null || config.hydratedFlag === false) {
```
rwaskiewicz, line 19: Feel free to push back on this.

src/compiler/config/validate-service-worker.ts
```diff
@@ -17,9 +17,7 @@ export const validateServiceWorker = (config: d.ValidatedConfig, outputTarget: d
     return;
   }
 
-  if (outputTarget.serviceWorker === true) {
```
rwaskiewicz, line 20: Feel free to push back on this.

src/compiler/config/validate-hydrated.ts
tanner-reits: Same thoughts as above

src/compiler/config/validate-dev-server.ts
rwaskiewicz: This logic actually is the validation on the field 😉 

I agree, this may cause issues. However, we can't coerce `historyApiFallback` to a boolean with a double-bang operator, because that unfortunately changes the logic of the conditional as well; instead of  running the consequence of the if-statement whenever `historyApiFallback` is not `false` (prior to my proposed changed) to running the consequence of the if-statement whenever `historyApiFallback` is `true` - since this is an untyped field, that means that if `historyApiFallback === 1` for instance, we've changed the fallback logic.

In my view, it comes down to how many guard rails we want to put up for folks. With my proposed change, we take some down in the name of type safety. We could also do something like this:
```diff
- if (devServer.historyApiFallback !== null && devServer.historyApiFallback !== false) {
+ // this field could be anything, let's only act on it if and only if it's truthy
+ if (devServer.historyApiFallback !== null && (devServer.historyApiFallback as unknown as boolean) !== false) {
```
which keeps the logic exactly the same, but is still weird if `devServer.historyApiFallback === 1` as soon as we enter the consequence:
```ts
    // this is 1 (that doesn't seem right),
    devServer.historyApiFallback = devServer.historyApiFallback || {};

    if (!isString(devServer.historyApiFallback.index)) {
      devServer.historyApiFallback.index = 'index.html';
    }
```

I think there are a few subtle bugs in this logic, that's "break" a config if it's not typed properly. Perhaps the best thing to do is if `historyApiFallback` is not null, coerce it to an object literal if it's not one already:
```diff
   if (devServer.historyApiFallback !== null) {
-    devServer.historyApiFallback = devServer.historyApiFallback || {};
+    if (Array.isArray(devServer.historyApiFallback) || typeof devServer.historyApiFallback !== 'object') {
+      devServer.historyApiFallback = {};
+    }

```

rwaskiewicz: Thoughts?

src/compiler/config/validate-hydrated.ts
rwaskiewicz: Similar to above, perhaps the check should become:
```diff
    *
    * See {@link HydratedFlag} for more details.
    */
-  if (config.hydratedFlag === null) {
+  if (config.hydratedFlag === null || (Array.isArray(config.hydratedFlag) && typeof config.hydratedFlag !== 'object')) {
     return undefined;
   }
```
Thoughts?

src/compiler/config/validate-dev-server.ts
tanner-reits: I kinda like that. We can't protect against _everything_ a user might do in their config, but having some type checks are an easy thing we can do.

src/compiler/config/validate-hydrated.ts
tanner-reits: Same here, but should all of the checks be `||`?

tanner-reits (approved):
lgtm!

rwaskiewicz:
Hmmm I think one of those two most recent commits broke the Karma tests.. Investigating

rwaskiewicz:
@alicewriteswrongs @tanner-reits turns out in https://github.com/ionic-team/stencil/pull/3863/commits/7068a6bbd90e213cbdca144692570720c6af5690 I was removing 'known' functionality for Stencil. It looks like (based on the Karma tests), we need to keep the existing behavior for validating `[]` and `true` values for the `HydratedFlag` field. I've updated the code+unit tests appropriately to catch this sooner in the pipeline https://github.com/ionic-team/stencil/pull/3863/commits/dc0f3df642ac46ef59ecc0f3a180a310f418b51c

src/compiler/config/validate-hydrated.ts
```diff
@@ -12,11 +12,12 @@ export const validateHydrated = (config: UnvalidatedConfig): HydratedFlag | unde
   /**
    * If `config.hydratedFlag` is set to `null` that is an explicit signal that we
    * should _not_ create a default configuration when validating and should instead
-   * just return `undefined`.
+   * just return `undefined`. It may also have been set to `false`; this is an invalid
+   * value as far as the type system is concerned, but users may ignore its.
```
alicewriteswrongs, line 16: I think this is a little typo?

```suggestion
   * value as far as the type system is concerned, but users may ignore this.
```

alicewriteswrongs (approved):
I think this looks good! Just noticed a small typo in a comment

tanner-reits (approved):

src/compiler/config/validate-service-worker.ts
rwaskiewicz: You know, after the last round of fixes, I wonder if we should just go ahead and restore this in a similar fashion to https://github.com/ionic-team/stencil/pull/3863/files#diff-d9e941746821b7e6229ddee0bdbf0fca566b115a9e6b75bc4d3bcaa88c598443R20 - that's where I lean currently (and add tests!). Thoughts?

rwaskiewicz: Err NVM - we have tests for this already. This works 👍

PR #3864, opened 2022-12-01
chore(deps): upgrade rollup a bit

alicewriteswrongs:
This upgrades Rollup to the highest version which we can trivially run. It's not that high! With any version `^2.57` I am currently running into an issue with Jest which I'm still investigating.

For now, this at least gets us on a slightly newer version.

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Rollup is quite out of date!

## What is the new behavior?

Rollup is upgraded a bit. In order to get the upgrade to work we needed to make a few adjustments to our node module polyfills (adding methods to `"url"` for instance).

- [ ] Yes
- [x] No

## Testing

I think it's a good idea to try to build and `npm start` in Framework, possibly in a little test project too.


package.json
```diff
@@ -117,13 +118,14 @@
     "prettier": "2.8.0",
     "prompts": "2.4.2",
     "puppeteer": "~10.0.0",
-    "rollup": "2.42.3",
+    "rollup": "^2.56.3",
```
tanner-reits, line 121: Should we hard-lock this version?

src/compiler/sys/modules/url.ts
```diff
@@ -1 +1,6 @@
-export const URL = globalThis.URL as any;
+import { basicURLParse, serializePath, serializeURL } from 'whatwg-url';
```
tanner-reits, line 1: Some explanation of what's going on in this file and how it relates to Rollup would be super helpful.

tanner-reits (changes requested):

src/compiler/sys/modules/url.ts
alicewriteswrongs: Yes! I should probably have gone ahead and called this out.

So the files in this directory (`src/compiler/sys/modules/`) are basically polyfills for node.js modules that we maintain. We have a Rollup plugin called `sys-modules-plugin` which patches module resolution during the build of Stencil to intercept any requires of node.js built-in modules like `fs`, `os`, etc and change them to import our corresponding polyfill modules instead.

This will apply to imports of node modules which occur in any of our code, but crucially _also_ imports which occur in any of our dependencies. This is done so that Stencil can run in the browser, and if our polyfills correctly polyfill the node built-ins they correspond to then even Stencil's dependencies which are written for a node.js environment and have no browser support will work.

This actually works at present, but as you might imagine it is also a bit yucky.

The reason I had to make a change here in this file was that upgrading Rollup resulted in either Rollup or a dependency of Rollup trying to import and then call functions from the `url` module which we weren't polyfilling, so I had to add implementations here to fix the error.

Here is the plugin code, which might help illustrate some of how this all works:

[scripts/bundles/plugins/sys-modules-plugin.ts:L1-L31](https://github.com/ionic-team/stencil/blob/main/scripts/bundles/plugins/sys-modules-plugin.ts#L1-L31)
```ts
import path from 'path';
import type { Plugin } from 'rollup';

/**
 * Modules that are polyfilled by Stencil, rather than using the default Node implementation.
 */
const modules = new Set(['crypto', 'events', 'fs', 'module', 'os', 'path', 'stream', 'url', 'util']);

/**
 * Rollup plugin that aids in resolving various system level (sys) modules properly.
 * @param inputDir the directory from which modules should be resolved. Care should be taken to observe this value at
 * build to verify where modules are being resolved from.
 * @returns the rollup plugin that resolves Stencil `sys` modules
 */
export function sysModulesPlugin(inputDir: string): Plugin {
  return {
    name: 'sysModulesPlugin',
    /**
     * A rollup build hook for resolving Stencil's environment agnostic packages
     * [Source](https://rollupjs.org/guide/en/#resolveid)
     * @param importee the importee exactly as it is written in an import statement in the source code
     * @returns a string that resolves an import to some id, null otherwise
     */
    resolveId(importee: string): string | null {
      if (modules.has(importee)) {
        return path.join(inputDir, 'sys', 'modules', `${importee}.js`);
      }
      return null;
    },
  };
}
```

scripts/test/validate-build.ts
```diff
@@ -335,7 +335,7 @@ async function validateModuleTreeshake(opts: BuildOptions, moduleName: string, e
 
   const bundle = await rollup({
     input: virtualInputId,
-    treeshake: true,
+    treeshake: 'smallest',
```
alicewriteswrongs, line 338: A change was [introduced in rollup@2.52.0](https://github.com/rollup/rollup/blob/master/CHANGELOG.md#2520) which adds preset options that can be set here, see the [PR which introduced the options](https://github.com/rollup/rollup/pull/4131) for more on this.

Without this change I get a build error when doing `node scripts/build --validate-build`

alicewriteswrongs (commented):
just pointing something out

PR #3895, opened 2022-12-08
chore(config): mark `config.extras.dynamicImportShim` as deprecated

alicewriteswrongs:
This marks the config field as deprecated and also adds TODO comments for STENCIL-661, the ticket for removing the code itself.

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

We have support for a shim of `import()` which we'd like to remove at some point in the future.

## What is the new behavior?

The `dynamicImportShim` is now marked as deprecated, preparing the ground for us to remove it entirely later (and delete the code that implements the shim and so on).

- [x] Yes
- [ ] No

### DIFF
--- a/BREAKING_CHANGES.md
+++ b/BREAKING_CHANGES.md
@@ -14,7 +14,6 @@ This is a comprehensive list of the breaking changes introduced in the major ver
   * [New Configuration Defaults](#new-configuration-defaults)
     * [SourceMaps](#sourcemaps)
     * [`dist-custom-elements` Type Declarations](#dist-custom-elements-type-declarations)
-  * [Backwards-compat Configuration Fields Deprecated](#backwards-compat-configuration-fields-deprecated)
   * [Legacy Browser Support Fields Deprecated](#legacy-browser-support-fields-deprecated)
     * [`dynamicImportShim`](#dynamicimportshim)
   * [Deprecated `assetsDir` Removed from `@Component()` decorator](#deprecated-assetsdir-removed-from-component-decorator)
--- a/BREAKING_CHANGES.md
+++ b/BREAKING_CHANGES.md
@@ -12,8 +12,10 @@ This is a comprehensive list of the breaking changes introduced in the major ver
 
 * [General](#general)
   * [New Configuration Defaults](#new-configuration-defaults)
-  * [SourceMaps](#sourcemaps)
-  * [`dist-custom-elements` Type Declarations](#dist-custom-elements-type-declarations)
+    * [SourceMaps](#sourcemaps)
+    * [`dist-custom-elements` Type Declarations](#dist-custom-elements-type-declarations)
+  * [Backwards-compat Configuration Fields Deprecated](#backwards-compat-configuration-fields-deprecated)
+    * [dynamicImportShim](#dynamicimportshim)
   * [Deprecated `assetsDir` Removed from `@Component()` decorator](#deprecated-assetsdir-removed-from-component-decorator)
   * [Drop Node 12 Support](#drop-node-12-support)
   * [Strongly Typed Inputs](#strongly-typed-inputs)
@@ -66,6 +68,30 @@ export const config: Config = {
     // ...
   ],
   // ...
+};
+```
+
+#### Backwards-compat Configuration Fields Deprecated
+
+Several configuration options related to support for Safari <11, IE11, and Edge
+<19 have been marked as deprecated, and will be removed entirely in a future
+version of Stencil.
+
+##### dynamicImportShim
+
+The `extras.dynamicImportShim` option causes Stencil to include a polyfill for
+the `import()` function for use at runtime. The field is renamed to
+`__deprecated__dynamicImportShim` to indicate deprecation. To retain the
+prior behavior the new option can be set in your project's `stencil.config.ts`:
+
+```ts
+// stencil.config.ts
+import { Config } from '@stencil/core';
+
+export const config: Config = {
+  extras: {
+    __deprecated__dynamicImportShim: true
+  }
 };
 ```
 
--- a/BREAKING_CHANGES.md
+++ b/BREAKING_CHANGES.md
@@ -82,7 +82,7 @@ version of Stencil.
 
 The `extras.dynamicImportShim` option causes Stencil to include a polyfill for
 the [dynamic `import()`
-function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import)
+function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import)
 for use at runtime. The field is renamed to `__deprecated__dynamicImportShim`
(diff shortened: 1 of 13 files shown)


src/declarations/stencil-public-compiler.ts
```diff
@@ -273,11 +273,12 @@ export interface ConfigExtras {
    */
   cssVarsShim?: boolean;
 
+  // TODO(STENCIL-661): Remove code related to the dynamic import shim
   /**
    * Dynamic `import()` shim. This is only needed for Edge 18 and below, and Firefox 67
    * and below. Defaults to `false`.
    */
```
rwaskiewicz, line 280: Can we add a [`@deprecated` JSDoc tag](https://jsdoc.app/tags-deprecated.html) to the public compiler interface here? 
```suggestion
   * @deprecated Since Stencil v3.0.0. IE 11, Edge <= 18, and old Safari versions are no longer supported.
   */
```


That way Intellisense(s) can pick this up in editor (although the name should be _pretty_ self explanatory)


Screenshot 2022-12-09 at 9 25 43 AM

src/compiler/output-targets/dist-lazy/generate-esm-browser.ts
```diff
@@ -24,7 +24,8 @@ export const generateEsmBrowser = async (
       preferConst: true,
       sourcemap: config.sourceMap,
     };
-    if (config.extras.dynamicImportShim) {
+    // TODO(STENCIL-661): Remove code related to the dynamic import shim
+    if (config.extras.__deprecated__dynamicImportShim) {
```
rwaskiewicz, line 28: I _think_ that `src/utils/util.ts#getDynamicImportFunction` can be marked as deprecated too. You probably saw this as a part of your digging/tracing here, but can we add an explicit TODO (STENCIL-661) to remove it as well? It has 2 callers, here and in https://github.com/ionic-team/stencil/pull/3895/files#diff-5e031103fd3832f45d5667ca75259450d5e42d1fe24dfb7aabd0bd6c143e038fR88 (whose caller you also marked to remove) - I just don't want to forget about this, since there's gonna be a lot of moving pieces 😆

rwaskiewicz (changes requested):
All in all looks good to me. My comments on the code are super minor.

The bigger question I wanted to ask is how you want to handle landing pieces of deprecation - do you want to do this as putting these PRs (that deprecate a single option) into a feature branch, or merge them into `v3.0.0-dev` one at a time? I'm fine with either, but I think the strategy we choose here has ramifications on:
- docs - how many PRs we have (one per Stencil Core PR, or one big one)
- pr naming - this PR is named as `chore`, but I'd like to call it a `feat` to make sure it shows up in the changelog if we merge it individually. Alternatively, we can use the [breaking change](https://github.com/ionic-team/stencil/blob/main/.github/CONTRIBUTING.md?plain=1#LL189C34-L189C43) syntax to get this to be rendered in a `# Breaking Changes` section of the changelog. Either works, I just don't want these changes to be forgotten about/skipped over when the changelog is generated like a Dependabot bump would

src/compiler/output-targets/dist-lazy/generate-esm-browser.ts
alicewriteswrongs: yeah it looks like the only codepaths in which it will ever get called are when `dynamicImportShim` is `true`

alicewriteswrongs: And I agree, it's better for a ticket like STENCIL-661 to be really straightforward and basically a checklist of 'delete all this'

rwaskiewicz:
LGTM! We have a few minor doc-related items that we're discussing offline as a team - once that's wrapped up this should be good to go

tanner-reits (approved):

alicewriteswrongs (commented):
just pointing one thing out

BREAKING_CHANGES.md
```diff
@@ -69,6 +71,30 @@ export const config: Config = {
...
+Several configuration options related to support for Safari <11, IE11, and Edge
+<19 have been marked as deprecated, and will be removed entirely in a future
+version of Stencil.
+
+##### dynamicImportShim
+
+The `extras.dynamicImportShim` option causes Stencil to include a polyfill for
+the `import()` function for use at runtime. The field is renamed to
+`__deprecated__dynamicImportShim` to indicate deprecation. To retain the
+prior behavior the new option can be set in your project's `stencil.config.ts`:
```
rwaskiewicz, line 85: We may want to clarify that this is for the dynamic `import()` fn, as opposed to the static variant - even though the name implies just that, I want to make sure we're as clear as possible/that folks don't think we dropped static imports (which would be _wild_ 😆)
```suggestion
The `extras.dynamicImportShim` option causes Stencil to include a polyfill for
the [dynamic `import()` function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import) for use at runtime. The field is renamed to
`__deprecated__dynamicImportShim` to indicate deprecation. To retain the
prior behavior the new option can be set in your project's `stencil.config.ts`:
```

BREAKING_CHANGES.md
```diff
@@ -69,6 +71,30 @@ export const config: Config = {
 };
 ```
 
+#### Backwards-compat Configuration Fields Deprecated
+
```
rwaskiewicz, line 75: Can we make spell this out in full? 
```suggestion
#### Backwards Compatible Configuration Fields Deprecated

```

Maybe something slightly different to draw attention to the fact these are browser-specific?
```suggestion
#### Legacy Browser Support Fields Deprecated

```
Thoughts?

rwaskiewicz (changes requested):

BREAKING_CHANGES.md
alicewriteswrongs: Sure, that makes sense to me

definitely don't want people thinking we've decided to only support commonjs!

BREAKING_CHANGES.md
alicewriteswrongs: ooh yeah legacy browser gets at it well 👍

BREAKING_CHANGES.md
alicewriteswrongs: haha no worries, just updated

rwaskiewicz:
LGTM! @alicewriteswrongs can you put up/link the documentation PRs for the Stencil Site to this PR? I think once those are up I think I'd be good to approve this

alicewriteswrongs:
Here's a PR for the V3 upgrade notes: https://github.com/ionic-team/stencil-site/pull/948

rwaskiewicz (approved):

alicewriteswrongs:
and a PR to change the documentation for `extras` on the v3 dev branch: https://github.com/ionic-team/stencil-site/pull/949

PR #3897, opened 2022-12-09
feat(compiler): remove inlineDynamicImports from custom elements targets

tanner-reits:
## Pull request checklist

## Pull request type

 

 

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

Currently, when using `inlineDynamicImports` for custom element output targets, a build error will be thrown. This is due to constraints Rollup enforces around the number of "input" sources allowed when using this option. Since it is almost always guaranteed to be at least two input sources, this option serves no purpose on the custom element targets

GitHub Issue Number: N/A

## What is the new behavior?

Removes `inlineDynamicImports` from `dist-custom-elements` and `dist-custom-elements-bundle` configs

- [X] Yes
- [ ] No

I think we'll consider this a breaking change since it possibly changes Stencil configs... even though using this option would result in broken builds so nobody is using it currently.

## Testing

Removed existing unit tests for this option

## Other information

Follow-up PR to original #3789


tanner-reits:
@rwaskiewicz @alicewriteswrongs Closed out the original PR dealing with this option (#3789) since we changed direction a bit. Also have this going into v3 since we could probably consider removing that option a breaking change. I'll open up PRs for doc changes here shortly. 

Also worth noting, if we run into the deprecation of the root-level `inlineDynamicImports` option during the Rollup upgrades, we can cherry-pick commit [fc904ed](https://github.com/ionic-team/stencil/pull/3897/commits/fc904edc2ac47c053988d9664d9dac4a5800f8c6)

src/compiler/bundle/bundle-output.ts
```diff
@@ -96,6 +96,9 @@ export const getRollupOptions = (
   const afterPlugins = config.rollupPlugins.after || [];
   const rollupOptions: RollupOptions = {
     input: bundleOpts.inputs,
+    output: {
+      inlineDynamicImports: bundleOpts.inlineDynamicImports,
```
rwaskiewicz, line 100: Minor, can we add JSDoc to `src/compiler/bundle/bundle-interface.ts` where [this field is declared](https://github.com/ionic-team/stencil/blob/b3eedae849c02969a9628f3ea6131d8242222c92/src/compiler/bundle/bundle-interface.ts#L38)?

rwaskiewicz, line 100: This code is also reached by the `worker-plugin` and `lazy-output` - how does that affect those two code paths?

tanner-reits: I might be misunderstanding what you're asking, but:
1. Rollup just moved the option to the `output` section, behavior doesn't change
2. Even though we previously didn't have one before, I added a `false` fallback value for the option if it's not defined on the passed `BundleOptions`. This will ensure there is always a value provided

rwaskiewicz (changes requested):
Thanks! Only requests for change are doc related

alicewriteswrongs (approved):
I second @rwaskiewicz's point re `BREAKING_CHANGES.md` but this looks good to me otherwise

BREAKING_CHANGES.md
```diff
@@ -10,36 +10,41 @@ This is a comprehensive list of the breaking changes introduced in the major ver
 
 ## Stencil v3.0.0
 
-* [General](#general)
```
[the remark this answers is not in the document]
tanner-reits: Ugh stupid prettier. I have format-on-save turned on so it always screws w/ markdown formatting. My bad

rwaskiewicz:
@tanner-reits I think you removed the content of `BREAKING_CHANGES.md` related to this change in https://github.com/ionic-team/stencil/pull/3897/commits/60fe536457b590c49f6e01af1829b3e632a1a3c0#diff-2ea154b8b1accf354a338d439aa57c81c10618f59e7d3241682609d475a2e7abL236

rwaskiewicz (approved):

PR #3932, opened 2022-12-20
fix some strictNullChecks errors

alicewriteswrongs:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

GitHub Issue Number: N/A

## What is the new behavior?

This just fixes a few strictNullChecks errors, selected just by running the TypeScript compiler with the option turned on and fixing a few errors that popped up.

- [ ] Yes
- [ ] No

### DIFF
--- a/src/cli/run.ts
+++ b/src/cli/run.ts
@@ -1,4 +1,5 @@
 import { hasError, isFunction, shouldIgnoreError } from '@utils';
+import { join } from 'path';
 
 import { dependencies } from '../compiler/sys/dependencies.json';
 import { createLogger } from '../compiler/sys/logger/console-logger';
@@ -129,13 +130,15 @@ export const runTask = async (
   sys?: d.CompilerSystem
 ): Promise<void> => {
   const logger = config.logger ?? createLogger();
+  const rootDir = config.rootDir ?? '/';
   const strictConfig: ValidatedConfig = {
     ...config,
     flags: createConfigFlags(config.flags ?? { task }),
     hydratedFlag: config.hydratedFlag ?? null,
     logger,
     outputTargets: config.outputTargets ?? [],
-    rootDir: config.rootDir ?? '/',
+    packageJsonFilePath: join(rootDir, 'package.json'),
+    rootDir,
     sys: sys ?? config.sys ?? coreCompiler.createSystem({ logger }),
     testing: config.testing ?? {},
   };
--- a/src/cli/run.ts
+++ b/src/cli/run.ts
@@ -1,5 +1,4 @@
 import { hasError, isFunction, shouldIgnoreError } from '@utils';
-import { join } from 'path';
 
 import { dependencies } from '../compiler/sys/dependencies.json';
 import { createLogger } from '../compiler/sys/logger/console-logger';
@@ -131,15 +130,16 @@ export const runTask = async (
 ): Promise<void> => {
   const logger = config.logger ?? createLogger();
   const rootDir = config.rootDir ?? '/';
+  sys = sys ?? config.sys ?? coreCompiler.createSystem({ logger });
   const strictConfig: ValidatedConfig = {
     ...config,
     flags: createConfigFlags(config.flags ?? { task }),
     hydratedFlag: config.hydratedFlag ?? null,
     logger,
     outputTargets: config.outputTargets ?? [],
-    packageJsonFilePath: join(rootDir, 'package.json'),
+    packageJsonFilePath: sys.platformPath.join(rootDir, 'package.json'),
     rootDir,
-    sys: sys ?? config.sys ?? coreCompiler.createSystem({ logger }),
+    sys,
     testing: config.testing ?? {},
   };
 
--- a/src/cli/run.ts
+++ b/src/cli/run.ts
@@ -130,16 +130,16 @@ export const runTask = async (
 ): Promise<void> => {
   const logger = config.logger ?? createLogger();
   const rootDir = config.rootDir ?? '/';
-  sys = sys ?? config.sys ?? coreCompiler.createSystem({ logger });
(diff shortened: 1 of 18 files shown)


src/declarations/stencil-public-compiler.ts
```diff
@@ -2265,17 +2265,11 @@ export interface Diagnostic {
...
-    lineNumber: number;
-    text?: string;
-    errorCharStart: number;
-    errorLength?: number;
-  }[];
+  absFilePath?: string | undefined;
+  relFilePath?: string | undefined;
+  lineNumber?: number | undefined;
+  columnNumber?: number | undefined;
+  lines: PrintLine[];
```
alicewriteswrongs, line 2272: does this count as the type of change where we'd need to make it a breaking change? if so I'll close this PR and open against the v3 dev branch instead

alicewriteswrongs (commented):
one Q for anyone reading

src/declarations/stencil-public-compiler.ts
rwaskiewicz: Yeah, I think we'd need to target v3.0.0 since this is in `stencil-public-compiler` (more specifically, is exported by it). I don't have a list on hand of who is using `Diagnostic` in their APIs, but I wouldn't be surprised if some nx-related tooling was. I'd target v3 to be on the safe side

alicewriteswrongs: that makes sense, just updated it to be against `v3.0.0-dev` instead

src/declarations/stencil-public-compiler.ts
```diff
@@ -424,7 +424,7 @@ type RequireFields<T, K extends keyof T> = T & { [P in K]-?: T[P] };
 /**
  * Fields in {@link Config} to make required for {@link ValidatedConfig}
  */
-type StrictConfigFields = 'flags' | 'logger' | 'outputTargets' | 'rootDir' | 'sys' | 'testing';
+type StrictConfigFields = 'flags' | 'logger' | 'outputTargets' | 'rootDir' | 'sys' | 'testing' | 'packageJsonFilePath';
```
alicewriteswrongs, line 427: this adds `packageJsonFilePath` to the required fields, which requires changes every time we instantiate a `ValidatedConfig` and drops down a fair number of SNC errors just by itself.

src/cli/run.ts
```diff
@@ -129,13 +129,16 @@ export const runTask = async (
   sys?: d.CompilerSystem
 ): Promise<void> => {
   const logger = config.logger ?? createLogger();
+  const rootDir = config.rootDir ?? '/';
+  sys = sys ?? config.sys ?? coreCompiler.createSystem({ logger });
```
alicewriteswrongs, line 133: We have to do this convoluted thing here because we need access to a `CompilerSystem` below to use `sys.platformPath`. I ran into a bunch of issues with just doing `import { join } from 'path'` with treeshaking, module resolution, and whatnot, so I think this is the best option.

src/utils/logger/logger-typescript.ts
```diff
@@ -108,16 +108,18 @@ export const loadTypeScriptDiagnostic = (tsDiagnostic: Diagnostic): d.Diagnostic
     header: 'TypeScript',
     code: tsDiagnostic.code.toString(),
     messageText: flattenDiagnosticMessageText(tsDiagnostic, tsDiagnostic.messageText),
-    relFilePath: null,
-    absFilePath: null,
+    columnNumber: undefined,
+    lineNumber: undefined,
+    relFilePath: undefined,
+    absFilePath: undefined,
```
alicewriteswrongs, line 114: this and a few other changes are because of the slightly stricter typing of the `d.Diagnostic` interface

alicewriteswrongs (commented):
just noting a few things

tanner-reits (approved):
I don't see anything concerning here. Nice job and a good chunk of violations dropped!

src/utils/util.ts
```diff
@@ -189,7 +189,7 @@ export const readPackageJson = async (config: d.ValidatedConfig, compilerCtx: d.
 export type ParsePackageJsonResult = {
   diagnostic: d.Diagnostic | null;
   data: any | null;
-  filePath: string;
+  filePath: string | undefined;
```
rwaskiewicz, line 192: See my other comment in this file - if we can remove the changes proposed there, I _think_ we won't need this either.

rwaskiewicz (changes requested):

src/cli/run.ts
alicewriteswrongs: oh good point! I haven't done any testing, but I think a little defensive programming here couldn't hurt. I'll do something like this probably:

```ts
const configSys = sys ?? config.sys ?? coreCompiler.createSystem({ logger });
```

and then

```ts
{ ..., sys: configSys, ... }
```

below

src/utils/util.ts
```diff
@@ -198,7 +198,7 @@ export type ParsePackageJsonResult = {
  * @param pkgJsonFilePath the path to the already read `package.json` file
  * @returns the results of parsing the provided contents of the `package.json` file
  */
-export const parsePackageJson = (pkgJsonStr: string, pkgJsonFilePath: string): ParsePackageJsonResult => {
+export const parsePackageJson = (pkgJsonStr: string, pkgJsonFilePath: string | undefined): ParsePackageJsonResult => {
```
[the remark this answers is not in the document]
alicewriteswrongs: tbh I don't remember why I changed this haha, but I think you're right that we don't need the `| undefined`

rwaskiewicz (changes requested):
@alicewriteswrongs I think we can safely delete that test I posted a snipped of as well https://github.com/ionic-team/stencil/pull/3932#discussion_r1081725279

rwaskiewicz (approved):

PR #3954, opened 2023-01-05
chore(config): add `hydratedFlag` to ValidatedConfig

alicewriteswrongs:
This adds the `hydratedFlag` property to the `ValidatedConfig` object, meaning that the field must be present.

Note that this required a change to use null to signify a missing hydrateflag.
this is because when we are building the `ValidatedConfig` type we do
the following:

```ts
type RequireFields<T, K extends keyof T> = T & { [P in K]-?: T[P] };
```

If there is a type `P` in `K extends keyof T` where the type of `T[P]` is
`Foo | undefined` then unfortunately when we do `[P in K]-?: T[P]` we'll
set the output type to just `Foo`, so the type returned by
`RequireFields` for the field in question won't be unioned with `undefined`,
which was previously what we were doing here to try to signify a missing
`HydratedFlag` value.

Instead we can just use `null` which makes it through the `-?` bit.

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

GitHub Issue Number: N/A

## What is the new behavior?

This should just fix a few SNC errors.

- [ ] Yes
- [x] No

It technically introduces one small change in a public type signature, from `HydratedFlag` to `HydratedFlag | null`. This could be grounds for making this change in v3.

### DIFF
--- a/src/compiler/config/validate-config.ts
+++ b/src/compiler/config/validate-config.ts
@@ -52,12 +52,12 @@ export const validateConfig = (
     ...config,
     // flags _should_ be JSON safe
     flags: JSON.parse(JSON.stringify(config.flags || {})),
+    hydratedFlag: validateHydrated(config),
     logger,
     outputTargets: config.outputTargets ?? [],
     rootDir: typeof config.rootDir === 'string' ? config.rootDir : '/',
     sys: config.sys ?? bootstrapConfig.sys ?? createSystem({ logger }),
     testing: config.testing ?? {},
-    hydratedFlag: validateHydrated(config),
   };
 
   // default devMode false
--- a/src/compiler/config/validate-config.ts
+++ b/src/compiler/config/validate-config.ts
@@ -57,6 +57,7 @@ export const validateConfig = (
     rootDir: typeof config.rootDir === 'string' ? config.rootDir : '/',
     sys: config.sys ?? bootstrapConfig.sys ?? createSystem({ logger }),
     testing: config.testing ?? {},
+    hydratedFlag: validateHydrated(config),
   };
 
   // default devMode false
@@ -149,9 +150,6 @@ export const validateConfig = (
 
   // testing
   validateTesting(validatedConfig, diagnostics);
-
-  // hydrate flag
-  validatedConfig.hydratedFlag = validateHydrated(validatedConfig);
 
   // bundles
   if (Array.isArray(validatedConfig.bundles)) {
--- a/src/compiler/sys/config.ts
+++ b/src/compiler/sys/config.ts
@@ -14,6 +14,7 @@ export const getConfig = (userConfig: d.Config): d.ValidatedConfig => {
     rootDir: userConfig.rootDir ?? '/',
     sys: userConfig.sys ?? createSystem({ logger }),
     testing: userConfig ?? {},
+    hydratedFlag: undefined,
   };
 
   setPlatformPath(config.sys.platformPath);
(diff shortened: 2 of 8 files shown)


tanner-reits (approved):
LGTM! Thanks for the explainer in the description

src/declarations/stencil-public-compiler.ts
```diff
@@ -410,7 +410,7 @@ type RequireFields<T, K extends keyof T> = T & { [P in K]-?: T[P] };
 /**
  * Fields in {@link Config} to make required for {@link ValidatedConfig}
  */
-type StrictConfigFields = 'flags' | 'logger' | 'outputTargets' | 'rootDir' | 'sys' | 'testing';
+type StrictConfigFields = 'flags' | 'logger' | 'outputTargets' | 'rootDir' | 'sys' | 'testing' | 'hydratedFlag';
```
rwaskiewicz, line 413: Minor, can we keep this list alphabetized?

src/testing/mocks.ts
```diff
@@ -39,6 +39,7 @@ export function mockValidatedConfig(overrides: Partial<ValidatedConfig> = {}): V
     rootDir: path.resolve('/'),
     sys: createTestingSystem(),
     testing: {},
+    hydratedFlag: null,
```
rwaskiewicz, line 42: Minor, can we keep this list alphabetized?

src/compiler/sys/config.ts
```diff
@@ -14,6 +14,7 @@ export const getConfig = (userConfig: d.Config): d.ValidatedConfig => {
     rootDir: userConfig.rootDir ?? '/',
     sys: userConfig.sys ?? createSystem({ logger }),
     testing: userConfig ?? {},
+    hydratedFlag: userConfig.hydratedFlag ?? null,
```
rwaskiewicz, line 17: Minor, can we keep this list alphabetized?

src/compiler/config/validate-config.ts
```diff
@@ -57,6 +57,7 @@ export const validateConfig = (
     rootDir: typeof config.rootDir === 'string' ? config.rootDir : '/',
     sys: config.sys ?? bootstrapConfig.sys ?? createSystem({ logger }),
     testing: config.testing ?? {},
+    hydratedFlag: validateHydrated(config),
```
rwaskiewicz, line 60: Minor, can we keep this list alphabetized?

rwaskiewicz (commented):
Looks good to me, nothing but nitpicks on my end.

rwaskiewicz (changes requested):
@alicewriteswrongs looks like we introduce 2 new strict null checks violations here as well (the readout from the GH Action only compares against `main`). Can you TAL at that as well?

rwaskiewicz (approved):

alicewriteswrongs:
Ok so I don't _think_ there are any new SNC errors introduced by the PR. To verify that I did the following

### get null errors on branch

```sh
git checkout ap/validate-hydrate-flag
npm ci && npm run build
npx tsc --strictNullChecks --noEmit --pretty false > /tmp/null_errors_pr.txt
```

### get null errors on V3

```sh
git checkout v3.0.0-dev
npm ci && npm run build
npx tsc --strictNullChecks --noEmit --pretty false > /tmp/null_errors_v3.txt
```

### compare

```sh
cd /tmp
diff null_errors_pr.txt null_errors_v3.txt
```

and I didn't see any differences in the files. also if I do:

```sh
cat /tmp/null_errors_pr.txt
cat /tmp/null_errors_v3.txt
```

I get 

```
2919
2919
```

PR #3955, opened 2023-01-05
feat(runtime): updates nonce fallback to use meta tag instead of window

tanner-reits:
This commit updates our CSP nonce support logic to allow implementers to leverage a meta tag in the DOM head for setting nonce values during the Stencil runtime rather than pulling the value off of the global window object

## Pull request checklist

- [X] Docs have been reviewed and added / updated if needed (for bug fixes / features)
  - https://github.com/ionic-team/stencil-site/pull/966/files

## Pull request type

 

 

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

With the initial implementation, the fallback mechanism for setting nonce values on elements during runtime relied on pulling the nonce off of the global `window` object. This was brittle and a bad DX

GitHub Issue Number: N/A

## What is the new behavior?

Now, implementers will be able to inject a `meta` tag in the DOM head which will contain the nonce value to be applied on each element generated by the Stencil runtime. 

- [ ] Yes
- [X] No

## Testing

Tested using [this branch](https://github.com/tanner-reits/stencil-csp-demo/tree/meta-fallback) in the CSP demo app.

**Steps:**
1. Make sure to be using the `meta-fallback` branch
2. `npm i`
3. `npm run start:server-valid` & navigate to `localhost:3333/home`
4. "Hello world" text should be red
5. Comment out `<meta name="csp-nonce">` from `www/index.html`
6. Refresh page
7. "Hello world" text will now be white and an additional CSP violation error appears


rwaskiewicz (changes requested):

rwaskiewicz (approved):

src/utils/query-nonce-meta-tag-content.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Helper method for querying a `meta` tag that contains a nonce value
+ * out of a DOM's head.
+ *
+ * @param doc The DOM containing the `head` to query against
+ * @returns The content of the meta tag representing the nonce value, `undefined` if no tag exists, or
+ * `null` if the tag has no content.
```
alicewriteswrongs, line 7: is there value for the user in having a distinction in the return value between the situation where the tag exists with no content and the case where it isn't there at all? perhaps so, but if we don't think so it might be good for us to just coerce down to a single nullish value with a `?? undefined` (or `?? null`) at the end

rwaskiewicz (approved):
LGTM - @tanner-reits is this good to merge into the parent?

PR #3956, opened 2023-01-05
feat(compiler): add flag for auto exporting custom types

tanner-reits:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

Currently, Framework encounters issues migrating to Stencil V3 due to a circular type definition. This happens because the types referenced by their components live in a root `interface.d.ts` file. This file is manually maintained and exports the contents of `components.d.ts` file, which is what results in the circular dependency.

GitHub Issue Number: N/A

## What is the new behavior?

This commit adds `autoExportCustomTypes` as a config flag to control whether or not custom types used for component props/events get exported from the `components.d.ts` file. This flag defaults `true` and was immediately marked as deprecated as it was only added as an interim fix for the Ionic Framework team's migrations to Stencil V3

- [ ] Yes
- [X] No

## Testing

- Unit tests were added to the config validation test suite for the new field
- Tested Framework with the local build and got the same output in `components.d.ts` as V2 with the flag disabled.

### DIFF
--- a/src/cli/run.ts
+++ b/src/cli/run.ts
@@ -137,6 +137,7 @@ export const runTask = async (
     rootDir: config.rootDir ?? '/',
     sys: sys ?? config.sys ?? coreCompiler.createSystem({ logger }),
     testing: config.testing ?? {},
+    autoExportCustomTypes: config.autoExportCustomTypes ?? true,
   };
 
   switch (task) {
--- a/src/cli/run.ts
+++ b/src/cli/run.ts
@@ -131,13 +131,13 @@ export const runTask = async (
   const logger = config.logger ?? createLogger();
   const strictConfig: ValidatedConfig = {
     ...config,
+    autoExportCustomTypes: config.autoExportCustomTypes ?? true,
     flags: createConfigFlags(config.flags ?? { task }),
     logger,
     outputTargets: config.outputTargets ?? [],
     rootDir: config.rootDir ?? '/',
     sys: sys ?? config.sys ?? coreCompiler.createSystem({ logger }),
     testing: config.testing ?? {},
-    autoExportCustomTypes: config.autoExportCustomTypes ?? true,
   };
 
   switch (task) {
--- a/src/declarations/stencil-public-compiler.ts
+++ b/src/declarations/stencil-public-compiler.ts
@@ -436,13 +436,13 @@ type RequireFields<T, K extends keyof T> = T & { [P in K]-?: T[P] };
  * Fields in {@link Config} to make required for {@link ValidatedConfig}
  */
 type StrictConfigFields =
+  | 'autoExportCustomTypes'
   | 'flags'
   | 'logger'
   | 'outputTargets'
   | 'rootDir'
   | 'sys'
-  | 'testing'
-  | 'autoExportCustomTypes';
+  | 'testing';
 
 /**
  * A version of {@link Config} that makes certain fields required. This type represents a valid configuration entity.
--- a/src/declarations/stencil-public-compiler.ts
+++ b/src/declarations/stencil-public-compiler.ts
@@ -246,6 +246,17 @@ export interface StencilConfig {
   watchIgnoredRegex?: RegExp | RegExp[];
   excludeUnusedDependencies?: boolean;
   stencilCoreResolvedId?: string;
+
+  /**
+   * @deprecated will not be supported in a future version of Stencil.
+   *
+   * When `true`, any custom types used as a part of a component `@Prop` or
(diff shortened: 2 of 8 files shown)


src/cli/run.ts
```diff
@@ -137,6 +137,7 @@ export const runTask = async (
     rootDir: config.rootDir ?? '/',
     sys: sys ?? config.sys ?? coreCompiler.createSystem({ logger }),
     testing: config.testing ?? {},
+    autoExportCustomTypes: config.autoExportCustomTypes ?? true,
```
rwaskiewicz, line 140: Minor, can we keep this list alphabetized, even though this flag is temporary?

src/declarations/stencil-public-compiler.ts
```diff
@@ -246,6 +246,17 @@ export interface StencilConfig {
...
+  /**
+   * @deprecated will not be supported in a future version of Stencil.
+   *
+   * When `true`, any custom types used as a part of a component `@Prop` or
+   * `@Event` will be automatically exported from the generated `components.d.ts` file
+   * in the compiled output.
+   *
+   * @default true
+   */
+  autoExportCustomTypes?: boolean;
```
rwaskiewicz, line 259: From the PR summary:

> was immediately marked as deprecated as it was only added as an interim fix for the Ionic Framework team's migrations to Stencil V3

Do we have an idea as to how long this flag will exist? I.E. can we delete it before Stencil v3.0.0 is released? If the answer is no, can we be more explicit about steering folks away from this flag by renaming it something like:
```suggestion
  __ionicFrameworkInternalAutoExportCustomTypes?: boolean;
```

src/declarations/stencil-public-compiler.ts
```diff
@@ -424,7 +435,14 @@ type RequireFields<T, K extends keyof T> = T & { [P in K]-?: T[P] };
...
  */
-type StrictConfigFields = 'flags' | 'logger' | 'outputTargets' | 'rootDir' | 'sys' | 'testing';
+type StrictConfigFields =
+  | 'flags'
+  | 'logger'
+  | 'outputTargets'
+  | 'rootDir'
+  | 'sys'
+  | 'testing'
+  | 'autoExportCustomTypes';
```
rwaskiewicz, line 445: Minor, same comment regarding alphabetization order

src/testing/mocks.ts
```diff
@@ -39,6 +39,7 @@ export function mockValidatedConfig(overrides: Partial<ValidatedConfig> = {}): V
     rootDir: path.resolve('/'),
     sys: createTestingSystem(),
     testing: {},
+    autoExportCustomTypes: true,
```
rwaskiewicz, line 42: Minor, same comment regarding alphabetization order

rwaskiewicz (changes requested):
I think we talked a bit about this offline. I'd like to understand how long this flag will exist in the long term. I have some reservations around it existing in v3.0.0 GA. If this will be removed by then, I don't have a problem merging this. If not, I think we need to have a bigger sync with the Framework team to discuss our options/path forward

src/declarations/stencil-public-compiler.ts
tanner-reits: @rwaskiewicz @alicewriteswrongs It won't be removed before V3. Do we want to follow the `deprecated` naming convention used for the `extras` options Alice recently worked on?

tanner-reits:
> I have some reservations around it existing in v3.0.0 GA. If this will be removed by then, I don't have a problem merging this. If not, I think we need to have a bigger sync with the Framework team to discuss our options/path forward

@rwaskiewicz Yeah, so this would not be removed by the v3 release. Happy to talk through this with you and anyone else we wanna pull in. Wanna bring this up in standup, or should we dedicate some other time?

rwaskiewicz:
@tanner-reits Yes, let's talk about this at standup and move it a dedicated meeting if we can't run this to ground in 5 min or so

src/declarations/stencil-public-compiler.ts
rwaskiewicz: It's that's the case, I'd like to make this _extra_ volatile 😈 Something like React's ` __SECRET_DOM_DO_NOT_USE_OR_YOU_WILL_BE_FIRED` so that we can remove this in a minor version of Stenci - hence the 'ionicFrameworkInternal' part. If anyone states removing that flag broke them we can say:
1. it was marked as internal only
2. it used `__` as a prefix (designating it as private)
3. AND it wasn't documented

tanner-reits:
Closing this out based on team discussion

PR #3976, opened 2023-01-12
chore(ci): test component starter

rwaskiewicz:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes

N/A - non-existent 

GitHub Issue Number: N/A

## What is the new behavior?

this commit adds a new workflow, test-component-starter, to the ci
workflow. it's purpose is to smoke test proposed changes in a pull
request in a stencil project. it accomplishes this by testing the
proposed changes, build stencil, install it in an instance of the
starter template, and run the template's build/test steps.

- [ ] Yes
- [x] No

## Testing

Logs tailed to verify the behavior https://github.com/ionic-team/stencil/actions/runs/3903522456/jobs/6668094925
## Other information

I'll need to update the required checks should we accept this

### DIFF
--- a/.github/workflows/test-component-starter.yml
+++ b/.github/workflows/test-component-starter.yml
@@ -1,0 +1,68 @@
+name: Component Starter Smoke Test
+
+on:
+  workflow_call:
+    # Make this a reusable workflow, no value needed
+    # https://docs.github.com/en/actions/using-workflows/reusing-workflows
+
+jobs:
+  analysis_test:
+    name: (${{ matrix.os }}.${{ matrix.node }})
+    strategy:
+      fail-fast: false
+      matrix:
+        node: ['14', '16']
+        os: ['ubuntu-latest', 'windows-latest']
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # v3.3.0
+
+      - name: Get Core Dependencies
+        uses: ./.github/workflows/actions/get-core-dependencies
+
+      - name: Use Node ${{ matrix.node }}
+        uses: actions/setup-node@64ed1c7eab4cce3362f8c340dee64e5eaeef8f7c # v3.6.0
+        with:
+          node-version: ${{ matrix.node }}
+          cache: 'npm'
+
+      - name: Download Build Archive
+        uses: ./.github/workflows/actions/download-archive
+        with:
+          name: stencil-core
+          path: .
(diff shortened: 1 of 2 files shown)


.github/workflows/test-component-starter.yml
```diff
@@ -0,0 +1,68 @@
...
+          name: stencil-core
+          path: .
+          filename: stencil-core-build.zip
+
+      - name: Pack the Build Archive
+        run: npm pack
+        shell: bash
+
+      - name: Move the Stencil Build Artifact
+        # there isn't a great way to get the output of `npm pack`, just grab the most recent and hope for the best
```
alicewriteswrongs, line 42: I think the `.tgz` is named predictably based on the version in `package.json`, but probably mapping `version -> filename` is more fragile than what you have here

.github/workflows/test-component-starter.yml
```diff
@@ -0,0 +1,68 @@
...
+        run: npm init stencil component tmp-component-starter
+        shell: bash
+
+      - name: Install Component Starter Dependencies
+        run: npm install
+        working-directory: ./tmp-component-starter
+        shell: bash
+
+      - name: Install Stencil Eval
+        run: npm i ../stencil-eval.tgz
```
alicewriteswrongs, line 56: although, would we have wildcard expansion here? if so, could we do `npm i ../*.tgz` ?

alicewriteswrongs (changes requested):
great idea to add this test, just had like one question

.github/workflows/test-component-starter.yml
rwaskiewicz: I'm not sure I understand - wouldn't changing this to `../*.tgz` introduce wildcard expansion?

tanner-reits: There should, theoretically, only ever be one build artifact, right? So this _should_ work. I don't really have a preference either way

alicewriteswrongs: should have phrased my question better! Was wondering if the wildcard in `npm i ../*.tgz` would be expanded properly, if the command is running in `bash` I would think so but I don't think I've tried that before in gh actions. In any case, I think this can be merged as-is and we can fiddle with that later if we want

alicewriteswrongs (approved):

.github/workflows/test-component-starter.yml
tanner-reits: Works for me!

tanner-reits (approved):

PR #3991, opened 2023-01-18
refactor(vdom): improve type annotations in some virtual DOM code

alicewriteswrongs:
This improves the typings of some virtual DOM related code, mainly correcting the types of some function parameters which take virtual DOM nodes to be accurate and then changing other type declarations in order to agree with that.

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

GitHub Issue Number: N/A

## What is the new behavior?

-
-
-

- [ ] Yes
- [x] No

### DIFF
--- a/src/declarations/stencil-public-runtime.ts
+++ b/src/declarations/stencil-public-runtime.ts
@@ -561,12 +561,12 @@ export interface FunctionalComponent<T = {}> {
  * {@link FunctionalUtilities}).
  */
 export interface ChildNode {
-  vtag?: string | number | Function;
-  vkey?: string | number;
-  vtext?: string;
-  vchildren?: VNode[];
+  vtag?: string | number | Function | null;
+  vkey?: string | number | null;
+  vtext?: string | null;
+  vchildren?: VNode[] | null;
   vattrs?: any;
-  vname?: string;
+  vname?: string | null;
 }
 
 /**
@@ -620,13 +620,13 @@ export declare function h(sel: any, data: VNodeData | null, children: VNode): VN
  */
 export interface VNode {
   $flags$: number;
-  $tag$: string | number | Function;
+  $tag$: string | number | Function | null;
   $elm$: any;
-  $text$: string;
-  $children$: VNode[];
+  $text$: string | null;
+  $children$: VNode[] | null;
   $attrs$?: any;
-  $name$?: string;
-  $key$?: string | number;
+  $name$?: string | null;
+  $key$?: string | number | null;
 }
 
 export interface VNodeData {
--- a/src/runtime/vdom/h.ts
+++ b/src/runtime/vdom/h.ts
@@ -109,7 +109,7 @@ Empty objects can also be the cause, look for JSX comments that became objects.`
  * @param text possible text content for the node
  * @returns a newly-minted virtual DOM node
  */
-export const newVNode = (tag: string, text: string) => {
+export const newVNode = (tag: string | null, text: string | null): d.VNode => {
   const vnode: d.VNode = {
     $flags$: 0,
     $tag$: tag,
@@ -137,7 +137,7 @@ export const Host = {};
  * @param node the virtual DOM node to check
  * @returns whether it's a Host node or not
  */
-export const isHost = (node: any): node is d.VNode => node && node.$tag$ === Host;
+export const isHost = (node: d.VNode): boolean => node && node.$tag$ === Host;
 
 /**
  * Implementation of {@link d.FunctionalUtilities} for Stencil's VDom.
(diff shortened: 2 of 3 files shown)


src/runtime/vdom/vdom-render.ts
```diff
@@ -806,7 +811,10 @@ export const renderVdom = (hostRef: d.HostRef, renderFnResults: d.VNode | d.VNod
   const hostElm = hostRef.$hostElement$;
   const cmpMeta = hostRef.$cmpMeta$;
   const oldVNode: d.VNode = hostRef.$vnode$ || newVNode(null, null);
-  const rootVnode = isHost(renderFnResults) ? renderFnResults : h(null, null, renderFnResults as any);
+  const rootVnode: d.VNode =
+    !Array.isArray(renderFnResults) && isHost(renderFnResults)
```
rwaskiewicz, line 815: One thing we should verify - I'm oversimplifying the issue a little bit, but in TS it's possible to assign a property to an array and have the type checker be OK with it:
```ts
const foo: any = [];
foo.$tag$ = 'bar';
``` 

With our change to `isHost`, the previous implementation (of `isHost`) _could_ have returned `true` and narrowed the type of `renderFnResults` (since the original implementation was a type guard), even if `renderFnResults` was an array (assuming it had a `$tag$` property on it, which i believe only could happen in the situation I described above.

Can you do me a favor and verify the call chain to `renderVdom` to verify that something like this shouldn't happen? I'm not sure if there's any type erasure or widening in that chain that could lead to an issue here, but probably best be safe than sorry.

rwaskiewicz (changes requested):

src/runtime/vdom/h.ts
```diff
@@ -109,7 +109,7 @@ Empty objects can also be the cause, look for JSX comments that became objects.`
  * @param text possible text content for the node
  * @returns a newly-minted virtual DOM node
  */
-export const newVNode = (tag: string, text: string) => {
+export const newVNode = (tag: string | null, text: string | null): d.VNode => {
```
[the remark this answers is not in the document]
alicewriteswrongs: Yeah, that makes sense. I made the change basically because we are explicitly passing `null` to both of these parameters in a few places, e.g. here:

https://github.com/ionic-team/stencil/blob/9ace5cd87cd1ed1380087fedd8d7c53e043afc12/src/runtime/vdom/h.ts#L90-L90

and here:

https://github.com/ionic-team/stencil/blob/9ace5cd87cd1ed1380087fedd8d7c53e043afc12/src/runtime/vdom/h.ts#L44-L44

So I figured the typing for the parameters should be inline with that! I'm going to adjust the types for `VNode` and `ChildNode` to match

src/runtime/vdom/vdom-render.ts
alicewriteswrongs: It's hard to say 100% because unfortunately we more or less have typechecking turned off in `update-component.ts` (lots of `any` scattered around 😞) but looking through the call sites for `renderVdom`, `callRender`, and `updateComponent` I don't see anything like that going on. The reason I opened this PR in the first place was in part that I get a little scared of how un-checked some of these files are...

rwaskiewicz: Fair, so long as we're fairly confident here. Worst case scenario, we need to revert this (and if there is a rendering bug, we know where to look first)

src/declarations/stencil-public-runtime.ts
```diff
@@ -561,12 +561,12 @@ export interface FunctionalComponent<T = {}> {
...
  */
 export interface ChildNode {
-  vtag?: string | number | Function;
-  vkey?: string | number;
-  vtext?: string;
-  vchildren?: VNode[];
+  vtag?: string | number | Function | null;
+  vkey?: string | number | null;
+  vtext?: string | null;
+  vchildren?: VNode[] | null;
```
rwaskiewicz, line 567: Does widening this to `null` have any affect on Functional Components? Namely, does this affect the `FunctionUtilities#forEach` function? I think that was originally written under the assumption that this property would always be an array (even if it were empty), but I'm not 100% sure of the fallout of making this `null`.

rwaskiewicz (changes requested):
We're widening a few more types that I expected we would in these revisions. Can you confirm that you traced through the usages of these fields for cases where the TypeScript compiler might not have caught typing issues (since there are so many `any`s in here 😢 )?

Also, it looks like we're +5 for SNC validation failures, any way we can drive that down?

tanner-reits (changes requested):
I'm not familiar enough w/ all the vDOM code to know of any wider-impact these changes may have, but the changes seem fine from what I know. Only thing I'd like to see is if we can remove the newly-introduced strict null checks

alicewriteswrongs:
Just closing this as it's been open w/out review for ~6 months

PR #4042, opened 2023-02-07
feat(compiler): transform module aliases in emitted js, typedefs 

alicewriteswrongs:
This implements a transformer which is very similar to the one added
in https://github.com/ionic-team/stencil/pull/3523 but which is run earlier in the build process and which can
therefore rewrite aliased paths in both emitted JS and typedef files.

This matters if the user has the `generateTypeDeclarations` option set
to `true` on one of their output targets.

The new behavior implemented here, however, is no longer specific to a
particular output target, and applies to all TypeScript code which runs
through Stencil. Accordingly, the behavior is opt-in, and is controlled
by a new configuration value, `transformAliasedImportPaths` which
defaults to `false`.

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Currently if you use `paths` in `tsconfig.json` to set up module aliases then these are not transpiled in the `.d.ts` files that Stencil emits for individual components. This is unfortunate! This PR introduces a transformer into the typescript compiler which will transform these paths from something like `@utils` to a relative path from the importing module to the importee. This will allow the build output to be used directly without any need for further transformation. 

GitHub Issue Number: #3788

## What is the new behavior?

Now such imports will be transformed at build time, so that emitted JS files and `.d.ts` files will have a resolved relative path to the module they're importing.

Because this behavior 

- [ ] Yes
- [x] No

## Testing

The best way to test this is to check out the reproduction case on the linked issue and try building it with this change installed. You'll also need to add `transformAliasedImportPaths` on the stencil config.

With that reproduction it might be helpful to also do something like:

```sh
cd /tmp
mkdir stencil-alias-test && cd stencil-alias-test
git init
```

then copy the contents of `dist` over and commit them before installing Stencil built from this branch. Then you can install Stencil from this branch and repeat the procedure to see a diff of the changes its produced.

### DIFF
--- a/src/compiler/config/transpile-options.ts
+++ b/src/compiler/config/transpile-options.ts
@@ -123,6 +123,7 @@ export const getTranspileConfig = (input: TranspileOptions) => {
   const config: Config = {
     rootDir: compileOpts.currentDirectory,
     srcDir: compileOpts.currentDirectory,
+    transformAliasedImportPaths: true,
     devMode: true,
     minifyCss: true,
     minifyJs: false,
--- a/src/compiler/config/transpile-options.ts
+++ b/src/compiler/config/transpile-options.ts
@@ -12,7 +12,7 @@ import type {
 } from '../../declarations';
 import { STENCIL_INTERNAL_CLIENT_ID } from '../bundle/entry-alias-ids';
 import { IS_NODE_ENV, requireFunc } from '../sys/environment';
-import {createInMemoryFs, InMemoryFileSystem} from '../sys/in-memory-fs';
+import { createInMemoryFs, InMemoryFileSystem } from '../sys/in-memory-fs';
 import { createSystem } from '../sys/stencil-sys';
 import { parseImportPath } from '../transformers/stencil-import-path';
 
@@ -42,10 +42,10 @@ export const getTranspileResults = (code: string, input: TranspileOptions) => {
 const transpileCtx = { sys: null as CompilerSystem };
 
 interface TranspileConfig {
-  config: Config,
-  inMemoryFs: InMemoryFileSystem,
-  compileOpts: TranspileOptions,
-  transformOpts: TransformOptions
+  compileOpts: TranspileOptions;
+  config: Config;
+  inMemoryFs: InMemoryFileSystem;
+  transformOpts: TransformOptions;
 }
 
 export const getTranspileConfig = (input: TranspileOptions): TranspileConfig => {
@@ -147,8 +147,8 @@ export const getTranspileConfig = (input: TranspileOptions): TranspileConfig => 
   return {
     compileOpts,
     config,
+    inMemoryFs,
     transformOpts,
-    inMemoryFs
   };
 };
 
--- a/src/compiler/config/transpile-options.ts
+++ b/src/compiler/config/transpile-options.ts
@@ -40,12 +40,22 @@ export const getTranspileResults = (code: string, input: TranspileOptions) => {
 
 const transpileCtx = { sys: null as CompilerSystem };
 
+/**
+ * Configuration necessary for transpilation
+ */
 interface TranspileConfig {
   compileOpts: TranspileOptions;
   config: Config;
   transformOpts: TransformOptions;
 }
 
+/**
+ * Get configuration necessary to carry out transpilation, including a Stencil
+ * configuration, transformation options, and transpilation options.
+ *
+ * @param input options for Stencil's transpiler (string-to-string compiler)
+ * @returns the options and configuration necessary for transpilation
+ */
 export const getTranspileConfig = (input: TranspileOptions): TranspileConfig => {
   if (input.sys) {
     transpileCtx.sys = input.sys;
--- a/src/compiler/config/transpile-options.ts
+++ b/src/compiler/config/transpile-options.ts
@@ -12,6 +12,7 @@ import type {
 } from '../../declarations';
 import { STENCIL_INTERNAL_CLIENT_ID } from '../bundle/entry-alias-ids';
 import { IS_NODE_ENV, requireFunc } from '../sys/environment';
+import {createInMemoryFs, InMemoryFileSystem} from '../sys/in-memory-fs';
 import { createSystem } from '../sys/stencil-sys';
 import { parseImportPath } from '../transformers/stencil-import-path';
 
@@ -40,7 +41,14 @@ export const getTranspileResults = (code: string, input: TranspileOptions) => {
 
 const transpileCtx = { sys: null as CompilerSystem };
 
-export const getTranspileConfig = (input: TranspileOptions) => {
+interface TranspileConfig {
+  config: Config,
+  inMemoryFs: InMemoryFileSystem,
+  compileOpts: TranspileOptions,
+  transformOpts: TransformOptions
+}
+
+export const getTranspileConfig = (input: TranspileOptions): TranspileConfig => {
   if (input.sys) {
     transpileCtx.sys = input.sys;
   } else if (!transpileCtx.sys) {
@@ -134,10 +142,13 @@ export const getTranspileConfig = (input: TranspileOptions) => {
     tsCompilerOptions,
   };
 
+  const inMemoryFs = createInMemoryFs(transpileCtx.sys);
+
   return {
     compileOpts,
     config,
     transformOpts,
+    inMemoryFs
   };
 };
 
--- a/src/compiler/config/transpile-options.ts
+++ b/src/compiler/config/transpile-options.ts
@@ -44,7 +44,6 @@ const transpileCtx = { sys: null as CompilerSystem };
 interface TranspileConfig {
   compileOpts: TranspileOptions;
   config: Config;
-  inMemoryFs: InMemoryFileSystem;
   transformOpts: TransformOptions;
 }
 
@@ -142,12 +141,9 @@ export const getTranspileConfig = (input: TranspileOptions): TranspileConfig => 
     tsCompilerOptions,
   };
 
-  const inMemoryFs = createInMemoryFs(transpileCtx.sys);
-
   return {
     compileOpts,
     config,
-    inMemoryFs,
     transformOpts,
   };
 };
--- a/src/compiler/config/transpile-options.ts
+++ b/src/compiler/config/transpile-options.ts
@@ -12,7 +12,6 @@ import type {
 } from '../../declarations';
 import { STENCIL_INTERNAL_CLIENT_ID } from '../bundle/entry-alias-ids';
 import { IS_NODE_ENV, requireFunc } from '../sys/environment';
-import { createInMemoryFs, InMemoryFileSystem } from '../sys/in-memory-fs';
 import { createSystem } from '../sys/stencil-sys';
 import { parseImportPath } from '../transformers/stencil-import-path';
 
@@ -128,17 +127,17 @@ export const getTranspileConfig = (input: TranspileOptions): TranspileConfig => 
   };
 
   const config: Config = {
-    rootDir: compileOpts.currentDirectory,
-    srcDir: compileOpts.currentDirectory,
-    transformAliasedImportPaths: true,
+    _isTesting: true,
     devMode: true,
+    enableCache: false,
     minifyCss: true,
     minifyJs: false,
-    _isTesting: true,
-    validateTypes: false,
-    enableCache: false,
+    rootDir: compileOpts.currentDirectory,
+    srcDir: compileOpts.currentDirectory,
     sys: transpileCtx.sys,
+    transformAliasedImportPaths: input.transformAliasedImportPaths,
     tsCompilerOptions,
+    validateTypes: false,
   };
 
   return {
--- a/src/compiler/sys/typescript/typescript-resolve-module.ts
+++ b/src/compiler/sys/typescript/typescript-resolve-module.ts
@@ -100,7 +100,10 @@ export const patchedTsResolveModule = (
     let resolvedFileName = join(containingDir, moduleName);
     resolvedFileName = normalizePath(ensureExtension(resolvedFileName, containingFile));
 
-    if (isAbsolute(resolvedFileName) && !inMemoryFs.accessSync(resolvedFileName)) {
+    // in some cases `inMemoryFs` will not be defined here, so we should use
+    // `accessSync` on `config.sys` instead.
+    const accessSync = inMemoryFs?.accessSync ?? config.sys.accessSync;
+    if (isAbsolute(resolvedFileName) && !accessSync(resolvedFileName)) {
       return null;
     }
 
--- a/src/compiler/sys/typescript/typescript-resolve-module.ts
+++ b/src/compiler/sys/typescript/typescript-resolve-module.ts
@@ -19,6 +19,7 @@ import {
 } from '../resolve/resolve-utils';
 import { patchTsSystemFileSystem } from './typescript-sys';
 
+// TODO(STENCIL-728): fix typing of `inMemoryFs` parameter in `patchTypescript`, related functions
 export const patchTypeScriptResolveModule = (config: d.Config, inMemoryFs: InMemoryFileSystem) => {
   let compilerExe: string;
   if (config.sys) {
@@ -100,8 +101,23 @@ export const patchedTsResolveModule = (
     let resolvedFileName = join(containingDir, moduleName);
     resolvedFileName = normalizePath(ensureExtension(resolvedFileName, containingFile));
 
-    // in some cases `inMemoryFs` will not be defined here, so we should use
-    // `accessSync` on `config.sys` instead.
+    // In some cases `inMemoryFs` will not be defined here, so we should use
+    // `accessSync` on `config.sys` instead. This is because this function is
+    // called by `patchTypeScriptResolveModule` which is then in turn called by
+    // `patchTypescript`. If you check out that function it takes an
+    // `InMemoryFileSystem` as its second parameter:
+    //
+    // https://github.com/ionic-team/stencil/blob/5b4bb06a4d0369c09aeb63b1a626ff8df9464117/src/compiler/sys/typescript/typescript-sys.ts#L165-L175
+    //
+    // but if you look at its call sites there are a few where we pass `null`
+    // instead, eg:
+    //
+    // https://github.com/ionic-team/stencil/blob/5b4bb06a4d0369c09aeb63b1a626ff8df9464117/src/compiler/transpile.ts#L42-L44
+    //
+    // so in short the type for `inMemoryFs` here is not accurate, so we need
+    // to add a runtime check here to avoid an error.
+    //
+    // TODO(STENCIL-728): fix typing of `inMemoryFs` parameter in `patchTypescript`, related functions
     const accessSync = inMemoryFs?.accessSync ?? config.sys.accessSync;
     if (isAbsolute(resolvedFileName) && !accessSync(resolvedFileName)) {
       return null;
--- a/src/compiler/transformers/rewrite-aliased-paths.ts
+++ b/src/compiler/transformers/rewrite-aliased-paths.ts
@@ -8,19 +8,20 @@ import { retrieveTsModifiers } from './transform-utils';
  * Transform module import paths aliased with `paths` in `tsconfig.json` to
  * relative imported in `.d.ts` files.
  *
- * @returns a TypeScript transformer factory
+ * @param transformCtx a TypeScript transformation context
+ * @returns a TypeScript transformer
  */
-export function rewriteAliasedDTSImportPaths(): ts.TransformerFactory<ts.Bundle | ts.SourceFile> {
-  return (transformCtx: ts.TransformationContext) => {
-    const compilerHost = ts.createCompilerHost(transformCtx.getCompilerOptions());
+export function rewriteAliasedDTSImportPaths(
+  transformCtx: ts.TransformationContext
+): ts.Transformer<ts.Bundle | ts.SourceFile> {
+  const compilerHost = ts.createCompilerHost(transformCtx.getCompilerOptions());
 
-    return (tsBundleOrSourceFile) => {
-      const fileName = ts.isBundle(tsBundleOrSourceFile)
-        ? tsBundleOrSourceFile.getSourceFile().fileName
-        : tsBundleOrSourceFile.fileName;
+  return (tsBundleOrSourceFile) => {
+    const fileName = ts.isBundle(tsBundleOrSourceFile)
+      ? tsBundleOrSourceFile.getSourceFile().fileName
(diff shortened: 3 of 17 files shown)


src/compiler/transformers/rewrite_aliased_paths.ts
```diff
@@ -0,0 +1,152 @@
...
+ * module which is not 1) not external (i.e. not a dependency) and 2) is not
+ * already a relative, file-path based import. It then replaces the module
+ * identifier with the relative path from the importer to the importee.
+ *
+ * @param compilerHost a TS compiler host
+ * @param transformCtx a TS transformation context
+ * @param sourceFilePath the path to the source file being visited
+ * @returns a visitor which takes a node and optionally transforms imports
+ */
+const visit = (compilerHost: ts.CompilerHost, transformCtx: ts.TransformationContext, sourceFilePath: string) => {
```
alicewriteswrongs, line 103: this 'visitor' is closely based on the one found in https://github.com/ionic-team/stencil/blob/main/src/compiler/transformers/map-imports-to-path-aliases.ts. It's basically just be modified to not be dependent on output target-specific stuff.

alicewriteswrongs (commented):
one little note

package.json
```diff
@@ -114,7 +114,7 @@
     "path-browserify": "^1.0.1",
     "pixelmatch": "5.3.0",
     "postcss": "^8.2.8",
-    "prettier": "2.8.3",
+    "prettier": "2.8.4",
```
tanner-reits, line 117: Not sure how this change found it's way into this PR 🤔

src/compiler/transformers/rewrite_aliased_paths.ts
```diff
@@ -0,0 +1,152 @@
+import { normalizePath } from '@utils';
```
tanner-reits, line 1: Not about this line, just a general comment about the file. Can we update the name of this file to be kebab-case instead of snake-case (i.e. `rewrite-aliased-paths.ts`)

src/compiler/transformers/rewrite_aliased_paths.ts
```diff
@@ -0,0 +1,152 @@
...
+
+import { retrieveTsModifiers } from './transform-utils';
+
+/**
+ * Transform modules aliased with `paths` in `tsconfig.json` to relative
+ * imported in `.d.ts` files.
+ *
+ * @returns a TypeScript transformer factory
+ */
+export function rewriteAliasedDTSPaths(): ts.TransformerFactory<ts.Bundle | ts.SourceFile> {
```
tanner-reits, line 13: Just a thought, but what do you think about a function name like `rewriteAliasedDTSImportPaths()`?

src/compiler/transformers/rewrite_aliased_paths.ts
```diff
@@ -0,0 +1,152 @@
...
+  };
+}
+
+/**
+ * Transform modules aliased with `paths` in `tsconfig.json` to relative
+ * imported in source files.
+ *
+ * @returns a TypeScript transformer factory
+ */
+export const rewriteAliasedSourceFilePaths = (): ts.TransformerFactory<ts.SourceFile> => {
```
tanner-reits, line 33: Same idea as above: `rewriteAliasedSourceFileImportPaths`

Also, just feels strange one of these is a `function` and one a `const`... makes no functional difference, just something I noticed

tanner-reits (changes requested):
Overall, things are looking good and I'm on board with doing these transformations globally (if they don't break any output we generate obviously). Called out a couple small things (mostly suggestions/thoughts).

I'll test this out on a few things locally too.

Do we need the specific transformer from #3523 anymore with this change? I guess that one would still have an impact if you set `transformAliasedImportPaths: false` in the Stencil config, but had `transformAliasedImportPaths: true` in the `dist-collection` output target config. We may want to think about that behavior a bit since it feels confusing to have the same flag/control available at different levels

package.json
rwaskiewicz: Probably a case of v3.1-feature being created before we updated Prettier, but this feature branch being created with it?

rwaskiewicz: But I agree, we should rm this (likely by rebasing the 3.1 feature branch on `main`, then this off the 3.1 feature branch)

alicewriteswrongs: oops haha

I should inspect the diff before I open the PR!

src/compiler/transformers/rewrite_aliased_paths.ts
alicewriteswrongs: sure, consistency is fine by me!

src/compiler/transformers/rewrite_aliased_paths.ts
alicewriteswrongs: It's a bit on the long side, but I'm fine with it 👍

alicewriteswrongs:
ok I had some scratch work on this branch about trying to do the same thing in tests, and when trying to remove that I messed up the rebase (oops) so the easiest thing was to just re-push as a single new commit - sorry for losing the history!

@tanner-reits I made the changes you suggested

alicewriteswrongs:
Also to answer your question, I believe if you have the global `transformAliasedImportPaths` set to true then yeah, the `dist-collection` one set to true (or false!) it would not do anything.

I think my preference would be to have one global option, since it's not clear to me that you'd want to set this differently on a per-output-target basis, but I think if we wanted to do that we would need to first mark the `dist-collection` specific option as deprecated and then remove it later on. Which is kind of a pain. But in any case, I didn't touch that code in this change because I didn't want to break any current behavior (for the same reason this is added as an opt-in behavior, rather than defaulting to doing this path transformation).

src/compiler/transformers/rewrite-aliased-paths.ts
```diff
@@ -0,0 +1,152 @@
...
+
+        const hasResolvedFileName = module.resolvedModule?.resolvedFileName != null;
+        const isModuleFromNodeModules = module.resolvedModule?.isExternalLibraryImport === true;
+        const shouldTranspileImportPath = hasResolvedFileName && !isModuleFromNodeModules;
+
+        if (shouldTranspileImportPath) {
+          // Create a regular expression that will be used to remove the last file extension
+          // from the import path
+          const extensionRegex = new RegExp(
+            Object.values(ts.Extension)
```
rwaskiewicz, line 126: This regex (I think) would look something like:
```
 .ts$|.tsx$|.d.ts$|.js$|.jsx$|.json$|.tsbuildinfo$|.mjs$|.mts$|.d.mts$|.cjs$|.cts$|.d.cts$
```

Which I think is fine, as we're using `ts.Extension` elsewhere in the codebase (the potential problem being that this enum is defined by TypeScript, and can change w/o us knowing). However, we've assumed that risk already, so I don't think that's much of a problem.

Does this current regex make sense for our purposes? I _think_ so, but wanted to double check with the team as a whole.

src/compiler/transpile/run-program.ts
```diff
@@ -50,10 +54,30 @@ export const runTsProgram = async (
...
 
-  // Emit files that changed
-  tsBuilder.emit(undefined, emitCallback, undefined, false, {
+  const transformers: ts.CustomTransformers = {
     before: [convertDecoratorsToStatic(config, buildCtx.diagnostics, tsTypeChecker)],
-  });
+    afterDeclarations: [],
+  };
+
+  if (config.transformAliasedImportPaths) {
```
rwaskiewicz, line 62: Can we add this to [`src/compiler/transpile/transpile-module.ts#transpileModule`](https://github.com/ionic-team/stencil/blob/main/src/compiler/transpile/transpile-module.ts#L121) as well for single string compilation?

rwaskiewicz (changes requested):
Had a few questions/clarifications - once those are resolved, can we add:

- Documentation in the `next` directory of docs
- Notion page for flags that we may flip in the future for a major version of Stencil (like we did for v3?)
- Tests for `rewrite-aliased-paths`
  - I _think_ using the `test/transpile` fn, we could just pass the new  transformer to the `transpileModule()` fn

src/compiler/transpile/run-program.ts
alicewriteswrongs: Sure, that makes sense to me

src/compiler/transformers/rewrite-aliased-paths.ts
```diff
@@ -0,0 +1,152 @@
...
+ *
+ * ```js
+ * // dist/importing.js
+ * import { myUtil } from "./utils";
+ * ```
+ *
+ * The TypeScript team have stated pretty unequivocally that they will not
+ * automatically resolve these identifiers to relative paths in output code
+ * {@see https://github.com/microsoft/TypeScript/issues/10866} and have
+ * said that resolving these module identifiers is the responsability of module
```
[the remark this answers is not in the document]
alicewriteswrongs: man I do that one all the time

idk why, I feel like it should be response-ability I guess haha

src/compiler/transformers/rewrite-aliased-paths.ts
```diff
@@ -0,0 +1,152 @@
...
+      return transformCtx.factory.updateImportDeclaration(
+        node,
+        retrieveTsModifiers(node),
+        node.importClause,
+        transformCtx.factory.createStringLiteral(importPath),
+        node.assertClause
+      );
+    }
+
+    return ts.visitEachChild(node, visit(compilerHost, transformCtx, sourceFilePath), transformCtx);
```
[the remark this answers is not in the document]
alicewriteswrongs: I poked around a bit and found that the following (contrived) example does seem to be valid TypeScript:

```ts
namespace Foo {
    export function foobar () {
        return 2;
    }
}

namespace Bar {
    import foo = Foo;

    export const baz = foo.foobar;
}
```

([TS playground link](https://www.typescriptlang.org/play?#code/HYQwtgpgzgDiDGEAEAxA9mpBvAUE-SEAHjGgE4AuSAZgK7DwUCWawNGARiGUgBQCU2PARFkIFWmTYAmANzD8AXxzKcoSLATIAQtyEimYUpXaYAvKgzyFhEuSrxWUKlwBeSC9QwA6L2i5k8opAA) for that snippet)

_However_, if you take a look at [the AST for that snippet](https://ts-ast-viewer.com/#code/HYQwtgpgzgDiDGEAEAxA9mpBvAUE-SEAHjGgE4AuSAZgK7DwUCWawNGARiGUgBQCU2PARFkIFWmTYAmANzD8AXxzKcoSLATIAQtyEimYUpXaYAvKgzyFhEuSrxWUKlwBeSC9QwA6L2i5k8opAA), happily the `import foo = Foo;` bit is not a `ts.ImportDeclaration`, but a `ts.ImportEqualsDeclaration`, and I don't think that TypeScript will allow a normal import declaration within a namespace (I could be wrong though). So that would suggest we could get by with filtering on the top-level statements in a `ts.SourceFile`.

Screenshot 2023-02-15 at 2 30 53 PM

In the future, however, if the [module declarations proposal](https://github.com/tc39/proposal-module-declarations) get accepted as part of the ECMAScript standard (it's [currently stage 2](https://github.com/tc39/proposals#stage-2)) then I think it would be possible to have something like this:

```ts
module Foo {
  import fs from 'fs';
  export const myFileSync = fs.readFileSync;
}
```

so in the interests of future-proofing we might want to keep the recursion around.

Thoughts?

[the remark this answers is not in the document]
rwaskiewicz: I struggle with this decision (or, my opinion on it 😆). IIUC, the module declarations proposal just hit Stage 2 this past November - I think I'd err on the side of [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) - and if we _do_ need it in the future:
1. This probably won't be the only change we'll need to make, and I can see value in looking at the solution for module declarations holistically
2. We have git history, and can recover this original impl and reuse it if needed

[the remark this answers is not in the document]
alicewriteswrongs: I think that makes sense, it's not going to be that difficult retrofit recursion back in if we turn out to need it.

I'm going to finish tests for the current implementation first and then look at swapping things out, just to have some assurance that things are working as I think they are.

src/compiler/transformers/rewrite-aliased-paths.ts
alicewriteswrongs: I think this should work alright for us, and I'd rather that the TypeScript devs maintain it than us I think!

src/compiler/transformers/test/rewrite-aliased-paths.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+      '@namespace/subdir': [path.join(config.rootDir, 'name/space/subdir.ts')],
+    },
+    declaration: true,
+  };
+
+  // we need to have files in the `inMemoryFs` which TypeScript
+  // can resolve, otherwise it won't find the module and won't
+  // transform the module ID
+  await compilerContext.fs.writeFile(path.join(config.rootDir, 'name/space.ts'), 'export const foo = x => x');
+  await compilerContext.fs.writeFile(path.join(config.rootDir, 'name/space/subdir.ts'), 'export const bar = x => x;');
```
alicewriteswrongs, line 25: in order for these tests to work TypeScript needs to be able to resolve the files we're importing from, so here in this little helper function we instantiate an in-memory FS and write two little modules to it. TypeScript is patched to read from the in-memory FS, so when we try to resolve modules in our transformer it will see these.

src/compiler/transformers/test/transpile.ts
```diff
@@ -15,53 +17,67 @@ import { getScriptTarget } from '../transform-utils';
  * @param compilerCtx a compiler context to use in the transpilation process
  * @param beforeTransformers TypeScript transformers that should be applied before the code is emitted
  * @param afterTransformers TypeScript transformers that should be applied after the code is emitted
+ * @param afterDeclarations TypeScript transformers that should be applied
+ * after declarations are generated
+ * @param tsConfig optional typescript compiler options to use
```
alicewriteswrongs, line 22: I had to add these two options to 1) set the `afterDeclarations` transformers and 2) override some of the TS compiler options set before the transpilation happens.

src/compiler/transformers/test/transpile.ts
```diff
@@ -15,53 +17,67 @@ import { getScriptTarget } from '../transform-utils';
...
+    outFile: undefined,
+    paths: undefined,
+    removeComments: false,
+    rootDirs: undefined,
+    suppressOutputPathCheck: true,
+    target: getScriptTarget(),
+    types: undefined,
+    // add in possible default config overrides
+    ...tsConfig,
+  };
```
alicewriteswrongs, line 62: this is just the same as what was here before just alphabetized, reformatted as an object literal, and adding in the `tsConfig` param.

src/compiler/transformers/test/transpile.ts
```diff
@@ -129,32 +141,43 @@ export function transpileModule(
...
-    events,
-    event,
-    methods,
-    method,
-    elementRef,
-    legacyContext,
-    legacyConnect,
+    states,
+    tagName,
+    virtualProperties,
```
alicewriteswrongs, line 167: alphabetization here too, only addition is `declarationOutputText`

alicewriteswrongs (commented):
just added tests, pointing out a few things with them

src/compiler/transformers/rewrite-aliased-paths.ts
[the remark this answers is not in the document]
alicewriteswrongs: alright so I just pushed a change that did a bit of a refactor on the `rewrite-aliased-paths` module to avoid recurring down the whole tree, lmk what you think!

rwaskiewicz:
@alicewriteswrongs can you take a look at the unit tests again? Looks like they're failing in CI ATM

alicewriteswrongs:
ah looks like there's something that doesn't work correctly in node 14. I installed v14.21.3 locally and I can make the tests pass. I'll try to do some `git bisect` to figure out what's up

alicewriteswrongs:
I'm not sure what's going on, I checked out `main` and changed the node version to `14.21.3` and got the same test failures when running the unit tests there, something strange is afoot here...

alicewriteswrongs:
definitely funky, so the tests on this branch all pass 100% (at least locally on my machine) with node v14.21.0, but if I use v14.21.1 I get test failures similar to the ones observable in the CI.

The strange thing is, it looks like 14.21.1 was released in November: https://nodejs.org/en/download/releases/ (2022-11-04 to be specific) so I'm a little puzzled why we wouldn't have seen these issues before 🤔

alicewriteswrongs:
actually scratch that, something is wrong on this branch - I wasn't doing a `jest --clearCache` in between runs so some cached stuff was hanging around and causing problems.

alicewriteswrongs:
alright sorry for all the noise - turns out this was a probably with too-eagerly-enabling the transformer in `transpileModule`, I added a flag to the transpilation options to control it and that should address the issue

alicewriteswrongs:
Looks like there is still a test failure but this time only on Windows, trying to dig into that a bit more. I'll ping when that's resolved.

alicewriteswrongs:
Hard to track down windows-specific failures without windows!

I think I've cleared it up now, should be ready for re-review

src/compiler/transformers/test/parse-events.spec.ts
```diff
@@ -105,7 +108,7 @@ describe('parse events', () => {
       references: {
         Mode: {
           location: 'local',
-          path: 'module.tsx',
+          path: IS_WINDOWS_ENV ? normalizePath('D:/module.tsx') : normalizePath('/module.tsx'),
```
rwaskiewicz, line 111: At first glance, I'm not sure I fully understand why we had to make this change. Is this due to [this change](https://github.com/ionic-team/stencil/pull/4042/files#diff-94c5db3ee58dfccae212efa88a639467f0838e1021abe6cc20cd5b2fdef7a86bR67)?

src/compiler/transformers/test/rewrite-aliased-paths.spec.ts
```diff
@@ -0,0 +1,134 @@
+import { CompilerCtx } from '@stencil/core/declarations';
+import { mockCompilerCtx, mockValidatedConfig } from '@stencil/core/testing';
+import path from 'path';
+import ts from 'typescript';
+
+import { patchTypescript } from '../../sys/typescript/typescript-sys';
+import { rewriteAliasedDTSImportPaths, rewriteAliasedSourceFileImportPaths } from '../rewrite-aliased-paths';
+import { transpileModule } from './transpile';
+
+async function pathTransformTranspile(component: string) {
```
rwaskiewicz, line 10: Can we JSDoc this helper function please?

src/compiler/sys/typescript/typescript-resolve-module.ts
```diff
@@ -100,7 +100,10 @@ export const patchedTsResolveModule = (
     let resolvedFileName = join(containingDir, moduleName);
     resolvedFileName = normalizePath(ensureExtension(resolvedFileName, containingFile));
 
-    if (isAbsolute(resolvedFileName) && !inMemoryFs.accessSync(resolvedFileName)) {
+    // in some cases `inMemoryFs` will not be defined here, so we should use
+    // `accessSync` on `config.sys` instead.
+    const accessSync = inMemoryFs?.accessSync ?? config.sys.accessSync;
```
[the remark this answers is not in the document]
alicewriteswrongs: Yeah, so this function will be called by `patchTypeScriptResolveModule` which is then in turn called by `patchTypescript`. If you check out that function it takes an `InMemoryFileSystem` as its second parameter:

https://github.com/ionic-team/stencil/blob/5b4bb06a4d0369c09aeb63b1a626ff8df9464117/src/compiler/sys/typescript/typescript-sys.ts#L165-L175

but if you look at its call sites there are a few where we pass `null` instead, eg:

https://github.com/ionic-team/stencil/blob/5b4bb06a4d0369c09aeb63b1a626ff8df9464117/src/compiler/transpile.ts#L42-L44

and

https://github.com/ionic-team/stencil/blob/5b4bb06a4d0369c09aeb63b1a626ff8df9464117/src/compiler/transpile.ts#L74-L76

I'm not sure exactly _why_ that's the case, but I did try making a change to ensure that at those two call sites it was called with an `InMemoryFileSystem` instead, and I ran into some module resolution errors and other things. So there may be a good reason that it's that way, but it does mean that unfortunately in the call tree underneath `patchTypescript` we can't count on an `InMemoryFileSystem` being there.

I'll expand the comment here to that effect. I was planning to circle around to fix this up, just made a JIRA ticket for it (STENCIL-728) and I'll add TODO's in some of the places I noticed.

src/compiler/transformers/test/parse-events.spec.ts
alicewriteswrongs: Yes, although I just realized there's a better approach which removes the need for this

src/compiler/config/transpile-options.ts
```diff
@@ -40,7 +40,13 @@ export const getTranspileResults = (code: string, input: TranspileOptions) => {
 
 const transpileCtx = { sys: null as CompilerSystem };
 
-export const getTranspileConfig = (input: TranspileOptions) => {
+interface TranspileConfig {
```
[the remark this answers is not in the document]
rwaskiewicz: Minor - unaddressed comment

src/compiler/config/transpile-options.ts
```diff
@@ -40,7 +40,13 @@ export const getTranspileResults = (code: string, input: TranspileOptions) => {
...
+  compileOpts: TranspileOptions;
+  config: Config;
+  transformOpts: TransformOptions;
+}
+
+export const getTranspileConfig = (input: TranspileOptions): TranspileConfig => {
```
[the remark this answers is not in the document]
rwaskiewicz: Minor - unaddressed comment

rwaskiewicz:
@alicewriteswrongs I think there are a few outstanding comments (minor JSDoc ones). It also looks like lint/format is failing in CI

src/compiler/config/transpile-options.ts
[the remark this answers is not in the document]
alicewriteswrongs: oops, sorry I missed this

rwaskiewicz:
LGTM - Final approval contingent on
Had a few questions/clarifications - once those are resolved, can we add:

- Documentation in the `next` directory of docs
- Notion page for flags that we may flip in the future for a major version of Stencil (like we did for v3?)
- Validation with Ionic Framework team

tanner-reits (approved):
LGTM. I like it!

alicewriteswrongs:
Opened a PR for the documentation changes here: https://github.com/ionic-team/stencil-site/pull/1046

rwaskiewicz (approved):

PR #4090, opened 2023-02-22
feat(testing): add support for transforming path aliases in spec tests

### LINKED ISSUE #3826: bug: tsconfig path does not work with spec tests
### Prerequisites

### Stencil Version

v2.19.3

### Current Behavior

Running a spec test on a component that imports something from a path define in tsconfig results in a "Cannot find module" error:

`Cannot find module '@utils/utils' from 'src/components/my-component/my-component.tsx'`

### Expected Behavior

I expect the tsconfig path to be respected when running spec tests.

### Steps to Reproduce

1. Clone repo.
2. Run `npm install`.
3. Run `npm run test`.
4. Observe that the following error is logged:

```
  ● Test suite failed to run

    Cannot find module '@utils/utils' from 'src/components/my-component/my-component.tsx'

    Require stack:
      src/components/my-component/my-component.tsx
      src/components/my-component/my-component.spec.ts

      1 | import { Component, Prop, h } from '@stencil/core';
    > 2 | import { format } from '@utils/utils';
        | ^
      3 |
      4 | @Component({
      5 |   tag: 'my-component',

      at Resolver.resolveModule (node_modules/jest-resolve/build/resolver.js:324:11)
      at Object.<anonymous> (src/components/my-component/my-component.tsx:2:1)
```

### Code Reproduction URL

https://github.com/liamdebeasi/blank-stencil-repro/tree/tsconfig-test

### Additional Information

_No response_

alicewriteswrongs:
This adds support for transforming path aliases in spec tests, using the transformer which was added in #4042. Adding such support also revealed a bug with the existing path transformer, notably that it did not produce a valid path for sibling modules. See the diff for a comment explaining the details.

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Imports from modules aliases with `paths` in `tsconfig.json` don't work in spec tests.

GitHub Issue Number: fixes #3826

## What is the new behavior?

They do! This change reads the `transformAliasedImportPaths` option from the Stencil config and, if it is set to true, will transform such aliased modules to relative paths when the transpiler is being run by Jest. If the user does not set that config value to `true` then things should be as they are.

- [ ] Yes
- [x] No

## Testing

There is a reproduction case here: https://github.com/liamdebeasi/blank-stencil-repro/tree/tsconfig-test

Confirm that it breaks:

- as-is
- with this change but without the `transformAliasImportPaths` option set to `true`

and confirm that if you set that option to `true` it will all work correctly. To test the sibling path problem mentioned in a comment on the PR you can apply the following change:

```diff
diff --git a/src/utils/utils.spec.ts b/src/utils/utils.spec.ts
index 81ff970..a7e2938 100644
--- a/src/utils/utils.spec.ts
+++ b/src/utils/utils.spec.ts
@@ -1,4 +1,4 @@
-import { format } from './utils';
+import { format } from '@utils/utils';
 
 describe('format', () => {
   it('returns empty string for no names defined', () => {
```

### DIFF
--- a/src/compiler/sys/typescript/typescript-sys.ts
+++ b/src/compiler/sys/typescript/typescript-sys.ts
@@ -12,7 +12,7 @@ import { patchTypeScriptResolveModule } from './typescript-resolve-module';
 export const patchTsSystemFileSystem = (
   config: d.Config,
   compilerSys: d.CompilerSystem,
-  inMemoryFs: InMemoryFileSystem | null,
+  inMemoryFs: InMemoryFileSystem,
   tsSys: ts.System
 ): ts.System => {
   const realpath = (path: string) => {
--- a/src/compiler/sys/typescript/typescript-sys.ts
+++ b/src/compiler/sys/typescript/typescript-sys.ts
@@ -12,7 +12,7 @@ import { patchTypeScriptResolveModule } from './typescript-resolve-module';
 export const patchTsSystemFileSystem = (
   config: d.Config,
   compilerSys: d.CompilerSystem,
-  inMemoryFs: InMemoryFileSystem,
+  inMemoryFs: InMemoryFileSystem | null,
   tsSys: ts.System
 ): ts.System => {
   const realpath = (path: string) => {
@@ -55,8 +55,13 @@ export const patchTsSystemFileSystem = (
   };
 
   tsSys.directoryExists = (p) => {
-    const s = inMemoryFs.statSync(p);
-    return s.isDirectory;
+    if (inMemoryFs) {
+      const s = inMemoryFs.statSync(p);
+      return s.isDirectory;
+    } else {
+      const s = compilerSys.statSync(p);
+      return s.isDirectory;
+    }
   };
 
   tsSys.exit = compilerSys.exit;
@@ -68,8 +73,13 @@ export const patchTsSystemFileSystem = (
       filePath = getTypescriptPathFromUrl(config, tsSys.getExecutingFilePath(), p);
     }
 
-    const s = inMemoryFs.statSync(filePath);
-    return !!(s && s.isFile);
+    if (inMemoryFs) {
+      const s = inMemoryFs.statSync(filePath);
+      return !!(s && s.isFile);
+    } else {
+      const s = compilerSys.statSync(filePath);
+      return !!(s && s.isFile);
+    }
   };
 
   tsSys.getCurrentDirectory = compilerSys.getCurrentDirectory;
@@ -79,8 +89,13 @@ export const patchTsSystemFileSystem = (
   tsSys.getDirectories = (p) => {
     const items = compilerSys.readDirSync(p);
     return items.filter((itemPath) => {
-      const s = inMemoryFs.statSync(itemPath);
-      return !!(s && s.exists && s.isDirectory);
+      if (inMemoryFs) {
+        const s = inMemoryFs.statSync(itemPath);
+        return !!(s && s.exists && s.isDirectory);
+      } else {
+        const s = compilerSys.statSync(itemPath);
+        return !!(s && s.isDirectory);
+      }
     });
   };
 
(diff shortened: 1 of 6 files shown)


src/compiler/transformers/rewrite-aliased-paths.ts
```diff
@@ -167,7 +167,37 @@ function rewriteAliasedImport(
...
+      // ```
+      //
+      // you'd think that in this case `relative` would return `'./bar.ts'` as
+      // a correct relative path to `bar.ts` from the `foo` directory, but
+      // actually in this case it returns just `bar.ts`. So since when updating
+      // import paths we're only concerned with `paths` aliases that should be
+      // transformed to relative imports anyway, we check to see if the new
+      // `importPath` starts with `'.'`, and add `'./'` if it doesn't, since
+      // otherwise nodejs will interpret `bar` as a module name, not a relative
+      // path.
```
rwaskiewicz, line 198: Good stuff 👍 

It might be worth throwing in this comment that we assume we have already dealt with other relative paths ([here](https://github.com/ionic-team/stencil/pull/4090/files#diff-42aa79700f64320ff179029d36e22035fa882b37bd80b8ef4aec7361759a7f60R138-R142))

src/compiler/sys/typescript/typescript-sys.ts
```diff
@@ -55,8 +55,13 @@ export const patchTsSystemFileSystem = (
   };
 
   tsSys.directoryExists = (p) => {
-    const s = inMemoryFs.statSync(p);
-    return s.isDirectory;
+    if (inMemoryFs) {
```
rwaskiewicz, line 58: I'm not entirely sure I understand why we need to fallback to `compilerSys` for this PR based on the existing comments/PR summary. Can you explain that/add it to either please?

rwaskiewicz (changes requested):

src/compiler/sys/typescript/typescript-sys.ts
alicewriteswrongs: yes, sorry should have probably left on draft while I was sorting just that out

alicewriteswrongs: The reason basically has to do with these calls to `patchTypescript`, here:

https://github.com/ionic-team/stencil/blob/f0781d29160d794fcd1315c2631a286437d71fb7/src/compiler/transpile.ts#L43-L43

and here:

https://github.com/ionic-team/stencil/blob/f0781d29160d794fcd1315c2631a286437d71fb7/src/compiler/transpile.ts#L75

Because we pass in `null` here we actually can't count on the `inMemoryFs` argument to `patchTsFileSystem` being defined in here :(

I was considering addressing STENCIL-728 as part of this PR but thought it better to save that for another, however I did have to make these edits in `typescript-sys` because these methods are called by TypeScript when transpiling a spec file.

alicewriteswrongs: I added some comments to this file (as well as TODOs for STENCIL-728) just noting why we need to make the check

src/testing/jest/jest-runner.ts
```diff
@@ -13,6 +13,7 @@ export async function runJest(config: d.ValidatedConfig, env: d.E2EProcessEnv) {
     const emulateConfigs = getEmulateConfigs(config.testing, config.flags);
     env.__STENCIL_EMULATE_CONFIGS__ = JSON.stringify(emulateConfigs);
     env.__STENCIL_ENV__ = JSON.stringify(config.env);
+    env.__STENCIL_TRANSPILE_PATHS__ = !!config.transformAliasedImportPaths;
```
alicewriteswrongs, line 16: this is admittedly funky but I didn't see another clean way to get a Stencil config value into the Jest preprocessor, and it seems like we do this already for some things.

Anywho - basically the intent is just to get the value of `transformAliasedImportPaths` into the Jest context so we can set the corresponding option correctly on the options object passed in to the transpilation functions.

tanner-reits (approved):
LGTM! The preprocessor stuff is a bit outside my understanding of the test infrastructure still, so I'll defer to Ryan on that. Good catch with the sibling imports too!

alicewriteswrongs:
@rwaskiewicz just pushed those changes we just went through, and confirm in the test case that toggling `transformAliasedImportPaths` on and off will cause the `@utils/` import to work or not, respectively, in the reproduction case

rwaskiewicz (approved):

rwaskiewicz:
@alicewriteswrongs 🚢

Issue #3828, opened 2022-11-17
bug: CLI Jest `-u` argument is not passed


duhem-s:
**Stencil version:**
```
 @stencil/core@2.13.0
```

**Current behavior:**
When I run the `test` script, I can not pass anymore the `-u` argument to update my snapshots.

**Expected behavior:**
Being able to update my snapshot using the CLI.

**GitHub Reproduction Link:**
https://github.com/duhem-s/stencil-cli

**Other information:**
Using the command `npm run test -u` with stencil@2.12.1 :
![image](https://user-images.githubusercontent.com/113420581/202398737-1545ad8a-7267-4659-825e-706a5cda273f.png)

With stencil@2.13.0 :
![image](https://user-images.githubusercontent.com/113420581/202399382-12f0acc3-c9b0-4159-9729-a26353957382.png)

tanner-reits:
Hey @duhem-s 👋 

Thanks for reporting this issue and providing a reproduction! I hopped in and took a peek at the problem you encountered across a couple different versions of Stencil (2.12.1, 2.13.0, and 2.19.3 (latest)). Definitely some inconsistencies between the versions. So, I'll get this pulled into our backlog for us to take a look at!

In the meantime, if you're running Stencil 2.13.0, I was able to get the snapshots to update in the reproduction repo using the un-aliased Jest CLI flag:
`npm run test -- --updateSnapshot`

Hopefully that can keep you moving for now!

rwaskiewicz:
@duhem-s Can you do us a favor and let us know what versions of node and npm you're using? From your project's directory, run:
```
npm --version
node --version
```
Thanks!

duhem-s:
Hi @rwaskiewicz !

I'm using npm@6.14.17 and node@14.19.3

alicewriteswrongs:
Hey @duhem-s, we now have a PR which I believe fixes this problem (#4124). I just published a dev build with those changes which could be installed like so:

```
npm install @stencil/core@3.1.0-dev.1678201586.203c481
```

after installing it in your reproduction case I'm able to do

```
npm test -- -u
```

and the snapshots update!

Would you mind trying that out to confirm that it works for you as well?

rwaskiewicz:
The PR associated with this issue has landed and is included in today's [v3.2.0 release](https://github.com/ionic-team/stencil/releases/tag/v3.2.0). Should this issue reappear, please feel free to open a new issue. Thanks again!

PR #4124, opened 2023-03-03
fix(cli): support Jest-specific CLI flag aliases

alicewriteswrongs:
This adds support for the following Jest-specific CLI flag aliases:

- "b" for "bail"
- "e" for "expand"
- "w" for "maxWorkers"
- "o" for "onlyChanged"
- "f" for "onlyFailures"
- "i" for "runInBand"
- "t" for "testNamePattern"
- "u" for "updateSnapshot"

Two additional flags supported by Jest ('c' for 'config' and 'h' for 'help') were already supported by Stencil.

This also makes a small refactor to the CLI argument parser, mainly building up the `flags.unknownArgs` array proactively whenever we're in a situation where we don't know what to do, rather than filtering all the arguments after the fact.

## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Stencil supports a lot of jest arguments, but Jest's particular aliases are not supported. So, for instance, the `-u` shortcut for `updateSnapshot` (the issue called out in #3828) doesn't work.

GitHub Issue Number: #3828 

## What is the new behavior?

This adds Jest-supported aliases to our list of supported aliases, and makes a few tweaks to the CLI parser to support parsing those flags correctly.

- [ ] Yes
- [x] No

## Testing

We have some pretty extensive testing around the parser, so I'm fairly confident in the changes there. However! It is really a good idea to test this out, especially in the repro case on #3828 to make sure things work as they should.


src/cli/config-flags.ts
```diff
@@ -213,13 +213,25 @@ export const CLI_FLAG_ALIASES: AliasMap = {
   h: 'help',
   p: 'port',
   v: 'version',
+
+  // JEST SPECIFIC CLI FLAGS
+  // these are defined in
+  // https://github.com/facebook/jest/blob/4156f86/packages/jest-cli/src/args.ts
```
alicewriteswrongs, line 219: I pulled these out of this file by running the following (hacky!) script:

```js
const { readFileSync } = require('fs');

// this is a commonjs module, so the stuff in it will be stuck on `exports`
eval(String(readFileSync('./node_modules/jest-cli/build/cli/args.js')));

const aliases = {}

Object.entries(exports.options).forEach(([k,v]) => {
  if (typeof v.alias === 'string') {
    aliases[v.alias] = k
  }
})

console.log(JSON.stringify(aliases))
```

alicewriteswrongs: (in the Stencil repo, but this would work in any npm project where Jest is installed)

src/cli/config-flags.ts
```diff
@@ -213,13 +213,25 @@ export const CLI_FLAG_ALIASES: AliasMap = {
...
+  t: 'testNamePattern',
+  u: 'updateSnapshot',
 };
 
 /**
  * A regular expression which can be used to match a CLI flag for one of our
  * short aliases.
  */
-export const CLI_FLAG_REGEX = new RegExp(`^-[chpv]{1}$`);
+export const CLI_FLAG_REGEX = new RegExp(`^-[chpvbewofitu]{1}$`);
```
alicewriteswrongs, line 234: note the update to the CLI flag regex here - I dislike this, but [as noted previously](https://github.com/ionic-team/stencil/pull/3765#discussion_r1009465356), the obvious thing you'd like to do here is the following:


```suggestion
export const CLI_FLAG_REGEX = new RegExp(`^-[${
  Object.keys(CLI_FLAG_ALIASES).join("")
}]{1}$`);
```

but unfortunately that causes some tree-shaking issues because (I speculate) Rollup can't determine that if nothing imports this value it will be Ok to remove both it and `CLI_FLAG_ALIASES` from the bundle. Bah!

src/cli/parse-flags.ts
```diff
@@ -42,14 +42,6 @@ export const parseFlags = (args: string[]): ConfigFlags => {
     }
   }
 
-  // to find unknown / unrecognized arguments we filter `args`, including only
-  // arguments whose normalized form is not found in `knownArgs`. `knownArgs`
-  // is populated during the call to `parseArgs` above. For arguments like
-  // `--foobar` the string `"--foobar"` will be added, while for more
-  // complicated arguments like `--bizBoz=bop` or `--bizBoz bop` just the
-  // string `"--bizBoz"` will be added.
-  flags.unknownArgs = flags.args.filter((arg: string) => !flags.knownArgs.includes(parseEqualsArg(arg)[0]));
```
alicewriteswrongs, line 51: I decided to remove this filter here and to instead proactively build up the `unknownArgs` array as we're parsing. You'll see below that when we get into the `else` clauses of both `parseCLITerm` and `setCLIArg` there is new code to stick any argument which we're not handling into `flags.unknownArgs`.

Many tests cases which look for the presence or absence of values in this array continue to pass with these changes, so I believe they are correct to do.

But why do them?

Well, take a look at this filtering code. In `main` right now we're doing:

```ts
flags.unknownArgs = flags.args.filter((arg: string) => !flags.knownArgs.includes(parseEqualsArg(arg)[0]));
```

Basically, we declare the `unknownArgs` to be any arguments in the `args` where a 'normalized' version of the arg as calculated by `parseEqualsArg` isn't found in `knownArgs`.

This breaks down for handling aliased Jest args, in particular because I found that in order to reliably pass these arguments to Jest it is best to pass the 'desugared'[^1] form of the alias to Jest (through the `knownArgs`) rather than just the alias.

So in order to get that to work we want to make sure that the argument we're pushing onto `knownArgs` is full argument name, but this means that the normalized form of the _raw argument_ (i.e. the flag as set at the CLI by the user) will no longer be present in the `knownArgs` array, so this filter isn't reliable anymore.

As an example, if I do

```sh
npx stencil test --spec -u
```

that will result in a `knownArgs` that looks like

```ts
[ "--unknownArgs" ]
```

rather than

```ts
[ "-u" ]
```

Please let me know if that doesn't make sense, it's a bit hard to explain well!

[^1]: My  overloaded term for the full config flag name that an alias refers to, so e.g. `'-c'` will desugar to `'--config'`, `'-u'` to `'--updateSnapshot'`, etc. Also by 'best' I mean that without desugaring these aliases there are difficult-to-pin-down circumstances under which they are picked up by Jest _and_ circumstances under which they _aren't_. I stopped investigating once I realized that passing the desugared form always works.

src/cli/parse-flags.ts
```diff
@@ -454,22 +457,35 @@ const throwNumberParsingError = (flag: string, value: string) => {
...
- * passed string unmodified.
+ * If no expansion is found for the possible alias we just return the passed
+ * string unmodified.
  *
  * @param maybeAlias a string which _could_ be an alias to a full flag name
  * @returns the full aliased flag name, if found, or the passed string if not
  */
-const dereferenceAlias = (maybeAlias: string): string => {
-  const possibleDereference = CLI_FLAG_ALIASES[maybeAlias];
+const desugarAlias = (maybeAlias: string): string => {
```
alicewriteswrongs, line 469: I renamed this because I think desugaring is a more apt analogy than dereferencing

alicewriteswrongs (commented):
pointing out a few things

src/cli/parse-flags.ts
```diff
@@ -320,6 +314,14 @@ const setCLIArg = (flags: ConfigFlags, rawArg: string, normalizedArg: string, va
...
       throwCLIParsingError(rawArg, 'expected to receive a valid log level but received nothing');
     }
+  } else {
+    // we haven't found this flag in any of our lists of arguments, so we
+    // should put it in our list of unknown arguments
+    flags.unknownArgs.push(rawArg);
+
+    if (typeof value === 'string') {
+      flags.unknownArgs.push(value);
+    }
```
tanner-reits, line 324: Do we prevent duplicate args either up or down stream? Or do we just not care and let the respective CLI handle that (i.e. Jest can filter those out)?

tanner-reits (approved):
LGTM. Left a non-blocking question

rwaskiewicz (approved):
One non-block ask

src/cli/parse-flags.ts
```diff
@@ -164,13 +156,15 @@ const parseCLITerm = (flags: ConfigFlags, args: string[]) => {
   // SimpleArg → "--" ArgName ( " " CLIValue )? ;
   else if (arg.startsWith('--') && arg.length > '--'.length) {
     setCLIArg(flags, arg, normalizeFlagName(arg), parseCLIValue(args));
+  } else {
+    // if we get here the argument w is not an argument in our list of
```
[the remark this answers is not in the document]
alicewriteswrongs: not sure how that snuck in tbh 🤔

PR #4173, opened 2023-03-16
chore(release): separate prompts from release tasks

rwaskiewicz:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

We're in the process of enabling CI-based deployments. Today, this is a manual process that (at a high level), requires two steps:
1. Preparing the release - where we bump the version, generate the changelog, run the tests, etc.
2. Run the release - publish the commit to GitHub, push tags to GH+NPM, create a GH release, etc.

As a part of allowing publishes to occur in CI, those two steps need to be combined. However, the second step ('Run the release') requires a user to answer a few prompts in a terminal window. This limits our options for code reuse when we programmatically prepare a release.

This commit assumes that https://github.com/ionic-team/stencil/pull/4172 has landed, which solve the same problem for the pre-release step

GitHub Issue Number: N/A

## What is the new behavior?

this commit separates the existing prompts (questions asked to an
engineer during the release process) that expect an interactive
terminal to be available. this is the second step that we'll take
in order to enable ci to publish stencil, rather than it being a manual
process. the motivation behind this change is that we would like to
reuse the existing release logic. however, in order to do so, we
must separate the prompting (which won't be available in ci) from the
necessary logic to initiate a release.

- [ ] Yes
- [x] No

## Testing

1. Unit tests pass
2. `npm run release.prepare -- --any-branch --dry-run`, then `npm run release -- --any-branch-dry-run` will test _most_ of the functionality on this branch. The version validation, git checks (tag, that the local branch is clean/has an upstream) will be skipped due to the `dry-run` flag. Running this, one should observe that the changelog, package.json, and package-lock.json all get bumped. Nothing will get pushed to NPM/github

### Workflow isn't Optimal

The steps in 'release' aren't "optimal"/a little confusing. I've added tests to add some clarification, but these scripts were written with mutating arguments in mind. I did not attempt to "fix" all of these in one PR. The focus here is to separaterelease prompts from the `release` function, so that we can do both that cleanup and get CI publishes one step closer to a reality

### Why not use `np`?

When we're writing tests for something like our release scripts over several hack days, it's a valid question to ask. The Stencil prod build pipeline relies on switching a lot of generated values out of the bundle. While I think it _could_ be feasible with `np`, I held off for two reasons:
1. To better understand the pre-existing steps we have. Otherwise, I wouldn't know what I was replacing `np` with 😆 
2. To avoid shoehorning our process into `np` prematurely. I'm preferring running the steps we run today (locally) in CI (eventually, that's the general plan), over a whole new pipeline. Once dev/nightly/prod are online and we have all their use cases accounted for, we can reconsider.

### DIFF
--- a/scripts/prompts.ts
+++ b/scripts/prompts.ts
@@ -2,7 +2,13 @@ import color from 'ansi-colors';
 import inquirer from 'inquirer';
 
 import { BuildOptions } from './utils/options';
-import { getNewVersion, isValidVersionInput, prettyVersionDiff, SEMVER_INCREMENTS } from './utils/release-utils';
+import {
+  getNewVersion,
+  isPrereleaseVersion,
+  isValidVersionInput,
+  prettyVersionDiff,
+  SEMVER_INCREMENTS,
+} from './utils/release-utils';
 
 /**
  * A type describing the answers to prompts for preparing a release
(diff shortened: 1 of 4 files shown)


scripts/prompts.ts
```diff
@@ -87,3 +93,115 @@ export async function promptPrepareRelease(opts: BuildOptions): Promise<PrepareR
   }
   return answers;
 }
+
+/**
+ * A type describing the answers to prompts for performing a release
+ */
+export type ReleasePromptAnswers = {
```
tanner-reits, line 100: Just a thought: Might be nice to break these types out into a separate file now that there are two type definitions and function definitions

tanner-reits (approved):
One non-blocking thought

alicewriteswrongs (approved):
this all looks straightforward enough to me 👍

PR #4187, opened 2023-03-21
chore(build): build the CLI module with esbuild

alicewriteswrongs:
This adds a support for building Stencil with [esbuild](https://esbuild.github.io/). This comprises:

- a new subdirectory `scripts/esbuild` which holds all the esbuild-related stuff
- support for bundling the `cli/` bundle with esbuild
- new `package.json` scripts `build.esbuild` and `build.esbuild.watch` for doing a "hybrid" build
- new CI workflows to build Stencil with esbuild and then run the existing unit and e2e test suites against the build artifact

Right now this new build / bundle setup exists parallel to the existing, rollup-based build setup. The idea is that we'll continue to use rollup for building and releasing, but we'll gradually increase the scope of the esbuild-based pipeline until it can build all of stencil. As noted above, this PR starts with just the `cli/` module.

The esbuild-based pipeline is also set up to first call the existing rollup-based build. It then deletes the `cli/` module built by rollup and replaces it with its own output.

Esbuild has a few notable differences from rollup which are relevant here:

- It can directly bundle and transform TypeScript, so we can go from TypeScript to bundled JS in one step (rather than with our Rollup pipeline, where we run `tsc` and then run Rollup on the JS emitted by `tsc`). This decreases a bit of build complexity because for instance esbuild can read the `paths` configuration in `tsconfig.json`, so we no longer need something analogous to our [alias plugin](https://github.com/ionic-team/stencil/blob/b97dadc967b1fde892cb75a544b1eecd2361b194/scripts/bundles/plugins/alias-plugin.ts) which normally brings those aliases into the Rollup build.
- It's very fast! the `cli/` module builds in about 13ms.
- It also strips documentation comments from the code, I think because it does an AST based transform and discards comments as part of parsing (or doesn't make an effort to keep them around). This results in smaller bundle sizes. It does have support for retaining [certain specially-formatted comments](https://esbuild.github.io/api/#legal-comments) which is a good thing because we use a few magic comments for Vite support.
- OOB it supports pretty much all of what we're currently doing in our Rollup-based build pipeline, so we should be able to gradually port everything over.

Once this is merged we can start using esbuild for local development and also start to port over other modules. I have a working build already for the `compiler/` directory, getting that set up for local development will really speed things up!

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes

no esbuild, rollup slow

## What is the new behavior?

yes esbuild, rollup still slow!

- [ ] Yes
- [x] No

## Testing

I've built and packed this and then run it in several little example projects and in Framework with no problems.

### DIFF
--- a/.github/workflows/build-esbuild.yml
+++ b/.github/workflows/build-esbuild.yml
@@ -1,0 +1,41 @@
+name: Build Stencil with Esbuild
+
+on:
+  workflow_call:
+  # Make this a reusable workflow, no value needed
+  # https://docs.github.com/en/actions/using-workflows/reusing-workflows
+
+jobs:
+  build_core:
+    name: Core
+    runs-on: 'ubuntu-22.04'
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@8ade135a41bc03ea155e62e844d188df1ea18608 # v4.1.0
+        with:
+          # the pull_request_target event will consider the HEAD of `main` to be the SHA to use.
+          # attempt to use the SHA associated with a pull request and fallback to HEAD of `main`
+          ref: ${{ github.event_name == 'pull_request_target' && format('refs/pull/{0}/merge', github.event.number) || '' }}
+          persist-credentials: false
+
+      - name: Get Core Dependencies
+        uses: ./.github/workflows/actions/get-core-dependencies
+
+      - name: Core Build
+        run: npm run build.esbuild -- --ci
+        shell: bash
+
+      - name: Validate Build
+        run: npm run test.dist
+        shell: bash
+
+      - name: Validate Testing
+        run: npm run test.testing
+        shell: bash
+
+      - name: Upload Build Artifacts
+        uses: ./.github/workflows/actions/upload-archive
+        with:
+          name: stencil-core-esbuild
+          output: stencil-core-build.zip
+          paths: build cli compiler dev-server internal mock-doc scripts/build screenshot sys testing
--- a/package.json
+++ b/package.json
@@ -24,6 +24,8 @@
   ],
   "scripts": {
     "build": "npm run tsc.scripts && npm run tsc.prod && npm run rollup.prod.ci",
+    "build.esbuild": "npm run clean && npm run build && node scripts/build/esbuild/build.js",
+    "build.esbuild.watch": "npm run clean && npm run build && node scripts/build/esbuild/build.js --watch",
     "clean": "rm -rf build/ cli/ compiler/ dev-server/ internal/ mock-doc/ sys/ testing/ && npm run clean-scripts",
     "clean-scripts": "rm -rf scripts/build",
     "lint": "eslint 'bin/*' 'scripts/*.ts' 'scripts/**/*.ts' 'src/*.ts' 'src/**/*.ts' 'src/**/*.tsx'",
@@ -86,6 +88,7 @@
     "conventional-changelog-cli": "^4.0.0",
     "cspell": "^7.0.1",
     "dts-bundle-generator": "~8.0.0",
+    "esbuild": "^0.19.4",
     "eslint": "^8.23.1",
     "eslint-config-prettier": "^9.0.0",
     "eslint-plugin-jest": "^27.0.4",
--- a/scripts/esbuild/cli.ts
+++ b/scripts/esbuild/cli.ts
@@ -1,0 +1,82 @@
+import type { BuildOptions as ESBuildOptions } from 'esbuild';
+import fs from 'fs-extra';
+import { join } from 'path';
+
+import { getBanner } from '../utils/banner';
+import { BuildOptions } from '../utils/options';
+import { writePkgJson } from '../utils/write-pkg-json';
+import { getBaseEsbuildOptions, getEsbuildAliases, getEsbuildExternalModules, runBuilds } from './util';
+
+/**
+ * Runs esbuild to bundle the `cli` submodule
+ *
+ * @param opts build options
+ * @returns a promise for this bundle's build output
+ */
+export async function buildCli(opts: BuildOptions) {
+  // clear out rollup stuff
+  await fs.emptyDir(opts.output.cliDir);
+
+  const inputDir = join(opts.srcDir, 'cli');
+  const buildDir = join(opts.buildDir, 'cli');
+
+  const outputDir = opts.output.cliDir;
+  const esmFilename = 'index.js';
+  const cjsFilename = 'index.cjs';
+
+  const dtsFilename = 'index.d.ts';
+
+  const cliAliases = getEsbuildAliases();
+
+  const external = getEsbuildExternalModules(opts, opts.output.cliDir);
+
+  const cliEsbuildOptions = {
+    ...getBaseEsbuildOptions(),
+    alias: cliAliases,
+    entryPoints: [join(inputDir, 'index.ts')],
+    external,
+    platform: 'node',
+    sourcemap: 'linked',
+  } satisfies ESBuildOptions;
+
+  // ESM build options
+  const esmOptions: ESBuildOptions = {
+    ...cliEsbuildOptions,
+    outfile: join(outputDir, esmFilename),
+    format: 'esm',
+    banner: {
+      js: getBanner(opts, `Stencil CLI`, true),
+    },
+  };
+
+  // CommonJS build options
+  const cjsContext: ESBuildOptions = {
+    ...cliEsbuildOptions,
+    outfile: join(outputDir, cjsFilename),
+    format: 'cjs',
+    banner: {
+      js: getBanner(opts, `Stencil CLI (CommonJS)`, true),
+    },
+  };
+
+  // create public d.ts
+  let dts = await fs.readFile(join(buildDir, 'public.d.ts'), 'utf8');
+  dts = dts.replace('@stencil/core/internal', '../internal/index');
+  await fs.writeFile(join(opts.output.cliDir, dtsFilename), dts);
+
+  // copy config-flags.d.ts
+  let configDts = await fs.readFile(join(buildDir, 'config-flags.d.ts'), 'utf8');
+  configDts = configDts.replace('@stencil/core/declarations', '../internal/index');
+  await fs.writeFile(join(opts.output.cliDir, 'config-flags.d.ts'), configDts);
+
+  // write cli/package.json
+  writePkgJson(opts, opts.output.cliDir, {
+    name: '@stencil/core/cli',
+    description: 'Stencil CLI.',
+    main: cjsFilename,
+    module: esmFilename,
+    types: dtsFilename,
+  });
+
+  return runBuilds([esmOptions, cjsContext], opts);
+}
--- a/scripts/esbuild/util.ts
+++ b/scripts/esbuild/util.ts
@@ -1,0 +1,124 @@
+import type { BuildOptions as ESBuildOptions, BuildResult as ESBuildResult } from 'esbuild';
+import * as esbuild from 'esbuild';
+import { join } from 'path';
+
+import { BuildOptions } from '../utils/options';
+
+/**
+ * Aliases which map the module identifiers we set in `paths` in `tsconfig.json` to
+ * bundles (built either with esbuild or with rollup).
+ *
+ * @returns a map of aliases to bundle entry points, relative to the root directory
+ */
+export function getEsbuildAliases(): Record<string, string> {
+  return {
+    // node module redirection
+    chalk: 'ansi-colors',
+
+    // mapping aliases to top-level bundle entrypoints
+    '@stencil/core/testing': './testing/index.js',
+    '@stencil/core/compiler': './compiler/stencil.js',
+    '@stencil/core/dev-server': './dev-server/index.js',
+    '@stencil/core/mock-doc': './mock-doc/index.cjs',
+    '@stencil/core/internal/testing': './internal/testing/index.js',
+
+    // mapping node.js module names to `sys/node` "cache"
+    //
+    // these allow us to bundle and ship a dependency (like `prompts`) as part
+    // of the Stencil distributable but also have our separate bundles
+    // reference the same file
+    prompts: './sys/node/prompts.js',
+  };
+}
+
+/**
+ * Node modules which should be universally marked as external
+ *
+ * Note that we should not rely on this to mark node.js built-in modules as
+ * external. Doing so will override esbuild's automatic marking of those modules
+ * as side-effect-free, which allows imports from them to be properly
+ * tree-shaken.
+ */
+const externalNodeModules = [
+  '@jest/core',
+  '@jest/reporters',
+  '@microsoft/typescript-etw',
+  'assert',
+  'buffer',
+  'child_process',
+  'console',
+  'constants',
+  'expect',
+  'fsevents',
+  'inspector',
+  'jest',
+  'jest-cli',
+  'jest-config',
+  'jest-message-id',
+  'jest-pnp-resolver',
+  'jest-runner',
+  'net',
+  'puppeteer',
+  'puppeteer-core',
+  'readline',
+  'stream',
+  'tty',
+  'url',
+  'util',
+  'vm',
+  'yargs',
+  'zlib',
+];
+
+/**
+ * Get a manifest of modules which should be marked as external for a given
+ * esbuild bundle
+ *
+ * @param opts options for the current build
+ * @param ownEntryPoint the entry point alias of the current module
+ * @returns a list of modules which should be marked as external
+ */
+export function getEsbuildExternalModules(opts: BuildOptions, ownEntryPoint: string): string[] {
+  const bundles = Object.values(opts.output);
+  const externalBundles = bundles.filter((outdir) => outdir !== ownEntryPoint).map((outdir) => join(outdir, '*'));
+
+  return [...externalNodeModules, ...externalBundles];
+}
+
+/**
+ * A helper which runs an array of esbuild, uh, _builds_
+ *
+ * This accepts an array of build configurations and will either run a synchronous
+ * build _or_ run them all in watch mode, depending on the {@link BuildOptions['isWatch']}
+ * setting.
+ *
+ * @param builds the array of outputs to build
+ * @param opts Stencil build options
+ * @returns a Promise representing the execution of the builds
+ */
+export function runBuilds(builds: ESBuildOptions[], opts: BuildOptions): Promise<(void | ESBuildResult)[]> {
+  if (opts.isWatch) {
+    return Promise.all(
+      builds.map(async (buildConfig) => {
+        const context = await esbuild.context(buildConfig);
+        return context.watch();
+      }),
+    );
+  } else {
+    return Promise.all(builds.map(esbuild.build));
+  }
+}
+
+/**
+ * Get base esbuild options which we want to always have set for all of our
+ * bundles
+ *
+ * @returns a base set of options
+ */
+export function getBaseEsbuildOptions(): ESBuildOptions {
+  return {
+    bundle: true,
+    legalComments: 'inline',
+    logLevel: 'info',
+  };
+}
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -22,6 +22,7 @@
     "paths": {
       "@app-data": ["src/app-data/index.ts"],
       "@app-globals": ["src/app-globals/index.ts"],
+      "@environment": ["src/compiler/sys/environment.ts"],
       "@dev-server-process": ["src/dev-server/server-process.ts"],
       "@hydrate-factory": ["src/hydrate/runner/hydrate-factory.ts"],
       "@platform": ["src/client/index.ts"],
--- a/.github/workflows/main-esbuild.yml
+++ b/.github/workflows/main-esbuild.yml
@@ -1,11 +1,9 @@
-name: CI
+name: esbuild CI
 
 on:
-  merge_group:
   push:
     branches:
       - 'main'
-      - 'stencil/v4-dev'
   pull_request:
     branches:
       - '**'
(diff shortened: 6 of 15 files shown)


scripts/esbuild/cli.ts
```diff
@@ -0,0 +1,77 @@
...
+  const cjsFilename = join(outputDir, 'index.cjs');
+
+  const dtsFilename = 'index.d.ts';
+
+  const cliAliases = getEsbuildAliases(opts);
+
+  const external = getEsbuildExternalModules(opts, opts.output.cliDir);
+
+  const cliEsbuildOptions: ESBuildOptions = {
+    entryPoints: [join(inputDir, 'index.ts')],
```
alicewriteswrongs, line 35: note that the entry point here is the typescript index file for the bundle, not the JS output by `tsc`

scripts/esbuild/cli.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  // create public d.ts
+  let dts = await fs.readFile(join(buildDir, 'public.d.ts'), 'utf8');
+  dts = dts.replace('@stencil/core/internal', '../internal/index');
+  await fs.writeFile(join(opts.output.cliDir, dtsFilename), dts);
+
+  // copy config-flags.d.ts
+  let configDts = await fs.readFile(join(buildDir, 'config-flags.d.ts'), 'utf8');
+  configDts = configDts.replace('@stencil/core/declarations', '../internal/index');
+  await fs.writeFile(join(opts.output.cliDir, 'config-flags.d.ts'), configDts);
```
alicewriteswrongs, line 67: these bits for copying `.d.ts` files are copied pretty much verbatim from an equivalent bit in the Rollup file:

https://github.com/ionic-team/stencil/blob/8301b79be1b2aab18f377b6518442be572d39fa3/scripts/bundles/cli.ts#L44-L52

scripts/esbuild/cli.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  // write @stencil/core/compiler/package.json
+  writePkgJson(opts, opts.output.cliDir, {
+    name: '@stencil/core/cli',
+    description: 'Stencil CLI.',
+    main: cjsFilename,
+    module: esmFilename,
+    types: dtsFilename,
+  });
```
alicewriteswrongs, line 76: likewise, this is copied too:

https://github.com/ionic-team/stencil/blob/8301b79be1b2aab18f377b6518442be572d39fa3/scripts/bundles/cli.ts#L55-L61

scripts/esbuild/util.ts
```diff
@@ -0,0 +1,88 @@
...
+  return {
+    // node module redirection
+    chalk: 'ansi-colors',
+
+    // mapping aliases to top-level bundle entrypoints
+    '@stencil/core/testing': './testing/index.js',
+    '@stencil/core/compiler': './compiler/stencil.js',
+    '@stencil/core/dev-server': './dev-server/index.js',
+    '@stencil/core/mock-doc': './mock-doc/index.cjs',
+    '@stencil/core/internal/testing': './internal/testing/index.js',
```
alicewriteswrongs, line 21: set up aliases which will point to the entry points of other bundles. Note that in combination with `getEsbuildExternalModules`, below, these will be marked as external and thus imported or required instead of being bundled in

scripts/esbuild/util.ts
```diff
@@ -0,0 +1,88 @@
...
+    // correctly
+    '@environment': join(opts.srcDir, 'compiler/sys/environment.ts'),
+  };
+}
+
+/**
+ * Node modules which should be universally marked as external
+ *
+ */
+const externalNodeModules = [
```
alicewriteswrongs, line 39: the CLI module doesn't pull most of these in, but they are used by other bundles

alicewriteswrongs (commented):
just noting a few things

esbuild.js
```diff
@@ -0,0 +1,13 @@
+const { getOptions } = require('./scripts/build/utils/options.js');
```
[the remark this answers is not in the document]
alicewriteswrongs: sure I think it could be helpful for you two to start and then we can go from there

and yeah, this is the same `getOptions`, I figured we already have a good data structure there for this so no need to reinvent the wheel

scripts/esbuild/cli.ts
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * Runs esbuild to bundle the `cli` submodule
+ *
+ * @param opts build options needed to generate the rollup configuration
+ * @returns an array containing the generated rollup options
+ */
+export async function buildCli(opts: BuildOptions) {
+  // clear out rollup stuff
+  fs.emptyDir(opts.output.cliDir);
```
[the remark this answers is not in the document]
alicewriteswrongs: In a post-rollup world we'd probably want to do this for all the of top-level bundles, as we do now for the Rollup bundle to ensure we're always producing clean output:

https://github.com/ionic-team/stencil/blob/1dd6323d2012b0ed81facbbd8b7c8a221b93974e/scripts/build.ts#L48-L56

but right now it's mainly here to ensure that the existing stuff from the Rollup build gets removed.

scripts/esbuild/util.ts
alicewriteswrongs: it may not be totally exhaustive - on my other branch this was the external modules I collated while getting the `cli`, `compiler`, `testing`, and `mock-doc` bundles building with Esbuild, so when we go to tackle the others maybe we'll need a few more!

tsconfig.json
```diff
@@ -76,5 +77,5 @@
     "src/testing/index.ts",
     "src/testing/platform/index.ts"
   ],
-  "include": ["types/*.d.ts", "src/declarations/*.ts", "src/**/*.spec.ts"]
+  "include": ["types/*.d.ts", "src/declarations/*.ts", "src/**/*.spec.ts", "src/utils/**/*.ts" ]
```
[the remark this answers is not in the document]
alicewriteswrongs: I don't recall if this change was necessary for the `cli` module or for the `compiler` one, but basically there were some files in here that weren't being included in the TypeScript build and that was creating some issue...thb I don't remember what though! So we may not need it after all

src/cli/load-compiler.ts
```diff
@@ -1,9 +1,9 @@
 import type { CompilerSystem } from '../declarations';
 
-export const loadCoreCompiler = async (sys: CompilerSystem): Promise<CoreCompiler> => {
-  await sys.dynamicImport(sys.getCompilerExecutingPath());
+export const loadCoreCompiler = async (_sys: CompilerSystem): Promise<CoreCompiler> => {
+  const coreCompiler = await import('@stencil/core/compiler');
 
-  return (globalThis as any).stencil;
+  return coreCompiler;
 };
```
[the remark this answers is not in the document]
alicewriteswrongs: in an Esbuild world this actually doesn't need to be a dynamic import (`import` or `sys.dynamicImport`) at all - it can instead just be a normal `import` declaration, and as long as we have our `external` and `aliases` configurations set up correctly for Esbuild then it will compile down to either a `require` call or an `import` declaration, depending on the output format

I don't believe IE11 support (or in-browser support) is relevant here since we're in the CLI module, which I don't think we've ever supported using in a browser (unless I'm mistaken).

scripts/esbuild/util.ts
```diff
@@ -0,0 +1,89 @@
...
+ *
+ * @param opts options for the current build
+ * @returns a map of aliases to bundle entry points, relative to the root directory
+ */
+export function getEsbuildAliases(opts: BuildOptions): Record<string, string> {
+  return {
+    // node module redirection
+    chalk: 'ansi-colors',
+
+    // mapping aliases to top-level bundle entrypoints
```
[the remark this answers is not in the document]
alicewriteswrongs: yeah, I think it will need to grow a bit. Right now it covers the ones I needed to add to get the `cli` and then `compiler` bundles building correctly and externalizing each other.

scripts/esbuild/util.ts
```diff
@@ -0,0 +1,89 @@
...
+ * Aliases which map the module identifiers we set in `paths` in `tsconfig.json` to
+ * bundles (built either with esbuild or with rollup).
+ *
+ * @param opts options for the current build
+ * @returns a map of aliases to bundle entry points, relative to the root directory
+ */
+export function getEsbuildAliases(opts: BuildOptions): Record<string, string> {
+  return {
+    // node module redirection
+    chalk: 'ansi-colors',
```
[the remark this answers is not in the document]
alicewriteswrongs: this isn't cli-specific, but actually from `testing`:

https://github.com/ionic-team/stencil/blob/91737597ad3b5edbce6427643da133c4f7b60d27/scripts/bundles/testing.ts#L91-L93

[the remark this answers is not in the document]
alicewriteswrongs: possibly in a no-browser-compilation world we don't need to worry about some of these things anymore (I hope so!)

src/cli/load-compiler.ts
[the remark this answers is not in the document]
rwaskiewicz: Oh, right 🤦 NVM, carry on. Shoulda read the file name

esbuild.js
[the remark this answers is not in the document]
alicewriteswrongs: I just took a look back at this and ended up deciding to move this to `scripts/esbuild/build.ts` instead of having it live in the root as a `.js` file - that way we can use ESM imports and have some type-checking and whatnot, and we can just run the built output in `scripts/build/esbuild/build.js`

package.json
```diff
@@ -24,6 +24,7 @@
   ],
   "scripts": {
     "build": "npm run tsc.scripts && npm run tsc.prod && npm run rollup.prod.ci",
+    "build:esbuild": "git clean -fd && npm run clean && npm run build && node esbuild.js",
```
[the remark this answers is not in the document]
alicewriteswrongs: haha good point - I'm good w/ `build.esbuild` although I must admit to an aesthetic preference for `:` - maybe because it's taller and looks more balanced? not important in any case 😆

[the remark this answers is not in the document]
alicewriteswrongs: no I don't think so, just being extra cautious I think

scripts/esbuild/cli.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  const cliAliases = getEsbuildAliases(opts);
+
+  const external = getEsbuildExternalModules(opts, opts.output.cliDir);
+
+  const cliEsbuildOptions: ESBuildOptions = {
+    entryPoints: [join(inputDir, 'index.ts')],
+    bundle: true,
+    platform: 'node',
+    outfile: esmFilename,
```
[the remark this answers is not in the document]
alicewriteswrongs: It's not a required property on the `BuildOptions` interface for `esbuild` - if you don't set it then it will infer a sensible output filename for you, like here if I omit it then esbuild will write to `index.js`, which makes sense given the `index.ts` entrypoint

scripts/esbuild/cli.ts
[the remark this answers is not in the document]
rwaskiewicz: True, I think I'd be OK with that (having just unit + e2e) to start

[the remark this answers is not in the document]
alicewriteswrongs: cool, I already got an action stood up on this branch to run the esbuild-based build, I can work today on feeding that into a run of the unit and e2e tests

alicewriteswrongs:
Alright so I've just gotten some CI stuff together so that we now have essentially a parallel CI to our normal one which instead starts with the esbuild-based build and then runs the e2e, unit tests, etc (everything except browserstack)

scripts/esbuild/cli.ts
```diff
@@ -0,0 +1,75 @@
...
+import { join } from 'path';
+
+import { BuildOptions } from '../utils/options';
+import { writePkgJson } from '../utils/write-pkg-json';
+import { getEsbuildAliases, getEsbuildExternalModules } from './util';
+
+/**
+ * Runs esbuild to bundle the `cli` submodule
+ *
+ * @param opts build options needed to generate the rollup configuration
```
tanner-reits, line 13:
```suggestion
 * @param opts build options needed to generate the esbuild configuration
```

scripts/esbuild/cli.ts
```diff
@@ -0,0 +1,75 @@
...
+ * @returns an array containing the generated rollup options
```
tanner-reits, line 14: Doesn't look like this actually returns anything

scripts/esbuild/cli.ts
```diff
@@ -0,0 +1,75 @@
...
+    outfile: join(outputDir, esmFilename),
+    format: 'esm',
+  });
+
+  // CommonJS build
+  await esbuild.build({
+    ...cliEsbuildOptions,
+    outfile: join(outputDir, cjsFilename),
+    format: 'cjs',
+  });
```
tanner-reits, line 55: Do these need to run sequentially, or can they run in parallel? i.e. using `await Promise.all(...)`

scripts/esbuild/cli.ts
tanner-reits: 👍 Bet we could do this in parallel also, _but_ I also bet the time delta is negligible for just writing a file 😂

scripts/esbuild/util.ts
tanner-reits: Man, I really wish there was a way to automatically generate this list (and the list of module aliases above). Feels like something easy to overlook.

Did you come up with all these by trial and error (like, does the build fail if we're missing one of these), or did you have to scan the code for each bundle to make sure you had all of them?

tanner-reits (changes requested):
Took a look at most of this. Skipped over the actions stuff for now, but can review that if it's in a good state (wasn't sure if it's still a wip)

scripts/esbuild/util.ts
alicewriteswrongs: I based this on the existing `external` lists used for the rollup builds (see [here](https://github.com/ionic-team/stencil/blob/a0dc40512c47982c1f933f59f71206cef4d1923c/scripts/bundles/testing.ts#L33-L64), [here](https://github.com/ionic-team/stencil/blob/a0dc40512c47982c1f933f59f71206cef4d1923c/scripts/bundles/cli.ts#L66-L66), and [here](https://github.com/ionic-team/stencil/blob/a0dc40512c47982c1f933f59f71206cef4d1923c/scripts/bundles/screenshot.ts#L38-L38) for instance) and then I think possibly had to add one or two things that weren't already present - it is a bit of an annoying thing, if we accidentally forget to add something we do want to have externalized to the list then esbuild (and rollup too!) will happily just pull it into the bundle.

scripts/esbuild/cli.ts
alicewriteswrongs: oh yeah they could be in parallel I think, although I don't know what the performance implications of that would be. Esbuild is implemented in Go so the call to `esbuild.build` is actually going to call out to a native node extension which will do all the hard work (there's basically a small JS shim around the Go code afaik). The extension being written in Go means that it has its own internal threads and whatnot - I think it will basically saturate your CPU provided the job takes long enough. It could even be that if you try to fire off two esbuild processes at the same time they are slightly slower because you get some resource contention between them, but in this case the execution times are so brief that who knows.

If we want to we could do some empirical testing to get an idea of whether we see a significant difference from using `Promise.all` here

scripts/esbuild/cli.ts
alicewriteswrongs: 🙃 right you are! I copied this JSDoc from the `cli` function for Rollup

scripts/esbuild/util.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    // this has to be added here pointing to the `.ts` file so that the `import
+    // { IS_NODE_ENV } from '@environment';` in the `typescript.js` file we
+    // bundle will resolve correctly
+    //
+    // esbuild will automatically resolve aliases from `paths` in
+    // `tsconfig.json` but only for `.ts` files, so when we bundle `typescript`
+    // into e.g. the compiler bundle we'll need this so things resolve
+    // correctly
+    '@environment': join(opts.srcDir, 'compiler/sys/environment.ts'),
```
rwaskiewicz, line 32: Do we still need this, now that we removed in-browser compilation? I see some usages of `@environment`, but I _think_ those are all infrastructure (and not actual imports that use the alias).

scripts/esbuild/cli.ts
```diff
@@ -0,0 +1,74 @@
...
+  const esmFilename = 'index.js';
+  const cjsFilename = 'index.cjs';
+
+  const dtsFilename = 'index.d.ts';
+
+  const cliAliases = getEsbuildAliases(opts);
+
+  const external = getEsbuildExternalModules(opts, opts.output.cliDir);
+
+  const cliEsbuildOptions: ESBuildOptions = {
```
rwaskiewicz, line 32: Is there an option that is equivalent to Rollup's `preferConst`? It looks like entries get compiled down to `var`s:
Screenshot 2023-10-10 at 1 31 32 PM

rwaskiewicz (changes requested):
One overarching question - how do we want to handle the size investigation? ESBuild is producing output that is >2x the size of the rollup build. I don't recall if we discussed this explicitly the last time or not

scripts/esbuild/util.ts
alicewriteswrongs: oh yeah we probably can get rid of that! I think there might be one or two other things that can be simplified. I'll do a bit of an audit for that

.github/workflows/build-esbuild.yml
```diff
@@ -0,0 +1,41 @@
...
+
+on:
+  workflow_call:
+  # Make this a reusable workflow, no value needed
+  # https://docs.github.com/en/actions/using-workflows/reusing-workflows
+
+jobs:
+  build_core:
+    name: Core
+    runs-on: 'ubuntu-20.04'
```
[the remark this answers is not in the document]
alicewriteswrongs: makes sense! you can tell the PR was opened a while ago haha

scripts/esbuild/cli.ts
alicewriteswrongs: I did a little googling around and it seems that this is not something that esbuild supports, and for some of what seem like decent reasons. They have an FAQ entry about it here: <https://esbuild.github.io/faq/#top-level-var>

The tl;dr as I understand it is that esbuild will sometimes need to lazily initialize an imported module[^1] and for performance reasons when it does so it will separate variable declaration from initialization so it of course can't use `const` to declare those variables since initialization has to happen at variable declaration with `const`. Then I think esbuild prefers `var` over `let` because of those TDZ-check related performance issues.

I don't really know enough about the mechanics of bundling or about JS VM performance to fully assess the validity of the argument, but that's what they're saying as to why there's no equivalent of a `preferConst` option.

[^1]: I think especially with an `import()` call but I couldn't _quite_ get this to happen in a small example so I'm not sure exactly what they're referring to

scripts/esbuild/cli.ts
```diff
@@ -0,0 +1,76 @@
...
+    entryPoints: [join(inputDir, 'index.ts')],
+    bundle: true,
```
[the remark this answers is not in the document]
tanner-reits: Resurfacing this as a possibility of bundle size increase

scripts/esbuild/cli.ts
```diff
@@ -0,0 +1,76 @@
...
+    platform: 'node',
+    outfile: esmFilename,
+    logLevel: 'info',
+    sourcemap: 'linked',
+    external,
+    alias: cliAliases,
+  };
```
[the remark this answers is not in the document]
tanner-reits: Resurfacing this as well

tanner-reits (commented):
I remember feeling like this was in a decent state when we first looked through a long time ago. I think the most important aspects are:
1. Bundle size should definitely be investigated before we rely on this full-time. If we wanna merge just to have CI run against this in the meantime I think that's fine
2. Is there anything left to remove from the removal of in-browser compilation

rwaskiewicz:
@alicewriteswrongs Does that mean that `/*!__STENCIL_STATIC_IMPORT_SWITCH__*/` also gets removed (a part of our support for Vite)?

alicewriteswrongs:
I don't see a usage of that comment in the code that will be pulled into the `cli` module so I can't test this but I believe that can be retained because esbuild [recognizes it as a "legal comment"](https://esbuild.github.io/api/#legal-comments). It looks like we can configure the behavior we want, i.e. retaining it precisely where it is or the default behavior which is to move it to the end of the bundled file (I don't know exactly what that comment does so I'm not sure how we need to treat it).

scripts/esbuild/util.ts
```diff
@@ -0,0 +1,124 @@
...
+    return Promise.all(
+      builds.map(async (buildConfig) => {
+        const context = await esbuild.context(buildConfig);
+        return context.watch();
+      }),
+    );
+  } else {
+    return Promise.all(builds.map(esbuild.build));
+  }
+}
```
alicewriteswrongs, line 110: this lil helper will take an array of esbuild configs and our own Stencil-specific `BuildOptions` object and if the `isWatch` property is set on the `BuildOptions` it will run a watch-mode build using [`esbuild.context`](https://esbuild.github.io/api/#watch), otherwise it will just use `esbuild.build` to do a single build.

The idea is that each of our bundle functions would create a series of esbuild configurations, do any other setup that is necessary, and then return the result of calling this function with those build configs. Then in our main entry point we use `Promise.all` to await the result of all the builds.

also kind of funny that the esbuild interface is _also_ called `BuildOptions`, so I've taken to aliasing it as `ESBuildOptions`

scripts/esbuild/cli.ts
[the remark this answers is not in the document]
alicewriteswrongs: just did this!

rwaskiewicz (approved):

tanner-reits (approved):
:shipit:

PR #4199, opened 2023-03-24
refactor(utils): add a little bit of type safety around JSON.stringify

alicewriteswrongs:
This adds two things, the first is a type, `JsonSafe`, which takes a type parameter and will basically evaluate (at the type level) to `never` if the supplied type can't be serialized to JSON. This allows us to add constraints throughout the codebase wherever we want to assert that a value must be serializable (or, well, that the type annotated on or inferred for a value is serializable).

Then this type is used to implement a type-safe `JSON.stringify`. This is exported from a new `utils/` module, `src/utils/json.ts`. If you call it with type which can't be safely (meaning losslessly) serialized to JSON you'll get a type error.

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

We call `JSON.stringify` quite a bit throughout the codebase. Unfortunately, `JSON.stringify` has its first parameter typed as `any` so we don't have much in the way of assurance that the value we pass is going to serialize well.

## What is the new behavior?

No real code changes, just adding some type-system level assurance that values we pass to `JSON.stringify` are serializable.

- [ ] Yes
- [x] No

## Testing

I ran the tests and checked that Framework built correctly.

### DIFF
--- a/src/utils/json.ts
+++ b/src/utils/json.ts
@@ -1,23 +1,23 @@
 import { JsonSafe } from '@stencil/core/declarations';
 
 /**
- * A thin wrapper around JSON.stringify which ensures that the value to be
+ * A minimal wrapper around JSON.stringify which ensures that the value to be
  * stringified is a valid JSON-serializable value, as codified by the
  * {@link JsonSafe} type. If you try to serialize a type which isn't valid
  * you'll get a type error.
  *
  * **NOTE**: This is _only_ a type-level safety guarantee. No runtime check is
- * made at all that the passed value has a serializable runtime type.
+ * made at all that the passed value is serializable.
  *
- * @param toStringify a value to be serialized
+ * @param value a value to be serialized
  * @param replacer an optional replacer (same as JSON.stringify)
  * @param space an  optional space to use (same as JSON.stringify)
  * @returns a serialized string
  */
 export function safeJSONStringify<T>(
-  toStringify: T & JsonSafe<T>,
+  value: T & JsonSafe<T>,
   replacer?: (this: any, key: string, value: any) => any,
   space?: string | number
 ): string {
-  return JSON.stringify(toStringify, replacer, space);
+  return JSON.stringify(value, replacer, space);
 }


src/cli/telemetry/telemetry.ts
```diff
@@ -195,21 +197,31 @@ const CONFIG_PROPS_TO_ANONYMIZE: ReadonlyArray<ConfigStringKeys> = [
...
   'env',
   'logger',
+  'nodeResolve', // this contains regexes which won't serialize to JSON
   'rollupConfig',
   'sys',
   'testing',
   'tsCompilerOptions',
-];
+  'watchIgnoredRegex', // this contains regexes which won't serialize to JSON
+] satisfies ReadonlyArray<keyof d.Config>;
```
alicewriteswrongs, line 216: the new `satisfies` keyword here is super handy - basically what it does is if you do something like

```ts
const Obj satisfies MyType;
```

it will check to ensure that `Obj` _satisfies_ the `MyType` type but it will _not_ change the type of `Obj` to be type `MyType` -- instead, it will keep whatever type the type checker is able to infer for `Obj`. This could be the same type as `MyType`, but it could also be a more specific type that we'd like to keep around.

In this case here that's exactly what we want. We want to check that all of the strings in our `ReadonlyArray` here are keys of `d.Config`, so checking that the value satisfies `ReadonlyArray<keyof d.Config>` gives us that check, but we want to keep the type as `Array<"commonjs" | "devServer"...>` i.e. keep the type as being derived from the actual strings that are here.

Super handy!

src/declarations/stencil-private.ts
```diff
@@ -2644,3 +2644,25 @@ export interface TelemetryConfig {
...
+
+/**
+ * Wrap a type with `JsonSafe` to ensure that it can be safely serialized
+ * to/from JSON. In particular, this basically checks that every property of the
+ * type is a `JsonObject`, a `JsonArray`, or a `JsonPrimitive`.
+ *
+ * In particular this will exclude any type which contains a function, since
+ * functions don't have a valid representation in JSON.
+ */
+export type JsonSafe<T> = T extends JsonPrimitive
```
alicewriteswrongs, line 2660: We could also call this `JsonSerializable` if we wanted, or `ToJson`, `AsJson`, something like that. Up to us!

src/utils/json.ts
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * **NOTE**: This is _only_ a type-level safety guarantee. No runtime check is
+ * made at all that the passed value has a serializable runtime type.
+ *
+ * @param toStringify a value to be serialized
+ * @param replacer an optional replacer (same as JSON.stringify)
+ * @param space an  optional space to use (same as JSON.stringify)
+ * @returns a serialized string
+ */
+export function safeJSONStringify<T>(
```
alicewriteswrongs, line 17: we could change this name, I do think this type-level assurance is a helpful thing to add but wouldn't want to convey more safety than there is here (i.e. as the JSDoc notes there is no runtime check at all, it's just a type-level check)

alicewriteswrongs (commented):
noting a few things for posterity

alicewriteswrongs:
closing in on a year with no review, so going to go ahead and close

Issue #3165, opened 2021-12-01
Bundling warning SOURCEMAP_ERROR


dgonzalezr:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/.github/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

2.11.0

### Current Behavior

When enabling source maps on `stencil.config.ts` a warning message is shown during the build:

![CleanShot 2021-12-01 at 20 09 40](https://user-images.githubusercontent.com/328492/144289783-089b676b-6228-44eb-ab8b-6302d1920e78.png)

### Expected Behavior

Source maps get built without issues.

### Steps to Reproduce

1. Setup a stencil component project.
2. Enable source maps on `stencil.config.ts`: `sourceMap: true`.
3. Build the project with `npm run build`.

### Code Reproduction URL

https://github.com/dgonzalezr/stencil-tailwind

rwaskiewicz:
Looks like this was introduced in v2.10.0, but is not present in v2.9.0. We'll take a further look!

dgonzalezr:
Any update about this issue? It is still happening with the latest @stencil/core version 😐

rwaskiewicz:
I did some preliminary digging on this issue. It appears to:

- occur in v2.9.0 (which is the opposite of what I mentioned [here](https://github.com/ionic-team/stencil/issues/3165#issuecomment-983958568))
- manifest only when more than one output target is used (e.g. `www` and `dist-custom-elements`)

This leads me to believe that rollup is either reading a file that's not in a 'good state' (because it's being modified by another output target) or something similar.  I don't have a definitive timeline for when we'll be able to fix this, but I promise it hasn't been forgotten about 🙂

pratheeshp007:
is there any workaround for this issue?

jepiqueau:
@dgonzalezr @pratheeshp007 This seems to have been fixed in 2.18.0. I update jeep-sqlite this morning with 2.18.0 and the  warning message disappears.

jepiqueau:
@dgonzalezr @pratheeshp007 @rwaskiewicz  Well in fact after having done a npm start when you do a npm run build the warning message is still there

[...]

PR #4200, opened 2023-03-24
fix(compiler): sourcemap for dist-custom-elements generation

rwaskiewicz:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

GitHub Issue Number: https://github.com/ionic-team/stencil/issues/3165

## What is the new behavior?

this commit updates an issue where sourcemap generation in a stencil
project would result in an error:

```
[ WARN  ]  Bundling Warning SOURCEMAP_ERROR
           Error when using sourcemap for reporting an error: Can't resolve original location of error.
```

in order to trigger this warning, the following conditions needed to be
true:
1. `sourceMaps` had to be enabled in `stencil.config.ts` (this is the
   default value starting with stencil v3.0+)
2. both a lazy build (e.g. `www`) and `dist-custom-elments` had to be
   specified in a project's `stencil.config.ts#outputTargets` field. the
   aforementioned output targets are provided by default in the current
   stencil-component-starter, provided by the create-stencil cli.

this error would occur due to how the import statement for
`attachShadow` would be added to stencil's intermediate representation
of a component. prior to this commit, the intent to later add an import
for `attachShadow` during the component generation process would be
expressed by adding it to a `Module` instance's `coreRuntimeApis`
collection. `Module#coreRuntimeApis` however, is a data structure that
is shared during the generation of all output targets. mutations to this
data structure could be picked up by other output target generation
steps.

the reason that changes to `Module#coreRuntimeApis` can be found across
the generation of multiple output targets is that stencil:
1. attempts to parallelize output target generation by generating all
   output targets specified in `stencil.config.ts` at once
2. no copy of `Module` objects are made for each output target
   generation, to avoid a performance hit in allocating the same object
   for each output target specified in a a project's config

`lazy` output targets have their own mechanism for calling
`attachShadow` that doesn't require importing a proxy for it.

this all resulted in a race condition where the api for `attachShadow`
would be added to `Module#coreRuntimeApis` before a lazy build got to
the sourcemap generation step. once the lazy build attempted to generate
sourcemaps, it saw the import for `attachShadow`, which it could not
reconcile against the original source file, causing the warning message
for SOURCMAP_ERROR above.

this commit fixes this bug by introducing the concept if 'output target
specific imports'. for imports that are not meant to be used across
multiple output targets, `coreRuntimeApis` can be added using a stringly
typed identifier (the name of the output target being generated). this
allows for apis to be added while:
- keeping the `addImports` function signature the same
- not changing the type of `Module#coreRuntimeApis`
- avoiding creating several separate copies of data for each output
  target
- avoiding a 'copy and restore' procedure where `coreRuntimeApis` are
  copied at the beginning of output target generation, and restored at
  the end of such generation
- the added data structure allows `Module` to continue to be safely
  serialized using `JSON.stringify`

this does however, have a few implications:

1. named import order changes

   named imports that occur in the same path will now have an ordering
   where `coreRuntimeApis` occur first in the order they were added,
   followed by output target specific named imports occur second in the
   order that they were added. this reordering occurs in a single import
   statement for generated code, and no issues are foreseen with regard
   to import side effect orderings or lint errors (i.e. alphabetized
   named imports)

2. the shape of `Module` is changed

   it is assumed that a `Module`, a private data structure, can be
   safely changed.

this commit was a part of a larger investigation into the bug. as a
result, several functions/modules were provided jsdoc that were
read/re-read in pursuit of tracking down this bug. they are included in
this commit for posterity.

- [ ] Yes
- [x] No

## Testing

Unit tests added + continue to pass after updates

I built this branch:
```terminal
npm ci && /
npm run clean && /
npm run build && /
npm pack
```
then installed it in the output of `npm init stencil component hello-world`

I then started the dev server, and used devtools to verify breakpoints for the lazy build would be hit correctly.

I also used a third party tool (https://sokra.github.io/source-map-visualization/#custom) to validate the `dist-custom-elements` sourcemaps


src/compiler/transformers/component-native/native-constructor.ts
```diff
@@ -32,15 +33,17 @@ export const updateNativeConstructor = (
   if (cstrMethodIndex >= 0) {
     // add to the existing constructor()
     const cstrMethod = classMembers[cstrMethodIndex] as ts.ConstructorDeclaration;
+    // a constructor may not have a body (e.g. in the case of constructor overloads)
+    const cstrBodyStatements: ts.NodeArray<ts.Statement> = cstrMethod.body?.statements ?? ts.factory.createNodeArray();
```
rwaskiewicz, line 37: This function largely _assumes_ that there won't be any overloaded ctor's in a Stencil component. While I don't know if that will always hold true, I don't think it's something we need to worry about this second.

src/compiler/output-targets/dist-custom-elements/index.ts
```diff
@@ -48,7 +48,7 @@ export const outputCustomElements = async (
     return;
   }
 
-  const bundlingEventMessage = 'generate custom elements';
+  const bundlingEventMessage = `generate custom elements${config.sourceMap ? ' + source maps' : ''}`;
```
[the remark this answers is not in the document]
alicewriteswrongs: consistency is good!

src/compiler/transformers/transform-utils.ts
```diff
@@ -165,6 +165,12 @@ export const createStaticGetter = (propName: string, returnExpression: ts.Expres
   );
 };
 
+/**
+ * Retrieves a value represented by TypeScript's syntax tree by name of a static getter. The value is transformed to a TypeScript value.
```
alicewriteswrongs, line 169: Should we possibly call a "TypeScript Value" a "runtime value" here, for less ambiguity?

alicewriteswrongs (changes requested):
no real concerns on my end, just had one or two questions

src/compiler/transformers/transform-utils.ts
rwaskiewicz: Good call - 5ebde1635

src/compiler/transpile/transpiled-module.ts
```diff
@@ -3,11 +3,26 @@ import ts from 'typescript';
...
+ * @param staticSourceFile the TypeScript representation of the source file. This may not be the original
+ * representation of the file, but instead a new `SourceFile` created using the TypeScript API
+ * @param staticSourceFileText the text from the `SourceFile`. This text may originate from the original representation
+ * of the file.
+ * @param emitFilepath the path of the JavaScript file that should be emitted after transpilation
+ * @returns the created `Module`
+ */
 export const createModule = (
-  staticSourceFile: ts.SourceFile, // this is NOT the original
+  staticSourceFile: ts.SourceFile, // this may NOT the original
```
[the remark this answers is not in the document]
rwaskiewicz: Fixed in 6667944a3 👍

src/compiler/transformers/core-runtime-apis.ts
```diff
@@ -29,12 +29,44 @@ export const RUNTIME_APIS = {
...
+ * already exist.
+ * This allows Stencil to keep `moduleFile` easily serializable, where this helper function treats the data structure
+ * that stores {@link Module#outputTargetCoreRuntimeApis} similar to a JS `Set`.
+ * @param moduleFile the Module entity to update
+ * @param outputTarget the output target to assign the provided runtime api under
+ * @param coreRuntimeApi the API to add to the provided module
+ */
+export const addOutputTargetCoreRuntimeApi = (
+  moduleFile: d.Module,
+  outputTarget: string,
```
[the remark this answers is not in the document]
rwaskiewicz: Good point! Fixed in d3ee6355b

src/compiler/transformers/component-native/tranform-to-native-component.ts
```diff
@@ -9,15 +10,29 @@ import { updateStyleImports } from '../style-imports';
...
       const moduleFile = getModuleFromSourceFile(compilerCtx, tsSourceFile);
 
-      const visitNode = (node: ts.Node): any => {
+      /**
+       * Helper function that recursively walks the concrete syntax tree. Upon finding a class declaration that Stencil
+       * recognizes as a component, update the component class
+       * @param node the current node in the tree being inspected
+       * @returns the updated component class, or the unchanged node
+       */
+
```
alicewriteswrongs, line 34: I think extra blank line here

src/declarations/stencil-private.ts
```diff
@@ -1398,19 +1404,34 @@ export interface MinifyJsResult {
...
+  /**
+   * A collection of modules that a component will need. The modules in this list must have import statements generated
+   * in order for the component to function.
+   */
   coreRuntimeApis: string[];
+  /**
+   * A collection of modules that a component will need for a specific output target. The modules in this list must
+   * have import statements generated in order for the component to function, but only for a specific output target.
+   */
+  outputTargetCoreRuntimeApis: Record<string, string[]>;
```
alicewriteswrongs, line 1434: just realized this, we might want to type this `Record<OutputTarget['type'], string[]>` here just to prevent any typos in the future

src/declarations/stencil-private.ts
```diff
@@ -1398,19 +1404,34 @@ export interface MinifyJsResult {
...
+ */
 export type ModuleMap = Map<string, Module>;
 
 /**
  * Stencil's Intermediate Representation (IR) of a module, bundling together
  * various pieces of information like the classes declared within it, the path
  * to the original source file, HTML tag names defined in the file, and so on.
  *
- * Note that this gets serialized/parsed as JSON and therefore cannot be a
+ * Note that this gets serialized/parsed as JSON and therefore cannot contain a
```
alicewriteswrongs, line 1420: if you're concerned about JSON have I got a PR for you!

alicewriteswrongs (changes requested):
just noticed a few more things

src/declarations/stencil-private.ts
alicewriteswrongs: makes sense!

alicewriteswrongs (approved):
I think this all looks good!

rwaskiewicz:
@tanner-reits fixed in https://github.com/ionic-team/stencil/pull/4200/commits/9c034fe48596100ebb0719d318093bb9f50aecb7

tanner-reits (approved):
🚢

PR #4098, opened 2023-02-24
chore(release): update build id, version generation

rwaskiewicz:
## PR Train 🚂 

This PR requires the following PR(s) to be approved first:
- https://github.com/ionic-team/stencil/pull/4097

## Pull request checklist

Please check if your PR fulfills the following requirements:
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [x] Build (`npm run build`) was run locally and any changes were pushed
- [x] Unit tests (`npm test`) were run locally and passed
- [ ] E2E Tests (`npm run test.karma.prod`) were run locally and passed
- [x] Prettier (`npm run prettier`) was run locally and passed

## Pull request type

 

 

Please check the type of change your PR introduces:
- [ ] Bugfix
- [ ] Feature
- [ ] Refactoring (no functional changes, no api changes)
- [x] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

## What is the current behavior?

Local Stencil builds (aka "Dev Builds") have some metadata associated with them that I think could be more valuable in terms of getting development builds into the hands of Stencil Users, the Ionic Framework, etc.

GitHub Issue Number: N/A

## What is the new behavior?

this commit changes the method used by the release scripts to generate
the build id and version number for "development builds", where a
"development build" is considered a build of stencil performed outside
using the `npm run release.*` scripts. this includes running `npm run
build` locally and in ci.

build id generation has changed from a date format 'yyyyMMddHHmmss', to
be epoch time in seconds. this allows us to easily convert the build
time to a human readable time if needed, with the added benefits of
easier de/serialization and a more compact representation. there exists
no place in the stencil codebase where this value is compared today, and
given this change only applies to dev builds, we can safely make this
change without breaking stencil or downstream consmers.

version generation has changed from '0.0.0-dev.BUILD_ID` to use the
current version stored in stencil's `package.json` as a base (so that it
is easily identifiable the stable version of stencil a build derives
from, the new build id, and a seven character sha to pinpoint the base
commit of the development build.

this commit is motivated by an effort to get development builds working
in a github actions ci environment. we wish to have this information
generated by stencil, rather than the ci process, such that commands
like `npx stencil info` returns the same version info as what's
published to the npm registry

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Testing

I tested three pathways here. 
1. With none of these changes applied
[...]

PR #4217, opened 2023-03-31
chore(compiler): add debug statements for file watching

rwaskiewicz:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

We have no debug logging for watching builds (which is why I'm adding it 😉) 
GitHub Issue Number: N/A

## What is the new behavior?

this commit is a follow up to https://github.com/ionic-team/stencil/pull/4146. when debugging the file watcher, it's valuable to know what files have been perceived as added/updated/etc. for #4146, we created dev-builds with additional logging statements to help diagnose the issue. in this commit, we add more refined debugging statements that attempt to balance providing useful information while not spamming the output with logging statements.

when starting a dev server, the version of stencil has been added to the `[LOCAL DEV]`. starting with https://github.com/ionic-team/stencil/pull/4098, stencil has useful information in both dev and prod builds of the project that can is used here. this was also added to avoid a call to stencil's info task - most of the information from the info task is already provided in the output of starting of the dev server. by adding only the version here, we avoid redundant logging statements.

the watch build task has been updated to log the files changed (added, updated, deleted, written). these statements can get slightly verbose, with at least one line per category per rebuild. however, this information was immensely helpful in #4146, and was deemed worth the extra output. the output is only provided when the logging level of the stencil logger is set to "debug", and won't be enabled for normal runs of the compiler.

similar to the watch build task, additional debug-only logging was added to the node-sys watch file-related functions. debug statements in this file are not guarded with a conditional statement, as we do not have any guarantees/information around the logging level that has been set when a node-sys instance is created. in order to log correctly, an optional logger (optional as to not break the contract of this public api) has been added to the creation of node-sys. should one not be provided, debug logging will not apply.

- [ ] Yes
- [x] No

## Testing

- Create a new Stencil project with `npm init stencil component debug-log-test`
- In that project, install a dev build of this project: `npm i @stencil/core@3.2.0-dev.1680277977.315afe6`
- For the following commands, one should _not_ see the debug output:
    - `npm run build`
    - `npm run build -- --debug`
    - `npm start`
- For `npm start -- --debug` one _should_ see the newly created debug statements
## Other information

There are less debug statements in this commit/pr than there were from our original logging branches. I tried to balance spewing _tons_ of info to the console with the value provided. While I think for our original logging/debugging branch that was OK (since we were trying to debug a serious issue), it's not necessarily required for _all_ debugging scenarios in Stencil

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1270,9 +1270,9 @@
       }
     },
     "node_modules/@puppeteer/browsers": {
-      "version": "0.3.1",
-      "resolved": "https://registry.npmjs.org/@puppeteer/browsers/-/browsers-0.3.1.tgz",
-      "integrity": "sha512-WTFVqWY7ipI+CsJRb6vEddYJ/lMl2bHYP5BjFDBM9JMKPjJNT4psXz6DLuJpi3EkG80Q0ef5Azzv3gKu9MvMUg==",
+      "version": "0.3.2",
+      "resolved": "https://registry.npmjs.org/@puppeteer/browsers/-/browsers-0.3.2.tgz",
+      "integrity": "sha512-YAwrqy68PskkCDnSsv0afR8l8u90KhodcIFXCage5NLAoeWm0STGmtVugLBp2wc5PnhnCapCKkMjp9fYjvJl2g==",
       "dev": true,
       "dependencies": {
         "debug": "4.3.4",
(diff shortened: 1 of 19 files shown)


alicewriteswrongs (approved):
looks good to me!

tanner-reits (approved):
Feels good to me. Only thing we _may_ want to change is the timestamp when a change is detected since spitting out the time since epoch isn't the most legible thing in the world, but I don't think date strings would be any better 😂

rwaskiewicz:
@tanner-reits that's fair with respect to legibility - ATM something like `toUTCString` returns something like:
```
const event = new Date('14 Jun 2017 00:00:00 PDT');

console.log(event.toUTCString());
// Expected output: "Wed, 14 Jun 2017 07:00:00 GMT"
```
where we don't necessarily want/need the date (otherwise we run the risk of overflowing the current row in the terminal by exceeding the calculated column count), and it doesn't have the required (IMO) level of precision (it stops at seconds). At that point, I kind of figured all I want is to see if the number increments as new files change, and just went for the epoch date.

PR #4225, opened 2023-04-04
refactor(utils): move output-target related utils to @utils

alicewriteswrongs:
This moves a bunch of utility functions (and some constants) from src/compiler/output-targets/output-utils.ts to instead live in a new file in the `@utils` directory (`src/utils`). This is done to reflect the fact that functions in the file are used outside of the `src/compiler/` bundle so they should be pulled out into a common location.

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

GitHub Issue Number: N/A

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/scripts/bundles/internal-platform-hydrate.ts
+++ b/scripts/bundles/internal-platform-hydrate.ts
@@ -39,6 +39,29 @@ export async function internalHydrate(opts: BuildOptions) {
       banner: getBanner(opts, 'Stencil Hydrate Platform'),
       preferConst: true,
     },
+    /**
+     * The `internal-platform-hydrate` bundle depends on functions which use
+     * the output-target utils which, in turn, imports `path`. However, none of
+     * the functions which it uses actually _use_ functioned exported by `path`,
+     * so when bundling it doesn't actually import anything from `path` but if
+     * you don't set this option Rollup will leave an unqualified import like
+     * so:
+     *
+     * ```ts
+     * import "path";
+     * ```
+     *
+     * If you say `moduleSideEffects: false` you're basically saying 'dont assume
+     * that imported a module has a desired side effect' which allows an import that
+     * doesn't result in any symbols being pulled into the importing module to be
+     * eliminated.
+     *
+     * Another way to say this is that without this change the treeshaking
+     * validation for `test.dist` doesn't pass for this bundle.
+     */
+    treeshake: {
+      moduleSideEffects: false,
+    },
     plugins: [
       {
         name: 'internalHydratePlugin',
@@ -65,6 +88,17 @@ export async function internalHydrate(opts: BuildOptions) {
       file: join(outputInternalHydrateDir, 'runner.js'),
       banner: getBanner(opts, 'Stencil Hydrate Runner'),
       preferConst: true,
+    },
+    /**
+     * See above for an explanation, this is necessary to eliminate an
+     * unqualified import of `"path"`, like so:
+     *
+     * ```ts
+     * import "path";
+     * ```
+     */
+    treeshake: {
+      moduleSideEffects: false,
     },
     plugins: [
       aliasPlugin(opts),
--- a/src/cli/task-docs.ts
+++ b/src/cli/task-docs.ts
@@ -1,4 +1,5 @@
-import { isOutputTargetDocs } from '../compiler/output-targets/output-utils';
+import { isOutputTargetDocs } from '@utils';
+
 import type { ValidatedConfig } from '../declarations';
 import type { CoreCompiler } from './load-compiler';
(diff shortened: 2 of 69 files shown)


src/cli/task-docs.ts
```diff
@@ -1,4 +1,5 @@
-import { isOutputTargetDocs } from '../compiler/output-targets/output-utils';
```
alicewriteswrongs, line 1: when I was fiddling with build stuff I noticed this relative import from `src/compiler/` which was causing code from within the compiler bundle be also pulled into the `cli` bundle. I considered a bit how to address it - we could export the output targets related utils (and constants) from the compiler bundle, making this into something like

```ts
import { isOutputTargetDocs } from '@stencil/core'
```

but I thought that wasn't too great for a few reasons, mainly that it changes the public API of `@stencil/core`. I also thought just for organization, if the output target related utils are used across the codebase they should really live in the `src/utils` directory and therefore be importable from the `@utils` namespace. So I thought we should look at moving the functions to `src/utils/output-target.ts` and moving the constants to `src/utils/constants.ts`

scripts/bundles/internal-platform-hydrate.ts
```diff
@@ -39,6 +39,9 @@ export async function internalHydrate(opts: BuildOptions) {
       banner: getBanner(opts, 'Stencil Hydrate Platform'),
       preferConst: true,
     },
+    treeshake: {
+      moduleSideEffects: false,
+    },
```
alicewriteswrongs, line 44: the `internal-platform-hydrate` bundle depends on functions which use the output-target utils which, in turn, imports `path`. However, none of the functions which it uses actually _use_ functioned exported by `path`, so when bundling it doesn't actually import anything from `path` but if you don't set this option Rollup will leave an unqualified import like so:

```ts
import "path";
```

if you say `moduleSideEffects: false` you're basically saying 'dont assume that imported a module has a desired side effect' which allows an import that doesn't result in any symbols being pulled into the importing module to be eliminated.

Another way to say this is that without this change the treeshaking validation for `test.dist` doesn't pass for this bundle.

alicewriteswrongs (commented):
just noting a few things

scripts/bundles/internal-platform-hydrate.ts
rwaskiewicz: Can we capture this info in a comment for future us?

rwaskiewicz (commented):
Had one minor comment/ask for JSDoc

src/compiler/config/outputs/validate-dist.ts
```diff
@@ -11,7 +8,10 @@ import {
   DIST_TYPES,
   getComponentsDtsTypesFilePath,
   isOutputTargetDist,
-} from '../../output-targets/output-utils';
+} from '@utils';
```
tanner-reits, line 11: Can we combine this with the existing `@utils` import?

src/compiler/config/outputs/validate-www.ts
```diff
@@ -1,15 +1,8 @@
 import { buildError, isBoolean, isString } from '@utils';
+import { COPY, DIST_GLOBAL_STYLES, DIST_LAZY, isOutputTargetDist, isOutputTargetWww, WWW } from '@utils';
```
tanner-reits, line 2: Can we combine this with the existing @utils import?

src/compiler/fs-watch/fs-watch-rebuild.ts
```diff
@@ -1,8 +1,8 @@
 import { isString, unique } from '@utils';
+import { isOutputTargetDocsJson, isOutputTargetDocsVscode, isOutputTargetStats } from '@utils';
```
tanner-reits, line 2: Can we combine this with the existing @utils import?

tanner-reits (changes requested):
Couple small things (sorry for the repetition). Also agree w/ Ryan's comment about adding docs/comment for the `moduleSideEffects` flag.

src/compiler/config/outputs/validate-dist.ts
alicewriteswrongs: oops, thought I captured all of those!

alicewriteswrongs:
just added a comment explaining the `moduleSideEffects` option for Rollup and then also bundled up some imports

alicewriteswrongs:
@tanner-reits you inspired me: https://github.com/ionic-team/stencil/pull/4233

tanner-reits (approved):
Looks good!

rwaskiewicz (approved):

Issue #4053, opened 2023-02-13
bug: Unexpected behavior with spec test executing code after test statement


sean-perkins:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/.github/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

v3.0.0

### Current Behavior

When a component has a `setTimeout` constructed in the `connectedCallback`, that is later cleared in the `disconnectedCallback`; writing a spec test for that component will lead to code executing after the test has completed. This results in different errors. 

Using a log statement will provide:
```
Cannot log after tests are done. Did you forget to wait for something async in your test?
```

Depending on the complexity of the code, you can either receive an error pointing to the problematic code:
```
 RUNS  src/components/my-component/my-component.spec.tsx
/Users/sean/Documents/ionic/issues/stencil-jest-instance-value/src/components/my-component/my-component.tsx:21
    const options = this.col.options;
```

or you may receive an error for the `$instanceValues$`:
```
TypeError: Cannot read properties of undefined (reading '$instanceValues$')
```

This occurs in the Ionic Framework, with the `ion-picker-column` component (observed here: https://github.com/ionic-team/ionic-framework/pull/26749). In that PR, I had to split the test criteria into two separate files, to workaround the issue. The provided repository below is a slimmed down example of what the `ion-picker-column` component is doing. 

### Expected Behavior

When a test has finished, the component should be "disconnected"/torn down and no code should continue to execute.

### System Info

```shell
Node: v16.15.1
npm: v8.19.1
```

### Steps to Reproduce

1. Clone the reproduction example
2. Build the reproduction example `npm run build`
3. Run the spec tests: `npm run test -- --no-build`
4. Observe: 
    - `my-component-a.spec.tsx` passes, with an error
    - `my-component-b.spec.tsx` does not run
    - `my-component.spec.tsx` runs, but throws an exception
6. In `my-component-a.spec.tsx`, comment in line 23, which manually calls the `disconnectedCallback` (after the test has performed its expect statement)
7. Run the spec tests: `npm run test -- --no-build`
8. Observe: 
   - `my-component-a.spec.tsx` passes, without an error
[...]

PR #4269, opened 2023-04-14
fix(testing): jest component disconnected callback

tanner-reits:
## Pull request checklist

- [X] Docs have been reviewed and added / updated if needed (for bug fixes / features)

## Pull request type

 

 

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

GitHub Issue Number: #4053 

The `disconnectedCallback` will not be called for a component after the test has executed.

## What is the new behavior?

This change will call the `disconnectedCallback` as a part of the `afterEach` for each test that is defined when constructing the Jest test environment.

**NOTE:** This behavior will not call the `disconnectedCallback` for child components. For some reason, the MockDoc implementation does not have reference to child/nested components. So, only the `disconnectedCallback` for `my-component` would be called in the following example:

```html
// my-component.tsx

<div>
  <p>My component</p>
  <my-child-component></my-child-component>
</div>
```

- [ ] Yes
- [X] No

## Testing

Tested this with the reproduction from the original Github issue. Hard to write tests around this since this is a change to the testing ecosystem itself.

### DIFF
--- a/src/testing/jest/jest-setup-test-framework.ts
+++ b/src/testing/jest/jest-setup-test-framework.ts
@@ -7,7 +7,7 @@ import {
   setErrorHandler,
   stopAutoApplyChanges,
 } from '@stencil/core/internal/testing';
-import { setupGlobal, teardownGlobal } from '@stencil/core/mock-doc';
+import { MockDocument, MockNode, MockWindow, setupGlobal, teardownGlobal } from '@stencil/core/mock-doc';
 
 import { expectExtend } from '../matchers';
 import { setupMockFetch } from '../mock-fetch';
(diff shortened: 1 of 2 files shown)


alicewriteswrongs (changes requested):
I think this looks good! I tried it out with the minimal repro locally and afaict it fixes the issue. It would be good to have _some_ sort of test - could we possibly include a modified version of one of the tests in the reproduction case as a regression test?

src/testing/jest/jest-setup-test-framework.ts
rwaskiewicz, line 1: I wonder if we could add a test for this either in `end-to-end` or `karma` - thoughts?

rwaskiewicz (changes requested):
Functionality here appears to work as expected 👍 Had a few asks here before we move forward with this.

src/testing/jest/jest-setup-test-framework.ts
tanner-reits: Chatted with @rwaskiewicz about this and came to the conclusion that there isn't a good, reliable way to test the teardown behavior since that logic executes in a global `afterEach()` callback. Hence, trying to spy on any component's `disconnectedCallback` or public class member won't work since a component's teardown wouldn't happen until after the test has finished and any assertions have already run.

alicewriteswrongs (approved):
lgtm!

rwaskiewicz (approved):

PR #4356, opened 2023-05-04
feat(testing): support puppeteer's 'headless': 'new'

rwaskiewicz:

GitHub Issue Number: N/A

## What is the new behavior?

- 

- [ ] Yes
- [ ] No


src/cli/config-flags.ts
```diff
@@ -193,6 +207,7 @@ export type StringCLIFlag = ArrayValuesAsUnion<typeof STRING_CLI_FLAGS>;
 export type StringArrayCLIFlag = ArrayValuesAsUnion<typeof STRING_ARRAY_CLI_FLAGS>;
 export type NumberCLIFlag = ArrayValuesAsUnion<typeof NUMBER_CLI_FLAGS>;
 export type StringNumberCLIFlag = ArrayValuesAsUnion<typeof STRING_NUMBER_CLI_FLAGS>;
+type BooleanStringCLIFlag = ArrayValuesAsUnion<typeof BOOLEAN_STRING_CLI_FLAGS>;
```
rwaskiewicz, line 210: I don't believe we need to export these anymore, but please LMK if that understanding is incorrect

tanner-reits (approved):
LGTM

src/cli/config-flags.ts
alicewriteswrongs: yeah I don't see any references to them outside of the `config-flags` module - possibly we can stop exporting all of them. I looked back on the PR where they were introduced and I think it was just a 'I guess we should export these?' sort of moment

src/cli/parse-flags.ts
```diff
@@ -300,6 +305,21 @@ const setCLIArg = (flags: ConfigFlags, rawArg: string, normalizedArg: string, va
...
   }
 
+  // We're setting a value which could be either a boolean _or_ a string
+  else if (readOnlyArrayHasStringMember(BOOLEAN_STRING_CLI_FLAGS, normalizedArg)) {
+    const derivedValue =
+      typeof value === 'string'
+        ? value
+          ? value // use the supplied value if it's a non-empty string
+          : false // otherwise, default to false for the empty string
+        : true; // no value was supplied, default to true
```
alicewriteswrongs, line 315: nested ternaries make my head spin haha

basically we want to coerce `""` to `false`, keep any truthy string, and use `true` otherwise, since we handle the `--no-` case above?

src/cli/test/parse-flags.spec.ts
```diff
@@ -132,6 +132,39 @@ describe('parseFlags', () => {
...
+      expect(flags.knownArgs).toEqual(['--headless', 'new']);
+    });
+
+    it("parses a boolean-string flag as a string with 'new' arg using equality", () => {
+      const args = ['--headless=new'];
+      const flags = parseFlags(args);
+      expect(flags.headless).toBe('new');
+      expect(flags.knownArgs).toEqual(['--headless', 'new']);
+    });
+  });
```
alicewriteswrongs, line 166: I know we only have one flag in this array right now, but what would you think about factoring this test as a `describe.each` with `BOOLEAN_STRING_CLI_FLAGS`, like we do for the other lists of flags? see [here](https://github.com/ionic-team/stencil/blob/9690a8226424db4bf2da4540bfdf826980c916ac/src/cli/test/parse-flags.spec.ts#L302-L309) for instance. I know we only have the one flag right now, but it's just nice to have assurance that if we add another in the future we'll always have full coverage

alicewriteswrongs (changes requested):
just a few small things

src/cli/parse-flags.ts
rwaskiewicz: Yah, exactly. I couldn't come up with a "nicer" way to express this without doing something like:
```
let derivedValue = true;
if (typeof value === 'string') {
  if (value) {
    derivedValue = value;
  } else {
    derivedValue = false;
  }
}
```
which felt a little less than clear. IDK though, maybe that's just me (I also don't _love_ the ternary myself). LMK if you'd prefer the `if` statement style above.

src/cli/test/parse-flags.spec.ts
rwaskiewicz: I considered doing this too, I'm up for it!

alicewriteswrongs (approved):

PR #4359, opened 2023-05-05
chore(lint): turn on the 'no-cond-assign' rule

alicewriteswrongs:
Turn on the 'no-cond-assign' eslint rule which disallows doing assignment with a conditional, so you can't do something like

```ts
let foo = { bar : "baz" };
let keyValue = "bizboz";

if (foo.bar = keyValue) {
  otherStuff();
}
```

This is generally a good idea to avoid because assignment in JavaScript evaluates to the assigned value, so assigning to a truthy value will evaluate to a truthy value while assigning a falsy value will evaluate to a falsy value. Thus if you put an assignment in a conditional the value it evaluates to will depend on the value which is assigned, and _not_ on what the variable or property already holds.

In the example above, rather than checking that `foo.bar` is equal to the value `"bizboz"` we're actually checking that `"bizboz"` is truthy. In this case that means that we'll always call the `otherStuff` function regardless of the current value of `foo.bar`, _and_ we'll always have a side-effect of changing the value of `foo.bar`.

This is a bit unfortunate because assignment within a conditional looks very similar to an equality check within a conditional, but it does not evaluate to what you might think it does and also produces a side effect, so it's probably better to do something more explicit if you do actually want this behavior.

For instance, to get the same behavior as the above you could do:

```ts
let foo = { bar : "baz" };
let keyValue = "bizboz";

if (keyValue) {
  foo.bar = keyValue;
  otherStuff();
}
```

which accomplishes the same thing with just one more line of code and is more clear.

Happily at present we don't have any violations of this rule in the codebase, so we have no errors to correct.

GitHub Issue Number: N/A

## What is the new behavior?

- 

- [ ] Yes
- [ ] No

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -65,20 +65,21 @@ module.exports = {
         checkSetters: true,
       },
     ],
+    'jsdoc/require-param-description': ['error'],
     // rely on TypeScript types to be the source of truth, minimize verbosity in comments
     'jsdoc/require-param-type': ['off'],
-    'jsdoc/require-param-description': ['error'],
+    'jsdoc/require-returns': ['error'],
     'jsdoc/require-returns-check': ['error'],
     'jsdoc/require-returns-description': ['error'],
     // rely on TypeScript types to be the source of truth, minimize verbosity in comments
     'jsdoc/require-returns-type': ['off'],
-    'jsdoc/require-returns': ['error'],
-    'prefer-const': 'error',
+    'no-cond-assign': 'error',
     'no-var': 'error',
+    'prefer-const': 'error',
     'prefer-rest-params': 'error',
     'prefer-spread': 'error',
-    'simple-import-sort/imports': 'error',
     'simple-import-sort/exports': 'error',
+    'simple-import-sort/imports': 'error',
   },
   overrides: [
     {


.eslintrc.js
```diff
@@ -79,6 +79,7 @@ module.exports = {
     'prefer-spread': 'error',
     'simple-import-sort/imports': 'error',
     'simple-import-sort/exports': 'error',
+    'no-cond-assign': 'error',
```
rwaskiewicz, line 82: Non-blocking nit - can move this up to keep this list alphabetized?

rwaskiewicz (approved):

tanner-reits (approved):
Yes, thank you! One of my least favorite dev patterns

.eslintrc.js
```diff
@@ -66,20 +66,20 @@ module.exports = {
       },
     ],
     // rely on TypeScript types to be the source of truth, minimize verbosity in comments
-    'jsdoc/require-param-type': ['off'],
     'jsdoc/require-param-description': ['error'],
+    'jsdoc/require-param-type': ['off'],
```
rwaskiewicz, line 70: FYI - I think this comment https://github.com/ionic-team/stencil/pull/4359/files/45309a3da216b947d8954daf072ab24466bc736e..a87503149155bc39316303d29c574920cc8f569d#diff-e2954b558f2aa82baff0e30964490d12942e0e251c1aa56c3294de6ec67b7cf5R68 got detached from this rule

alicewriteswrongs: I think I pushed an intermediate commit that's not there anymore, afaict it's good now

rwaskiewicz: Hmm...I don't think I'm seeing it:

Screenshot 2023-05-05 at 10 22 43 AM

rwaskiewicz (approved):

Issue #4349, opened 2023-05-01
bug: Path aliases are present in the generated type declarations


sean-perkins:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

v3.2.2

### Current Behavior

When using path aliases, both the `src/components.d.ts` and `dist/types/components.d.ts` reference the alias path instead of the relative path.

```ts
// dist/types/components.d.ts
import { MaskVisibility } from "@utils/input-masking/types";
export { MaskVisibility } from "@utils/input-masking/types";
```

### Expected Behavior

The generated type declaration file should be relatively pathed and all import aliases should be replaced.

### System Info

```shell
System: node 16.15.1
    Platform: darwin (22.2.0)
   CPU Model: Apple M1 (8 cpus)
    Compiler: /Users/sean/Documents/ionic/issues/stencil-alias-imports/node_modules/@stencil/core/compiler/stencil.js
       Build: 1682947516
     Stencil: 3.2.2 🏒
  TypeScript: 4.9.5
      Rollup: 2.42.3
      Parse5: 7.1.2
      Sizzle: 2.42.3
      Terser: 5.17.1
```
```


### Steps to Reproduce

1. Add a prop to a Stencil component, that has a type from an import alias.

```tsx
import { MaskVisibility } from '@utils/input-masking/types';

...

@Prop() maskVisibility: MaskVisibility;
```

2. Run `stencil build` to generate the type declaration files.
3. Observe: The import aliases are still present in the compiled output.

### Code Reproduction URL

https://github.com/sean-perkins/stencil-named-imports

rwaskiewicz:
The fix for this has gone out in today's [Stencil v3.3.0 release](https://github.com/ionic-team/stencil/releases/tag/v3.3.0). As a result, I'm going to close out this ticket. If the issue should reappear, please feel free to open a new issue. Thanks!

PR #4365, opened 2023-05-08
fix(compiler): components typedef path aliases

tanner-reits:

We do not resolve any aliased paths when generating the `components.d.ts` file(s).

GitHub Issue Number: #4349 

## What is the new behavior?

Added module resolution logic for path aliases to the typedef file generation code. This will only attempt to resolve modules for _potential_ aliased paths via the TS API. We don't know if a module is aliased (i.e. we can't rely on anything with an `@` in front of it to be an alias) until TS resolves it and tells us whether it is an external module (something in `node_modules`) so there is some conditional logic to prevent turning those paths into relative paths.

- [ ] Yes
- [X] No

## Testing

Manually tested with a starter app/issue repro to verify aliased and non-aliased paths generate correctly.

Added a unit test to verify path aliases get resolved via TS. _However,_ we are limited in our abilities to test the actual module resolution behavior (same problem we encountered with this type of logic in the past), so we will eventually want to update this test with "real" behavior.


src/compiler/types/update-import-refs.ts
```diff
@@ -83,6 +91,21 @@ const updateImportReferenceFactory = (typeCounts: Map<string, number>, filePath:
...
+
+          // We only care to resolve any _potential_ aliased
+          // modules if we're not already certain the path isn't an alias
+          if (!importResolvedFile.startsWith('.')) {
+            const { resolvedModule } = ts.resolveModuleName(
+              typeReference.path,
+              filePath,
+              tsOptions,
+              ts.createCompilerHost(tsOptions)
+            );
```
tanner-reits, line 103: Just FYI, this logic will currently transform the path aliases for the `src/components.d.ts` file we automatically generate at build time as well. This doesn't really cause any problems, but if we think that's confusing to users or just want to leave aliases in that file, I can update this logic a bit to ignore this block of code when writing to the `src` dir.

Aliases can exist there since it's not distributed code and the TS language server/any IDE would be able to handle the aliases in there.

Open to whatever the team thinks is best.

tanner-reits (commented):
Just calling out one thing.

Also, I put the abstraction logic on the back-burner and created [STENCIL-833](https://ionic-cloud.atlassian.net/browse/STENCIL-833) to eventually circle-back to it. It was eating too much time and I kept running into issues with all the different nuanced use-cases and wanted to get some other tasks done before I'm out for a few days.

src/compiler/types/update-import-refs.ts
```diff
@@ -81,8 +89,23 @@ const updateImportReferenceFactory = (typeCounts: Map<string, number>, filePath:
           // If local then import location is the current file
         } else if (typeReference.location === 'local') {
           importResolvedFile = filePath;
-        } else if (typeReference.location === 'import') {
-          importResolvedFile = typeReference.path;
+        } else {
```
tanner-reits, line 92: I _believe_ it's okay to remove the `else if` block here since the only valid values for `location` are `local`, `global`, and `import`. But lemme know if anyone thinks this is risky: only did it to remove some SNC violations

src/compiler/types/tests/generate-app-types.spec.ts
```diff
@@ -1406,6 +1407,198 @@ declare module "@stencil/core" {
...
+
+  it('should transform aliased paths if the config option is true', async () => {
+    // TODO(STENCIL-223): remove spy to test actual resolution behavior
+    jest.spyOn(ts, 'resolveModuleName').mockReturnValue({
+      resolvedModule: {
+        isExternalLibraryImport: false,
+        extension: Extension.Dts,
+        resolvedFileName: 'utils',
+      },
+    });
```
alicewriteswrongs, line 1425: so for now the test isn't testing that the values in `paths` are used - I assume this was the stuff you had trouble with testing? when testing the `rewrite-aliased-paths` transformer I was able to get this to work by making sure I called `patchTypescript`, which then let me create files in the in-memory file system (see this setup function for instance: https://github.com/ionic-team/stencil/blob/39ab8d9c00be08216d82ce984bbf7a28c28a93f2/src/compiler/transformers/test/rewrite-aliased-paths.spec.ts#L21-L51)

Any chance a similar approach would work here?

alicewriteswrongs (changes requested):
just had one question, otherwise I think this makes sense generally

src/compiler/types/update-import-refs.ts
rwaskiewicz: Yah - I think this is OK as well.

src/compiler/types/tests/generate-app-types.spec.ts
tanner-reits: Ooh nifty! I'll check this out today.

rwaskiewicz (commented):
Had a few minor comments, will TAL again after we see what the outcome of the testing strategy Alice mentioned was/is

src/compiler/types/update-import-refs.ts
```diff
@@ -81,8 +89,25 @@ const updateImportReferenceFactory = (typeCounts: Map<string, number>, filePath:
...
+          importResolvedFile = typeReference.path!;
+
+          // We only care to resolve any _potential_ aliased
+          // modules if we're not already certain the path isn't an alias.
+          // We also don't want to transform aliases unless the user has enabled the behavior
+          // in their Stencil config
+          if (config.transformAliasedImportPaths && !importResolvedFile.startsWith('.')) {
+            const { resolvedModule } = ts.resolveModuleName(
```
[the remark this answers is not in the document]
tanner-reits: Yep. It should return
```
{
  resolvedModule: undefined;
}
```

But, it's also supposed to return the failed lookup locations. Doesn't look like this is included in the typedefs for whatever reason, but I _believe_ the type would be:
```
{
  resolvedModule: undefined;
  failedLookupLocations: _whateverShapeThisIs_[]
}
```
I have seen the API return the lookup locations, so not sure why it's not on the object, but it should be there if we wanted those for some reason

src/compiler/types/tests/generate-app-types.spec.ts
```diff
@@ -1406,6 +1409,198 @@ declare module "@stencil/core" {
...
         }
 }
+`,
+      {
+        immediateWrite: true,
+      }
+    );
+  });
+
+  it('should transform aliased paths if the config option is true', async () => {
```
rwaskiewicz, line 1419: nit - can we add the name of the config option in the test here?
```suggestion
  it('should transform aliased paths if transformAliasedImportPaths is true', async () => {
```

src/compiler/types/tests/generate-app-types.spec.ts
```diff
@@ -1406,6 +1409,198 @@ declare module "@stencil/core" {
...
+    buildCtx.components = [compilerComponentMeta];
+    config.tsCompilerOptions = {
+      paths: {
+        '@utils': [path.join(config.rootDir, 'some/stubbed/path/utils/utils.ts')],
+      },
+      declaration: true,
+    };
+    config.transformAliasedImportPaths = true;
+    // We need to have a file in the in-memory fs for the TS module resolution to succeed
+    originalWriteFile(path.join(config.rootDir, 'some/stubbed/path/utils/utils.ts'), '');
```
rwaskiewicz, line 1452: Technically, this method is async in TS' eyes - can we add `await` to it?
```suggestion
    await originalWriteFile(path.join(config.rootDir, 'some/stubbed/path/utils/utils.ts'), '');
```

src/compiler/types/tests/generate-app-types.spec.ts
```diff
@@ -1406,6 +1409,198 @@ declare module "@stencil/core" {
...
+
+    patchTypescript(config, compilerCtx.fs);
```
rwaskiewicz, line 1454: Can we 'unpatch' this in between tests?

rwaskiewicz (approved):
A few non-blocking nits/asks

src/compiler/types/tests/generate-app-types.spec.ts
tanner-reits: Easily? I don't think so. Looks like that utility overwrites a few methods on the TS object, but it saves the originals on properties we create (like `__resolveModuleName`). So, we _could_ create an `unpatchTypescript` method that reassigns the original definitions to their signatures.

alicewriteswrongs (approved):
this looks good to me!

src/compiler/types/tests/generate-app-types.spec.ts
rwaskiewicz: Yeah, you're right - let's punt on this for now

PR #4368, opened 2023-05-09
add support for node 20

alicewriteswrongs:
This adds support for node 20 by

- adding it to the node version matrix for CI tests
- bumping the version we specify via Volta in `package.json`
- bumping the version of `@types/node`
- editing the renovate configuration to allow for newer versions of `@types/node`

GitHub Issue Number: N/A

## What is the new behavior?

- 

- [ ] Yes
- [x] No

### DIFF
--- a/src/utils/test/util.spec.ts
+++ b/src/utils/test/util.spec.ts
@@ -158,7 +158,7 @@ describe('util', () => {
       const expectedDiagnostic: d.Diagnostic = stubDiagnostic({
         absFilePath: mockPackageJsonPath,
         header: 'Error Parsing JSON',
-        messageText: 'Unexpected string in JSON at position 13', // due to missing colon in input
+        messageText: expect.stringMatching(/.*in JSON at position 13$/),
         type: 'build',
       });
 
--- a/.github/workflows/test-analysis.yml
+++ b/.github/workflows/test-analysis.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node: ['14', '16', '18']
+        node: ['14', '16', '18', '20']
         os: ['ubuntu-20.04', 'windows-latest']
     runs-on: ${{ matrix.os }}
     steps:
--- a/.github/workflows/test-component-starter.yml
+++ b/.github/workflows/test-component-starter.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node: ['14', '16', '18']
+        node: ['14', '16', '18', '20']
         os: ['ubuntu-latest', 'windows-latest']
     runs-on: ${{ matrix.os }}
     steps:
--- a/.github/workflows/test-e2e.yml
+++ b/.github/workflows/test-e2e.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node: ['14', '16', '18']
+        node: ['14', '16', '18', '20']
         os: ['ubuntu-20.04', 'windows-latest']
     runs-on: ${{ matrix.os }}
     steps:
--- a/.github/workflows/test-unit.yml
+++ b/.github/workflows/test-unit.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node: ['14', '16', '18']
+        node: ['14', '16', '18', '20']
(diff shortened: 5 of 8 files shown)


src/utils/test/util.spec.ts
```diff
@@ -155,10 +155,16 @@ describe('util', () => {
       // improperly formatted JSON - note the lack of ':'
       const diagnostic = util.parsePackageJson('{ "someJson" "value"}', mockPackageJsonPath);
 
+      // due to missing colon in input
+      // the text of this message was changed in node version 20
+      const messageText = Number(process.versions.node.split(".")[0]) > 20
+        ? "Unexpected string in JSON at position 13"
+        : "Expected ':' after property name in JSON at position 13";
+
```
alicewriteswrongs, line 163: this is def weird, do we have an existing pattern for doing this / is there a better way?

The problem is just that the text of this error message is different in node 20 than it is in 14, 16, and 18.

alicewriteswrongs (commented):
just pointing out one funky bit

alicewriteswrongs (commented):
just pointing out one funky bit

src/utils/test/util.spec.ts
rwaskiewicz: Yeah, this is definitely weird. IDK if we have a better pattern/prior art for dealing with it.

Alternative to the ternary here, maybe we re-assess the value of the message itself (in part or as a whole) in this test. Maybe we consider asserting that this is _some_ string ending with "in JSON at position 13" or just a string as a whole?

tanner-reits: I would probably just check that is identifies the correct location/position

src/utils/test/util.spec.ts
```diff
@@ -155,10 +155,17 @@ describe('util', () => {
       // improperly formatted JSON - note the lack of ':'
       const diagnostic = util.parsePackageJson('{ "someJson" "value"}', mockPackageJsonPath);
 
+      // due to missing colon in input
+      // the text of this message was changed in node version 20
+      const messageText =
```
rwaskiewicz, line 160: Looks like this is failing CI... I hope we're using the version of Node we say we are 😬

src/utils/test/util.spec.ts
alicewriteswrongs: that sounds reasonable to me!

src/utils/test/util.spec.ts
```diff
@@ -155,18 +155,30 @@ describe('util', () => {
       // improperly formatted JSON - note the lack of ':'
       const diagnostic = util.parsePackageJson('{ "someJson" "value"}', mockPackageJsonPath);
 
+      // due to missing colon in input
+      const messageText =
```
rwaskiewicz, line 159: FWIW, I think that the build will fail due to this variable and `expectedDiagnostic` no longer being used.

src/utils/test/util.spec.ts
```diff
@@ -155,18 +155,30 @@ describe('util', () => {
...
-        messageText: 'Unexpected string in JSON at position 13', // due to missing colon in input
         type: 'build',
       });
 
-      expect(diagnostic).toEqual<util.ParsePackageJsonResult>({
-        diagnostic: expectedDiagnostic,
-        data: null,
-        filePath: mockPackageJsonPath,
-      });
+      expect(diagnostic.filePath).toBe(mockPackageJsonPath);
```
rwaskiewicz, line 170: I think we can avoid splitting all this out (and fixing the unused variable issues) if we use the [`expect.stringMatching` matcher:](https://jestjs.io/docs/expect#expectstringmatchingstring--regexp)

```
    it('returns a parse error if parsing cannot complete', () => {
      // improperly formatted JSON - note the lack of ':'
      const diagnostic = util.parsePackageJson('{ "someJson" "value"}', mockPackageJsonPath);

      const expectedDiagnostic: d.Diagnostic = stubDiagnostic({
        absFilePath: mockPackageJsonPath,
        header: 'Error Parsing JSON',
        messageText: expect.stringMatching(/.*in JSON at position 13$/),
        type: 'build',
      });

      expect(diagnostic).toEqual<util.ParsePackageJsonResult>({
        diagnostic: expectedDiagnostic,
        data: null,
        filePath: mockPackageJsonPath,
      });
    });
```
Thoughts?

rwaskiewicz (changes requested):

tanner-reits (approved):

PR #4373, opened 2023-05-11
chore(snc): fix some SNC errors in src/compiler/plugin/plugin.ts

alicewriteswrongs:
This just fixes some SNC errors in `src/compiler/plugin/plugin.ts`, mainly by switching a few type annotations to instead use the `satisfies` operator and also adding some null property access checks w/ the nullish coalescing operator.

GitHub Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/src/compiler/plugin/plugin.ts
+++ b/src/compiler/plugin/plugin.ts
@@ -6,7 +6,7 @@ import { PluginCtx, PluginTransformResults } from '../../declarations';
 import { parseCssImports } from '../style/css-imports';
 
 export const runPluginResolveId = async (pluginCtx: PluginCtx, importee: string) => {
-  for (const plugin of pluginCtx.config.plugins) {
+  for (const plugin of pluginCtx.config?.plugins ?? []) {
     if (isFunction(plugin.resolveId)) {
       try {
         const results = plugin.resolveId(importee, null, pluginCtx);
@@ -32,7 +32,7 @@ export const runPluginResolveId = async (pluginCtx: PluginCtx, importee: string)
 };
 
 export const runPluginLoad = async (pluginCtx: PluginCtx, id: string) => {
-  for (const plugin of pluginCtx.config.plugins) {
+  for (const plugin of pluginCtx.config?.plugins ?? []) {
     if (isFunction(plugin.load)) {
       try {
         const results = plugin.load(id, pluginCtx);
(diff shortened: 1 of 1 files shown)


src/compiler/plugin/plugin.ts
```diff
@@ -81,10 +81,11 @@ export const runPluginTransforms = async (
     return null;
   }
 
-  const transformResults: PluginTransformResults = {
+  const transformResults = {
     code: sourceText,
     id: id,
-  };
+    dependencies: [] as string[],
+  } satisfies PluginTransformResults;
```
alicewriteswrongs, line 88: the `satisfies` operator is my favorite addition to typescript in a while!

src/compiler/plugin/plugin.ts
```diff
@@ -198,13 +199,13 @@ export const runPluginTransformsEsmImports = async (
     diagnostics: [],
   };
 
-  const transformResults: PluginTransformResults = {
+  const transformResults = {
```
rwaskiewicz, line 202: FWIW, if we used a type annotation here, we wouldn't need to add type assertions:
```ts
  const transformResults: PluginTransformResults = {
    code,
    id,
    map: undefined,
    diagnostics: [],
    dependencies: [],
  }
```
but at the same time `satisfies` will yell if we do something like
```ts
  const transformResults = {
    code,
    id,
    map: undefined as string | undefined,
    diagnostics: [] as d.Diagnostic[] | number, // <- Hey now!
    dependencies: [] as string[],
  } satisfies PluginTransformResults;
```
So...no action required 👍

rwaskiewicz (approved):

src/compiler/plugin/plugin.ts
alicewriteswrongs: yeah, it's a bit funny because `never[]` for `.diagnostics` on the object literal (which on `PluginTransformResults` has type `Diagnostic[]`) will technically satisfy that type, and without an annotation if you just write `diagnostics: []` TypeScript infers that `transformResults.diagnostics` has type `never[]`....so in order to tell it later on that yeah, assigning a value of type `Diagnostic[]` to `.diagnostics` is 100% ok we need to do the type assertion. But yeah as you said, if the type assertion conflicts with the type you put after `satisfies` (pass to satisfies?) you'll get a type error.

I think with `satisfies` in general TypeScript will try to keep the widest possible type that it can infer for a given value. So like if you do

```ts
const foo = {
  ...
} satisfies MyInterface;
```

the type of `Foo` in code after that will be whatever is the most general or least restricted / constrained thing that TypeScript can infer based on the actual object literal there. It just _also_ happens to check that the object literal satisfies `MyInterface`. I really like it because it allows optional types on the interface _and_ correct inference that if I declare an object literal with a value for that property it _is_ defined in the code that immediately follows the declaration.

I think it will let us clear up a fair few more SNC errors here and there!

tanner-reits (approved):

PR #4375, opened 2023-05-11
refactor(validate-testing): coalesce testRegex tests

rwaskiewicz:

the tests for the `testRegex` property take up a large amount of space, when `it.each` will do the trick

GitHub Issue Number: N/A

## What is the new behavior?

this commit coalesces some of the testing around the `testRegex` property on the testing config by using `it.each` rather than an explicit test per case.

- [ ] Yes
- [x] No

## Testing

Unit tests continue to pass, there should be 1:1 parity between input/output between before and after

## Other information

1. for this refactoring, it was a deliberate decision to not use `describe.each` for [spec, test, e2e], with the hope/assumption that not using this api would allow for more flexibility between these file extensions in the future should we need it.

2. arguably the tests are a little harder to understand now that we don't spell out each individual case. however, i think this is a worthwhile tradeoff to remove some ~175 lines of mind-numbing duplication


alicewriteswrongs (approved):
brevity is best 😌

tanner-reits (approved):

PR #4381, opened 2023-05-12
chore(snc): fix SNC errors relating to `HostRef.$hostElement$`

alicewriteswrongs:
This makes the `$hostElement$` property required on the `HostRef` interface. I believe this is actually in line with how this property is already used, since I can't find an instance where we create a `HostRef` without supplying a value for `$hostElement$`. Since this property is accessed a lot throughout the runtime this drops a number of SNC errors.

GitHub Issue Number: N/A

## What is the new behavior?

- 

- [ ] Yes
- [ ] No

### DIFF
--- a/src/declarations/stencil-private.ts
+++ b/src/declarations/stencil-private.ts
@@ -1730,7 +1730,7 @@ export interface HostRef {
   $ancestorComponent$?: HostElement;
   $flags$: number;
   $cmpMeta$: ComponentRuntimeMeta;
-  $hostElement$?: HostElement;
+  $hostElement$: HostElement;
   $instanceValues$?: Map<string, any>;
   $lazyInstance$?: ComponentInterface;
   $onReadyPromise$?: Promise<any>;
--- a/src/runtime/profile.ts
+++ b/src/runtime/profile.ts
@@ -2,8 +2,6 @@ import { BUILD } from '@app-data';
 import { getHostRef, win } from '@platform';
 import { HOST_FLAGS } from '@utils';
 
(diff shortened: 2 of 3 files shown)


rwaskiewicz (commented):
Nice!

I think we can remove one additional type assertion as a result of this, in `src/runtime/profile.ts`:
```diff
diff --git a/src/runtime/profile.ts b/src/runtime/profile.ts
index 55bb3a383..a3f1324fc 100644
--- a/src/runtime/profile.ts
+++ b/src/runtime/profile.ts
@@ -2,8 +2,6 @@ import { BUILD } from '@app-data';
 import { getHostRef, win } from '@platform';
 import { HOST_FLAGS } from '@utils';
 
-import type * as d from '../declarations';
-
 let i = 0;
 
 export const createTime = (fnName: string, tagName = '') => {
@@ -44,7 +42,7 @@ const inspect = (ref: any) => {
     return undefined;
   }
   const flags = hostRef.$flags$;
-  const hostElement = hostRef.$hostElement$ as d.HostElement;
+  const hostElement = hostRef.$hostElement$;
   return {
     renderCount: hostRef.$renderCount$,
     flags: {

```

The file below is a patch that can be applied with `git patch`, git just doesn't all `.diff` extensions for...reasons? 🤷 
[rm-type-assert.txt](https://github.com/ionic-team/stencil/files/11466489/rm-type-assert.txt)

rwaskiewicz (approved):

tanner-reits (approved):

PR #4395, opened 2023-05-17
feat(compiler): primary package output target validation

tanner-reits:

Validation for `package.json` fields currently follows a hierarchy that is defined by Stencil. 

GitHub Issue Number: N/A

## What is the new behavior?

Stencil will now allow developers to designate a target to use for validation. This validation only applies to common `package.json` fields (currently only `module` and `types`).

- [ ] Yes
- [X] No

## Testing

Tested the different types on a starter project.

Added automated tests for various scenarios and output target types.


src/declarations/stencil-public-compiler.ts
```diff
@@ -2186,6 +2188,11 @@ export interface OutputTargetBase {
   type: string;
 }
 
+// TODO(NOW): rename
+export interface ValidatableOutputTarget extends OutputTargetBaseNext {
```
sean-perkins, line 2192: Maybe `OutputTargetValidationConfig`/`OutputTargetValidationOptions`? Or if we wanted to represent it as the "base options" for output target validation: `OutputTargetBaseValidationConfig` (although I dislike the length of this type at that point).

src/compiler/types/validate-primary-package-output-target.ts
```diff
@@ -0,0 +1,229 @@
...
+      return outputTargetConfig.generateTypeDeclarations
+        ? normalizePath(relative(rootDir, join(rootDir, 'dist/types/index.d.ts')))
+        : null;
+    },
+  },
+  'dist-types': {
+    getTypesPath: (rootDir: string, outputTargetConfig: d.OutputTargetDistTypes) =>
+      normalizePath(relative(rootDir, join(outputTargetConfig.typesDir, 'index.d.ts'))),
+  },
+};
```
alicewriteswrongs, line 38: you might get nicer typing here if you use my new absolute favorite thing in typescript, the `satisfies` keyword:

```suggestion
const PRIMARY_PACKAGE_TARGET_CONFIGS = {
  dist: {
    getModulePath: (rootDir: string, outputTargetDir: string) =>
      normalizePath(relative(rootDir, join(outputTargetDir, 'index.js'))),
    getTypesPath: (rootDir: string, outputTargetConfig: d.OutputTargetDist) =>
      normalizePath(relative(rootDir, join(outputTargetConfig.typesDir, 'index.d.ts'))),
  },
  'dist-collection': {
    getModulePath: (rootDir: string, outputTargetDir: string) =>
      normalizePath(relative(rootDir, join(outputTargetDir, 'index.js'))),
  },
  'dist-custom-elements': {
    getModulePath: (rootDir: string, outputTargetDir: string) =>
      normalizePath(relative(rootDir, join(outputTargetDir, 'index.js'))),
    getTypesPath: (rootDir: string, outputTargetConfig: d.OutputTargetDistCustomElements) => {
      return outputTargetConfig.generateTypeDeclarations
        ? normalizePath(relative(rootDir, join(rootDir, 'dist/types/index.d.ts')))
        : null;
    },
  },
  'dist-types': {
    getTypesPath: (rootDir: string, outputTargetConfig: d.OutputTargetDistTypes) =>
      normalizePath(relative(rootDir, join(outputTargetConfig.typesDir, 'index.d.ts'))),
  },
} satisfies Record<string, PrimaryPackageOutputTargetRecommendedConfig>;
```

then you'll still get type-checking that `getTypesPath` and `getModulePath` have the right type (if they're defined) but you'll _also_ still get a more specific type for the `PRIMARY_PACKAGE_TARGET_CONFIGS` than a `Record<...>`

alicewriteswrongs (changes requested):
just took a quick look through a few things, will take a more thorough look through in a bit!

src/compiler/types/validate-primary-package-output-target.ts
```diff
@@ -0,0 +1,229 @@
...
+    : null;
+
+  let warningMessage: string;
+  if (!isString(currentModulePath)) {
+    warningMessage = 'package.json "module" property is required when generating a distribution.';
+
+    if (recommendedModulePath != null) {
+      warningMessage += ` It's recommended to set the "module" property to: ${recommendedModulePath}`;
+    }
+    return;
```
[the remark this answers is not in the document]
tanner-reits: Ope yeah totally meant to remove that 😅

src/compiler/types/validate-primary-package-output-target.ts
tanner-reits: Oooh yeah this does play nicely with your first suggestion to get type-safety on the keys as well (which is something I was having trouble with)

src/compiler/types/package-json-log.utils.ts
tanner-reits, line 1: Just abstracted these helpers into a separate file since I needed to use them in two different files now

src/compiler/types/validate-primary-package-output-target.ts
```diff
@@ -0,0 +1,222 @@
...
+ * @param compilerCtx The project's compiler context.
+ * @param buildCtx The project's build context.
+ */
+export const validatePrimaryPackageOutputTarget = (
+  config: d.ValidatedConfig,
+  compilerCtx: d.CompilerCtx,
+  buildCtx: d.BuildCtx
+) => {
+  // TODO(NOW): should _all_ validation be gated behind the config flag, or just this
+  // new layer of "primary" target validation?
```
tanner-reits, line 53: I think this is the only outstanding question. The current implementation seems like the right approach, but we can gate _all_ validation behind the config flag if we wanted.

Also, at the time of writing this, I'm realizing I didn't leave the previous validation logic in place if the new config flag wasn't enabled. Do we want to have the old logic run if the new flag isn't enabled?

src/compiler/types/validate-primary-package-output-target.ts
tanner-reits, line 1: (Almost) everything is exported from this file simply for testing purposes

tanner-reits (commented):
Noting a few things

alicewriteswrongs (changes requested):
I think this looks good, just had a few questions (feel free to say 'no let's merge this' haha)

src/declarations/stencil-public-compiler.ts
```diff
@@ -2186,6 +2191,20 @@ export interface OutputTargetBase {
...
+ */
+interface OutputTargetValidationConfig extends OutputTargetBaseNext {
+  isPrimaryPackageOutputTarget?: boolean;
+}
+
+export type PrimaryPackageOutputTarget =
+  | OutputTargetDist
+  | OutputTargetDistCustomElements
+  | OutputTargetDistCollection
+  | OutputTargetDistTypes;
```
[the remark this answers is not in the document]
tanner-reits: I think that's a fair change! Naming has honestly been some of the hardest parts of this rework 😂

src/compiler/types/validate-primary-package-output-target.ts
```diff
@@ -0,0 +1,222 @@
...
+          `Your Stencil project has not assigned a primary package output target. Stencil recommends that you assign a primary output target so it can validate values for fields in your project's 'package.json'`
+        );
+      }
+    }
+
+    // Log a warning if any targets that cannot be validated were marked as "primary"
+    if (nonPrimaryTargets.length && nonPrimaryTargets.some((ref: any) => ref.isPrimaryPackageOutputTarget)) {
+      logValidationWarning(
+        buildCtx,
+        `Your Stencil project has assigned one or more un-validated output targets as the primary package output target. No validation will take place. Please remove the 'isPrimaryPackageOutputTarget' flag from the following output targets in your Stencil config: ${nonPrimaryTargets
```
rwaskiewicz, line 107: In my local testing, I made the following changes to my stencil config and ran build:

Screenshot 2023-05-24 at 1 06 58 PM

Should we suppress the first message if we print the second? The two seem to be conflicting with one another

rwaskiewicz (changes requested):

src/compiler/types/validate-build-package-json.ts
```diff
@@ -27,22 +19,20 @@ export const validateBuildPackageJson = async (
...
+  // Validate any output target that the user has designated as the "primary"
+  // target that is bundled with their distribution
+  validatePrimaryPackageOutputTarget(config, compilerCtx, buildCtx);
+
   const distCollectionOutputTargets = config.outputTargets.filter(isOutputTargetDistCollection);
-  const typesOutputTargets = config.outputTargets.filter(isOutputTargetDistTypes);
-  await Promise.all([
-    ...distCollectionOutputTargets.map((distCollectionOT) =>
+  await Promise.all(
+    distCollectionOutputTargets.map((distCollectionOT) =>
```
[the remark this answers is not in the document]
tanner-reits: There was a spread before because we also `await`ed the `validateTypes` and `validateModule` methods. Since that validation is now based on the "primary" target, I removed those calls so we only needed to `await` the validation for each of the other fields that may exist on `dist-collection` here now

src/compiler/types/package-json-log.utils.ts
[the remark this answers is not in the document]
tanner-reits: Yeah sorry, just a naming convention I use in other projects so hard to break the habit

src/compiler/types/tests/validate-primary-package-output-target.spec.ts
```diff
@@ -0,0 +1,296 @@
...
+
+    it('should log a warning if the types path does not have a ".d.ts" extension', () => {
+      buildCtx.packageJson.types = '/dist/types/index.ts';
+
+      validateTypesPath(config, compilerCtx, buildCtx, recommendedOutputTargetConfig, targetToValidate);
+
+      expect(buildCtx.diagnostics.length).toBe(1);
+      expect(buildCtx.diagnostics[0].level).toEqual('warn');
+      expect(buildCtx.diagnostics[0].messageText).toEqual(
+        `package.json "types" file must have a ".d.ts" extension: /dist/types/index.ts`
```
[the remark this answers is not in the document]
tanner-reits: It's actually just spitting back the path the user currently has set. This was an existing test that I just pulled over to the updated validation logic. I can update the message a bit to make that clearer

src/compiler/types/validate-primary-package-output-target.ts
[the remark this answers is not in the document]
tanner-reits: It does list each output target type that violates the rules, but I'll add a link to the docs

tanner-reits: I don't think there's necessarily anything wrong with logging both messages since there are technically two configuration errors. This is something I think we roll with for now (with the added link to docs in each message) and react if we hear feedback that the messages are confusing, conflicting, or whatever

src/compiler/config/test/validate-config.spec.ts
```diff
@@ -480,4 +480,20 @@ describe('validation', () => {
...
+
+      expect(config.validatePrimaryPackageOutputTarget).toBe(false);
+    });
+
+    it.each([true, false])('should set validatePrimaryPackageOutputTarget to %p', (transformAliasedImportPaths) => {
+      userConfig.transformAliasedImportPaths = transformAliasedImportPaths;
+
+      const { config } = validateConfig(userConfig, bootstrapConfig);
+
+      expect(config.transformAliasedImportPaths).toBe(transformAliasedImportPaths);
```
rwaskiewicz, line 496: I think this test is still using `transformAliasedImportPaths` instead of `validatePrimaryPackageOutputTarget` (Copy Pasta maybe?)

rwaskiewicz (approved):

alicewriteswrongs (approved):
I think this looks good!

PR #4401, opened 2023-05-18
chore(snc): remove in-memory-fs.spec snc violations

rwaskiewicz:

GitHub Issue Number: N/A

## What is the new behavior?

this commit removes the `strictNullChecks` violations found in `in-memory-fs.spec.ts`. there are two types of changes found within this file:
1. `fsItem`, a helper function in the suite, has been updated to set it's default values to `undefined` to adhere to `Partial<FsItem>`, fixing assignability issues with setting each field to null
2. retrieval of entries from the `items` map use optional chaining. this was deemed safe as we are not asserting against `undefined` in any of the `get` calls in said tests

- [ ] Yes
- [x] No

## Testing

Unit tests continue to pass


alicewriteswrongs (approved):
looks good!

PR #4415, opened 2023-05-22
chore(revert): "refactor(compiler): add a Result union type for polymorphic returns (#4281)"

rwaskiewicz:

GitHub Issue Number: N/A

## What is the new behavior?

This reverts commit 3811a4c4714f9fb0f88dc1f798959196721548e5.

This commit is being reverted due to an issue with running Stencil locally. Starting with the commit to be reverted, this is reproducible by:
```
npm run clean && /
npm ci && /
npm run build /
npm pack
```
then
```
cd /tmp && /
npm init stencil component happy-monday && /
npm i <TARBALL>
```

- [ ] Yes
- [x] No

## Testing

See steps above

## Other information

I didn't use `refactor()` in the semantic commit msg, since this only affects internal types (and `refactor()` gets put in the changelog)


tanner-reits (approved):
Looks good. Did a quick side-by-side comparison of this commit and the reverted commit. Everything seemed to match up 👍

PR #4418, opened 2023-05-24
feat(config): set new defaults for transformAliasedImportPaths

rwaskiewicz:

the `transformAliasedImportPaths` and `transformAliasedImportPathsInCollection` configuration fields are opt-in (set to `false`) by default.

GitHub Issue Number: N/A

## What is the new behavior?

this commit updates the default values for the
`transformAliasedImportPaths` and
`transformAliasedImportPathsInCollection` configuration fields. both are being updated to be set to `true` if neither a `true` nor `false` value is provided (previously `false` if `false` or nothing was provided).

the latter of the two is used to set the former for, hence its new default.

by setting both of these new defaults to `true`, the hope is that stencil configuration will be simplified for projects using `paths` in their `tsconfig.json`

- [x] Yes
- [ ] No

BREAKING CHANGE: `transformAliasedImportPaths` and `transformAliasedPathsInCollection` both default to `true`. To opt-out of these new defaults, both may be set to `false` in `stencil.config.ts`. See https://stenciljs.com/docs/introduction/upgrading-to-stencil-four#transformaliasedimportpaths and https://stenciljs.com/docs/introduction/upgrading-to-stencil-four#transformaliasedimportpathsincollection (respectively) for more information

## Testing

1. `npm ci && npm run clean && npm run build && npm pack` to generate a tarball with this build
2. Install the tarball on the Ionic Framework's core package (`npm i <PATH TO TARBALL>`)
3. `npm run build` and `npx stencil test --spec -- --no-cache` should pass
4. Update `stencil.config.ts` in the core package:
```diff
diff --git a/core/stencil.config.ts b/core/stencil.config.ts
index 52f7146406..e07f12693e 100644
--- a/core/stencil.config.ts
+++ b/core/stencil.config.ts
@@ -217,5 +217,4 @@ export const config: Config = {
   preamble: '(C) Ionic http://ionicframework.com - MIT License',
   globalScript: 'src/global/ionic-global.ts',
   enableCache: true,
-  transformAliasedImportPaths: true,
 };
```
5. `npm run build` and `npx stencil test --spec -- --no-cache` should pass

## Other information

The first commit in this PR are the changes associated with https://github.com/ionic-team/stencil/pull/4501. These are needed for this PR to pass. However,
- https://github.com/ionic-team/stencil/pull/4501 targets `main` and hasn't landed yet
- This PR targets v4 
Hence it's inclusion here. Once https://github.com/ionic-team/stencil/pull/4501 lands, I'll rebase `stencil/v4-dev` and this branch so that that commit goes away on the PR. Please review the subsequent commits. 

Docs PR: https://github.com/ionic-team/stencil-site/pull/1128

STENCIL-829

### DIFF
--- a/BREAKING_CHANGES.md
+++ b/BREAKING_CHANGES.md
@@ -12,6 +12,9 @@ This is a comprehensive list of the breaking changes introduced in the major ver
 ## Stencil v4.0.0
 
 - [General](#general)
+  - [New Configuration Defaults](#new-configuration-defaults)
+    - [transformAliasedImportPaths](#transformaliasedimportpaths)
+    - [transformAliasedImportPathsInCollection](#transformaliasedimportpathsincollection)
   - [In Browser Compilation Support Removed](#in-browser-compilation-support-removed)
   - [Legacy Context and Connect APIs Removed](#legacy-context-and-connect-APIs-removed)
   - [Legacy Browser Support Removed](#legacy-browser-support-removed)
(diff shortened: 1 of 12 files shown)


BREAKING_CHANGES.md
```diff
@@ -4,10 +4,91 @@ This is a comprehensive list of the breaking changes introduced in the major ver
...
 ## Versions
 
+- [Stencil 4.x](#stencil-v400)
 - [Stencil 3.x](#stencil-v300)
 - [Stencil 2.x](#stencil-two)
 - [Stencil 1.x](#stencil-one)
 
+## Stencil v4.0.0
+
+### General
```
rwaskiewicz, line 14: Note there's no table of contents. In fact, I've broken the one for v3 (there are two 'general' anchors). I need to work through how this ought to look, but don't wanna/think that should be a blocker

BREAKING_CHANGES.md
```diff
@@ -4,10 +4,91 @@ This is a comprehensive list of the breaking changes introduced in the major ver
...
+```ts title="stencil.config.ts - disabling 'transformAliasedImportPaths' in Stencil v4.0.0"
+import { Config } from '@stencil/core';
+
+export const config: Config = {
+  transformAliasedImportPaths: false,
+  // ...
+};
+```
+
+For more information on this flag, please see the [configuration documentation](https://stenciljs.com/docs/config#transformaliasedimportpaths)
```
rwaskiewicz, line 62: Question to the team - for betas/rcs, this will 404. Should it point to the 'next' docs explicitly, then we remove 'next' from these URLs as a part of the release (I'm thinking yes, but wanna make sure there isn't something I'm not considering before adding more mechanical changes)

src/compiler/config/outputs/validate-dist.ts
```diff
@@ -137,7 +137,9 @@ const validateOutputTargetDist = (config: d.ValidatedConfig, o: d.OutputTargetDi
     copy: validateCopy(o.copy ?? [], []),
     polyfills: isBoolean(o.polyfills) ? o.polyfills : undefined,
     empty: isBoolean(o.empty) ? o.empty : true,
-    transformAliasedImportPathsInCollection: o.transformAliasedImportPathsInCollection ?? false,
+    transformAliasedImportPathsInCollection: isBoolean(o.transformAliasedImportPathsInCollection)
+      ? o.transformAliasedImportPathsInCollection
+      : false,
```
tanner-reits, line 142: Should default `true` now, no?
```suggestion
    transformAliasedImportPathsInCollection: isBoolean(o.transformAliasedImportPathsInCollection)
      ? o.transformAliasedImportPathsInCollection
      : true,
```

tanner-reits (changes requested):
Just one flag flip and g2g

src/compiler/config/outputs/validate-dist.ts
rwaskiewicz: Yuppppp. Unsure how that slipped by, good catch!

tanner-reits (approved):

alicewriteswrongs (approved):

PR #4419, opened 2023-05-24
chore(compiler/runtime): remove CSS var shim & patchEsm()

tanner-reits:
Removes the CSS variable shim implementation that was marked as deprecated in v3

We currently polyfill the CSS variable behavior

GitHub Issue Number: N/A

## What is the new behavior?

No more polyfill

- [X] Yes
- [ ] No

Stencil apps will no longer polyfill CSS variable behavior. Will take effect in v4

## Testing

Compiling Stencil libraries appears to be working as expected and no noticeable impacts to the runtime


rwaskiewicz (changes requested):
❤️ Deleting Code 

All in all looks good, I found a few references that are all string-refs to things we're removing.

In addition to my inline comments, I also found some references to the css-shim in [internal-platform-client.ts bundling](https://github.com/ionic-team/stencil/blob/main/scripts/bundles/internal-platform-client.ts) and the [browser-compile stuff](https://github.com/ionic-team/stencil/blob/42314dc202e9832086207759565c6660d3746eff/test/browser-compile/src/utils/load-deps.ts#L16) that I _think_ can be removed as well

rwaskiewicz:
Ah - sorry, I forgot this the first time around - can you update `BREAKING_CHANGES.md` as well to record these changes? I think we can just pull the contents of https://github.com/ionic-team/stencil-site/pull/1129/files in there

alicewriteswrongs (approved):
code deletion!! 🎉🥳🗑️

PR #4434, opened 2023-05-30
refactor(snc): fix SNC errors in app-error.ts

alicewriteswrongs:

GitHub Issue Number: N/A

## What is the new behavior?

- 

- [ ] Yes
- [ ] No

### DIFF
--- a/src/dev-server/client/app-error.ts
+++ b/src/dev-server/client/app-error.ts
@@ -9,10 +9,15 @@ interface AppErrorData {
 
 type OpenInEditorCallback = (data: { file: string; line: number; column: number }) => void;
 
-export const appError = (data: AppErrorData) => {
-  const results = {
-    diagnostics: [] as any[],
-    status: null as string,
+interface AppErrorResults {
+  diagnostics: Diagnostic[];
+  status: null | string;
+}
+
+export const appError = (data: AppErrorData): AppErrorResults => {
+  const results: AppErrorResults = {
+    diagnostics: [] as Diagnostic[],
+    status: null,
   };
 
   if (data && data.window && Array.isArray(data.buildResults.diagnostics)) {
@@ -37,7 +42,7 @@ export const appError = (data: AppErrorData) => {
 
 const appendDiagnostic = (
   doc: Document,
-  openInEditor: OpenInEditorCallback,
+  openInEditor: OpenInEditorCallback | undefined,
   modal: HTMLElement,
   diagnostic: Diagnostic
 ) => {
@@ -46,7 +51,7 @@ const appendDiagnostic = (
 
   const masthead = doc.createElement('div');
   masthead.className = 'dev-server-diagnostic-masthead';
-  masthead.title = `${escapeHtml(diagnostic.type)} error: ${escapeHtml(diagnostic.code)}`;
+  masthead.title = `${escapeHtml(diagnostic.type)} error: ${escapeHtml(diagnostic.code ?? 'unknown error')}`;
   card.appendChild(masthead);
 
   const title = doc.createElement('div');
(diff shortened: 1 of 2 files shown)


src/dev-server/client/app-error.ts
```diff
@@ -139,18 +144,19 @@ const appendDiagnostic = (
       tdNum.className = 'dev-server-diagnostic-blob-num';
       if (l.lineNumber > 0) {
         tdNum.setAttribute('data-line-number', l.lineNumber + '');
-        tdNum.title = escapeHtml(diagnostic.relFilePath) + ', line ' + l.lineNumber;
+        tdNum.title = escapeHtml(diagnostic.relFilePath ?? '') + ', line ' + l.lineNumber;
 
-        if (canOpenInEditor) {
+        const maybeFile = diagnostic.absFilePath;
+        if (canOpenInEditor && maybeFile) {
```
tanner-reits, line 150: Do we need to update this check since `canOpenInEditor` already checks if the type is a string? Or maybe we should have: `canOpenInEditor && !!maybeFile` to ensure the file path is not an empty string?

src/dev-server/client/app-error.ts
```diff
@@ -199,12 +205,12 @@ const getDevServerModal = (doc: Document) => {
...
   outer.innerHTML = `<style>${appErrorCss}</style><div id="${DEV_SERVER_MODAL}-inner"></div>`;
 
-  return doc.getElementById(`${DEV_SERVER_MODAL}-inner`);
+  return doc.getElementById(`${DEV_SERVER_MODAL}-inner`) as HTMLElement;
 };
 
 export const clearAppErrorModal = (data: { window: Window }) => {
   const appErrorElm = data.window.document.getElementById(DEV_SERVER_MODAL);
-  if (appErrorElm) {
+  if (appErrorElm && appErrorElm.parentNode) {
```
tanner-reits, line 213: Not necessary, but could be coalesced
```suggestion
  if (appErrorElm?.parentNode) {
```

tanner-reits (changes requested):
One question, but otherwise think this looks good. Wish so many of the fields on `Diagnostic` weren't optional

src/dev-server/client/app-error.ts
alicewriteswrongs: I think we do since `canOpenInEditor` is just type `bool` and not `diagnostic.absFilePath is string` or `. I think typescript will narrow the type of a primitive when you do a check like that, like if you have

```ts
const foo: string | undefined = something();

if (typeof foo === "string") {
  // type of 'foo' is narrowed from `string | undefined` to `string`
}
```

but with objects like the `diagnostic` object they still keep their declared type which so, in this case, the `absFilePath` property still has type `string | undefined` even after you do a check that `.absFilePath` is of type string. Thus the type of the `maybeFile` variable here is also `string | undefined`.

we could add the `!!` but won't just putting the string in the conditional already coerce it to a boolean?

src/dev-server/client/app-error.ts
alicewriteswrongs: yea more concise!

src/dev-server/client/app-error.ts
tanner-reits: Oh yeah you're right cuz it's a conditional. My bad!

src/dev-server/client/app-error.ts
```diff
@@ -95,8 +100,8 @@ const appendDiagnostic = (
           openInEditor,
           fileHeader,
           diagnostic.absFilePath,
-          diagnostic.lineNumber,
-          diagnostic.columnNumber
+          diagnostic.lineNumber ?? 0,
+          diagnostic.columnNumber ?? 0
```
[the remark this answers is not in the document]
alicewriteswrongs: sure we can do that instead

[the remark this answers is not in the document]
alicewriteswrongs: unfortunately the `typeof line !== "number"` check and reassignment doesn't actually cause the type for the `line` parameter to be narrowed from `number | undefined` to `number`, so still throughout the whole body of the `addOpenInEditor` function it is `number | undefined`. Instead we can assign a new variable to do the narrowing and then pass that to `openInEditor`, I just pushed a change to do that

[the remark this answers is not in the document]
rwaskiewicz: Gahh, gotcha. Thanks!

rwaskiewicz (approved):
LGTM - only non-blocker is there are 2 lint failures causing CI to fail

rwaskiewicz:
Sorry @alicewriteswrongs, I had to remove this from the queue while 3.3.1 builds locally 😛

alicewriteswrongs:
oops!!

PR #4436, opened 2023-05-30
chore(ci): update workflows for merge queue

rwaskiewicz:

GitHub Issue Number: N/A

## What is the new behavior?

this commit updates all workflows whose status is reported to gate on `main` to include the `merge_group` event trigger. the github documentation[1] states:

> If your repository uses GitHub Actions to perform required checks on
> pull requests in your repository, you need to update the workflows
> to include the merge_group event as an additional trigger. Otherwise,
> status checks will not be triggered when you add a pull request to a
> merge queue. The merge will fail as the required status check will
> not be reported.

it's not abundantly clear if this is needed for reusable workflows explicitly. `merge_group` is added in this commit to avoid the need to configure this feature via multiple prs

[1]: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue#triggering-merge-group-checks-with-github-actions

- [ ] Yes
- [x] No

## Testing

No way to test 😢

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,7 @@
 name: Build Stencil
 
 on:
+  merge_group:
   workflow_call:
   # Make this a reusable workflow, no value needed
   # https://docs.github.com/en/actions/using-workflows/reusing-workflows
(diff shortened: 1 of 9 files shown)


tanner-reits (approved):
Makes sense. We'll merge it and find out

alicewriteswrongs (approved):

PR #4444, opened 2023-05-31
chore(ci): merge queue fixes

rwaskiewicz:

Our merge queues are running certain tasks 2x

GitHub Issue Number: N/A

## What is the new behavior?

this commit removes the `merge_group` event trigger from our reusable workflows. it turns out this is _not_ needed, as the `merge_group` in our `main.yml` properly propagates to all reusable workflows.

when these were applied, each workflow would be run 2x - once from `main.yml` and once independently. the latter would always fail, as they didn't have any notion of order to run in (so they'd often fail if there was no pre-built stencil artifact).

- [ ] Yes
- [x] No

## Testing

N/A, sadly

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,7 +1,6 @@
 name: Build Stencil
 
 on:
-  merge_group:
   workflow_call:
   # Make this a reusable workflow, no value needed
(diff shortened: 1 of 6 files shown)


alicewriteswrongs (approved):

tanner-reits (approved):
Makes sense!

Issue #4424, opened 2023-05-25
bug: Rollup: Parse Error: Duplicate export


danyball:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

3.3.0

### Current Behavior

Throws error on `npm run build`:

```
[ ERROR ]  Rollup: Parse Error: ./src/components/my-component/my-component.tsx:80:9
           Duplicate export 'MyComponent' (Note that you need plugins to import files that are not JavaScript) Error
           parsing: /Users/XXXX/stencilboilerplate/test/src/components/my-component/my-component.tsx,
           line: 80, column: 9

```

### Expected Behavior

No error.

### System Info

```shell
MacOS, NPM 8.15.0
```

### Steps to Reproduce

1. create project: `npm init stencil`
2. add `private static myVar = "input"` to `my-component.tsx`
3. also change this line: `return format(this.first, this.middle, MyComponent.myVar);`
4. run `npm run build`

### Code Reproduction URL

notyet

### Additional Information

Creating a static variable is common to add a native event to the documentation with `@event` in JSDoc comment. For example:
```
  /**
   * Emitted when a user clicks on the button.
   * @event
   */
  private static eventInput = "input"
```

rwaskiewicz:
Thanks @danyball! I've confirmed that this is related to the TypeScript 5 upgrade.

deleonio:
Yes, we have the same issue, since update stencil from 3.2 to 3.3.

https://github.com/public-ui/kolibri/actions/runs/5081216007/jobs/9129248718

rwaskiewicz:
Hey folks,

I have a dev build that I believe will fix the issue. Can anyone running into this problem please install it and provide the team and I with some feedback?
```
npm i @stencil/core@3.3.0-dev.1685651872.3091be9
```

Thanks!

danyball:
Is working, thanks @rwaskiewicz !

rwaskiewicz:
[...]

PR #4447, opened 2023-06-01
fix(compiler): handle ts 5.0 static members

rwaskiewicz:

A component like the following will fail to compile:
```tsx

@Component({
  tag: 'my-component',
})
export class MyComponent {
  private static myVar = "input";

  render() {
    console.log(MyComponent.myVar);
    return (<div></div>);
  }
}
```
With the following error:
```console
[ ERROR ]  Rollup: Parse Error: <FILE_PATH>
		   Duplicate export 'MyComponent' (Note that you need plugins to
		   import files that are not JavaScript) Error parsing: <FILE_PATH>
```

GitHub Issue Number: #4424

## What is the new behavior?

this commit fixes a bug where static class members that are initialized
on a stencil component result in 2 `export` statements of a component
class being generated. as a result of 2 `ExportDeclaration`s being
created, the compilation process will fail.

this fix involves detecting initialized static members of a class before
they go through the transpilation phase, because a class can be
transformed in such a way that we lose the ability to determine if
something like the following was originally authored:
```ts
class MyComponent {
  static myVar = 1;
}
```

as a result, whether or not this static initializer exists or not is
stuffed onto the component's metadata. this allows us to store if an
additional transformation is needed or not after transpilation has
occurred. specifically, should we add an `export` keyword to a modified
class declaration in the form:
```ts
// determine if this `VariableDeclaration` needs an `export`
const MyComponent = class {}
```

- [ ] Yes
- [x] No

## Testing

A special build of Stencil, `@stencil/core@3.3.0-dev.1685651872.3091be9` was created using the dev release process [LOGS](https://github.com/ionic-team/stencil/actions/runs/5148994502)

### Starter Project 
- Created a local Stencil project and smoke tested locally: `npm init stencil component tmp-component-starter && cd $_ && npm i @stencil/core@3.3.0`
- Apply the following patch (`git apply starter.txt`, [starter.txt](https://github.com/ionic-team/stencil/files/11629611/starter.txt)):
```diff
diff --git a/src/components/my-component/my-component.tsx b/src/components/my-component/my-component.tsx
index 56d51d9..8deec6e 100644
--- a/src/components/my-component/my-component.tsx
+++ b/src/components/my-component/my-component.tsx
@@ -7,6 +7,7 @@ import { format } from '../../utils/utils';
   shadow: true,
 })
 export class MyComponent {
+  static MyVar = 1;
   /**
    * The first name
    */
@@ -27,6 +28,6 @@ export class MyComponent {
   }
 
   render() {
-    return <div>Hello, World! I'm {this.getText()}</div>;
+    return <div>Hello, World! I'm {this.getText()} {MyComponent.MyVar}</div>;
   }
 }
```
- `npm run build`
- Build fails since the fix is not in Stencil v3.3.0:
```console
 ERROR ]  Rollup: Parse Error: ./src/components/my-component/my-component.tsx:82:9
           Duplicate export 'MyComponent' (Note that you need plugins to import files that are not JavaScript) Error
           parsing: /private/tmp/tmp-component-starter/src/components/my-component/my-component.tsx, line: 82, column:
           9

[45:10.0]  build failed in 1.25 s
```
- `npm i @stencil/core@3.3.0-dev.1685651872.3091be9 && npm run build && npm t`
- Build Passes 🎉 Tests Fail (for a good reason) 🎉 

### Ionic Framework
- I [created a branch](https://github.com/ionic-team/ionic-framework/compare/main...rwaskiewicz/nightly-and-more) in the Ionic Framework repo that points to my special branch/release, and [was able to kick off the "nightly" build step manually to run through Framework's CI step](https://github.com/ionic-team/ionic-framework/actions/runs/5149116085)
  - The first run passes, since Framework has no static members on any of its component classes
- Apply the [following patch](https://github.com/ionic-team/ionic-framework/commit/9ef11b82a65756e3875687b6e6e69e15b4b6101c) (`git apply framework.txt`, [framework.txt](https://github.com/ionic-team/stencil/files/11629900/framework.txt)):

```diff
diff --git a/core/src/components/accordion/accordion.tsx b/core/src/components/accordion/accordion.tsx
index fd65ffcd3e..8d87f9de60 100644
--- a/core/src/components/accordion/accordion.tsx
+++ b/core/src/components/accordion/accordion.tsx
@@ -42,6 +42,7 @@ export class Accordion implements ComponentInterface {
   private contentEl: HTMLDivElement | undefined;
   private contentElWrapper: HTMLDivElement | undefined;
   private headerEl: HTMLDivElement | undefined;
+  private static MyVar = 123;
 
   private currentRaf: number | undefined;
 
@@ -146,6 +147,7 @@ export class Accordion implements ComponentInterface {
   };
 
   private getSlottedHeaderIonItem = () => {
+    console.log(Accordion.MyVar);
     const { headerEl } = this;
     if (!headerEl) {
       return;
```
- Initiate another run, [see CI pass again](https://github.com/ionic-team/ionic-framework/actions/runs/5149345350)
- [Revert the change](https://github.com/ionic-team/ionic-framework/commit/0774104fa54d026a20f9d1b9e302e724af4e37e3) where we use the special dev build for the nightly job
- Kick off one final CI run, [which fails](https://github.com/ionic-team/ionic-framework/actions/runs/5149459834)

## Other information

Consider a TypeScript class that:
- has a static member that is defined on the class
- is exported from the file in which it is defined
```ts
export class MyComponent {
  static myVar = 1;
}
```

[In TypeScript v4.9 and lower](https://www.typescriptlang.org/play?ts=4.9.5#code/KYDwDg9gTgLgBAYwDYEMDOa4FkCeBhCAW0gDtgT4BvAKAEg0YUYBLBOQnANRSjgF44ARgDc1AL5A) the class is transpiled as such:
```js
export class MyComponent {
}
MyComponent.myVar = 1;
```

[Starting with TypeScript v5.0](https://www.typescriptlang.org/play?ts=5.0.4#code/KYDwDg9gTgLgBAYwDYEMDOa4FkCeBhCAW0gDtgT4BvAKAEg0YUYBLBOQnANRSjgF44ARgDc1AL5A), the class is transpiled like so:
```js
class MyComponent {
}
MyComponent.myVar = 1;
export { MyComponent };
```
Where the `export` occurs separate from the class's declaration:
```diff
- export class MyComponent {
+ class MyComponent {
}
MyComponent.myVar = 1;
+ export { MyComponent };
```

Note this only occurs when both conditions above are met. The following
code snippet compiles the same using TypeScript [v4.9](https://www.typescriptlang.org/play?ts=4.9.5#code/KYDwDg9gTgLgBAYwDYEMDOa4FkCeBhCAW0gDtgT4BvAKAF8g) and [v5.0](https://www.typescriptlang.org/play?ts=5.0.4&ssl=2&ssc=2&pln=1&pc=1#code/KYDwDg9gTgLgBAYwDYEMDOa4FkCeBhCAW0gDtgT4BvAKAF8g):
```ts
export class MyComponent {
}
// transpiles to
export class MyComponent {
}
```

The difference in compilation when both conditions are met leads to a bug in
Stencil v3.3.0, the first version of the library which supports TypeScript v5.0.

Stencil made the assumption that the `export` keyword continues to be inlined
with the class declaration, rather than a new `ExportDeclaration` instance
being created. In Stencil v3.3.0, the `export` keyword is added to the generated
code:
```ts
export class MyComponent {};
MyComponent.myVar = 1;
export { MyComponent }; // 2 exports causes an error!
```
Two exports of the class (`MyComponent`) causes errors following the TypeScript
compilation step in Stencil:
```console
[ ERROR ]  Rollup: Parse Error: <FILE_PATH>
		   Duplicate export 'MyComponent' (Note that you need plugins to
		   import files that are not JavaScript) Error parsing: <FILE_PATH>
```

In an ideal world, we delegate inserting `export` to the TypeScript compiler.
However, this is not possible as the `ExportDeclaration` is _only_ created for
classes that are exported and have static members. So while removing Stencil's
`export` keyword injection will work for:
```ts
export class MyComponent {
  static myVar = 1;
}
// becomes the following on TS 5.0 if we remove Stencil's `export` injection
class MyComponent {
}
MyComponent.myVar = 1;
export { MyComponent };
```
it will not work for:
```ts
export class MyComponent {
  myVar = 1;
}
// becomes the following on TS 5.0 if we remove Stencil's `export` injection
class MyComponent {
  constructor() {
    this.myVar = 1;
  }
}
```
as `export` is removed from `MyComponent` in the compilation process. This behavior is
controlled by the [`TransformOptions#componentExport` field](https://github.com/ionic-team/stencil/blob/bd1023a67c5224cbae3e445170cb13227a5cf8fb/src/declarations/stencil-public-compiler.ts#L2683).

Import injection should only occur when neither of two aforementioned criteria are met:
- has a static member that is defined on the class
- is exported from the file in which it is defined

### DIFF
--- a/src/compiler/transformers/decorators-to-static/convert-decorators.ts
+++ b/src/compiler/transformers/decorators-to-static/convert-decorators.ts
@@ -2,8 +2,14 @@ import { augmentDiagnosticWithNode, buildError } from '@utils';
 import ts from 'typescript';
 
 import type * as d from '../../../declarations';
-import { retrieveTsDecorators, retrieveTsModifiers } from '../transform-utils';
+import {
+  convertValueToLiteral,
+  createStaticGetter,
+  retrieveTsDecorators,
+  retrieveTsModifiers,
+} from '../transform-utils';
 import { componentDecoratorToStatic } from './component-decorator';
+import { hasStaticInitializerInClass } from './convert-static-members';
 import { isDecoratorNamed } from './decorator-utils';
 import {
   CLASS_DECORATORS_TO_REMOVE,
@@ -70,6 +76,14 @@ const visitClassDeclaration = (
     elementDecoratorsToStatic(diagnostics, decoratedMembers, typeChecker, filteredMethodsAndFields);
     watchDecoratorsToStatic(config, diagnostics, decoratedMembers, watchable, filteredMethodsAndFields);
     listenDecoratorsToStatic(diagnostics, decoratedMembers, filteredMethodsAndFields);
+  }
+
+  // Handle static members that are initialized
+  const hasStaticMembersWithInit = classMembers.some(hasStaticInitializerInClass);
+  if (hasStaticMembersWithInit) {
+    filteredMethodsAndFields.push(
+      createStaticGetter('stencilHasStaticMembersWithInit', convertValueToLiteral(hasStaticMembersWithInit))
+    );
   }
 
   // We call the `handleClassFields` method which handles transforming any
--- a/src/compiler/transformers/decorators-to-static/convert-static-members.ts
+++ b/src/compiler/transformers/decorators-to-static/convert-static-members.ts
@@ -1,0 +1,18 @@
+import ts from 'typescript';
+
+/**
+ * Helper function to detect if a class element fits the following criteria:
+ * - It is a property declaration (e.g. `foo`)
+ * - It has an initializer (e.g. `foo *=1*`)
+ * - The property declaration has the `static` modifier on it (e.g. `*static* foo =1`)
+ * @param classElm the class member to test
+ * @returns true if the class member fits the above criteria, false otherwise
+ */
+export const hasStaticInitializerInClass = (classElm: ts.ClassElement): boolean => {
+  return (
+    ts.isPropertyDeclaration(classElm) &&
+    classElm.initializer !== undefined &&
+    Array.isArray(classElm.modifiers) &&
+    classElm.modifiers!.some((modifier) => modifier.kind === ts.SyntaxKind.StaticKeyword)
+  );
+};
--- a/src/compiler/transformers/test/proxy-custom-element-function.spec.ts
+++ b/src/compiler/transformers/test/proxy-custom-element-function.spec.ts
@@ -83,6 +83,19 @@ describe('proxy-custom-element-function', () => {
 
       expect(formatCode(transpiledModule.outputText)).toContain(
         formatCode(
+          `const ${componentClassName} = /*@__PURE__*/ __stencil_proxyCustomElement(class ${componentClassName} extends HTMLElement {}, true);`
+        )
+      );
+    });
+
+    it('wraps an exported class initializer in a proxyCustomElement call', () => {
+      const code = `export const ${componentClassName} = class extends HTMLElement {};`;
+
+      const transformer = proxyCustomElement(compilerCtx, transformOpts);
+      const transpiledModule = transpileModule(code, null, compilerCtx, [], [transformer]);
+
+      expect(formatCode(transpiledModule.outputText)).toContain(
+        formatCode(
           `export const ${componentClassName} = /*@__PURE__*/ __stencil_proxyCustomElement(class ${componentClassName} extends HTMLElement {}, true);`
         )
       );
@@ -97,7 +110,7 @@ describe('proxy-custom-element-function', () => {
 
         expect(formatCode(transpiledModule.outputText)).toContain(
           formatCode(
-            `export const foo = 'hello world!', ${componentClassName} = /*@__PURE__*/ __stencil_proxyCustomElement(class ${componentClassName} extends HTMLElement {}, true);`
+            `const foo = 'hello world!', ${componentClassName} = /*@__PURE__*/ __stencil_proxyCustomElement(class ${componentClassName} extends HTMLElement {}, true);`
           )
         );
       });
@@ -110,7 +123,7 @@ describe('proxy-custom-element-function', () => {
 
         expect(formatCode(transpiledModule.outputText)).toContain(
           formatCode(
-            `export const ${componentClassName} = /*@__PURE__*/ __stencil_proxyCustomElement(class ${componentClassName} extends HTMLElement {}, true), foo = 'hello world!';`
+            `const ${componentClassName} = /*@__PURE__*/ __stencil_proxyCustomElement(class ${componentClassName} extends HTMLElement {}, true), foo = 'hello world!';`
           )
         );
       });
@@ -123,7 +136,7 @@ describe('proxy-custom-element-function', () => {
 
         expect(formatCode(transpiledModule.outputText)).toContain(
           formatCode(
-            `export const foo = 'hello world!', ${componentClassName} = /*@__PURE__*/ __stencil_proxyCustomElement(class ${componentClassName} extends HTMLElement {}, true), bar = 'goodbye?';`
+            `const foo = 'hello world!', ${componentClassName} = /*@__PURE__*/ __stencil_proxyCustomElement(class ${componentClassName} extends HTMLElement {}, true), bar = 'goodbye?';`
           )
         );
       });
--- a/src/compiler/transformers/test/transpile.ts
+++ b/src/compiler/transformers/test/transpile.ts
@@ -175,11 +175,19 @@ export function transpileModule(
  */
 const prettifyTSOutput = (tsOutput: string): string => tsOutput.replace(/\s+/gm, ' ');
 
-export function getStaticGetter(output: string, prop: string) {
-  const toEvaluate = `return ${output.replace('export', '')}`;
+/**
+ * Helper function for tests that converts stringified JavaScript to a runtime value.
+ * A value from the generated JavaScript is returned based on the provided property name.
+ * @param stringifiedJs the stringified JavaScript
+ * @param propertyName the property name to pull off the generated JavaScript
+ * @returns the value associated with the provided property name. Returns undefined if an error occurs while converting
+ * the stringified JS to JavaScript, or if the property does not exist on the generated JavaScript.
+ */
+export function getStaticGetter(stringifiedJs: string, propertyName: string): string | void {
+  const toEvaluate = `return ${stringifiedJs.replace('export', '')}`;
   try {
     const Obj = new Function(toEvaluate);
-    return Obj()[prop];
+    return Obj()[propertyName];
   } catch (e) {
     console.error(e);
     console.error(toEvaluate);
--- a/src/compiler/transformers/update-component-class.ts
+++ b/src/compiler/transformers/update-component-class.ts
@@ -3,11 +3,26 @@ import ts from 'typescript';
 import type * as d from '../../declarations';
 import { retrieveTsDecorators, retrieveTsModifiers } from './transform-utils';
 
+/**
+ * Transformation helper for updating how a Stencil component class is declared.
+ *
+ * Based on the output module type (CommonJS or ESM), the behavior is slightly different:
+ * - For CommonJS, the component class is left as is
+ * - For ESM, the component class is re-written as a variable statement
+ *
+ * @param transformOpts the options provided to TypeScript + Rollup for transforming the AST node
+ * @param classNode the node in the AST pertaining to the Stencil component class to transform
+ * @param heritageClauses a collection of heritage clauses associated with the provided class node
+ * @param members a collection of members attached to the provided class node
+ * @param moduleFile the Stencil intermediate representation associated with the provided class node
+ * @returns the updated component class declaration
+ */
 export const updateComponentClass = (
   transformOpts: d.TransformOptions,
   classNode: ts.ClassDeclaration,
   heritageClauses: ts.HeritageClause[] | ts.NodeArray<ts.HeritageClause>,
-  members: ts.ClassElement[]
+  members: ts.ClassElement[],
+  moduleFile: d.Module
 ): ts.ClassDeclaration | ts.VariableStatement => {
   let classModifiers = retrieveTsModifiers(classNode)?.slice() ?? [];
 
@@ -15,7 +30,8 @@ export const updateComponentClass = (
     // CommonJS, leave component class as is
 
     if (transformOpts.componentExport === 'customelement') {
-      // remove export from class
+      // remove export from class - it may already be removed by the TypeScript compiler in certain circumstances if
+      // this transformation is run after transpilation occurs
       classModifiers = classModifiers.filter((m) => {
         return m.kind !== ts.SyntaxKind.ExportKeyword;
       });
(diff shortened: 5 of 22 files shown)


src/compiler/transformers/decorators-to-static/convert-static-members.ts
```diff
@@ -0,0 +1,20 @@
...
+ * - It has an initializer (e.g. `foo *=1*`)
+ * - The property declaration has the `static` modifier on it (e.g. `*static* foo =1`)
+ * @param classElm the class member to test
+ * @returns true if the class member fits the above criteria, false otherwise
+ */
+export const hasStaticInitializerInClass = (classElm: ts.ClassElement): boolean => {
+  return (
+    classElm.kind === ts.SyntaxKind.PropertyDeclaration &&
+    // we use type assertions here as a `filter()` call for narrowing the type is an order of magnitude more
+    // expensive from a time perspective
```
alicewriteswrongs, line 15: does this refer to using one of the `ts.isSomeNodeType` method? (in this case obviously `ts.isPropertyDeclaration`)

that does give us type narrowing afterward if we do:

```suggestion
  return (
    ts.isPropertyDeclaration(classElm) &&
    // we use type assertions here as a `filter()` call for narrowing the type is an order of magnitude more
    // expensive from a time perspective
```

afaict `isPropertyDeclaration` basically just does this same check too, but with the addition that it's return type is `node is PropertyDeclaration`

src/compiler/transformers/decorators-to-static/convert-decorators.ts
```diff
@@ -72,6 +78,14 @@ const visitClassDeclaration = (
     listenDecoratorsToStatic(diagnostics, decoratedMembers, filteredMethodsAndFields);
   }
 
+  // Handle static members that are initialized
+  const hasStaticMembersWithInit = classMembers.some(hasStaticInitializerInClass);
+  if (hasStaticMembersWithInit) {
+    filteredMethodsAndFields.push(
+      createStaticGetter('stencilHasStaticMembersWithInit', convertValueToLiteral(hasStaticMembersWithInit))
```
alicewriteswrongs, line 85: would it be helpful to declare a constant for this string?

`createStaticGetter` doesn't seem to have any type safety or help to enforce a contract with a known list of static getters — the `propName` param just has type `string`, instead of `"stencilHasStaticMembersWithInit" | "is" | "watchers" | ...`

maybe something to file a ticket about and improve later

src/compiler/transformers/test/transpile.ts
```diff
@@ -175,11 +175,19 @@ export function transpileModule(
  */
 const prettifyTSOutput = (tsOutput: string): string => tsOutput.replace(/\s+/gm, ' ');
 
-export function getStaticGetter(output: string, prop: string) {
-  const toEvaluate = `return ${output.replace('export', '')}`;
+/**
+ * Helper function for tests that converts JavaScript represented as a string and converts the string to JavaScript.
```
alicewriteswrongs, line 179:
```suggestion
 * Helper function for tests that converts stringified JavaScript to a runtime value.
```

src/compiler/transformers/update-component-class.ts
```diff
@@ -31,25 +47,48 @@ export const updateComponentClass = (
   }
 
   // ESM with export
-  return createConstClass(transformOpts, classNode, heritageClauses, members);
+  return createConstClass(transformOpts, classNode, heritageClauses, members, moduleFile);
```
alicewriteswrongs, line 50: any idea why we do this in the esm case?

test/karma/test-app/static-members/static-members-separate-export.tsx
```diff
@@ -0,0 +1,22 @@
+import { Component, h } from '@stencil/core';
+
+@Component({
+  tag: 'static-members-separate-export',
+})
+class StaticMembersWithSeparateExport {
```
alicewriteswrongs, line 6: not to complicate things, but how does `export default` factor into this?

alicewriteswrongs: (like should we have a test for that too)

src/compiler/transformers/test/proxy-custom-element-function.spec.ts
```diff
@@ -97,7 +110,7 @@ describe('proxy-custom-element-function', () => {
 
         expect(formatCode(transpiledModule.outputText)).toContain(
           formatCode(
-            `export const foo = 'hello world!', ${componentClassName} = /*@__PURE__*/ __stencil_proxyCustomElement(class ${componentClassName} extends HTMLElement {}, true);`
+            `const foo = 'hello world!', ${componentClassName} = /*@__PURE__*/ __stencil_proxyCustomElement(class ${componentClassName} extends HTMLElement {}, true);`
```
alicewriteswrongs, line 113: just curious, and I'm probably just not fully following the logic in the diff here, but why did this test change? is there an `export ${componentClassName}` elsewhere in `transpiledModule.outputText`?

src/compiler/transformers/test/convert-static-members.ts
```diff
@@ -0,0 +1,120 @@
...
+            ts.factory.createIdentifier('propertyName'),
+            undefined,
+            undefined,
+            ts.factory.createStringLiteral('initial value')
+          ),
+        ]
+      );
+      expect(classWithStaticMembers.members.some(hasStaticInitializerInClass)).toBe(false);
+    });
+  });
```
alicewriteswrongs, line 119: is it worthwhile to add a test case here for the `private static` case?

src/compiler/transformers/update-component-class.ts
```diff
@@ -3,19 +3,35 @@ import ts from 'typescript';
...
+ * - For CommonJS, the component class is left as is
+ * - For ESM, the component class is re-written as a variable statement
+ *
+ * @param transformOpts the options provided to TypeScript + Rollup for transforming the AST node
+ * @param classNode the node in the AST pertaining to the Stencil component class to transform
+ * @param heritageClauses a collection of heritage clauses associated with the provided class node
+ * @param members a collection of members attached to the provided class node
+ * @param moduleFile the Stencil intermediate representation associated with the provided class node
+ * @returns the updated component class declaration
+ */
```
alicewriteswrongs, line 19: comprehension check with what's going on here:

We create the static getter in the `convertDecoratorsToStatic` transform that is passed to typescript and run through our initial pass through the typescript compiler.

We need to calculate and store that value (via the static getter) because we need that information later on in the compilation process, when we're marshalling components to pass them to Rollup in the output targets.

The reason we need to do the calculation in the `convertDecoratorsToStatic` is because if we defer until the separate transformers that we run as part of the output targets (where the call chain for this `updateComponentClass` function originates) then TypeScript has already made the transformation that we're trying to pre-empt, so we can no longer find a `static` property on our class and figure out if we need to do our own transformation.

Thus we need to hook into the compilation process early on:

https://github.com/ionic-team/stencil/blob/3091be91c009e3c98295221cb56980ab64fba61e/src/compiler/transpile/run-program.ts#L64-L67

then store then information so we can act on it later on when dealing with output target specific code.

Is that on the right track at least?

src/compiler/transformers/decorators-to-static/convert-decorators.ts
```diff
@@ -72,6 +78,14 @@ const visitClassDeclaration = (
     listenDecoratorsToStatic(diagnostics, decoratedMembers, filteredMethodsAndFields);
   }
 
+  // Handle static members that are initialized
+  const hasStaticMembersWithInit = classMembers.some(hasStaticInitializerInClass);
```
alicewriteswrongs, line 82: I have a question as to whether this falls within the ambit of this transformer - as-named it's concerned with transforming properties decorated with 'stencil decorators' to static getters on the class.

Is it possible that the handling of `static` members should live in a separate transformer?

tanner-reits (changes requested):
Just one small thing on top of Alice's thoughts

src/compiler/transformers/test/convert-static-members.ts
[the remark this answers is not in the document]
rwaskiewicz: Good catch! Done in 8d9ebea0c

src/compiler/transformers/decorators-to-static/convert-decorators.ts
rwaskiewicz: Possibly - but since we're going to partially revert this once we upgrade to TS 5.1 (and remove this metadata), I'd prefer keeping it for now. LMK if today's 5.1 revelation changes your opinion from an hour ago 😆

src/compiler/transformers/update-component-class.ts
rwaskiewicz: Yep! Exactly - LMK if I didn't answer this/fill in additional details during our sync this morning

src/compiler/transformers/test/convert-static-members.ts
rwaskiewicz: 5fc75ab0939966f40dc62ec6be1782be321c0904 👍

src/compiler/transformers/test/proxy-custom-element-function.spec.ts
rwaskiewicz: I don't think I explained this part well in my novella of a PR summary - this is due to the changes here - https://github.com/ionic-team/stencil/pull/4447/files#diff-5a3f4b576667cc1940602dba4a00df74180725754979fb8a88e1e51e3761d87eL87

Previously we were always assuming we should just tack on an `export` keyword, which I think was actually a really subtle bug that never managed to bite us until now. With the linked change, we just carry the modifiers over from the variable statement and let the compiler determine where those should be used elsewhere

test/karma/test-app/static-members/static-members-separate-export.tsx
rwaskiewicz: I actually had written one, and removed it (the test). Stencil won't allow default exports, even if a file consists of a single class component that is the default export.

For a project with the following component:
```tsx

@Component({
  tag: 'my-component',
})
export default class MyComponent {
  private static myVar = "input";

  render() {
    console.log(MyComponent.myVar);
    return (<div></div>);
  }
}
```
running `npx stencil build` yields:
```console
> stencil build --docs

[23:51.2]  @stencil/core
[23:51.5]  [LOCAL DEV] v3.3.0-dev.1685644211.49532eb 🔭
[23:52.3]  build, dupe-export, prod mode, started ...
[23:52.3]  transpile started ...
[23:53.0]  transpile finished in 696 ms

[ ERROR ]  ./src/components/my-component/my-component.tsx:3:1
           To allow efficient bundling, modules using @Component() can only have a single export which is the component
           class itself. Any other exports should be moved to a separate file. For further information check out:
           https://stenciljs.com/docs/module-bundling

      L3:  @Component({                                                                                                                                                                                 
      L4:    tag: 'my-component',

[23:53.0]  build failed in 709 ms
```

I had tried a few other legal (TypeScript/JavaScript-legal) combinations, but they all yielded the same error:
```tsx
// this one has a separate export default at the end of the file

@Component({
  tag: 'my-component',
})
export class MyComponent {
  private static myVar = "input";

  render() {
    console.log(MyComponent.myVar);
    return (<div></div>);
  }
}

export default MyComponent;
```
```tsx
// don't use export on the class declaration, but export at the very bottom

@Component({
  tag: 'my-component',
})
class MyComponent {
  private static myVar = "input";

  render() {
    console.log(MyComponent.myVar);
    return (<div></div>);
  }
}

export default MyComponent;
```

src/compiler/transformers/update-component-class.ts
rwaskiewicz: AFAIK, for "optimizations" in https://github.com/ionic-team/stencil/commit/438c6b310f24d93dc82aba1614e0f9d91dbccd7a. The issue that commit links to in the TS repo implies that classes were getting wrapped in IIFE's if they had static members - a few things aren't immediately clear to me:
1. Whether the optimization is purely from a TS perspective once that commit is applied again, or has something to do with how Rollup interacts with variable declarations 
2. What was broken in the CJS export to begin with

I believe it was attempted to do this for both CJS and ESM, but it was partially reverted in https://github.com/ionic-team/stencil/commit/4eb5b421cfdff5a73b96837656997a06180de806

src/compiler/transformers/test/transpile.ts
rwaskiewicz: 👍 Done in 326964c2f

src/compiler/transformers/decorators-to-static/convert-decorators.ts
rwaskiewicz: Yeah, I'm not sure why it isn't like that currently. It definitely doesn't help the searchability of this constant (or any of it's siblings, `is`, `properties`, etc.). I stubbed out STENCIL-856 in JIRA and added a TODO in 3c87ded35

src/compiler/transformers/decorators-to-static/convert-static-members.ts
rwaskiewicz: Oooh, yes! That helps us get rid of all those type assertions! Done in c3ab21a4b

rwaskiewicz:
For posterity, we've determined that we'll roll this back partially when we upgrade to TS 5.1 🤣

tanner-reits (approved):
LGTM since we'll be reverting part of this in the near-ish future

src/compiler/transformers/decorators-to-static/convert-decorators.ts
alicewriteswrongs: haha yeah, that makes sense I think

src/compiler/transformers/update-component-class.ts
alicewriteswrongs: Nope, all makes sense!

src/compiler/transformers/test/proxy-custom-element-function.spec.ts
alicewriteswrongs: That seems like what should have been the behavior all along, so I think that makes sense 👍

test/karma/test-app/static-members/static-members-separate-export.tsx
alicewriteswrongs: Ah I wasn't aware that Stencil had that limitation! Obvs fine to leave out in that case 👍

src/compiler/transformers/update-component-class.ts
alicewriteswrongs: ah, optimizations...

ok sounds like we probably don't want to mess with this now, I was just wondering if you had more context on it 👍

alicewriteswrongs (approved):
I think this looks good to me, I'm ready to :shipit:

PR #4456, opened 2023-06-05
chore(ci): remove auto-request-reviews action

rwaskiewicz:

GitHub Issue Number: N/A

## What is the new behavior?

this commit is a followup to 25d691d (#4448), where in we removed/commented out the events tied to this action. the hope in that commit was to temporarily disable auto request reviews while we worked through round robin testing.

however, this action still attempts to run on merge queue actions (more specifically, GH parses the action and see that it doesn't have an event `on` section and sends an error email). to put a stop to that, remove the action.

should we decide to re-enable this action, this and 25d691d should be reverted in the same commit/pr

- [ ] Yes
- [x] No

## Testing

N/A - this shouldn't run/no errors should be raised (either in GH or in my inbox)


rwaskiewicz:
Closing to see what round robin does

tanner-reits (approved):
bye bye 👋

Issue #4459, opened 2023-06-06
bug: lifecycle hooks are not awaited when zone.js has patched Promise constructor


marcincichocki:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

3.2.2

### Current Behavior

Async lifecycle methods (specifically https://tc39.es/ecma262/#prod-AsyncMethod not only methods which return instance of the `Promise`) are not "awaited" due to latest changes: https://github.com/ionic-team/stencil/pull/4250 when zone.js is preset on the page.

Because zone.js is patching `Promise` constructor, but async methods are unaffected (they still return native `Promise` instance) stencil is now comparing instance of a native promise to a patched promise with instanceof check (`nativePromiseFromAsyncMethod instanceof ZoneAwarePromise`). This will always resolve to `false` and lifecycle method will be treated as synchronous, causing all sorts of problems, primarily that render method will not wait for promise to resolve or reject.

### Expected Behavior

Lifecycle methods like `componentWillLoad` and others that can be "awaited" should be if custom promise is used.

### System Info

```shell
System: node 18.16.0
    Platform: darwin (22.5.0)
   CPU Model: Apple M2 Pro (12 cpus)
    Compiler: /Users/marcin.cichocki/dev/playground/dist-bug/node_modules/@stencil/core/compiler/stencil.js
       Build: 1682947516
     Stencil: 3.2.2 🏒
  TypeScript: 4.9.5
      Rollup: 2.42.3
      Parse5: 7.1.2
      Sizzle: 2.42.3
      Terser: 5.17.1
```

### Steps to Reproduce

Run demo with `npm run start` and open console. The order of logs should be 1,2,3, but it is 1,3,2

### Code Reproduction URL

https://github.com/marcincichocki/stencil-promise-issue

### Additional Information

The workaround would be to not use async methods, and return `Promise.resolve()` but that would prevent from using await keyword, which is annoying when working with complex async code (hello promise hell).

Also as a note why we use zone.js: This is a direct dependency of angular, which we use side by side with stencil.

alicewriteswrongs:
[...]

PR #4250, opened 2023-04-10
refactor(runtime): document Promise-based queueing in dispatchHooks

alicewriteswrongs:
This documents and slightly simplifies code which uses `Promise.then` to enqueue a series of rerendering-related actions which need to take place when a component gets updated. In particular, the previous implementation used a helper function called `then` which was a) `any`-typed and b) not necessary.

This also adds JSDoc to `dispatchHooks` and some explanatory inline comments.

## Pull request checklist

Please check if your PR fulfills the following requirements:
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [ ] Build (`npm run build`) was run locally and any changes were pushed
- [ ] Unit tests (`npm test`) were run locally and passed
- [ ] E2E Tests (`npm run test.karma.prod`) were run locally and passed
- [ ] Prettier (`npm run prettier`) was run locally and passed

## Pull request type

 

 

Please check the type of change your PR introduces:
- [ ] Bugfix
- [ ] Feature
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

## What is the current behavior?

GitHub Issue Number: N/A

## What is the new behavior?

This shouldn't change anything about how components are rendered, but is basically just a small simplification of a bit of code I read through recently and found confusing.

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

PR #4460, opened 2023-06-06
fix(runtime): issue with update-component and patched Promise

alicewriteswrongs:
This fixes an issue with a check used in `updateComponent` to enqueue asynchronous updates (implemented with async functions) to components. In particular, if the global `Promise` object is patched such that it is not equal to the native `Promise` object returned by, for instance, an `async` function, which will return the native, unpatched object even if `window.Promise` has been overwritten.

A previous change in #4250 made a change such that we did a check like this:

```ts
maybePromise instanceof Promise
```

The problem is that if `maybePromise` is the return value of an async function, like

```ts
const foo = await something();
```

and `window.Promise` has been overwritten then this check will unfortunately fail unexpectedly.

The fix is just to first check `instanceof` and then add a duck-type-ey fallback as well.

## Testing

Check out the reproduction case linked in #4459. Without this fix it will log `1,3,2` in the console, and with it, `1,2,3`.

### DIFF
--- a/src/runtime/update-component.ts
+++ b/src/runtime/update-component.ts
@@ -118,7 +118,21 @@ const dispatchHooks = (hostRef: d.HostRef, isInitialLoad: boolean): Promise<void
  * @returns either a `Promise` or the return value of the provided function
  */
 const enqueue = (maybePromise: Promise<void> | undefined, fn: () => Promise<void>): Promise<void> | undefined =>
-  maybePromise instanceof Promise ? maybePromise.then(fn) : fn();
+  isPromisey(maybePromise) ? maybePromise.then(fn) : fn();
+
+/**
+ * Check that a value is a `Promise`. To check, we first see if the value is an
+ * instance of the `Promise` global. In a few circumstances, in particular if
+ * the global has been overwritten, this is could be misleading, so we also do
+ * a little 'duck typing' check to see if the `.then` property of the value is
+ * defined and a function.
+ *
+ * @param maybePromise it might be a promise!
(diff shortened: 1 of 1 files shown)


src/runtime/update-component.ts
```diff
@@ -118,7 +118,11 @@ const dispatchHooks = (hostRef: d.HostRef, isInitialLoad: boolean): Promise<void
  * @returns either a `Promise` or the return value of the provided function
  */
 const enqueue = (maybePromise: Promise<void> | undefined, fn: () => Promise<void>): Promise<void> | undefined =>
-  maybePromise instanceof Promise ? maybePromise.then(fn) : fn();
+  isPromisey(maybePromise) ? maybePromise.then(fn) : fn();
+
+const isPromisey = (maybePromise: Promise<void> | unknown): maybePromise is Promise<void> =>
```
rwaskiewicz, line 123: There's a lot of good detail about this change in the pr summary - could we capture the cliff notes of it for `isPromisey` to describe why we're doing this structural type checking?

rwaskiewicz (changes requested):
Can you do me a favor and add testing steps to the PR summary please?

src/runtime/update-component.ts
alicewriteswrongs: ah made that change locally but forgot to push it! haha

rwaskiewicz (approved):
TY!

tanner-reits (approved):
Makes sense!

JessicaSachs (approved):

Issue #4462, opened 2023-06-07
bug: Rollup: Missing Export in 3.3.1, compiled ok in 3.3.0 and earlier


goulu:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

3.3.1

### Current Behavior

After updating 3.31, I suddenly get this build error on a custom component library that compiles in ealier versions :

Rollup: Missing Export 'ScreenTools' is not exported by ./src/components/test/screen-tools/screen-tools.tsx, imported by screen-tools.entry 1: export { ScreenTools as screen_tools } from 'C:/dev/sphere-stencil/src/components/test/screen-tools/screen-tools.tsx'; ^

The missing 

### Expected Behavior

compile ok as before ...

### System Info

```shell
System: node 19.8.1
    Platform: windows (10.0.19044)
   CPU Model: 11th Gen Intel(R) Core(TM) i7-11800H @ 2.30GHz (16 cpus)
    Compiler: C:\dev\sphere-stencil\node_modules\@stencil\core\compiler\stencil.js
       Build: 1685732213
     Stencil: 3.3.1
  TypeScript: 5.0.4
      Rollup: 2.42.3
      Parse5: 7.1.2
      Sizzle: 2.42.3
      Terser: 5.17.6


BTW I notice TypeScript version is here 5.04 while I set it to "typescript": "<5.0" in package.json because "jest": "^27" wants TS <5 ... conflict ?
```

### Steps to Reproduce

no idea, sorry. the library is proprietary so  I cannot share the code. Maybe just the beginning of screen-tools.tsx which is the screen allowing to set some parameters for the app :
```
import { Component, Event, EventEmitter, h, State } from '@stencil/core';
import { FieldInterface, logging, isMetric } from '@bobst/notre-lib';
import { Metric, Imperial, UnitSystem } from '../../../utils/units';
import { state } from '../../../utils/store';

import { logger } from '../../../utils/sphere';
const log = logger.getLogger('ScreenTools');

const skins = ['sphere', 'ionic', 'none'];

const unitSystems = [Metric, Imperial];

@Component({
[...]

PR #4463, opened 2023-06-07
fix(compiler): handle static members with stencil decorators

rwaskiewicz:

Stencil fails to compile a component that has a static class member that is both initialized and is decorated with Stencil's `@Prop` or `@State` decorators

GitHub Issue Number: #4462 

## What is the new behavior?

this commit fixes an edge case to the fix introduced in 6dbe9a5 (#4447). the mechanism for checking if there is a statically intiialized class member has been updated in order to exclude such members that are decorated with Stencil's `@Prop()` and `@State()` decorators.

- [ ] Yes
- [x] No

## Testing

Unit/Karma tests updated/added

I also tested this against #4462's provided repo, as well as a slightly larger component that still compiles/uses more Stencil decorators:
```tsx
import { Component, Event, EventEmitter,  Method, Listen, Prop, State } from '@stencil/core';

@Component({
  tag: 'screen-tools',
})
export class ScreenTools {
  @Prop() static skin = 12
  @State() static skin1 = 12

  @Event() static skinChange: EventEmitter;
  @Listen('todoCompleted')
  static todoCompletedHandler(event: CustomEvent<any>) {
    console.log('Received the custom todoCompleted event: ', event.detail);
  }

  @Method()
  static async changeSkin(skin: string) {
    console.log('skin', skin)
  }
}
```

### DIFF
--- a/src/compiler/transformers/decorators-to-static/convert-static-members.ts
+++ b/src/compiler/transformers/decorators-to-static/convert-static-members.ts
@@ -1,10 +1,14 @@
+import { readOnlyArrayHasStringMember } from '@utils';
 import ts from 'typescript';
+
+import { CONSTRUCTOR_DEFINED_MEMBER_DECORATORS } from './decorators-constants';
 
 /**
  * Helper function to detect if a class element fits the following criteria:
  * - It is a property declaration (e.g. `foo`)
  * - It has an initializer (e.g. `foo *=1*`)
  * - The property declaration has the `static` modifier on it (e.g. `*static* foo =1`)
+ * - The property declaration does not include a Stencil @Prop or @State decorator
  * @param classElm the class member to test
  * @returns true if the class member fits the above criteria, false otherwise
  */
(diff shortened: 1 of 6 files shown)


src/compiler/transformers/decorators-to-static/convert-static-members.ts
```diff
@@ -13,6 +16,25 @@ export const hasStaticInitializerInClass = (classElm: ts.ClassElement): boolean
...
+ * @returns true if the node is a decorator with the name 'Prop' or 'State', false otherwise
+ */
+const isStencilStateOrPropDecorator = (modifier: ts.ModifierLike): boolean => {
+  if (ts.isDecorator(modifier)) {
+    const decoratorName =
+      ts.isCallExpression(modifier.expression) &&
+      ts.isIdentifier(modifier.expression.expression) &&
+      modifier.expression.expression.text;
+    return (
+      typeof decoratorName !== 'boolean' && CONSTRUCTOR_DEFINED_MEMBER_DECORATORS.some((val) => val === decoratorName)
```
alicewriteswrongs, line 36: are you doing the `.some(...)` stuff to get around the typing for `.includes` ? I created this dumb little [`readOnlyArrayHasStringMember`](https://github.com/ionic-team/stencil/blob/29ae12061984eadd46fe2a73a56a7245e4271cee/src/utils/util.ts#L234-L237) function which could be used for that purpose (it has to do a typecast :/ )

alicewriteswrongs (changes requested):
looks straightforward enough and I believe the karma test!

just had like two little questions

alicewriteswrongs (changes requested):
looks straightforward enough and I believe the karma test!

just had like two little questions

alicewriteswrongs (changes requested):
looks straightforward enough and I believe the karma test!

just had like two little questions

alicewriteswrongs (changes requested):
looks straightforward enough and I believe the karma test!

just had like two little questions

alicewriteswrongs (approved):

test/karma/test-app/static-members/karma.spec.ts
```diff
@@ -15,6 +15,12 @@ describe('static-members', function () {
     expect(cmp.textContent.trim()).toBe('This is a component with static public and private members');
   });
 
+  it('renders properly with initialized, decorated static members', async () => {
+    const cmp = app.querySelector('static-decorated-members');
+
+    expect(cmp.textContent.trim()).toBe('This is a component with static a Stencil decorated member');
```
tanner-reits, line 21: non-blocker:

```suggestion
    expect(cmp.textContent.trim()).toBe('This is a component with a static Stencil decorated member');
```

tanner-reits (approved):
Changes look good. Allowing a `static` modified on `Prop`s and `State`s seems a little bizarre to me, but maybe I don't understand the use case 🤷‍♂️

JessicaSachs (approved):

PR #4464, opened 2023-06-07
remove unused dependencies after in-browser compilation delete

alicewriteswrongs:
This removes some dependencies which I believe to be unused and / or unnecessary after the in-browser compilation support delete (#4317)

## Testing

Good to make sure that Stencil builds and can be used to build projects like Framework

### DIFF
--- a/scripts/bundles/plugins/typescript-source-plugin.ts
+++ b/scripts/bundles/plugins/typescript-source-plugin.ts
@@ -139,8 +139,6 @@ async function bundleTypeScriptSource(tsPath: string, opts: BuildOptions): Promi
 
   const o: string[] = [];
   o.push(`// TypeScript ${opts.typescriptVersion}`);
-  o.push(`import { IS_NODE_ENV } from '@environment';`);
-  o.push(`process.browser = !IS_NODE_ENV;`);
   o.push(code);
   o.push(`export default ts;`);
   code = o.join('\n');
--- a/package-lock.json
+++ b/package-lock.json
@@ -67,7 +67,6 @@
         "open": "^9.0.0",
         "open-in-editor": "2.2.0",
         "parse5": "7.1.2",
-        "path-browserify": "^1.0.1",
         "pixelmatch": "5.3.0",
         "postcss": "^8.2.8",
         "prettier": "2.8.8",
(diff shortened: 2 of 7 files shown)


scripts/bundles/plugins/typescript-source-plugin.ts
```diff
@@ -140,7 +140,6 @@ async function bundleTypeScriptSource(tsPath: string, opts: BuildOptions): Promi
   const o: string[] = [];
   o.push(`// TypeScript ${opts.typescriptVersion}`);
   o.push(`import { IS_NODE_ENV } from '@environment';`);
-  o.push(`process.browser = !IS_NODE_ENV;`);
```
rwaskiewicz, line 143: Does `process.browser` get relied on somehow? E.g. in the TS source code? I'm curious what this was doing beforehand.

Also, do we need the import on `IS_NODE_ENV` still?

rwaskiewicz (changes requested):

scripts/bundles/plugins/typescript-source-plugin.ts
alicewriteswrongs: Oh yeah we could remove that here too (the `IS_NODE_ENV` import)

I'm going to do a separate PR to remove the `IS_NODE_ENV` and `IS_BROWSER_ENV` variables entirely, it's a little more involved for that though

alicewriteswrongs: and I don't believe that we have any dependency on `process.browser`

rwaskiewicz: FWIW, TS has the following in its source:
```
  function isNodeLikeSystem() {
    return typeof process !== "undefined" && process.nextTick && !process.browser && typeof module === "object";
  }
```
It _may_ make sense to just set this value to `false` explicitly, although IDK if there's another "thing" that could be setting it

alicewriteswrongs: I don't think we need to do that, most people using typescript in a node environment don't end up needing to do anything in particular to tell typescript 'this is node, do node things' so I don't think we have a particular need to either - this stuff was I think here purely for indicating the opposite case (in appropriate conditions)

rwaskiewicz: Fair, that all makes sense to me!

rwaskiewicz (approved):

tanner-reits (approved):

Issue #2882, opened 2021-04-13
Property 'autoComplete' does not exist on type 'TextareaHTMLAttributes<HTMLTextAreaElement>'


joao-m-santos:
**Stencil version:**

```
 @stencil/core@2.5.0
```

**I'm submitting a:**

[x] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**
Setting the `autoComplete` or `autocomplete` attributes of a `<textarea>` causes the error:
```Property 'autoComplete' does not exist on type 'TextareaHTMLAttributes<HTMLTextAreaElement>'```

**Expected behavior:**
There should be no error popping up.

**Related code:**

```tsx
<textarea autoComplete="on"></textarea>
```

**Other information:**
Typescript version: 3.9.9

rwaskiewicz:
The fix for this issue has been [officially released as a part of Stencil 3.4.0](https://github.com/ionic-team/stencil/releases/tag/v3.4.0). Folks running into this issue should update to v3.4.0 at minimum. I'm going to close this out following the release, but please feel free to open a new issue should this reappear. Thanks!

PR #4465, opened 2023-06-07
fix(runtime): add autocomplete to textarea

rwaskiewicz:

Using 'autoComplete' attributes causes type errors like so:
Screenshot 2023-06-07 at 5 36 48 PM

GitHub Issue Number: https://github.com/ionic-team/stencil/issues/2882

## What is the new behavior?

we accept 'on' or 'off' for the attribute

- [ ] Yes
- [x] No

## Testing

I spun up a component library `npm init stencil component attr-test` and added the text-area elements you see in the screenshot above. I then built + packed this library, and installed it. After installing it, type errors only occur on invalid values:
Screenshot 2023-06-07 at 5 37 08 PM

### DIFF
--- a/src/declarations/stencil-public-runtime.ts
+++ b/src/declarations/stencil-public-runtime.ts
@@ -1257,6 +1257,8 @@ export namespace JSXBase {
   }
 
   export interface TextareaHTMLAttributes<T> extends HTMLAttributes<T> {
+    autoComplete?: string;
+    autocomplete?: string;
     autoFocus?: boolean;
     autofocus?: boolean | string;
     cols?: number;


src/declarations/stencil-public-runtime.ts
```diff
@@ -1257,6 +1257,8 @@ export namespace JSXBase {
   }
 
   export interface TextareaHTMLAttributes<T> extends HTMLAttributes<T> {
+    autoComplete?: 'on' | 'off';
+    autocomplete?: 'on' | 'off';
```
alicewriteswrongs, line 1261: I'm a little unsure of exactly what's the standard behavior here. The [MDN docs for <textarea>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea#attributes) imply that `'on'` and `'off'` are the only supported values for `autocomplete` on a `<textarea>`, but the [MDN page for `autocomplete`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete) shows a much wider range of possible values like `"given-name"`, `"email"`, etc.

I went at looked at [the whatwg standard document](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofilling-form-controls:-the-autocomplete-attribute) and I haven't found something explicitly saying 'these values should be allowed on a `<textarea>`' _but_ some of the examples show other values being passed to autocomplete, like in this example code:

```html
<fieldset>
 <legend>Ship the blue gift to...</legend>
 <p> <label> Address:     <textarea name=ba autocomplete="section-blue shipping street-address"></textarea> </label>
 <p> <label> City:        <input name=bc autocomplete="section-blue shipping address-level2"> </label>
 <p> <label> Postal Code: <input name=bp autocomplete="section-blue shipping postal-code"> </label>
</fieldset>
<fieldset>
 <legend>Ship the red gift to...</legend>
 <p> <label> Address:     <textarea name=ra autocomplete="section-red shipping street-address"></textarea> </label>
 <p> <label> City:        <input name=rc autocomplete="section-red shipping address-level2"> </label>
 <p> <label> Postal Code: <input name=rp autocomplete="section-red shipping postal-code"> </label>
</fieldset>
```

So anyway this is a long way to say we might have to worry about a wider type here - perhaps we should just do `string` and let the browser worry about doing the right thing? Alternatively we could maintain a string union type that captures all these possible values, like

```ts
type AutocompleteOption =
  | "on"
  | "off"
  | "email"
  | "cc-number"
  | ... ; // etc
```

alicewriteswrongs (changes requested):
just had one question about how this should be typed

src/declarations/stencil-public-runtime.ts
sean-perkins: I would recommend a `string` type to align with the implementation for `HTMLInputElement`, etc.: https://github.com/TypeStrong/tscs/blob/master/lib/node_modules/TypeScript/bin/lib.dom.d.ts#L6814-L6817

JessicaSachs: Agreed. Should keep it in line with the `HTMLInputElement` spec. Better to make it wider rather than receive another bug report about the type being too narrow.

JessicaSachs (changes requested):
Same feedback as Alice. Let's widen the type to `string`.

JessicaSachs (approved):

sean-perkins (approved):

PR #4471, opened 2023-06-12
test(delegatesFocus): skip delegatesfocus test

rwaskiewicz:
starting on 2023.06.12, browserstack started to use microsoft edge 114.0.1823.37. a test that had previously passed began to fail. the test that is skipped in this commit is that test. we were unable to reproduce this locally on macos and windows with the latest version of edge (114.0.1823.42). the hope is that browserstack will move to a newer version of edge in the next day or so to render this change obsolete

GitHub Issue Number: N/A

## What is the new behavior?

- 

- [ ] Yes
- [ ] No

### DIFF
--- a/test/karma/test-app/delegates-focus/karma.spec.ts
+++ b/test/karma/test-app/delegates-focus/karma.spec.ts
@@ -9,7 +9,8 @@ describe('delegates-focus', function () {
   });
   afterEach(tearDownDom);
 
-  it('should delegate focus', async () => {
+  // TODO(STENCIL-862): Re-enable this test
+  xit('should delegate focus', async () => {
     const button = app.querySelector('button');
     const delegateFocusElm = app.querySelector('delegates-focus');
     const noDelegateFocusElm = app.querySelector('no-delegates-focus');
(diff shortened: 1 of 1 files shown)


alicewriteswrongs (approved):
lgtm :shipit: 

should we merge this to the v4 branch too?

rwaskiewicz:
Perhaps - my thought was to kind of "queue it up" for tomorrow if we needed it, but that was assuming:
1. that BS wouldn't stabilize between now and tomorrow - it still hasn't, but if I could wait ~24 hours and have BS "just work" without needing to merge and later revert this PR, that'd be ideal
2. I assumed there wouldn't be _additional_ BS flakiness, which there is 😢 

My current plan is to evaluate the state of things tomorrow morning. 
1. If everything is stable (without this PR) I'll close this PR and get everything into the queue prior to/for the release
2. If everything is stable but _only_ with this PR, I'll merge this into `main`, get everything into the queue prior to/for the release, cut the release in the afternoon, and rebase the v4 branch on `main`
3. If nothing is stable, well I'm going to have to figure something else out 😬 

🤞 for option 1 or 2!

PR #4472, opened 2023-06-12
chore(engines): drop Node 14 support

tanner-reits:

Stencil currently support Node 14

GitHub Issue Number: N/A

## What is the new behavior?

Minimum version Stencil supports is now Node 16 🎉  

- [X] Yes
- [ ] No


package.json
```diff
@@ -125,8 +125,8 @@
     "ws": "8.13.0"
   },
   "engines": {
-    "node": ">=14.10.0",
-    "npm": ">=6.0.0"
+    "node": ">=16.0.0",
+    "npm": ">=7.10.0"
```
tanner-reits, line 129: Pulled these from Node's [previous releases matrix](https://nodejs.org/en/download/releases)

tanner-reits (commented):
Just noting a few things. Need CI to run to make sure things are looking good and need to test the `fs` changes still. 

Also, the docs site was already updated for v4 node support so don't need to make any changes over there

alicewriteswrongs (approved):

test/karma/package.json
```diff
@@ -45,7 +45,7 @@
     "workbox-build": "4.3.1"
   },
   "volta": {
-    "node": "14.21.1",
-    "npm": "6.14.18"
+    "node": "16.20.0",
+    "npm": "8.19.4"
```
rwaskiewicz, line 49: Can we bump `@types/node` while we're in here? ATM, we're pinned at v14

rwaskiewicz: In doing this (bump) I believe that'll update the `package-lock.json` to v2 of the package lock format. If that's the case, can you bump the `package-lock.json` files of under `test/karma` too? I think it's just:
- `test/karma/test-sibling`
- `test/karma/test-invisible-prehydration`

rwaskiewicz (changes requested):
Can you do me a favor and fill out the testing steps section of this PR for me please?

rwaskiewicz (approved):

PR #4483, opened 2023-06-13
chore(snc): fix SNC errors stemming from `src/compiler/cache.ts`

alicewriteswrongs:
This just fixes a few SNC errors that stem from code in `src/compiler/cache.ts`.

- [ ] Yes
- [x] No

## Testing

I think that if unit tests and CI is all passing then we're probably good here

### DIFF
--- a/src/compiler/build/test/build-stats.spec.ts
+++ b/src/compiler/build/test/build-stats.spec.ts
@@ -1,16 +1,17 @@
 import type * as d from '@stencil/core/declarations';
-import { mockBuildCtx, mockCompilerCtx, mockConfig } from '@stencil/core/testing';
+import { mockBuildCtx, mockCompilerCtx, mockValidatedConfig } from '@stencil/core/testing';
 import { result } from '@utils';
 
 import { generateBuildResults } from '../build-results';
 import { generateBuildStats } from '../build-stats';
 
 describe('generateBuildStats', () => {
-  const config = mockConfig();
+  let config: d.ValidatedConfig;
   let compilerCtx: d.CompilerCtx;
   let buildCtx: d.BuildCtx;
 
   beforeEach(() => {
+    config = mockValidatedConfig();
     compilerCtx = mockCompilerCtx(config);
     buildCtx = mockBuildCtx(config, compilerCtx);
   });
--- a/src/compiler/cache.ts
+++ b/src/compiler/cache.ts
@@ -11,7 +11,7 @@ export class Cache implements d.Cache {
   private buildCacheDir: string;
 
   constructor(
-    private config: d.Config,
+    private config: d.ValidatedConfig,
     private cacheFs: InMemoryFileSystem,
   ) {
     this.sys = config.sys;
@@ -55,7 +55,7 @@ export class Cache implements d.Cache {
       return null;
     }
 
-    let result: string;
+    let result: string | null;
     try {
       result = await this.cacheFs.readFile(this.getCacheFilePath(key));
       this.failed = 0;
@@ -73,17 +73,13 @@ export class Cache implements d.Cache {
       return false;
     }
 
-    let result: boolean;
-
     try {
       await this.cacheFs.writeFile(this.getCacheFilePath(key), value);
-      result = true;
+      return true;
     } catch (e: unknown) {
       this.failed++;
-      result = false;
+      return false;
     }
-
-    return result;
   }
 
   async has(key: string) {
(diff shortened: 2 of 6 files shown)


src/testing/mocks.ts
```diff
@@ -200,14 +200,12 @@ export function mockBuildCtx(config?: Config, compilerCtx?: CompilerCtx): BuildC
 }
 
 export function mockCache(config?: Config, compilerCtx?: CompilerCtx) {
-  if (!config) {
-    config = mockConfig();
-  }
+  const validatedConfig = mockValidatedConfig((config as Partial<ValidatedConfig>) ?? {});
```
rwaskiewicz, line 203: Is the type assertion here necessary? I would have thought TS woulda been happy about this, but maybe I'm missing something!

rwaskiewicz (commented):
Two small questions - I'm not entirely convinced they're valid questions, feel free to push back

src/testing/mocks.ts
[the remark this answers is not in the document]
alicewriteswrongs: I'm not terribly worried about making the change right now, as it is passing CI and working and whatnot, but I wouldn't be opposed to making the `Config` passed in to `CompilerCtx` and so on a `ValidatedConfig` which would let us do this (pass down a single object that is shared between the `CompilerCtx` and the `Cache`). I started to do so but backed it out in the interest of keeping this diff a bit smaller, but I'll take a stab at it and see how it ends up. Usually swapping the two out gets us a few more SNC violations dropped anyway!

alicewriteswrongs: yeah it is unfortunately, I thought that the typing of `config` as `d.Config` would satisfy `Partial<ValidatedConfig>` but it did not :/

I think we now have a few fields where the difference between them isn't just that they're optional in `Config` and required on `ValidatedConfig`

src/compiler/cache.ts
```diff
@@ -158,15 +158,12 @@ export class Cache implements d.Cache {
     }
   }
 
-  private getCacheFilePath(key: string) {
+  private getCacheFilePath(key: string): string {
     return join(this.config.cacheDir, key) + '.log';
   }
 
-  getMemoryStats() {
-    if (this.cacheFs != null) {
```
sean-perkins, line 166: There is a number of checks around `cacheFs` being defined. Is it safe to assume it is always defined (meaning we can simplify the other checks) or is it always guaranteed to be defined in the execution when this method would be called?

alicewriteswrongs: I believe it is guaranteed to be defined, I can't find anywhere where it's reassigned after the initial assignment in the constructor so I think the checks are unnecessary -- I'm going to try removing them and see if anything breaks!

alicewriteswrongs: nvm did some testing and we need those checks!

src/compiler/build/test/build-stats.spec.ts
```diff
@@ -1,15 +1,16 @@
 import type * as d from '@stencil/core/declarations';
-import { mockBuildCtx, mockCompilerCtx, mockConfig } from '@stencil/core/testing';
+import { mockBuildCtx, mockCompilerCtx, mockValidatedConfig } from '@stencil/core/testing';
 
 import { generateBuildResults } from '../build-results';
 import { generateBuildStats } from '../build-stats';
 
 describe('generateBuildStats', () => {
-  const config = mockConfig();
+  let config: d.ValidatedConfig;
```
rwaskiewicz, line 8: 💯

src/compiler/cache.ts
rwaskiewicz: Does anything need to get pushed back up to GH/changed here if we need these checks?

alicewriteswrongs: yup, forgot to push reverting that! should be good now

src/testing/mocks.ts
```diff
@@ -230,12 +217,12 @@ export function mockCompilerSystem(): TestingSystem {
   return createTestingSystem();
 }
 
-export function mockDocument(html: string = null) {
+export function mockDocument(html: string) {
```
tanner-reits, line 220: Was the arg here intended not to be optional anymore? Looks like our internal implementations all provide an arg, but it this is also exported from our `testing` module. Also, `mockWindow` below had the same signature and that arg was kept optional.

tanner-reits (changes requested):
One tiny question

src/testing/mocks.ts
alicewriteswrongs: mmm tbh I don't remember why I did this! I'll change it so the type is `string | null`

tanner-reits (approved):

PR #4501, opened 2023-06-21
fix(compiler): reorder tsconfig#path transforms 

rwaskiewicz:
<!-- Please refer to our contributing documentation for any questions on submitting a pull request, or let us know here if you need any help: https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md -->


<!-- Please describe the current behavior that you are modifying, or link to a relevant issue. -->

In very specific circumstances, Stencil might not detect the correct `h` and `Host` functions as a result of transforming import paths specified in `tsconfig.json#paths` (see testing section).

GitHub Issue Number: N/A


## What is the new behavior?
<!-- Please describe the behavior or changes that are being added by this PR. -->

this commit reorders the transformations run for rewriting paths to
modules in import statements.

previously, modules in import statements would be rewritten based on
the contents of `tsconfig.json#paths` before any other transformation
was run. this would conflict with the `updateStencilCoreImports`
transformation if `@stencil/core` (or any of its valid subpaths) were
listed in `tsconfig.json#paths`. this would cause the wrong versions of
helpers such as `h` and `Host` to be picked up when a project used the
`--prerender` flag on a component that imported either of the above
helpers (the `lazy` output target version would be inlined to the output
rather than the `hydrate` version).

this commit removes the transformation `run-program` and places it
(conditionally) in the "before" transformers list _after_ stencil
imports are transformed. 

- [ ] Yes
- [x] No

<!-- If this introduces a breaking change, please describe the impact and migration path for existing applications below. -->

## Testing

I generated a dev build of this branch - `@stencil/core@3.4.0-dev.1687377141.615ce97`

### Locally

The underlying issue can be reproduced by using https://github.com/rwaskiewicz/transform-import-paths

this reproduction expects that the location of a built stencil compiler to exist at `~/workspaces/stencil/bin/stencil`.
if your local stencil exists at `~/dev/username/projects/stencil`, update occurences of `~/workspaces` with the parent directory of your stencil repo.

this project contains `build` and `build-local` scripts.
the former will run a build against your local stencil project, the latter against the version of stencil installed by npm.
when running without this PR built locally, we expect both to fail.
once this PR is built, `build` ought to work, while `build-local` continues to fail
if `npm i @stencil/core@3.4.0-dev.1687377141.615ce97` is run, `build-local` will pass

likewise, without building this pr, `npm run prerender` will fail.
upon building it, `npm run prerender` will pass

### Ionic Framework

Ionic Framework already [uses the `transformAliasedImportPaths`](https://github.com/ionic-team/ionic-framework/blob/f29c66aee20573b9aab30811b46b689c49d29524/core/stencil.config.ts#L220) and [has a few `paths` in `tsconfig.json` set](https://github.com/ionic-team/ionic-framework/blob/f29c66aee20573b9aab30811b46b689c49d29524/core/tsconfig.json#L30-L33). With that in mind, the test here is that existing aliasing doesn't break when we exercise the entirety of the CI pipeline in Framework.

I created a temporary branch in the Ionic Framework repo that would run the `stencil-nightly` job with my dev build. Specifically, I applied this patch:
```diff
diff --git a/.github/workflows/stencil-nightly.yml b/.github/workflows/stencil-nightly.yml                                                                                     
index 404cb22e5b..c82f0e092a 100644                                                                                                                                            
--- a/.github/workflows/stencil-nightly.yml                                                                                                                                    
+++ b/.github/workflows/stencil-nightly.yml                                                                                                                                    
@@ -24,7 +24,7 @@ jobs:                                                                                                                                                        
       - uses: actions/checkout@v3                                                                                                                                             
       - uses: ./.github/workflows/actions/build-core-stencil-prerelease                                                                                                       
         with:                                                                                                                                                                 
-          stencil-version: nightly                                                                                                                                            
+          stencil-version: 3.4.0-dev.1687377141.615ce97                                                                                                                       
                                                                                                                                                                               
   test-core-clean-build:                                                                                                                                                      
     needs: [build-core-with-stencil-nightly]                                                                                                                                  
```
and manually kicked off the workflow (using my branch).

The results of that build [can be found here](https://github.com/ionic-team/ionic-framework/actions/runs/5338399217/jobs/9675723302)

## Other information

<!-- Any other information that is important to this PR such as screenshots of how a component looks before and after the change. -->

There are 2 small refactors in this PR:
1. this commit renames the `BundleOptions#customTransfomers` field to
`customBeforeTransformers` as to clarify the usage of the transformers
placed in the collection assoicated with this field
2. JSDoc was added to the `BundleOptions` fields for transformers

### DIFF
--- a/src/compiler/bundle/typescript-plugin.ts
+++ b/src/compiler/bundle/typescript-plugin.ts
@@ -57,7 +57,9 @@ export const typescriptPlugin = (compilerCtx: d.CompilerCtx, bundleOpts: BundleO
           const tsResult = ts.transpileModule(mod.staticSourceFileText, {
             compilerOptions: config.tsCompilerOptions,
             fileName: mod.sourceFilePath,
-            transformers: { before: bundleOpts.customTransformers },
+            transformers: {
+              before: bundleOpts.customBeforeTransformers ?? [],
+            },
           });
           const sourceMap: d.SourceMap = tsResult.sourceMapText ? JSON.parse(tsResult.sourceMapText) : null;
           return { code: tsResult.outputText, map: sourceMap };
--- a/src/compiler/output-targets/dist-custom-elements/index.ts
+++ b/src/compiler/output-targets/dist-custom-elements/index.ts
@@ -20,6 +20,7 @@ import { addDefineCustomElementFunctions } from '../../transformers/component-na
 import { proxyCustomElement } from '../../transformers/component-native/proxy-custom-element-function';
 import { nativeComponentTransform } from '../../transformers/component-native/tranform-to-native-component';
 import { removeCollectionImports } from '../../transformers/remove-collection-imports';
+import { rewriteAliasedSourceFileImportPaths } from '../../transformers/rewrite-aliased-paths';
 import { updateStencilCoreImports } from '../../transformers/update-stencil-core-import';
 import { getCustomElementsBuildConditionals } from './custom-elements-build-conditionals';
 
@@ -75,7 +76,7 @@ export const getBundleOptions = (
   id: 'customElements',
   platform: 'client',
   conditionals: getCustomElementsBuildConditionals(config, buildCtx.components),
-  customTransformers: getCustomElementCustomTransformer(config, compilerCtx, buildCtx.components, outputTarget),
+  customBeforeTransformers: getCustomBeforeTransformers(config, compilerCtx, buildCtx.components, outputTarget),
   externalRuntime: !!outputTarget.externalRuntime,
   inlineWorkers: true,
   inputs: {
@@ -314,7 +315,7 @@ export const generateEntryPoint = (
  * @param outputTarget the output target configuration
  * @returns a list of transformers to use in the transpilation process
  */
-const getCustomElementCustomTransformer = (
+const getCustomBeforeTransformers = (
   config: d.ValidatedConfig,
   compilerCtx: d.CompilerCtx,
   components: d.ComponentCompilerMeta[],
@@ -329,11 +330,19 @@ const getCustomElementCustomTransformer = (
     style: 'static',
     styleImportData: 'queryparams',
   };
-  return [
+  const customBeforeTransformers = [
     addDefineCustomElementFunctions(compilerCtx, components, outputTarget),
     updateStencilCoreImports(transformOpts.coreImportPath),
+  ];
+
+  if (config.transformAliasedImportPaths) {
+    customBeforeTransformers.push(rewriteAliasedSourceFileImportPaths);
+  }
+
+  customBeforeTransformers.push(
     nativeComponentTransform(compilerCtx, transformOpts),
     proxyCustomElement(compilerCtx, transformOpts),
-    removeCollectionImports(compilerCtx),
-  ];
+    removeCollectionImports(compilerCtx)
+  );
+  return customBeforeTransformers;
 };
--- a/src/compiler/output-targets/dist-lazy/lazy-output.ts
+++ b/src/compiler/output-targets/dist-lazy/lazy-output.ts
@@ -1,5 +1,6 @@
 import { catchError, isOutputTargetDist, isOutputTargetDistLazy, sortBy } from '@utils';
 import MagicString from 'magic-string';
+import * as ts from 'typescript';
 
 import type * as d from '../../../declarations';
 import type { BundleOptions } from '../../bundle/bundle-interface';
@@ -17,6 +18,7 @@ import { generateComponentBundles } from '../../entries/component-bundles';
 import { generateModuleGraph } from '../../entries/component-graph';
 import { lazyComponentTransform } from '../../transformers/component-lazy/transform-lazy-component';
 import { removeCollectionImports } from '../../transformers/remove-collection-imports';
+import { rewriteAliasedSourceFileImportPaths } from '../../transformers/rewrite-aliased-paths';
 import { updateStencilCoreImports } from '../../transformers/update-stencil-core-import';
 import { generateCjs } from './generate-cjs';
 import { generateEsm } from './generate-esm';
@@ -42,7 +44,7 @@ export const outputLazy = async (
       id: 'lazy',
       platform: 'client',
       conditionals: getLazyBuildConditionals(config, buildCtx.components),
-      customTransformers: getLazyCustomTransformer(config, compilerCtx),
+      customBeforeTransformers: getCustomBeforeTransformers(config, compilerCtx),
       inlineWorkers: config.outputTargets.some(isOutputTargetDist),
       inputs: {
         [config.fsNamespace]: LAZY_BROWSER_ENTRY_ID,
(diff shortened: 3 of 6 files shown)


src/compiler/bundle/typescript-plugin.ts
```diff
@@ -57,7 +57,10 @@ export const typescriptPlugin = (compilerCtx: d.CompilerCtx, bundleOpts: BundleO
           const tsResult = ts.transpileModule(mod.staticSourceFileText, {
             compilerOptions: config.tsCompilerOptions,
             fileName: mod.sourceFilePath,
-            transformers: { before: bundleOpts.customTransformers },
+            transformers: {
+              before: bundleOpts.customBeforeTransformers,
```
alicewriteswrongs, line 61: should this `before` have a `?? []` too?

src/compiler/output-targets/dist-custom-elements/index.ts
```diff
@@ -314,7 +319,7 @@ export const generateEntryPoint = (
  * @param outputTarget the output target configuration
  * @returns a list of transformers to use in the transpilation process
  */
-const getCustomElementCustomTransformer = (
+const getBeforeCustomTransformers = (
```
alicewriteswrongs, line 322: should this function possibly be called `getCustomBeforeTransformers`? that would be consistent with the property added to `BundleOptions` I think

src/compiler/output-targets/dist-custom-elements/index.ts
```diff
@@ -329,11 +334,47 @@ const getCustomElementCustomTransformer = (
     style: 'static',
     styleImportData: 'queryparams',
   };
-  return [
+  const beforeCustomTransformers = [
```
alicewriteswrongs, line 337: likewise possibly this array should be `customBeforeTransformers`

src/compiler/output-targets/dist-custom-elements/index.ts
```diff
@@ -329,11 +334,47 @@ const getCustomElementCustomTransformer = (
...
     addDefineCustomElementFunctions(compilerCtx, components, outputTarget),
     updateStencilCoreImports(transformOpts.coreImportPath),
+  ];
+
+  if (config.transformAliasedImportPaths) {
+    beforeCustomTransformers.push(rewriteAliasedSourceFileImportPaths);
+  }
```
alicewriteswrongs, line 344: checking my understanding: this is because the `updateStencilCoreImports` transformer has to run before the `rewriteAliasedSourceFileImportPaths` transformer in order to avoid the reported bug?

src/compiler/output-targets/dist-custom-elements/index.ts
```diff
@@ -329,11 +334,47 @@ const getCustomElementCustomTransformer = (
...
+ * See here: https://github.com/itsdouges/typescript-transformer-handbook#transforms
+ * and here: https://github.com/microsoft/TypeScript/pull/23946
+ *
+ * This quirk is not terribly well documented, unfortunately.
+ *
+ * @param config the Stencil configuration associated with the current build
+ * @returns a collection of transformations that should be applied to the source code, intended for the
+ * `afterDeclarations` part of the pipeline
+ */
+const getAfterDeclarationsCustomTransformers = (
```
alicewriteswrongs, line 372: similarly I'm wondering if this should be `getCustomAfterDeclarationsTransformers`

src/compiler/output-targets/dist-lazy/lazy-output.ts
```diff
@@ -107,11 +125,45 @@ const getLazyCustomTransformer = (config: d.ValidatedConfig, compilerCtx: d.Comp
...
+ * See here: https://github.com/itsdouges/typescript-transformer-handbook#transforms
+ * and here: https://github.com/microsoft/TypeScript/pull/23946
+ *
+ * This quirk is not terribly well documented, unfortunately.
+ *
+ * @param config the Stencil configuration associated with the current build
+ * @returns a collection of transformations that should be applied to the source code, intended for the
+ * `afterDeclarations` part of the pipeline
+ */
+const getAfterDeclarationsCustomTransformers = (
```
alicewriteswrongs, line 159: question: we have this same function copy-pasted essentially. does it make sense to try to carve out a `getSharedCustomAfterDeclarationsTransformers` function that would pull out the transfomers which are common to all of our output targets? Likewise we could maybe pull out shared logic for setting up the transformers which are shared by output targets for the `before` bit. Might be more trouble than it's worth though...

alicewriteswrongs (changes requested):
code looks straightforward, just had a few questions about it

src/compiler/output-targets/dist-custom-elements/index.ts
rwaskiewicz: Yup - exactly.

In the future it might be interesting to try to remove the need for `updateStencilCoreImports` and avoid subtle-ish bugs like this. I thought about pulling that thread, but felt that would have been much bigger in scope 😅

src/compiler/output-targets/dist-lazy/lazy-output.ts
rwaskiewicz: I thought about that - and I almost did it. @tanner-reits noticed that type transformations aren't working in all cases in Framework. I _think_ I'm going to end up putting the `afterDeclarations` stuff back in `run-program.ts`, but need to do some testing first before I commit to that

rwaskiewicz:
@tanner-reits 

>    (I think this might actually be behavior we want, just noting the difference). With this change, aliased paths aren't re-written for the dist-collection target when transformAliasedImportPaths is set. Instead, these paths only get transformed when transformAliasedImportPathsInCollection is set on the dist target or transformAliasedImportPaths is set directly on the dist-collection target.

That seems right to be, since we set that flag for `DIST_COLLECTION` when an OT has `collectionDir` set:
https://github.com/ionic-team/stencil/blob/1a2c1608a41bbe0420ec066d2bf56f32ed6613b8/src/compiler/config/outputs/validate-dist.ts#L66-L72

>    Doesn't look like aliased imports in typedefs are getting transformed. I made a change in framework to add an aliased path to an interface used in a component but the path does not transform in the output in the dist/types directory like it does in v3.3.0

Good catch! Fixed in 9b9eb7e3c

src/compiler/output-targets/dist-custom-elements/index.ts
rwaskiewicz: I ended up removing these functions in https://github.com/ionic-team/stencil/pull/4501/commits/9b9eb7e3c8df296b28ea8e64b23aeecc1b6c6227

src/compiler/output-targets/dist-lazy/lazy-output.ts
rwaskiewicz: Yup - removing these in https://github.com/ionic-team/stencil/pull/4501/commits/9b9eb7e3c8df296b28ea8e64b23aeecc1b6c6227

src/compiler/output-targets/dist-custom-elements/index.ts
alicewriteswrongs: it makes sense that you had to put that back into the earlier pre-output target transformation since iirc it's in that run of the typescript compiler that we do `declaration: true` if you have certain options set, so I think adding a transformer for `afterDeclarations` wouldn't have an effect if it's added in the output target code

rwaskiewicz:
@tanner-reits @alicewriteswrongs Thanks for the feedback! Everything should be fixed/addressed!

src/compiler/bundle/typescript-plugin.ts
rwaskiewicz: https://github.com/ionic-team/stencil/pull/4501/commits/273f9caefb04da6757cc2e56ffd043bb2be0ab07

src/compiler/output-targets/dist-custom-elements/index.ts
rwaskiewicz: https://github.com/ionic-team/stencil/pull/4501/commits/fc54d37d4df6de33391442271b2f4828ff5fc2dd

src/compiler/output-targets/dist-custom-elements/index.ts
rwaskiewicz: https://github.com/ionic-team/stencil/pull/4501/commits/89f9ea20c7caa4d90e1a755e329e47ca10ec1803

alicewriteswrongs (approved):
all good! confirmed the issue w/ the reproduction (after a little trouble 😅) and then confirmed that the fix eliminates the issue

tanner-reits:
Typedefs are working as expected again. 🚢  it!

tanner-reits (approved):

PR #4503, opened 2023-06-21
fix(compiler): re-enable build caching

tanner-reits:

Caching was disabled in [this commit](https://github.com/ionic-team/stencil/commit/63595bc540db9aa85295f5a8dc8738beb2a37611#diff-a2a3c503f6a2b9ee197d01f14954b0f59a14820ed2b30cd6046e92d5d233e446). 

GitHub Issue Number: N/A

## What is the new behavior?

This commit re-enables build caching and makes a few updates to not conflict with caching when using a screenshot connector.

- [ ] Yes
- [X] No

## Testing

Automated unit and e2e tests continue to pass.

Did quite a bit of manual benchmarking against Ionic framework. Results can be found [here](https://ionic-cloud.atlassian.net/browse/STENCIL-848?focusedCommentId=21008)

## Other information

Huge thanks to @George-Payne for pointing this out and doing the work to get this moving in [his PR](#3456)!

This branch/PR was created as an easy means of porting his work over to an updated version of Stencil

### DIFF
--- a/src/compiler/cache.ts
+++ b/src/compiler/cache.ts
@@ -8,7 +8,7 @@ export class Cache implements d.Cache {
   private skip = false;
   private sys: d.CompilerSystem;
   private logger: d.Logger;
-  private cacheDir: string;
+  private buildCacheDir: string;
 
   constructor(private config: d.Config, private cacheFs: InMemoryFileSystem) {
     this.sys = config.sys;
@@ -20,7 +20,7 @@ export class Cache implements d.Cache {
       return;
     }
 
-    this.cacheDir = this.config.cacheDir;
+    this.buildCacheDir = join(this.config.cacheDir, '.build');
 
     if (!this.config.enableCache || !this.cacheFs) {
       this.config.logger.info(`cache optimizations disabled`);
@@ -28,10 +28,10 @@ export class Cache implements d.Cache {
       return;
     }
 
-    this.config.logger.debug(`cache enabled, cacheDir: ${this.cacheDir}`);
+    this.config.logger.debug(`cache enabled, cacheDir: ${this.buildCacheDir}`);
 
     try {
-      const readmeFilePath = join(this.cacheDir, '_README.log');
+      const readmeFilePath = join(this.buildCacheDir, '_README.log');
       await this.cacheFs.writeFile(readmeFilePath, CACHE_DIR_README);
     } catch (e) {
       this.logger.error(`Cache, initCacheDir: ${e}`);
@@ -126,8 +126,8 @@ export class Cache implements d.Cache {
       }
 
       const fs = this.cacheFs.sys;
-      const cachedFileNames = await fs.readDir(this.cacheDir);
-      const cachedFilePaths = cachedFileNames.map((f) => join(this.cacheDir, f));
+      const cachedFileNames = await fs.readDir(this.buildCacheDir);
+      const cachedFilePaths = cachedFileNames.map((f) => join(this.buildCacheDir, f));
 
       let totalCleared = 0;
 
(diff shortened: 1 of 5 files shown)


src/compiler/cache.ts
```diff
@@ -19,16 +20,18 @@ export class Cache implements d.Cache {
       return;
     }
 
+    this.buildCacheDir = join(this.config.cacheDir, '.build');
```
tanner-reits, line 23: Decided to just make changes here so that the `cacheDir` specified in the Stencil config has a sub-directory for all the build caching files. This prevents us from needing to add or change the config property and eliminates confusion on if that field represented the actual directory where files were written or a parent directory. Also added some JSDoc to the config field to help add clarity

src/declarations/stencil-public-compiler.ts
```diff
@@ -42,6 +42,40 @@ export interface StencilConfig {
...
+   *     ├── .build <-- Where build related file caching is written
+   *     |
+   *     └── screenshot-cache.json <-- Where screenshot caching is written
+   * ```
+   */
+  cacheDir?: string;
+  /**
+   * @deprecated this flag is no longer used by Stencil and can be safely removed.
+   */
+  enableCacheStats?: boolean;
```
tanner-reits, line 78: Do we want to just remove this now since we're just adding this change to v4? It's not used anywhere so the only "breaking change" is that TS will throw an error if it was in a project's config

src/screenshot/connector-base.ts
```diff
@@ -42,8 +42,8 @@ export class ScreenshotConnector implements d.ScreenshotConnector {
     this.buildAuthor = opts.buildAuthor;
     this.buildUrl = opts.buildUrl;
     this.previewUrl = opts.previewUrl;
-    (this.buildTimestamp = typeof opts.buildTimestamp === 'number' ? opts.buildTimestamp : Date.now()),
-      (this.cacheDir = opts.cacheDir);
+    this.buildTimestamp = typeof opts.buildTimestamp === 'number' ? opts.buildTimestamp : Date.now();
+    this.cacheDir = opts.cacheDir;
```
tanner-reits, line 46: This is just fixing formatting while I noticed it.

tanner-reits (commented):
Noting a few things. 

I'm going to open a PR for the docs to add the `cacheDir` property.

@rwaskiewicz I just ended up creating a new branch and cherry-picking commits from the branch you created. Git history got whacky when I tried to rebase and point the other branch over to v4. If that's okay, we can close the other branch.

src/declarations/stencil-public-compiler.ts
alicewriteswrongs: I'm down 👍

src/compiler/cache.ts
alicewriteswrongs: makes sense 👍

alicewriteswrongs (approved):
tried it out locally and I'm seeing a build-time reduction in framework! so looks good to me

src/declarations/stencil-public-compiler.ts
rwaskiewicz: I'm OK with it, so long as we can get everything documented/landed by Monday

rwaskiewicz (approved):

Tallyb:
are you sure this is actually enabled in 4.0?

tanner-reits:
Hi @Tallyb 👋 

Yes, we re-enabled this for v4. Just did a quick double-check and testing and was seeing the build increase we're expecting. For now, we only expect to see ~10% build-time increase, so, depending on the size of your Stencil project, might not be all that noticeable. Caching improvements are an area the team would like to investigate in the future to hopefully improve that number!

If you are experiencing problems or unexpected behavior, could you please open an issue and provide a reproduction case so we can take a look?

Thanks!

Issue #4489, opened 2023-06-15
bug: MediaQueryList.addEventListener('change') stopped working in tests in 3.2.1


asik:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

3.2.1

### Current Behavior

Repro is simple, just `yarn create stencil` and add this code to `my-component.tsx`:
```ts
  #mediaQueryList: MediaQueryList = null;

  componentWillRender() {
    this.#mediaQueryList = window.matchMedia(`(max-width: 4px)`);
    this.#mediaQueryList.addEventListener('change', () => {});
  }
```

Then run `yarn test`.

You get this error:

```
 FAIL  src/components/my-component/my-component.spec.ts
  ● my-component › renders

    TypeError: Cannot create property '__listeners' on string 'change'

      31 |   componentWillRender() {
      32 |     this.#mediaQueryList = window.matchMedia(`(max-width: 4px)`);
    > 33 |     this.#mediaQueryList.addEventListener('change', () => {});
         |                          ^
      34 |   }
      35 |
      36 |   render() {

      at Object.addEventListener (node_modules/@stencil/core/mock-doc/index.cjs:685:24)
      at MyComponent.componentWillRender [as __componentWillRender] (src/components/my-component/my-component.tsx:33:26)
      at MyComponent.e.componentWillRender (node_modules/@stencil/core/testing/index.js:4255:20)
      at safeCall (node_modules/@stencil/core/internal/testing/index.js:491:14)
      at node_modules/@stencil/core/internal/testing/index.js:428:56
      at enqueue (node_modules/@stencil/core/internal/testing/index.js:430:52)
      at dispatchHooks (node_modules/@stencil/core/internal/testing/index.js:428:38)
      at a (node_modules/@stencil/core/internal/testing/index.js:419:18)
      at a (node_modules/@stencil/core/internal/testing/index.js:30:17)
```

This started happening in 3.2.1. 
Someone noticed this before: https://github.com/ionic-team/stencil/issues/4193#issuecomment-1521019769
A workaround is to use the `onchange` function instead.

### Expected Behavior

The tests should pass

[...]

PR #4509, opened 2023-06-26
fix(mock-doc): adjust matchMedia return

bdriguesdev:
Fixes https://github.com/ionic-team/stencil/issues/4489

The **matchMedia** mock is returning the functions from [events.ts](https://github.com/ionic-team/stencil/blob/main/src/mock-doc/event.ts) file:
- [addEventListener](https://github.com/ionic-team/stencil/blob/main/src/mock-doc/event.ts)
- [removeListener](https://github.com/ionic-team/stencil/blob/main/src/mock-doc/event.ts#L162)
- [dispatchEvent](https://github.com/ionic-team/stencil/blob/main/src/mock-doc/event.ts#L210)
These functions aren't supposed to be a "direct mock" of the methods used for handling element events. You can see that the expected params from these functions aren't the same as expected from `MediaQueryList.addEventListener`, because they aren't the same.
We could use the [addEventListener ](https://github.com/ionic-team/stencil/blob/main/src/mock-doc/window.ts#L95) from MockWindow, but I think that isn't what we want here, I think we just need to prevent the tests from breaking instead of implementing the matchMedia functionally. With that in mind, I just returned empty functions and also add [addListener](https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/addListener) and [removeListener](https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/removeListener), which are deprecated methods returned by matchMedia.

GitHub Issue Number: N/A

## What is the new behavior?

- 

- [ ] Yes
- [ ] No


rwaskiewicz (approved):
Thanks!

PR #4520, opened 2023-06-27
chore(repo): minimize component starter test pack

rwaskiewicz:

GitHub Issue Number: N/A

There was an issue that was encountered after landing https://github.com/ionic-team/stencil/commit/3811a4c4714f9fb0f88dc1f798959196721548e5 where Stencil would fail to build a starter project using local/nightly builds. In the interest of time, we ended up reverting the commit in https://github.com/ionic-team/stencil/pull/4415 .

This should not have passed the CI workflow [“Test Component Starter”](https://github.com/ionic-team/stencil/blob/7515c4b83659faee0c856f7877bde8fb564391f5/.github/workflows/test-component-starter.yml#L1).  This PR seeks to fix the detection of the error that we originally missed

## What is the new behavior?

this commit minimizes the number of files included in `npm pack` when
running the stencil component starter smoke tests. previously, files
would get added to the binary that would either be unnecessary, or would
not be representative of an actual stencil release. these include:

- `build/`, an artifact of building stencil itself. this isn't needed at
  all, and has been removed from the build archive
- `node_modules/` would get implicitly linked by `npm pack`, which get
  deleted now prior to running the command

a side effect of this work is that we now compartmentalize the building
of the stencil tarball (as opposed to generating a dev build for every
run of ci and pushing that to a registry). this makes the chances of
accidentally picking the wrong file when running `npm pack` and renaming
it lower.

- [ ] Yes
- [x] No

## Testing

I reapplied https://github.com/ionic-team/stencil/commit/3811a4c4714f9fb0f88dc1f798959196721548e5 and ran the tests against my CI fix. CI is now able to [detect the error correctly](https://github.com/ionic-team/stencil/actions/runs/5394561664/jobs/9795976750):
Screenshot 2023-06-27 at 5 07 09 PM

Reverting the commit makes CI green again

## Other information

STENCIL-850 Investigate Component Starter Issues in CI
STENCIL-688 Revisit Logic for Component Starter Tests


.github/workflows/actions/download-archive/action.yml
```diff
@@ -16,5 +16,5 @@ runs:
         path: ${{ inputs.path }}
 
     - name: Extract Archive
-      run: unzip -q -o ${{ inputs.path }}/${{ inputs.filename }}
+      run: unzip -q -o ${{ inputs.path }}/${{ inputs.filename }} -d ${{ inputs.path }}
```
rwaskiewicz, line 19: This is kind of funky. We were always using the 'path' for the location of the tarball, rather than the location of where to extract things to!

.github/workflows/test-component-starter.yml
```diff
@@ -27,21 +27,48 @@ jobs:
           node-version: ${{ matrix.node }}
           cache: 'npm'
 
+      - name: Create Pack Directory
```
rwaskiewicz, line 30: I tried to comment the config notes that I thought were worthwhile someone who was wrenching on this would want. LMK if I can clarify anything!

alicewriteswrongs (approved):

PR #4521, opened 2023-06-27
fix(compiler): address when a home module cannot be found

alicewriteswrongs:

A bit of the code added in #4212 incorrectly assumed a value wouldn't be falsy

- [ ] Yes
- [x] No

## Testing

You should be able to build and install this in framework and then run `npm run test.spec` without errors. additionally, the output from the `docs-json` output target in `docs/core.json` shouldn't change.

### DIFF
--- a/src/compiler/transformers/test/parse-props.spec.ts
+++ b/src/compiler/transformers/test/parse-props.spec.ts
@@ -34,6 +34,39 @@ describe('parse props', () => {
     expect(t.cmp?.hasProp).toBe(true);
   });
 
+  it('should correctly parse a prop with an unresolved type', () => {
+    const t = transpileModule(`
+    @Component({tag: 'cmp-a'})
+      export class CmpA {
+        @Prop() val?: Foo;
+      }
+    `);
+    expect(getStaticGetter(t.outputText, 'properties')).toEqual({
+      val: {
+        attribute: 'val',
+        complexType: {
+          references: {
+            Foo: {
+              id: 'global::Foo',
+              location: 'global',
+            },
+          },
+          resolved: 'Foo',
+          original: 'Foo',
+        },
+        docs: {
+          text: '',
+          tags: [],
+        },
+        mutable: false,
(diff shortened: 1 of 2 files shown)


src/compiler/transformers/transform-utils.ts
```diff
@@ -525,12 +525,14 @@ const getTypeReferenceLocation = (
...
-      id,
-    };
+    if (importHomeModule) {
+      const id = addToLibrary(type, originalTypeName, checker, normalizePath(importHomeModule.fileName, false));
+      return {
+        location: 'import',
+        path: localImportPath,
+        id,
+      };
+    }
```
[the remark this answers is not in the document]
alicewriteswrongs: no, since there's still the code below which deals with the case that this is a global type and so on

alicewriteswrongs:
@rwaskiewicz I added a test to the spec file for the prop decorator handling code which serves as a regression test for this

rwaskiewicz (approved):

src/compiler/transformers/test/parse-props.spec.ts
```diff
@@ -34,6 +34,39 @@ describe('parse props', () => {
...
+      }
+    `);
+    expect(getStaticGetter(t.outputText, 'properties')).toEqual({
+      val: {
+        attribute: 'val',
+        complexType: {
+          references: {
+            Foo: {
+              id: 'global::Foo',
+              location: 'global',
```
tanner-reits, line 51: So TS thinks any type it can't resolve is just a global? Or is this something we're doing?

tanner-reits (approved):
Think this looks good. One question for my own knowledge

src/compiler/transformers/test/parse-props.spec.ts
alicewriteswrongs: yeah basically, if a type isn't imported into the module or declared inside of it we can safely assume it's a global type

PR #4522, opened 2023-06-27
fix(compiler): normalize recommended `collection` path for `package.json` validation

tanner-reits:

The `package.json` validation for the `collection` field does not use a normalized path for the comparison. This can result in a warning message to set the field to the value that already exists (can be seen by using the stock Stencil component starter)

GitHub Issue Number: N/A

## What is the new behavior?

The recommended path gets normalized for the string comparison of the existing path and recommended path.

- [ ] Yes
- [X] No

## Testing

Tested using the component starter.
1. Create a new Stencil project, run `npm i` and `npm run build`. Notice the warning message related to the `collection` field recommends setting the value that already exists in the `package.json`
2. Build and pack this branch. Install in the starter project
3. Run `npm run build` and notice the warning message no longer appears.
4. You can also change the `collection` field in the `package.json` to a relative path like `./dist/collection/collection-manifest.json` and get no warning.
5. Changing the `collection` field to something like `collection/collection-manifest.json` will correctly throw a warning.

### DIFF
--- a/src/compiler/types/validate-build-package-json.ts
+++ b/src/compiler/types/validate-build-package-json.ts
@@ -149,11 +149,11 @@ export const validateCollection = (
   outputTarget: d.OutputTargetDistCollection
 ) => {
   if (outputTarget.collectionDir) {
-    const collectionRel = join(relative(config.rootDir, outputTarget.collectionDir), COLLECTION_MANIFEST_FILE_NAME);
-    if (
-      !buildCtx.packageJson.collection ||
-      normalizePath(buildCtx.packageJson.collection) !== normalizePath(collectionRel)
-    ) {
+    const collectionRel = normalizePath(
+      join(relative(config.rootDir, outputTarget.collectionDir), COLLECTION_MANIFEST_FILE_NAME),
+      false
+    );
+    if (!buildCtx.packageJson.collection || normalizePath(buildCtx.packageJson.collection, false) !== collectionRel) {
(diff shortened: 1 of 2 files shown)


alicewriteswrongs (approved):

src/compiler/types/validate-build-package-json.ts
```diff
@@ -150,7 +150,10 @@ export const validateCollection = (
 ) => {
   if (outputTarget.collectionDir) {
     const collectionRel = join(relative(config.rootDir, outputTarget.collectionDir), COLLECTION_MANIFEST_FILE_NAME);
-    if (!buildCtx.packageJson.collection || normalizePath(buildCtx.packageJson.collection) !== collectionRel) {
+    if (
+      !buildCtx.packageJson.collection ||
+      normalizePath(buildCtx.packageJson.collection) !== normalizePath(collectionRel)
```
rwaskiewicz, line 155: Can we add a unit test for this conditional? It looks like we have some testing of this function in:
https://github.com/ionic-team/stencil/blob/2d5d98e4f028a9a7dcb8001e2defd8a13f30d7d3/src/compiler/types/tests/validate-package-json.spec.ts#L101-L106

rwaskiewicz (changes requested):
Looks good! One small testing ask

rwaskiewicz (approved):
👍

rwaskiewicz:
@tanner-reits looks like you need to normalize the path in the jest tests as well - it's failing only for Windows 🙃

PR #4524, opened 2023-06-27
refactor(telemetry): remove browser-specific code from telemetry

alicewriteswrongs:

GitHub Issue Number: N/A

## What is the new behavior?

- 

- [ ] Yes
- [ ] No

### DIFF
--- a/src/cli/telemetry/telemetry.ts
+++ b/src/cli/telemetry/telemetry.ts
@@ -1,6 +1,5 @@
 import { isOutputTargetHydrate, WWW } from '@utils';
 
-import { IS_BROWSER_ENV } from '../../compiler/sys/environment';
 import type * as d from '../../declarations';
 import { readConfig, updateConfig, writeConfig } from '../ionic-config';
 import { CoreCompiler } from '../load-compiler';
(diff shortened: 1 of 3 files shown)


src/cli/telemetry/telemetry.ts
```diff
@@ -155,7 +153,7 @@ export const prepareData = async (
     cpu_model,
     duration_ms,
     has_app_pwa_config,
-    is_browser_env,
+    is_browser_env: false,
```
alicewriteswrongs, line 156: from this point on hard-coding this to `false` is going to be correct, but I did have the thought that we may want to remove the value entirely too

alicewriteswrongs (commented):
just noting one thing

src/cli/telemetry/telemetry.ts
rwaskiewicz: I don't think anything catastrophic will happen from a backend perspective if we do. I may get a warning of some kind - better to see it now I suppose! I vote in favor of removal so long as it doesn't cause a "breaking change" in the return type that's public facing

alicewriteswrongs: cool, went ahead and removed it. the relevant type is in `stencil-private.ts` so I think we're good on that front:

https://github.com/ionic-team/stencil/blob/17973029d915ca11bdfe7a36891b42f68f057e07/src/declarations/stencil-private.ts#L2633-L2656

tanner-reits (approved):

rwaskiewicz (approved):

PR #4527, opened 2023-06-29
fix(compiler): sourcemap errors for dist-custom-elements + dist-hydrate-script

rwaskiewicz:
## What is the current behavior?

see 'new behavior' and https://github.com/ionic-team/stencil/issues/4260

GitHub Issue Number: Fixes: #4260

## What is the new behavior?

this commit fixes an issue where a stencil project that use both the `dist-custom-elements` and `dist-hydrate-script` output targets would see the following error when building their project (with sourcemaps enabled):

```
[ WARN  ]  Bundling Warning SOURCEMAP_ERROR
           Error when using sourcemap for reporting an error: Can't resolve original location of error.
```

this commit resolves this error by ensuring the `HTMLElement` core runtime API used by `dist-custom-elements` no longer interferes with the generated code for `dist-hydrate-script`. this is accomplished by using the output-target-specific runtime api slots introduced in #4200. this prevents the `HTMLElement` import from polluting the sourcemaps/source used in the `dist-hydrate-script` target

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Testing

Unit tests have been updated to account for the change in the order of imports.

Pull down the reproduction case at https://github.com/rwaskiewicz/stencil-dist-hydrate-script-sourcemap-warn
1. `npm i @stencil/core@latest`
2. `npm run build`
3. Observe the following output:
```
> dist-hydrate-script-sourcemap-warn@0.0.1 build
> stencil build --docs

[58:06.7]  @stencil/core
[58:06.8]  [CURRENT_VERSION}
[58:07.6]  build, dist-hydrate-script-sourcemap-warn, prod mode, started ...
[58:07.6]  transpile started ...
[58:08.6]  transpile finished in 938 ms
[58:08.6]  generate custom elements + source maps started ...
[58:08.6]  generate hydrate app started ...
[58:08.9]  generate custom elements + source maps finished in 377 ms
[58:08.9]  generate hydrate app finished in 378 ms

[ WARN  ]  Bundling Warning SOURCEMAP_ERROR
           Error when using sourcemap for reporting an error: Can't resolve original location of error.

[58:08.9]  build finished in 1.33 s
```
4. Install the dev build: `npm i @stencil/core@4.0.1-dev.1688045194.5d2dda3`
5. `npm run build`
6. Observe that the issue no longer occurs
7. I also compared the outputs of `dist-hydrate` and `dist-custom-elements`, and found no meaningful differences between the dev build and v4.0.1

PR #4529, opened 2023-06-29
fix(compiler): sourcemap generation without ext runtime

rwaskiewicz:

See 'new behavior' section

GitHub Issue Number: N/A

## What is the new behavior?

this commit fixes a bug where projects using the `dist-custom-elements`
output target with `externalRuntime: false` would receive the following
error when building their project:
```
[ WARN  ]  Bundling Warning SOURCEMAP_ERROR
           Error when using sourcemap for reporting an error: Can't resolve original location of error.
```

the cause of this error was attempting to import a function,
`attachShadow` that is no longer exported from the runtime bundle (as of
https://github.com/ionic-team/stencil/pull/3117). to date, this has not
had an effect on stencil (as the import gets treeshaken away). however,
when trying to generate sourcemaps for a project using this
configuration would cause a mismatch between what was expected to be in
the produced output (the import statement) and what was really there (no
import statement)

this commit removes the RUNTIME_APIS.ATTACH_SHADOW field that is no
longer used in the codebase. it's only usage was in tests for adding
runtime apis, and has been replaced with another field from
`RUNTIME_APIS`

- [ ] Yes
- [x] No

## Testing

Tests for delegates-focus (unit and karma) continue to pass
## Other information

Relies on #4527 to land first

### DIFF
--- a/src/compiler/transformers/component-native/native-constructor.ts
+++ b/src/compiler/transformers/component-native/native-constructor.ts
@@ -1,8 +1,6 @@
-import { DIST_CUSTOM_ELEMENTS } from '@utils';
 import ts from 'typescript';
 
 import type * as d from '../../../declarations';
-import { addOutputTargetCoreRuntimeApi, RUNTIME_APIS } from '../core-runtime-apis';
 import { addCreateEvents } from '../create-event';
 import { retrieveTsModifiers } from '../transform-utils';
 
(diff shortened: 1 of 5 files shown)


tanner-reits (approved):
Tested on a component starter and confirmed the fix 👍

alicewriteswrongs (approved):

Issue #3354, opened 2022-05-04
Typescript interface `SVGAttributes` is not correct declared for attribute `color-interpolation-filters`


fbrandel:
**Stencil version:**
```
 @stencil/core@2.15.1
```

**Current behavior:**

Typescript interface `SVGAttributes` is not correct declared for attribute `color-interpolation-filters`. Contains values "s-rGB" and "linear-rGB" instead.

```
<filter color-interpolation-filters="sRGB" .../>

[ ERROR ]  TypeScript: ./src/components/...
           Type '{ id: string; x: string; y: string; width: string; height: string; filterUnits: string;
           "color-interpolation-filters": "sRGB"; }' is not assignable to type 'SVGAttributes<SVGElement>'.Types of
           property ''color-interpolation-filters'' are incompatible.Type '"sRGB"' is not assignable to type '"auto" |
           "inherit" | "s-rGB" | "linear-rGB"'. Did you mean '"s-rGB"'?
```

**Expected behavior:**
Attribute `color-interpolation-filters` is typed with correct values "sRGB" and "linearRGB"

rwaskiewicz:
I gave this a second read through, and it appears this is concerning the Mock Doc. Sorry about that! I'm going to label this to be ingested into the team's backlog

rwaskiewicz:
The fix for this issue has been released as a part of today's [Stencil v4.0.2 release](https://github.com/ionic-team/stencil/releases/tag/v4.0.2)

PR #4530, opened 2023-06-29
fix(runtime): properly type color-interpolation-filter

rwaskiewicz:

Stencil improperly types `'color-interpolation-filter'` attribute values. It causes TypeScript to error when 'sRGB' or 'linearRGB' are provided, both of which are valid.

GitHub Issue Number: Fixes: https://github.com/ionic-team/stencil/issues/3354

## What is the new behavior?

this commit updates the typings of
`SVGAttributes#color-interpolation-filter` to match the current spec, which states the only accepted values are 'auto', 'sRGB' and 'linearRGB'.

this commit renames 's-rGB' to 'sRGB' and 'linear-rGB' to 'linearRGB'. it also removes 'inherit'.

it is believed this was done by accident in a regex-based replacement of camelCased properties to kebab-cased attributes in 667597dd

- [ ] Yes
- [x] No - although folks using this may see TS errors upon upgrading

## Testing

Before applying this change:
Screenshot 2023-06-29 at 5 41 26 PM
After:
Screenshot 2023-06-29 at 5 52 36 PM

### DIFF
--- a/src/declarations/stencil-public-runtime.ts
+++ b/src/declarations/stencil-public-runtime.ts
@@ -1444,7 +1444,7 @@ export namespace JSXBase {
     clipPathUnits?: number | string;
     'clip-rule'?: number | string;
     'color-interpolation'?: number | string;
-    'color-interpolation-filters'?: 'auto' | 's-rGB' | 'linear-rGB' | 'inherit';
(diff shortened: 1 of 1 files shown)


tanner-reits (approved):
Double-checked the docs/spec for the attribute to make sure everything matches up 👍

alicewriteswrongs (approved):

Issue #4543, opened 2023-07-04
bug: typescript imports using windows paths


mrginglymus:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

4.0.1

### Current Behavior

On Windows, the `dist-custom-elements` target generates a `dist/components/index.d.ts` with some unescaped windows paths:

```ts
/* stencil custom elements */
export { Icon as VIcon } from '..\types\components\icons\icon';
export { defineCustomElement as defineCustomElementVIcon } from './v-icon';
```

### Expected Behavior

```ts
/* stencil custom elements */
export { Icon as VIcon } from '../types/components/icons/icon';
export { defineCustomElement as defineCustomElementVIcon } from './v-icon';
```

### System Info

```shell
System: node 16.20.0
    Platform: windows (10.0.19045)
   CPU Model: Intel(R) Xeon(R) Gold 5220 CPU @ 2.20GHz (8 cpus)
    Compiler: ...\playground\node_modules\.store\@stencil-core-npm-4.0.1-81248f31a7\node_modules\@stencil\core\compiler\stencil.js
       Build: 1687953951
     Stencil: 4.0.1
  TypeScript: 5.0.4
      Rollup: 2.42.3
      Parse5: 7.1.2
      Sizzle: 2.42.3
      Terser: 5.18.1
```

### Steps to Reproduce

On Windows, the `dist-custom-elements` target generates a `dist/components/index.d.ts` with some unescaped windows paths

### Code Reproduction URL

n/a

alicewriteswrongs:
@mrginglymus unfortunately I don't have a windows machine handy for testing so I have to rely on our windows CI but I think I have a PR up (#4545) which will address the issue somewhat comprehensively. Is there any chance you could try running a dev build of Stencil to see if it fixes the issue for you?

If so, you can install it like this:

```sh
npm install --save-dev @stencil/core@4.0.1-dev.1688586331.51327eb
```

mrginglymus:
Ah, that looks amazing, thanks so much for the quick and comprehensive response. I'll try and test this afternoon.

[...]

PR #4544, opened 2023-07-04
Use posix paths for type imports

mrginglymus:
Stencil 4.0.1 on Windows generates type imports with Windows paths instead of posix paths.

GitHub Issue Number: #4543

## What is the new behavior?
Type imports now use posix paths

- [ ] Yes
- [X] No

## Testing

Updated tests to hard-code expected output, decoupling them from the running platform.


mrginglymus:
Closing in favour of https://github.com/ionic-team/stencil/pull/4545

Issue #4528, opened 2023-06-29
bug: Element decorator does not work when targeting ES2022 and setting useDefineForClassFields to true


henning96:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

4.0.1

### Current Behavior

When targeting ES2022 and having useDefineForClassFields set to truthy, Stencil generates an additional class field for properties annotated with Element.
This results in both a class field and a getter for the property.
Stencil should only generate the getter.
This was already tackled for State and Prop decorated properties but not for Element: https://github.com/ionic-team/stencil/pull/3614.

### Expected Behavior

Stencil should handle the Element decorator in the same way than the other decorators such as State and Prop which are listed here: https://github.com/ionic-team/stencil/blob/main/src/compiler/transformers/decorators-to-static/decorators-constants.ts#L17.
There shouldn't be an additional class field being generated next to the getter.

### System Info

```shell
System: node 18.16.0
    Platform: darwin (22.5.0)
   CPU Model: Apple M1 (8 cpus)
    Compiler: /Users/henning/Projects/test-app/node_modules/@stencil/core/compiler/stencil.js
       Build: 1687953951
     Stencil: 4.0.1 ⛹
  TypeScript: 5.0.4
      Rollup: 2.42.3
      Parse5: 7.1.2
      Sizzle: 2.42.3
      Terser: 5.18.1
```

### Steps to Reproduce

To reproduce the bug, set the Typescript target to ES2022 and set the useDefineForClassFields setting to true.
The problem is actually pretty much the same than what was already tacked in this Pull Request: https://github.com/ionic-team/stencil/pull/3614.
But for some reason, it doesn't consider the Element decorator. When making use of the Element decorator in a Stencil component while having the mentioned TypeScript settings, Stencil generates both a class field for the Element and a getter. The Browser only evaluates the class field, so that the annotated property is always undefined during runtime.

It starts to work fine when the list of decorators to rewrite is extended with Element. I mean this constant here: https://github.com/ionic-team/stencil/blob/main/src/compiler/transformers/decorators-to-static/decorators-constants.ts#L17.

[...]

PR #4547, opened 2023-07-06
fix(compiler): only create one class member when transforming `@Element()` decorators

tanner-reits:

Use of the `@Element()` decorator when targeting ES2023 will result in two class members being created on the class declaration: a getter and a class field.

GitHub Issue Number: Fixes: #4528 

## What is the new behavior?

With this change, Stencil will only create the getter for grabbing the element reference. This change was made to both the lazy and native transformers.

- [ ] Yes
- [X] No

## Testing

Unit & e2e tests continue to pass.

Manual testing can be done for the lazy build using the reproduction in the associated issue and a local build of this branch.
Testing of the native build was done using a local lerna project with a Stencil component library and consuming application that leveraged the `dist-custom-elements` output.

## Other information

I also made sure to test this change in a component that also used at least one other decorator that gets transformed (`@State()` or `@Prop()`)

### DIFF
--- a/src/compiler/transformers/component-lazy/lazy-element-getter.ts
+++ b/src/compiler/transformers/component-lazy/lazy-element-getter.ts
@@ -14,20 +14,35 @@ export const addLazyElementGetter = (
   if (cmp.elementRef) {
     addCoreRuntimeApi(moduleFile, RUNTIME_APIS.getElement);
 
-    classMembers.push(
-      ts.factory.createGetAccessorDeclaration(
-        undefined,
-        cmp.elementRef,
-        [],
-        undefined,
-        ts.factory.createBlock([
-          ts.factory.createReturnStatement(
-            ts.factory.createCallExpression(ts.factory.createIdentifier(GET_ELEMENT), undefined, [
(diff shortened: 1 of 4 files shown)


src/compiler/transformers/component-lazy/lazy-element-getter.ts
tanner-reits, line 1: I didn't abstract the changes to a shared space since it was a small amount of code and I could see there maybe being future cases where these two implementations deviate from each other and we'd have to deal with unnecessary layers of abstraction then

alicewriteswrongs (approved):
tried out the fix locally and it looks good to me and everything makes sense and whatnot

alicewriteswrongs:
oh one additional question - would it be feasible to throw together a quick test? just maybe to serve as a regression test in the future. if it's a pain to get testing together for it no worries though

tanner-reits:
@alicewriteswrongs Good idea! Added a quick regression test for each transformer

alicewriteswrongs (approved):
💯 :shipit:

rwaskiewicz (approved):

Issue #3052, opened 2021-09-07
feat: Allow Host attributes to be overwritable prior to componentWillLoad


splitinfinities:
### Prequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/master/.github/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/master/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already include this feature request, without success.

### Describe the Feature Request

Host, when setting values on attributes within a render function - with the exception of class - will toss a value that's set on the html element prior to the lifecycle. Ideally, we can have some mechanism to allow all attributes to be overwritten - or a select few attributes.

### Describe the Use Case

Consider this naive render function: 

```jsx
<Host role="header">...</Host>
```

Right now, if users did:

```html
<ion-header role="somethingelse"></ion-header>
```
 that would not work and the `somethingelse` value would be lost. 

### Describe Preferred Solution

```tsx
@Component({
  host: { 
    mutableAttributes: true // or ["role"] 
  }
})
```

Which will allow attributes the element has set to be read and overwritten prior to when the lifecycle is called. 

### Describe Alternatives

```tsx
import {Host} from "@stencil/core"

Host.mutableAttributes = ["role"] // or true

export class Component {
  render() {
    return <Host role="header" />
  }
}
```

or

```tsx
import {Host} from "@stencil/core"

export class Component {
  render() {
    Host.inherit = ["role"];
    return <Host role="header" />
  }
}
```

### Additional Information

(Via @willmartian) 

For context, this is what we do right now: https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/input/input.tsx#L237

And then this: https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/input/input.tsx#L439

tanner-reits:
This was resolved in [Stencil v4.0.3](https://github.com/ionic-team/stencil/releases/tag/v4.0.3). Please re-open this issue if you are still experiencing issues after updating.

PR #4548, opened 2023-07-06
fix handling of host attributes

alicewriteswrongs:

GitHub Issue Number: N/A

## What is the new behavior?

This adds a runtime check for attributes set on the host element instance ('from outside') before the attributes set on the `Host` component ('from inside') are set on it, to allow developers to, for instance, override a `role` attribute set on `Host` (see #3052 for instance)

- [ ] Yes
- [x] No

## Testing

I have a reproduction case up here: https://github.com/alicewriteswrongs/stencil-host-override

If you build that project with this branch of Stencil you should see that the `role` set on the custom element in HTML wins out over the value set on the `Host` element.

### DIFF
--- a/src/declarations/stencil-private.ts
+++ b/src/declarations/stencil-private.ts
@@ -1261,6 +1261,10 @@ export interface EventEmitterData<T = any> {
   composed?: boolean;
 }
 
+/**
+ * An interface extending `HTMLElement` which describes the fields added onto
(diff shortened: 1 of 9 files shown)


tanner-reits (approved):
All looks good to me based on team conversation yesterday!

rwaskiewicz (approved):

PR #4551, opened 2023-07-07
refactor(environment): remove IS_NODE_ENV and delete dead code

alicewriteswrongs:
This removes the `IS_NODE_ENV` variable from our environment module which previously was used to switch between browser-specific and node-specific approaches to various things in the Stencil compilation process.

Following the merge of #4317 we only support one runtime, node, for compiling Stencil components, so we can assume that `IS_NODE_ENV` is `true` and delete code accordingly.

GitHub Issue Number: N/A

## What is the new behavior?

- 

- [ ] Yes
- [ ] No


tanner-reits (approved):
All looks good to me!

rwaskiewicz (approved):

PR #4569, opened 2023-07-12
refactor(compiler): propagate `ValidatedConfig`

alicewriteswrongs:
This propagates the `ValidatedConfig` in place of the `Config`, branching out from `src/compiler/transpile.ts`

Not enough `ValidatedConfig`

## What is the new behavior?

A bit more `ValidatedConfig`

- [ ] Yes
- [x] No

## Testing

I don't think this should affect anything, and should really only give us more safety, so just make sure that things are working as they should be!

### DIFF
--- a/src/compiler/transpile.ts
+++ b/src/compiler/transpile.ts
@@ -1,14 +1,15 @@
 import rollupPluginUtils from '@rollup/pluginutils';
 import type {
-  Config,
   TransformCssToEsmInput,
   TransformOptions,
   TranspileOptions,
   TranspileResults,
+  ValidatedConfig,
 } from '@stencil/core/internal';
 import { catchError, getInlineSourceMappingUrlLinker, isString } from '@utils';
 
 import { getTranspileConfig, getTranspileCssConfig, getTranspileResults } from './config/transpile-options';
+import { validateConfig } from './config/validate-config';
 import { transformCssToEsm, transformCssToEsmSync } from './style/css-to-esm';
 import { patchTypescript } from './sys/typescript/typescript-sys';
 import { getPublicCompilerMeta } from './transformers/add-component-meta-static';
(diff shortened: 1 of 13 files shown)


tanner-reits (approved):

src/compiler/transpile.ts
```diff
@@ -40,8 +41,9 @@ export const transpile = async (code: string, opts: TranspileOptions = {}): Prom
   try {
     if (shouldTranspileModule(results.inputFileExtension)) {
       const { config, compileOpts, transformOpts } = getTranspileConfig(opts);
-      patchTypescript(config, null);
-      transpileCode(config, compileOpts, transformOpts, results);
+      const validatedConfig = validateConfig(config, {}).config;
+      patchTypescript(validatedConfig, null);
+      transpileCode(validatedConfig, compileOpts, transformOpts, results);
```
rwaskiewicz, line 46: I'm hesitant we can safely make this change without it being potentially breaking, since this function is a part of our public Node API. By creating a `ValidatedConfig`, we end up providing a config object with additional properties on it that folks may be (un)intentionally not providing/relying upon.

src/compiler/transpile.ts
```diff
@@ -72,8 +74,9 @@ export const transpileSync = (code: string, opts: TranspileOptions = {}): Transp
...
```
rwaskiewicz, line 79: I'm hesitant we can safely make this change without it being potentially breaking, since this function is a part of our public Node API + used by Jest testing of Stencil projects. By creating a `ValidatedConfig`, we end up providing a config object with additional properties on it that folks may be (un)intentionally not providing/relying upon.

rwaskiewicz (changes requested):
I'm a little hesitant about a few of these changes, as they could be potentially breaking. Maybe we should save a portion of these changes for v5 - thoughts?

src/compiler/transpile.ts
alicewriteswrongs: Yeah I think it's conceivable that someone is depending on the implementation details of this but to me that doesn't fall within the bounds of what breaking changes are - I think it makes sense to guarantee that the function type signatures are not going to change, that the basic things they do will be stable, they won't be removed, etc, but implementation details can be changed without notice I think (that's kind of the point of a stable interface right? the stuff behind it can then vary). We could do it in a minor version if we wanted just to signal it's a slightly bigger change, but I personally am not worried about the impact, especially since I feel pretty confident that our config validation setup does "the right thing"

There's other changes we make where we ship them, someone finds a small bug or inconsistency, and then we make a fix and cut a quick point release. I think if we did get a report that this change broke someone's workflow we could evaluate whether we need to do something like that

PR #4572, opened 2023-07-13
fix(compiler): handle `@supports` blocks when scoping css

tanner-reits:

Builds for the `dist-hydrate-script` will fail with a PostCSS error when a component's styles contain an `@supports` block with a pseudo-selector:

```css
@supports selector(:host()) {
  :host() {
    color: red;
  }
}
```

GitHub Issue Number: N/A

## What is the new behavior?

Updates the Stencil [scopeCss()](https://github.com/ionic-team/stencil/blob/main/src/utils/shadow-css.ts#L418) method to not transform certain pseudo-selectors (`:host`, `:host-context`, `::slotted`) when they are a part of the CSS selector in an `@supports` block.

With this change, the block defined in the "current behavior" section will transform to something like the following:

```css
@supports selector(:host()) {
  .sc-my-cmp-h {
    color: red;
  }
}
```

- [ ] Yes
- [X] No

## Testing

**Automated Tests**
Unit tests continue to pass and new tests were added for the `@supports` cases.

**Manual Tests**
1. Clone [this](https://github.com/tanner-reits/stencil-prerender) repo and switch to [this](https://github.com/tanner-reits/stencil-prerender/tree/postcss-supports-issue) branch
2. From root, run `npx lerna bootstrap` to install dependencies
3. From root, run `cd packages/stencil-library && npm run build`
4. Observe build errors
5. From root, run `npm i ../../stencil-core-4.0.1.tgz -w stencil-library` to install the local build of this PR
6. From root, run `cd packages/stencil-library && npm run build`
7. Observe no build errors

You can also spin-up the node app in the repro to see the CSS getting applied to a rendered web component. For more information on making changes to the repro, see its [README](https://github.com/tanner-reits/stencil-prerender/tree/postcss-supports-issue#readme)

### DIFF
--- a/src/utils/shadow-css.ts
+++ b/src/utils/shadow-css.ts
@@ -80,7 +80,8 @@ const _polyfillHostRe = /-shadowcsshost/gim;
  * @param selector The CSS selector we want to match for replacement
  * @returns A look-behind regex containing the selector
  */
-const createSupportsRuleRe = (selector: string) => new RegExp(`(?<!(^@supports(.*)))(${selector}\\b)`, 'gim');
+const createSupportsRuleRe = (selector: string) =>
+  new RegExp(`((?<!(^@supports(.*)))|(?<=\{.*))(${selector}\\b)`, 'gim');
 const _colonSlottedRe = createSupportsRuleRe('::slotted');
 const _colonHostRe = createSupportsRuleRe(':host');
 const _colonHostContextRe = createSupportsRuleRe(':host-context');
--- a/src/utils/shadow-css.ts
+++ b/src/utils/shadow-css.ts
@@ -72,9 +72,18 @@ const _shadowDOMSelectorsRe = [/::shadow/g, /::content/g];
 
 const _selectorReSuffix = '([>\\s~+[.,{:][\\s\\S]*)?$';
 const _polyfillHostRe = /-shadowcsshost/gim;
-const _colonHostRe = /:host/gim;
-const _colonSlottedRe = /::slotted/gim;
-const _colonHostContextRe = /:host-context/gim;
+
+/**
+ * Little helper for generating a regex that will match a specified
+ * CSS selector when that selector is _not_ a part of a `@supports` rule.
+ *
+ * @param selector The CSS selector we want to match for replacement
+ * @returns A look-behind regex containing the selector
+ */
+const createSupportsRuleRe = (selector: string) => new RegExp(`^((?!@supports).)*${selector}\\b`, 'gim');
+const _colonSlottedRe = createSupportsRuleRe('::slotted');
+const _colonHostRe = createSupportsRuleRe(':host');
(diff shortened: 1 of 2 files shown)


src/utils/shadow-css.ts
```diff
@@ -72,9 +72,19 @@ const _shadowDOMSelectorsRe = [/::shadow/g, /::content/g];
...
 const _selectorReSuffix = '([>\\s~+[.,{:][\\s\\S]*)?$';
 const _polyfillHostRe = /-shadowcsshost/gim;
-const _colonHostRe = /:host/gim;
-const _colonSlottedRe = /::slotted/gim;
-const _colonHostContextRe = /:host-context/gim;
+
+/**
+ * Little helper for generating a regex that will match a specified
+ * CSS selector when that selector is _not_ a part of a `@supports` rule.
+ *
```
rwaskiewicz, line 79: Can we add a little more detail here as to what we're checking/how the regex works? ATM, it's not super clear to me why we need both the negative and positive look-behinds. Is the idea something like:
```suggestion
 *
 * We build a regex that checks for either: 
 * 1. A string that does not start with `@supports`, but contains the provided selector.
 *    For a provided selector of ':host',  the following will match:
 *    - ':host' returns a match of ':host'
 *    - ':HOST' returns a match of ':HOST'
 *    - '.a:host' returns a match of ':host'
 *    For a provided selector of ':host', the following will not match:
 *    - '@supports :host' will not match, as the string begins with '@supports'
 *    - ':hostname' will not match, as we require a word boundary after the selector
 * 2. A string that matches the selector immediately following the provided selector. 
 *    For a provided selector of ':host',  the following will match:
 *    - '{:host' returns a match of ':host'
 *    - '{:HOST' returns a match of ':HOST'
 *    For a provided selector of ':host', the following will not match:
 *    - '{:hostname' will not match, as we require a word boundary after the selector
 * The regex captures the provided selector
 * 
```

alicewriteswrongs (approved):
this all looks good to me, pending @rwaskiewicz's questions

src/utils/shadow-css.ts
tanner-reits: Added some more description and examples of what will and won't match

rwaskiewicz (approved):
Thanks!

toxaq:
The regular expression added here `createSupportsRuleRe` does not support older versions of iOs safari is so breaks existing apps on ios16.2 and earlier

https://github.com/ionic-team/ionic-framework/issues/30219

rwaskiewicz:
@toxaq can you please open a new issue with a minimal reproduction case? I'm not sure how well these older issues/PRs are monitored (and most of the team from this time period has moved on to other project/roles).

kensodemann:
@toxaq I created https://github.com/stenciljs/core/issues/6206 as I am working with someone else who is also running into this. Please add any comments over there if you have other use-cases or otherwise need to expand the discussion.

PR #4575, opened 2023-07-14
chore(ci): pin puppeteer for node 14 starter tests (v3)

rwaskiewicz:

Smoke tests for the component starter will fail under node 14 with the following:

```
[03:11.4]  @stencil/core
[03:12.2]  [LOCAL DEV] v3.4.1-dev.1688761098.7e0fe1b
[03:12.4]  testing e2e and spec files

[ ERROR ]  Unexpected token '??='
           D:\a\stencil\stencil\tmp-component-starter\node_modules\puppeteer-core\lib\cjs\puppeteer\common\util.js:399
```

GitHub Issue Number: N/A

## What is the new behavior?

this commit pins the version of puppeteer used by the component starter smoke tests. newer versions of puppeteer use the `??=` operator, which node 14 doesn't know how to parse. since the component starter's versioning is only expressed in git (and the cli doesn't know how to download a specific version backed by a git-sha), we pin the version of puppeteer to a known working version for node 14.

- [ ] Yes
- [x] No

## Testing

CI passes for all component starter flows 🎉 

See how we install puppeteer v19 only for Node 14:
Screenshot 2023-07-14 at 2 22 53 PM
Screenshot 2023-07-14 at 2 22 43 PM

### DIFF
--- a/.github/workflows/test-component-starter.yml
+++ b/.github/workflows/test-component-starter.yml
@@ -58,6 +58,15 @@ jobs:
         working-directory: ./tmp-component-starter
         shell: bash
 
+      # newer versions of puppeteer use the `??=` operator, which node 14 doesn't understand.
+      # puppeteer v20 dropped support for Node 14, pin a known working version of puppeteer v19 for cases where we need
+      # to land something to the v3-maintenance branch.
+      - name: Install Puppeteer (Node 14)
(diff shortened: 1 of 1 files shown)


tanner-reits (approved):

rwaskiewicz:
I'm going to merge this directly into the v3 branch, as we can't run browserstack tests w/o it.

Issue #4076, opened 2023-02-20
bug: Missing variable in hydration app


cptflammin:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/.github/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

3.0.1

### Current Behavior

While prerendering and hydration, stencil raises an error about an undefined `__DEV__` variable. This seems to be related to @apollo/client. Being able to define the variable in prerendering config would solve the issue. 

[19:35.9]  prerendering started ...
[19:36.8]  prerendering failed in 822 ms

[ ERROR ]  Hydrate Error
           ReferenceError: `__DEV__` is not defined at checkDEV
           (/Users/nico_fl/Development/_trials/stencil-prerender-notfoundvars/dist/hydrate/index.js:28002:5) at
           hydrateAppClosure

### Expected Behavior

Being able to define missing variable somewhere within app or config.
Defining var in app.ts does not work (variable is renamed with a $1 ending), index.js hydrated app:

`
const __DEV__$1 = false;
console.log('__DEV__', __DEV__$1);
`

### System Info

```shell
Mac OS Ventura
@apollo/client@3.7.8
@stencil/core@3.0.1
```

### Steps to Reproduce

stencil build --prerender

### Code Reproduction URL

https://github.com/cptflammin/stencil-prerender-notfoundvars

### Additional Information

`npm run build

> s3@0.0.1 build
> stencil build --prerender

[20:57.8]  @stencil/core
[20:58.1]  v3.0.1 🍒
[21:00.9]  build, app, prod mode, started ...
[21:01.0]  transpile started ...
[21:04.4]  transpile finished in 3.36 s
[21:04.4]  copy started ...
[21:04.4]  generate hydrate app started ...
[21:04.4]  generate lazy + source maps started ...
[21:06.6]  copy finished (1338 files) in 2.23 s
[21:23.1]  generate hydrate app finished in 18.78 s
[21:26.3]  generate lazy + source maps finished in 21.89 s

[ WARN  ]  Package Json: package.json
           package.json "module" property is required when generating a distribution.

[21:26.4]  build finished in 25.51 s

[21:30.7]  prerendering started ...
[...]

Issue #2429, opened 2020-05-11
ion-content fullscreen freezes the prerender process


HaythemSahbani:
**Stencil version:**

```
 @stencil/core@1.12.7
```

**I'm submitting a:**

[x] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**

Hi I am trying to prerender the ionic-stencil starter app. However when adding `fullscreen` to the `ion-content` element the prerender process becomes inresponsive and I have to manually kill the process.
**Expected behavior:**

The prerendering process doesn't freez
**Steps to reproduce:**

1. Run _npm init stencil_, (choose the ionic pwa)
2. in the generated home.tsx file add `fullscreen={true}` or `fullscreen` to the `ion-content`
3. Build with prerender, _npm run build -- --prerender_
**Related code:**

```tsx
// insert any relevant code here
```

**Other information:**

corysmc:
+1 I just ran into this issue myself! - repro steps outlined above

tricki:
Same here.
After some debugging I discovered  what seems to be an infinite loop that keeps rendering the ion-content indefinitely. Confirmed by logging calls to [`updateComponent`](https://github.com/ionic-team/stencil/blob/54ee75f6cb8130276a00af6fff479ea8e0741833/src/runtime/update-component.ts#L69).

cptflammin:
Same here. Really problematic as a pretty important property in an app design ! Any update ?

splitinfinities:
Could folks update to the latest versions of Stencil and Framework and help me understand if this still exists? It would be super helpful as well if we could get a GitHub repo with the recreation steps so we can quickly identify and resolve the issue.

tricki:
I just created a new project (`npm init stencil`), added `fullscreen={true}` and configured a base URL and can reproduce the bug. Here's the repo (with VS Code prerender launch configured): https://github.com/tricki/stencil-fullscreen-prerender-bug 
Just run `npm run build --prerender` to reproduce it.


I still see [`updateComponent`](https://github.com/ionic-team/stencil/blob/bf5f197910daab7f822a6e4c56f4f40a81c2ce7e/src/runtime/update-component.ts#L69) being called in an endless loop for `ion-content`. Here's the call stack (note `updateComponent` at the top and bottom):

[...]

PR #4591, opened 2023-07-18
fix(runtime): `forceUpdate` calls only execute when in a browser env

tanner-reits:

Using `forceUpdate` inside a component that gets pre-rendered (SSG) will result in an infinite loop in the pre-rendering process. This was observed in several bugs reported to the Ionic Framework and Stencil due to the `ion-content` component's `fullscreen` property. The issues in question:

- [#27411 ](https://github.com/ionic-team/ionic-framework/issues/27411) (Framework)
- #2429 (Stencil)
- #4076 (Stencil)

As a workaround, Framework wrapped the call to `forceUpdate` in the `ion-content` component in a check for a browser environment. That change can be seen [here](https://github.com/ionic-team/ionic-framework/pull/27440/files).

## What is the new behavior?

This commit introduces the workaround Framework made to solve their problem globally to the Stencil runtime. With this change, any call to the `forceUpdate` method will only execute the code to schedule a re-render if the project is running in a browser (`Build.isBrowser`). i.e. `forceUpdate` will perform no action in an SSR/SSG environment.

- [ ] Yes
- [X] No

## Testing

**Automated Tests**
Updated unit test conditionals for them to continue to pass

**Manual Tests**
1. Create a new component library (`npm init stencil`)
2. Add the following code snippet to the `my-component.tsx` file:
```tsx
componentDidRender() {
  forceUpdate(this);
}
```
_You'll also need to add `forceUpdate` to the import for `@stencil/core`_
3. Run `npm run build`. The build will hang during the pre-render process.
4. Build and install this branch in the project
5. Run `npm run build`. Pre-rendering will now complete successfully.

### DIFF
--- a/src/testing/platform/testing-build.ts
+++ b/src/testing/platform/testing-build.ts
@@ -2,7 +2,7 @@ import type * as d from '@stencil/core/internal';
 
 export const Build: d.UserBuildConditionals = {
   isDev: true,
-  isBrowser: false,
-  isServer: true,
+  isBrowser: true,
+  isServer: false,
   isTesting: true,
 };
--- a/bin/stencil
+++ b/bin/stencil
@@ -18,14 +18,14 @@ if (isNodeLT(minimumVersion)) {
       currentVersion +
       ', however Stencil requires v' +
       minimumVersion +
-      '.0 or greater. It is recommended to use an Active LTS version of Node (https://nodejs.org/en/about/releases/).\n'
+      '.0 or greater. It is recommended to use an Active LTS version of Node (https://nodejs.org/en/about/releases/).\n',
   );
   process.exit(1);
 }
 
 if (isNodeLT(futureDeprecationMinVersion)) {
   console.warn(
-    '\nIn an upcoming major release of Stencil, Node v' + recommendedVersion + '.0 or higher will be required.\n'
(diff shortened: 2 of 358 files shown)


src/testing/platform/testing-build.ts
```diff
@@ -2,7 +2,7 @@ import type * as d from '@stencil/core/internal';
 
 export const Build: d.UserBuildConditionals = {
   isDev: true,
-  isBrowser: false,
-  isServer: true,
+  isBrowser: true,
+  isServer: false,
```
tanner-reits, line 6: Needed to make these changes for unit tests leveraging `forceUpdate` to pass. No other tests started to fail with this change and the `newSpecPage` helper uses the lazy build which would traditionally have `isBrowser: true`. So, I believe this is a low-risk change

tanner-reits:
@alicewriteswrongs @rwaskiewicz 

I'm not sure of a great way to test these changes. In order to add more test cases in the unit tests we'd have to manipulate the build conditionals in the tests, which I'm not even sure will work or is worth testing. And for e2e tests, I'm not sure how we would be able to test that the call to `forceUpdate` does nothing.

If y'all have thoughts, I'm happy to take a stab at them, but I omitted any test additions for now.

alicewriteswrongs (approved):

rwaskiewicz (approved):

Issue #4589, opened 2023-07-18
bug: defineCustomElements return type is wrong


andrew9994:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

4.0.1

### Current Behavior

Currently the return type of `defineCustomElements` is defined as a `Promise<void>` on [this](https://github.com/ionic-team/stencil/blob/main/src/compiler/output-targets/output-lazy-loader.ts#L88C25-L88C45) line in the output-lazy-loader,
but the promise got removed in [this](https://github.com/ionic-team/stencil/commit/687592004e26be8a4ff3735b5dd7854440ff03aa#diff-aa243938e993487e7a3d8d8361f892f44009528eb48f7cf8028ab69b3ae3446bL193) commit. 

### Expected Behavior

The return value of `defineCustomElements` should be typed correctly. However, I am not sure whether this change should be marked as a breaking change.

### Steps to Reproduce

Import `defineCustomElements` from the loader of a package and try to do something after the promise gets resolved.

```
import { defineCustomElements } from 'stencil-package/loader';

defineCustomElements().then(() => {
  console.log('elements defined');
})
```

### Code Reproduction URL

n/a

alicewriteswrongs:
@andrew9994 thanks for reporting, I was just able to confirm that this is an issue by putting together a quick reproduction here (https://github.com/alicewriteswrongs/stencil-custom-elements-repro)

In the future when reporting issues if you have time to put together something along those lines it greatly helps us in diagnosing problems!

The fix is straightforward here so I'm going to get going on it. Thanks again for filing this!

andrew9994:
Thanks @alicewriteswrongs for jumping on this so quickly 😊 and sorry for not creating a repro repo, will do it next time.

PR #4592, opened 2023-07-18
fix(lazy): adjust the type of `defineCustomElements`

alicewriteswrongs:
This fixes an issue with the type for `defineCustomElements` that we set in output-lazy-loader.ts. In #4419 we changed the return type of the function to be `void` (at runtime) but we didn't make a corresponding adjustment to the type declaration for the function that we generate here:

https://github.com/ionic-team/stencil/blob/7d5dc6cf5e0d2020c513cc87b6b2e5b93eece9bc/src/compiler/output-targets/output-lazy-loader.ts#L88

This change adjusts that type declaration to have a `void` return value, instead of `Promise<void>`.

You'll get a type error when you're actually using things correctly, while code that type-checks will not work at runtime. Not good!

## What is the new behavior?

The runtime behavior is now accurate described by the type declaration.

- [ ] Yes
- [x] No

## Testing

On #4589 there is a reproduction case linked. If you clone that repro you can confirm the issue by following the steps described in the readme.

Then to confirm this fix works:

- check it out locally, build and pack stencil
- install the `.tgz` into the `stencil-project` package
- rebuild (`npm run build`) and then check that you'll get a type error in the editor if you open up `test.ts`. Adjust the code to fix the error and confirm that you have 1) no type errors and 2) no runtime errors!

### DIFF
--- a/src/compiler/output-targets/dist-lazy/lazy-output.ts
+++ b/src/compiler/output-targets/dist-lazy/lazy-output.ts
@@ -187,7 +187,7 @@ const getLazyEntry = (isBrowser: boolean): string => {
   } else {
     s.append(`import { globalScripts } from '${STENCIL_APP_GLOBALS_ID}';\n`);
     s.append(`export const defineCustomElements = (win, options) => {\n`);
-    s.append(`  if (typeof window === 'undefined') return Promise.resolve();\n`);
+    s.append(`  if (typeof window === 'undefined') return undefined;\n`);
     s.append(`  globalScripts();\n`);
     s.append(`  return bootstrapLazy([/*!__STENCIL_LAZY_DATA__*/], options);\n`);
     s.append(`};\n`);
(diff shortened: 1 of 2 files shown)


tanner-reits (approved):
Oops!

rwaskiewicz (commented):
This LGTM!

What're your thoughts on updating the function
https://github.com/ionic-team/stencil/blob/9d2d0a0612caafb88c559907c6dec1a39e760615/src/compiler/output-targets/dist-lazy/lazy-output.ts#L190
to return `undefined` here to align with the change?

alicewriteswrongs:
Sure, I'll do that real quick

rwaskiewicz (approved):

rwaskiewicz:
Looks like one of the runs for the component starter just timed out last night ([logs](https://github.com/ionic-team/stencil/actions/runs/5594180197/jobs/10228784302)). Re-added to the queue

deleonio:
This change breaks our theming engine in https://github.com/public-ui/kolibri.

We fixed the typings in our code base.

Issue #4609, opened 2023-07-24
bug: Stencil build generates local import


mlechler:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [x] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

4.0.1

### Current Behavior

When I'm running a stencil build I get the following import in my docs.d.ts
`import { ComponentCompilerEventComplexType, ComponentCompilerMethodComplexType, ComponentCompilerPropertyComplexType, ComponentCompilerReferencedType } from './stencil-private';`

Is this correct? The file is not found, cause it is obviously not there...

### Expected Behavior

Import from the package, e.g. @stencil/core/internal

### System Info

```shell
System: node 18.16.1
    Platform: darwin (21.6.0)
   CPU Model: Apple M1 Pro (10 cpus)
    Compiler: node_modules/@stencil/core/compiler/stencil.js
       Build: 1687953951
     Stencil: 4.0.1 ⛹
  TypeScript: 5.0.4
      Rollup: 2.42.3
      Parse5: 7.1.2
      Sizzle: 2.42.3
      Terser: 5.18.1
```

### Steps to Reproduce

Run following command
`stencil build`

### Code Reproduction URL

https://github.com/mlechler/stencil-build-bug

rwaskiewicz:
Hey @mlechler 👋 

Can you please create a code reproduction case for the team to take a look at? It's immensely helpful in understanding the issue that you're running into. Thanks!

mlechler:
Hey @rwaskiewicz,

I added a repository url in the issue.
Simply running `npm run build` will produce the error in the `dist/docs.d.ts`.

Maybe docs-json is the problem? I guess the d.ts file is generate because of the json?

alicewriteswrongs:
Hey @mlechler, I just published a dev build of Stencil which includes a fix for this issue. If you have time to install and try it out we can confirm the fix.

You can do so like this:

```sh
npm i @stencil/core@4.0.2-dev.1690405632.fa8e0c7 
```

mlechler:
Hey @alicewriteswrongs, works properly I guess 👍 

Thank you for this fast solution ❤️

[...]

PR #4619, opened 2023-07-25
fix(docs-json): use dts-bundle-generator to bundle types for docs-json

alicewriteswrongs:
This fixes an issue that crops up when building documentation using the
`docs-json` output target.

When https://github.com/ionic-team/stencil/pull/4212 was merged it included a change to
`src/declarations/stencil-public-docs.ts` which involved adding imports
from private Stencil type declarations. Previously this file had no
imports and so was "standalone", however, after this change it now
imported some types from `"./private"`.

This created a problem when using the `docs-json` output target because
when building a Stencil project with that output target enabled the
compiled typedef corresponding to `stencil-public-docs.ts` is copied out
of `node_modules` and into the users project. See here:

https://github.com/ionic-team/stencil/blob/43cfc584c8fbcb77122cbf2a813b1a0303085dce/src/compiler/docs/json/index.ts#L16-L18

This commit fixes the issue by using `dts-bundle-generator` to inline
the types imported by `stencil-public-docs.ts` so that the resulting
`.d.ts` file has no external dependencies and is therefore portable.

GitHub Issue Number: #4609

There's a build error because of an unresolved import when building with the `docs-json` output target.

- [ ] Yes
- [x] No

## Testing

The `test/docs-json` app is compiling with no problems. Check also that the reproduction on the repro is fixed.


scripts/utils/bundle-dts.ts
```diff
@@ -1,9 +1,21 @@
...
+ *
+ * **Note**: this file caches its output to disk, and will return any
+ * previously cached file if not in a prod environment!
+ *
+ * @param opts an object holding information about the current build of Stencil
+ * @param inputFile the path to the file which should be bundled
+ * @param ouputOptions options for bundling the file
+ * @returns a string containing the bundled typedef
+ */
+export async function bundleDts(opts: BuildOptions, inputFile: string, ouputOptions?: OutputOptions): Promise<string> {
```
alicewriteswrongs, line 18: I added the `outputOptions` optional parameter here because for bundling up `stencil-public-docs.d.ts` I wanted to be able to pass `noBanner: true` since we do our own banner already:

https://github.com/ionic-team/stencil/blob/43cfc584c8fbcb77122cbf2a813b1a0303085dce/src/compiler/docs/json/index.ts#L18-L26

and also I want to pass `exportReferencedTypes: false` to make sure that the inlined types aren't exported (this is the default behavior)

scripts/bundles/internal.ts
```diff
@@ -70,7 +70,12 @@ async function copyStencilInternalDts(opts: BuildOptions, outputInternalDir: str
   // @stencil/core/internal/stencil-public-docs.d.ts
   const docsDtsSrcPath = join(declarationsInputDir, 'stencil-public-docs.d.ts');
   const docsDtsDestPath = join(outputInternalDir, 'stencil-public-docs.d.ts');
-  const docsDts = cleanDts(await fs.readFile(docsDtsSrcPath, 'utf8'));
+  // We bundle with `dts-bundle-generator` here to ensure that when the `docs-json`
+  // OT writes a `docs.d.ts` file based on this file it is fully portable.
+  const docsDts = await bundleDts(opts, docsDtsSrcPath, {
+    noBanner: true,
+    exportReferencedTypes: false,
+  });
```
alicewriteswrongs, line 78: Instead of just copying over the file we now bundle it to ensure that it no longer has any imports

alicewriteswrongs (commented):
just noting a few things

rwaskiewicz (approved):
A few non-blocking asks that shouldn't affect the change itself. Feel free to accept/change what you'd like.

scripts/utils/bundle-dts.ts
```diff
@@ -1,9 +1,21 @@
...
+/**
+ * A thin wrapper for `dts-bundle-generator` which uses our build options to
+ * set a few things up
+ *
+ * **Note**: this file caches its output to disk, and will return any
+ * previously cached file if not in a prod environment!
+ *
+ * @param opts an object holding information about the current build of Stencil
+ * @param inputFile the path to the file which should be bundled
+ * @param ouputOptions options for bundling the file
```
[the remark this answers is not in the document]
alicewriteswrongs: haha I had actually caught this it just slipped my mind to rename it 🤪

PR #4624, opened 2023-07-26
chore(build): document build conditionals

rwaskiewicz:

How `BUILD` gets set isn't super clear.
GitHub Issue Number: N/A

## What is the new behavior?

- this commit adds a small explanation as to how BUILD conditionals get set at compile time of a Stencil project

- [ ] Yes
- [x] No

## Testing

I did tighten some types (`Config` -> `ValidatedConfig`). The type checker ought to catch any issues

### DIFF
--- a/src/app-data/index.ts
+++ b/src/app-data/index.ts
@@ -21,9 +21,6 @@ import type { BuildConditionals } from '@stencil/core/internal';
  * `STENCIL_APP_DATA_ID` uses it to replace these defaults with {@link BuildConditionals} that are derived from a
  * Stencil project's contents (i.e. metadata from the components). This replacement happens at a Stencil project's
  * compile time. Such code can be found at `src/compiler/app-core/app-data.ts`.
- *
- *
- * This collection is expected to be updated before used in compilation of a Stencil project.
  */
 export const BUILD: BuildConditionals = {
   allRenderFn: false,
--- a/src/app-data/index.ts
+++ b/src/app-data/index.ts
@@ -1,5 +1,30 @@
 import type { BuildConditionals } from '@stencil/core/internal';
 
+/**
+ * A collection of default build flags for a Stencil project.
+ *
+ * This collection can be found throughout the Stencil codebase, often imported from the `@app-data` module like so:
+ * ```ts
+ * import { BUILD } from '@app-data';
+ * ```
+ * and is used to determine if a portion of the output of a Stencil _project_'s compilation step can be eliminated.
+ *
+ * e.g. When `BUILD.allRenderFn` evaluates to `false`, the compiler will eliminate conditional statements like:
+ * ```ts
+ * if (BUILD.allRenderFn) {
+ *   // some code that will be eliminated if BUILD.allRenderFn is false
+ * }
+ * ```
+ *
+ * `@app-data`, the module that `BUILD` is imported from, is an alias for the `@stencil/core/internal/app-data`, and is
+ * partially referenced by {@link STENCIL_APP_DATA_ID}. The `src/compiler/bundle/app-data-plugin.ts` references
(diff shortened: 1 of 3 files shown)


src/app-data/index.ts
```diff
@@ -1,5 +1,30 @@
...
+ *
+ * `@app-data`, the module that `BUILD` is imported from, is an alias for the `@stencil/core/internal/app-data`, and is
+ * partially referenced by {@link STENCIL_APP_DATA_ID}. The `src/compiler/bundle/app-data-plugin.ts` references
+ * `STENCIL_APP_DATA_ID` uses it to replace these defaults with {@link BuildConditionals} that are derived from a
+ * Stencil project's contents (i.e. metadata from the components). This replacement happens at a Stencil project's
+ * compile time. Such code can be found at `src/compiler/app-core/app-data.ts`.
+ *
+ *
+ * This collection is expected to be updated before used in compilation of a Stencil project.
+ */
```
alicewriteswrongs, line 27:
```suggestion
 *
 * This collection should be updated before use in compilation of a Stencil project.
 */
```

I'm not sure what the last sentence there means exactly, does that change sort of get at it? I'm not clear maybe who is supposed to update it, so it's not clear to me what this refers to

src/compiler/app-core/app-data.ts
```diff
@@ -129,7 +137,15 @@ const getModuleImports = (moduleMap: ModuleMap, filePath: string, importedModule
   return importedModules;
 };
 
-export const updateBuildConditionals = (config: Config, b: BuildConditionals) => {
+/**
+ * For a provided Stencil project configuration, update provided the build conditionals
```
alicewriteswrongs, line 141:
```suggestion
 * Update the provided build conditionals object in-line with a provided Stencil project configuration
```

alicewriteswrongs: Sometimes I really wish TypeScript had a `mut` keyword that could be used to signal this (and to produce a type error if you try to mutate something without it). Ah well!

src/compiler/bundle/app-data-plugin.ts
```diff
@@ -182,7 +192,16 @@ const appendGlobalScripts = (globalScripts: GlobalScript[], s: MagicString) => {
   }
 };
 
-const appendBuildConditionals = (config: d.Config, build: d.BuildConditionals, s: MagicString) => {
+/**
+ * Generates the `BUILD` constant that is used at compile-time in a Stencil project
+ *
+ * **This function mutates the provided string argument**
```
alicewriteswrongs, line 198: should this call out that it's a `MagicString` ? I didn't read the type signature and I was confused because I was like 'hey, strings are pass-by-value in JS, what is this witchcraft?'

alicewriteswrongs (changes requested):

tanner-reits (approved):
Looks good aside from Alice's suggestions

src/compiler/app-core/app-data.ts
rwaskiewicz: Done in cafda3a87

> Sometimes I really wish TypeScript had a mut keyword that could be used to signal this (and to produce a type error if you try to mutate something without it). Ah well!

Oh, I'd love somethings like that! Terrified what edge cases _implementing_ something like that might look like though 😬 😆

src/app-data/index.ts
rwaskiewicz: I had this sentence in as an early placeholder before I added [this whole bit](https://github.com/ionic-team/stencil/pull/4624/files#diff-d01a21edbe6c4684280903b1fc7ef9d625c04ab58e14f0a0ae976276fabf1a63R6-R23). It think it probably makes sense to drop it at this point - 047bdb2f0

alicewriteswrongs (approved):
🚢

PR #4627, opened 2023-07-27
chore(repo): update renovate schedule

rwaskiewicz:

Renovate can be a little noisy, even with our current config

GitHub Issue Number: N/A

## What is the new behavior?

this commit updates the rate at which renovate runs. justification/details on the configuration can be found in the comments of the configuration file

**This is still up for discussion/alteration**. We talked about slowing this down in Slack, without specifying the degree to which we'll do it. LMK what you think!

- [ ] Yes
- [x] No

## Testing

Validated the cron schedule  - https://crontab.guru/#*_0-11_*_*_1,4

> At every minute past every hour from 0 through 11 on Monday and Thursday.

AFAICT, we follow the format correctly - https://docs.renovatebot.com/configuration-options/#schedule

> For Cron schedules, you must use the * wildcard for the minutes value, as Renovate doesn't support minute granularity.

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -94,8 +94,19 @@
   ],
   // Never rebase the branch or update it unless manually requested to avoid noisy PR emails
   rebaseWhen: 'never',
-  // Note: Timezone for the schedule is specified as UTC
-  schedule: ['every weekday before 11am'],
+  /**
+   * Run every Monday and Thursday between midnight at 11 AM (in UTC)
+   *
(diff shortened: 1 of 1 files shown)


tanner-reits (approved):
I'm okay with trying this schedule out. I could see this dropping down even more to bi-weekly or even monthly

alicewriteswrongs (approved):

PR #4648, opened 2023-08-01
docs(release): notify CS of enterprise fixes

tanner-reits:

N/A

GitHub Issue Number: N/A

## What is the new behavior?

With the decision to cut-down on the frequency of Stencil/CS syncs, we need to do our part to make sure the CS team is aware of enterprise fixes and feature requests that are included in our releases. This commit adds a step to our release process for this purpose.

- [ ] Yes
- [ ] No


RELEASE.md
```diff
@@ -71,4 +71,7 @@ manual release was performed.
       ```
 1. Let folks in Ionic know about the release in the #whathappened Slack channel
    with a link to the changelog.
+   1. Let the CS team know about any enterprise reported issues or feature requests that were included in the release. Jira
```
rwaskiewicz, line 74: Would be just be "at-ing" certain folks in a thread on #whathappened? Or via DM?

tanner-reits: I think that's still a little TBD. Right now, Marlon is the point-of-contact, but not sure if just a DM is the right move. We could thread the announcement if we end up doing one for every release, or just post in one of the Stencil channels.

rwaskiewicz: I think for now I'd prefer the latter (threading the announcement) - that way discussions don't get lost in DMs, hopefully 🤞 In my ideal world, we do this in a place/way that has high visibility, so that one individual (whomever releases) isn't the only one having conversations with CS. 

That being said, I should have some better clarity around what to do with the #whathappened channel later today, and can follow up here after the fact

rwaskiewicz: I have some better guidance around this 🎉 

Moving forward, we're only going to be using that channel for major releases and impactful minor/patch releases (e.g. we release a feature/patch a bug that reduces the Framework size by 20%).

So, with that mind, what do you think about something like:
```suggestion
   1. Let the CS team know by 'at-ing' them  about any enterprise reported issues or feature requests that were included in the release in a thread on your #whathappened post. Jira
```

rwaskiewicz (approved):

PR #4665, opened 2023-08-09
chore(compiler): remove some unused code

alicewriteswrongs:
I don't believe that these files are used anywhere, and Stencil builds just fine without them.

We have these two extra files hanging around! Nothing imports them, and as far as I can tell they aren't used for anything.

## What is the new behavior?

DELETE DELETE DELETE

- [ ] Yes
- [x] No

## Testing

I ran some builds of framework just to confirm that nothing was broken - good to test that thoroughly!


rwaskiewicz (approved):
Also ran `git rev-list --all | git grep "initWebWorkerThread"` && `git rev-list --all | git grep "createWebWorkerMainController"` on the exported variables/functions, and didn't find any other usages 🤔

tanner-reits (approved):
Took a quick look and didn't see these used anywhere either

PR #4673, opened 2023-08-10
chore(declarations): remove unused types

rwaskiewicz:

We have several types in the compiler that appear to be unused. This expands the size of the codebase, and makes it hard to reason about what's actually needed/used in modern Stencil.

GitHub Issue Number: N/A

## What is the new behavior?

this commit removes several types that are unused in the stencil codebase. these have previously gone undetected due to their being exported by `src/declarations/index.ts`.

- [ ] Yes
- [x] No

## Testing

I built this branch (`npm run clean && npm ci && npm run build && npm pack`) and installed it into the Framework's `feature-7.3` branch. I spun up the local dev server, and played around with datetime and accordion to see if anything was logged.

## Other information

these types are not used within the stencil codebase. _however_ it is possible that a consumer is reaching into our internal type declarations to use these types. those that are in this file are assumed to be 'private' (and therefore volatile). should a bug report come in regarding one of these types, the recommended recourse is that the team investigate/determine on a per-issue basis whether the type should be restored in `stencil-public.ts` or re-added to this file and deprecated (and removed in v5)

### DIFF
--- a/src/declarations/stencil-private.ts
+++ b/src/declarations/stencil-private.ts
@@ -6,7 +6,6 @@ import type {
   BuildLog,
   BuildResultsComponentGraph,
   CompilerBuildResults,
-  CompilerBuildStart,
   CompilerFsStats,
   CompilerRequestResponse,
   CompilerSystem,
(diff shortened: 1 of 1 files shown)


alicewriteswrongs (approved):
some much code to delete! 🤯

tanner-reits (approved):
Love to see it 🪓

PR #4675, opened 2023-08-10
chore(repo): create initial pull request action

rwaskiewicz:

our automated release process for production isn't compatible with the merge queue we have set up
GitHub Issue Number: N/A

## What is the new behavior?

this commit creates a new action that will (eventually) go on to create
pull requests for the stencil team when it is time to release a new
version of the library. the (eventual) intent is for stencil engineers
to run this workflow, which will generate a pull request that includes:
- version bumps in package.json & package-lock.json
- changelog, notice, and license generation

this commit currently only creates a new branch and commit. the branch
includes the version string, as well as the run/attempt in github
actions to tie us back to the run that created it.

at this point, the creation of the pull request is not included. the
intended library to use
(https://github.com/peter-evans/create-pull-request) does not work on
push/pull_request events in GHA without additional permissions. by
adding this action, we can test it without generating a new token.

- [ ] Yes
- [x] No

## Testing

In https://github.com/ionic-team/stencil/pull/4675/commits/1179138708c6fca0f8c05651fadb0891006e36e2, I had a `push` event configured for  this job. The logs are attached for those wanting to see the output. 
[1_Generate Stencil Release PR (1).txt](https://github.com/ionic-team/stencil/files/12314993/1_Generate.Stencil.Release.PR.1.txt)

I also was able to verify the branch was created successfully - 
Screenshot 2023-08-10 at 11 30 34 AM

## Other information

This commit does not remove 'extraneous' steps from the CI-based preparations. That is, it still builds Stencil, bundles it, runs karma/unit tests, etc. _outside of_ the CI workflow steps that we use in PRs. I'll update those in a separate PR. This is meant to just give me the ability to start hacking on PR creation

### DIFF
--- a/.github/workflows/create-production-pr.yml
+++ b/.github/workflows/create-production-pr.yml
@@ -1,0 +1,81 @@
+name: 'Stencil Release PR Creation'
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        required: true
+        type: choice
+        description: Which version should be published?
+        options:
+          - prerelease
+          - prepatch
+          - preminor
+          - premajor
+          - patch
+          - minor
+          - major
+
+jobs:
+  create-stencil-release-pull-request:
+    name: Generate Stencil Release PR
+    runs-on: ubuntu-latest
+    steps:
+      # Log the input from GitHub Actions for easy traceability
+      - name: Log GitHub Input
+        run: |
+          echo "Version: ${{ inputs.version }}"
+        shell: bash
+
+      - name: Checkout Code
+        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
+        with:
+          # A depth of 0 gets the entire git history.
+          # We need git history to generate the changelog; however, we don't know how deep to go.
+          # Since publishing is a one-off activity, just get everything.
+          fetch-depth: 0
+
+      - name: Get Core Dependencies
+        uses: ./.github/workflows/actions/get-core-dependencies
+
+      - name: Run Publish Preparation Script
+        run: npm run release.ci.prepare -- --version ${{ inputs.version }} --any-branch
+        shell: bash
+
+      - name: Log Generated Changes
+        run: git --no-pager diff
+        shell: bash
+
+      - name: Generate Version String and Branch Name
+        id: name_gen
+        run: |
+          VERSION_STR=$(jq '.version' package.json | sed s/\"//g)
+          echo "VERSION_STR=$VERSION_STR" >> "$GITHUB_OUTPUT"
+          echo "BRANCH_NAME=release/$VERSION_STR-run-${{ github.run_number }}-${{ github.run_attempt }}" >> "$GITHUB_OUTPUT"
+        shell: bash
+
+      - name: Print Version String and Branch Name
+        run: |
+          echo Version: ${{ steps.name_gen.outputs.VERSION_STR }}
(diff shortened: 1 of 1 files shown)


.github/workflows/create-production-pr.yml
```diff
@@ -0,0 +1,81 @@
...
+          echo "Version: ${{ inputs.version }}"
+        shell: bash
+
+      - name: Checkout Code
+        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
+        with:
+          # A depth of 0 gets the entire git history.
+          # We need git history to generate the changelog; however, we don't know how deep to go.
+          # Since publishing is a one-off activity, just get everything.
+          fetch-depth: 0
```
alicewriteswrongs, line 35: seems fine to me to do, we're not quite dealing with a repo where we need to start optimizing for clone times!

.github/workflows/create-production-pr.yml
```diff
@@ -0,0 +1,81 @@
...
+
+      - name: Print Version String and Branch Name
+        run: |
+          echo Version: ${{ steps.name_gen.outputs.VERSION_STR }}
+          echo Branch Name: ${{ steps.name_gen.outputs.BRANCH_NAME }}
+        shell: bash
+
+      - name: Create Git Branch
+        # Note: We'll let GitHub clean up the branch for us (this is configured in the repo settings on GitHub)
+        run: git checkout -b ${{ steps.name_gen.outputs.BRANCH_NAME }}
```
alicewriteswrongs, line 64: do we need to do any error handling to ensure the branch doesn't already exist?

alicewriteswrongs: I think if `$BRANCH` already exists then `git checkout -b $BRANCH` will have a non-zero exit code (looks like 128) so the workflow will fail at that point

.github/workflows/create-production-pr.yml
```diff
@@ -0,0 +1,81 @@
...
+
+      # Commit the CHANGELOG.md, NOTICE.md, LICENSE.md, package.json, etc.
+      - name: Commit Release Preparations
+        run: |
+          git config user.name "Stencil Release Bot (on behalf of ${{ github.actor }})"
+          git config user.email "stencil-release-bot@ionic.io"
+          git add .
+          git commit -m "v${{ steps.name_gen.outputs.VERSION_STR }}"
```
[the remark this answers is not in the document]
alicewriteswrongs: haha when you're piping `sed` output into `sed`...

alicewriteswrongs (changes requested):
just one question

.github/workflows/create-production-pr.yml
rwaskiewicz: Yeah - it should fail. `git checkout -b $BRANCH` _will_ return 0 though when a branch already exists if there aren't any commits on the base branch, but since we're grabbing `main` (and other established branches) that shouldn't be an issue

.github/workflows/create-production-pr.yml
rwaskiewicz: ....yet 😆

.github/workflows/create-production-pr.yml
alicewriteswrongs: ok cool that makes sense to me, if we see this fail we'll need to probably do some manual intervention because something got messed up anyway (i.e. we tried to create a release and closed the PR for some reason)

alicewriteswrongs (approved):

Issue #4667, opened 2023-08-09
bug: tsconfig include array triggers warning when path has leading ./


cam-narzt:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

4.0.4

### Current Behavior

if your tsconfig looks like this:

```
"include": [
        "./app/javascript"
    ],
```
you will get the very confusing warning:
```
[ WARN  ]  tsconfig.json "include" required
           In order for TypeScript to improve watch performance, it's recommended the "tsconfig.json" file should have
           the "include" property, with at least the app's "app/javascript" directory listed. For example: "include":
           ["app/javascript"]
```

### Expected Behavior

paths that are equivalent shouldn't raise this warning

### System Info

```shell
npx stencil info doesn't work on my system and isn't relevant in this case.
```

### Steps to Reproduce

clone linked repo
checkout `tsconfig_bug` branch
run `npm i`
run `npm run build`

### Code Reproduction URL

https://github.com/cam-narzt/stencilbug/tree/tsconfig_bug

rwaskiewicz:
Hey @cam-narzt 👋 

Thanks for bug report + reproduction case! I took a look at it this morning, and wasn't able to reproduce the error exactly on my end. Following the steps to repro, I:
- Cloned the repo
- Checked out `tsconfig_bug` (where the HEAD is c417fd76254031f390d0b564c18af36d1dce350e)
- Ran `npm ci` to get the same dependencies
- Ran `npm run build`

I _do_ get an error message, but it looks like it's properly listing the 'src' directory in the error message instead of 'app/javascript':
```
➜  stencilbug git:(tsconfig_bug) npm run build

> demo-error@0.0.1 build
> stencil build --docs

[56:01.5]  @stencil/core
[56:01.7]  v4.0.4 🍧

[ WARN  ]  tsconfig.json "include" required
           In order for TypeScript to improve watch performance, it's recommended the "tsconfig.json" file should have
           the "include" property, with at least the app's "src" directory listed. For example: "include": ["src"]
```

Does it look like I'm missing anything here?

rwaskiewicz:
[...]

PR #4676, opened 2023-08-10
fix(compiler): match tsconfig include paths properly

rwaskiewicz:

When `tsconfig.json#include` contains a relative path:
```
"include": [
  "./app/javascript"
],
``` 
and `tsconfig.json#srcDir` includes the non-relative version of that path:
```
srcDir: 'app/javascript',
```

then you receive a very confusing error:
```
[ WARN  ]  tsconfig.json "include" required
           In order for TypeScript to improve watch performance, it's recommended the "tsconfig.json" file should have
           the "include" property, with at least the app's "app/javascript" directory listed. For example: "include":
           ["app/javascript"]
```

GitHub Issue Number: #4667 

## What is the new behavior?

this commit fixes a bug where a user was unable to use relative paths in the `include` property of their `tsconfig.json` file if `srcDir` was also specified. 

- [ ] Yes
- [x] No

## Testing

- Unit tests have been added to the function in question
- Tested a dev build - `@stencil/core@v4.0.4-dev.1691704809.9fb61d7` - against the minimal reproduction case

### DIFF
--- a/src/compiler/sys/typescript/tests/typescript-config.spec.ts
+++ b/src/compiler/sys/typescript/tests/typescript-config.spec.ts
@@ -1,0 +1,35 @@
+import { hasSrcDirectoryInclude } from '../typescript-config';
+
+describe('typescript-config', () => {
+  describe('hasSrcDirectoryInclude', () => {
+    it('returns `false` for a non-array argument', () => {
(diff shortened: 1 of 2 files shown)


tanner-reits (approved):
Looks good!

alicewriteswrongs (approved):

PR #4682, opened 2023-08-15
fix(runtime): adds a testing check to the forceUpdate method

tanner-reits:

The changes to testing build conditionals made [here](https://github.com/ionic-team/stencil/commit/b203263482140fde31edfb7a91ac6054f5f98460#diff-4bb40df78659571d0bbbd5cd771c28484c1a1c9beefd451d6d004c7d4ed0c80f) were causing tests to fail in the Ionic Framework.

GitHub Issue Number: N/A

## What is the new behavior?

This commit adds a check for `isTesting` in addition to `isBrowser` so that the forceUpdate logic can run in test environments without needing to change the testing build conditional values. These changes were causing tests to fail in Framework.

- [ ] Yes
- [X] No

## Testing

Installed a build of this branch in the framework `feature-7.3` branch and confirmed the reported test failures are resolved. Also confirmed that all internal Stencil tests (unit & e2e) continue to pass.

### DIFF
--- a/src/runtime/test/globals.spec.tsx
+++ b/src/runtime/test/globals.spec.tsx
@@ -33,10 +33,10 @@ describe('globals', () => {
   });
 
   it('build values', () => {
-    expect(Build.isBrowser).toBe(true);
+    expect(Build.isBrowser).toBe(false);
     expect(Build.isDev).toBe(true);
     expect(Build.isTesting).toBe(true);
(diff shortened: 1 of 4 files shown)


rwaskiewicz (approved):

alicewriteswrongs (approved):
looks good!

PR #4711, opened 2023-08-21
chore(ci): create release pull requests

rwaskiewicz:

GitHub Issue Number: N/A

## What is the new behavior?

this commit adds the initial workflow step to create a pull request using the action. it currently has a few limitations regarding access tokens, but is usable with a small workaround (noted in RELEASE.md)

- [ ] Yes
- [x] No

## Testing

I ran the workflow per the instructions in `RELEASE.md`, whose results can be seen [here](https://github.com/ionic-team/stencil/actions/runs/5931698742).
[This created a PR](https://github.com/ionic-team/stencil/pull/4715), which I closed/re-opened to verify the workflows ran

## Other information

in that PR, you'll notice that some of the license text is removed. As far as I can tell, some of the TXT/MD files aren't getting picked up in GH Actions. I'm not sure if they're not there (I need to check), or if something else is awry. I'm also simultaneously trying to chase down what our actual license requirements are here. I think we'd be OK to merge this as-is, since we're still calling packages out and what their licenses are.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -6,18 +6,27 @@ Manual releases should only be performed when there are extenuating circumstance
 
 ## Automated Releases
 
+⚠️ Do not run this workflow at this time ⚠️
+
+⚠️ It is currently not in a working state ⚠️
+
 1. Call a `code-freeze` in the Stencil team channel
 1. Check that [Stencil's Merge
    Queue](https://github.com/ionic-team/stencil/queue/) is empty (nothing is
    queued for merge).
-1. Run the [Stencil Production Release](https://github.com/ionic-team/stencil/actions/workflows/release-production.yml)
+1. Run the [Stencil Production Release PR Creation Workflow](https://github.com/ionic-team/stencil/actions/workflows/create-production-pr.yml)
 in GitHub
    1. Run the workflow from the `main` branch, _unless_ the release is for a previous major version of Stencil.
    In that scenario, select the `v#-maintenance` branch corresponding to the version of Stencil being released.
    For example, `v3-maintenance` to release a new version of Stencil v3.
    1. Stencil follows semantic versioning. Select the appropriate version from the dropdown for this release.
-   1. Stencil should be published under the `latest` tag, _unless_ the release is for a previous major version of
-   Stencil.
+   1. Hit "Run Workflow" and wait for a new pull request to be created.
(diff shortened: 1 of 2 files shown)


RELEASE.md
```diff
@@ -6,18 +6,27 @@ Manual releases should only be performed when there are extenuating circumstance
...
    1. Stencil follows semantic versioning. Select the appropriate version from the dropdown for this release.
-   1. Stencil should be published under the `latest` tag, _unless_ the release is for a previous major version of
-   Stencil.
+   1. Hit "Run Workflow" and wait for a new pull request to be created.
+1. Open the pull request that was opened as a result of running the Stencil Production Release PR Creation Workflow.
+1. Complete the following (temporary) steps:
+   1. Close the pull request and reopen it. This allows actions that the team gates pull requests on to run.
+   1. Mark the pull request as ready for review.
+1. Ask the Stencil team for an approval on the PR
+1. Once approved, add it to the merge queue
```
tanner-reits, line 29: Not sure if we need to call this out since the queue _should_ be empty at this point, but should we say to add it to the front of the queue (I think there's an option for this, but could be wrong)?

RELEASE.md
```diff
@@ -6,18 +6,27 @@ Manual releases should only be performed when there are extenuating circumstance
...
    In that scenario, select the `v#-maintenance` branch corresponding to the version of Stencil being released.
    For example, `v3-maintenance` to release a new version of Stencil v3.
    1. Stencil follows semantic versioning. Select the appropriate version from the dropdown for this release.
-   1. Stencil should be published under the `latest` tag, _unless_ the release is for a previous major version of
-   Stencil.
+   1. Hit "Run Workflow" and wait for a new pull request to be created.
+1. Open the pull request that was opened as a result of running the Stencil Production Release PR Creation Workflow.
+1. Complete the following (temporary) steps:
+   1. Close the pull request and reopen it. This allows actions that the team gates pull requests on to run.
+   1. Mark the pull request as ready for review.
```
tanner-reits, line 27: So these won't be necessary once the permissions and licensing stuff is resolved?

tanner-reits (approved):
Everything in the action makes sense. Just had a few non-blocking comments

RELEASE.md
rwaskiewicz: It shouldn't be no

alicewriteswrongs (approved):

PR #4719, opened 2023-08-23
fix(runtime): patch methods for scoped slot `append`, `prepend`, and `insertAdjacent`

### LINKED ISSUE #2641: Elements rendered outside of slot when shadow not enabled

tanner-reits:

Presently, our pseudo shadow DOM behavior does not patch methods for `insertAdjacentText`, `insertAdjacentHtml`, or `insertAdjacentElement`. This results in nodes getting inserted into the DOM in the wrong location when using these methods and does not follow the expected slotting behavior you would see when using a true shadow DOM.

GitHub Issue Number: Fixes #2641 

## What is the new behavior?

- The aforementioned methods are now patched when using the `experimentalSlotFixes` flag and behavior when using `scoped: true` matches `shadow: true`
- This commit also patches our scoped behavior when using a custom elements build (i.e. adds the patch calls to `bootsrapCustomElements()`)
- Marks prior extras flags and build variables with "todo" comments to be removed at a later time

- [ ] Yes
- [X] No

We stated that fixes under the `experimentalSlotFixes` flag would not be considered breaking changes and do not need to follow semver

## Testing

**Automated Tests**

Automated unit and e2e tests continue to pass. Additional e2e tests were added for the expected behavior using any of the `insertAdjacent` methods as well as the `append`, `appendChild`, and `prepend` methods.

**Manual Tests**

Changes for lazy builds were tested using a build of this branch within the component starter.

Changes for the custom elements build were testing using a build of this branch in a component starter and loading the components into and Angular app


src/runtime/bootstrap-custom-element.ts
```diff
@@ -36,6 +43,25 @@ export const proxyCustomElement = (Cstr: any, compactMeta: d.ComponentRuntimeMet
     cmpMeta.$flags$ |= CMP_FLAGS.needsShadowDomShim;
   }
 
+  // TODO(STENCIL-914): this check and `else` block can go away and be replaced by just `BUILD.scoped` once we
+  // default our pseudo-slot behavior
+  if (BUILD.patchPseudoShadowDom && BUILD.scoped) {
+    patchPseudoShadowDom(Cstr.prototype, cmpMeta);
```
tanner-reits, line 49: I'm not sure if we'll need the `cmpMeta` arg passed in here when this behavior becomes default. Right now, it's only used by a few of these patch functions to prevent patching based on some niche condition.

src/runtime/bootstrap-custom-element.ts
```diff
@@ -36,6 +43,25 @@ export const proxyCustomElement = (Cstr: any, compactMeta: d.ComponentRuntimeMet
     cmpMeta.$flags$ |= CMP_FLAGS.needsShadowDomShim;
   }
 
+  // TODO(STENCIL-914): this check and `else` block can go away and be replaced by just `BUILD.scoped` once we
+  // default our pseudo-slot behavior
+  if (BUILD.patchPseudoShadowDom && BUILD.scoped) {
```
tanner-reits, line 48: I pulled over this `if-else` block directly from `bootstrapLazy`. Without it, we don't patch anything for scoped behavior when using a custom elements build.

test/karma/test-app/append-child/karma.spec.ts
tanner-reits, line 1: Killed this tests since I included it in the new `append-and-prepend` tests. Felt like they all belonged in the same test suite

tanner-reits (commented):
NOTE: When we merge, we also need to be sure to co-author the commit. I'm not sure how that works with merge queue or if we need to do it before merging 🤷‍♂️

alicewriteswrongs (commented):
To create a co-authored commit I've successfully followed these instructions in the past: https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors#creating-co-authored-commits-on-the-command-line

I had one non-blocking question about patching things, other than that I confirmed that the tests you added are broken without the Stencil config options, so I think things look good to me!

src/runtime/bootstrap-custom-element.ts
tanner-reits: I'd be surprised if that was the case. As a developer, I'd expect to have (mostly) the same runtime behavior for lazy and non-lazy bundles (with the obvious exceptions for how they load). Maybe the use-case isn't there for it, but seems like we should at least offer parity. If we offer `scoped` on custom elements builds, I'd think the down-stream features of that choice should be valid.

src/runtime/dom-extras.ts
```diff
@@ -65,6 +77,126 @@ export const patchSlotAppendChild = (HostElementPrototype: any) => {
...
+};
+
+/**
+ * Patches the `insertAdjacentHTML` method for a slotted node inside a scoped component. Specifically,
+ * we only need to patch the behavior for the specific `beforeend` and `afterbegin` positions so the element
+ * gets inserted into the DOM in the correct location.
+ *
+ * @param HostElementPrototype the `Element` to be patched
+ */
+export const patchSlotInsertAdjacentHTML = (HostElementPrototype: any) => {
```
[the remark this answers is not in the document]
alicewriteswrongs: ah `RenderNode`...always causing problems. sounds good!

src/runtime/bootstrap-custom-element.ts
alicewriteswrongs: yeah that makes sense to me 👍

alicewriteswrongs (changes requested):
one other question I just thought of

src/runtime/dom-extras.ts
```diff
@@ -65,6 +80,126 @@ export const patchSlotAppendChild = (HostElementPrototype: any) => {
...
+    if (position !== 'afterbegin' && position !== 'beforeend') {
+      return (this as any).__insertAdjacentElement(position, element);
+    }
+    if (position === 'afterbegin') {
+      this.prepend(element);
+    } else if (position === 'beforeend') {
+      this.append(element);
+    }
+  };
+};
```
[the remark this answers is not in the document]
alicewriteswrongs: awesome! I love it when we can avoid an `as any` haha

alicewriteswrongs (approved):

PR #4720, opened 2023-08-23
refactor(compiler): add types for supported decorators

alicewriteswrongs:
This adds a new `ReadonlyArray` called `STENCIL_DECORATORS` which holds a list of the named decorators supported in Stencil and also adds a new type derived from that value called `StencilDecorator`. This type is then used to annotate the `isDecoratorNamed` function as an additional guard against typos.

## What is the new behavior?

This just adds some type-level assurance that when we're checking for decorator names we're not making any typos.

- [ ] Yes
- [x] No

## Testing

I don't think this should have any runtime impact, so if the test suite passes I think we should be good!


rwaskiewicz (approved):
👍

tanner-reits (approved):

PR #4724, opened 2023-08-24
docs(compiler): add JSDoc to `component-decorator.ts`

alicewriteswrongs:
This adds JSDoc comments to functions responsible for transforming the `@Component` decorator to static which were previously undocumented. It also includes a few code changes to make it a bit more obvious what one function is doing (variable renames and the like).

These functions are just sitting there!

## What is the new behavior?

Just documenting a few functions!

- [ ] Yes
- [x] No

## Testing

The code changes are limited to renaming things and type-level changes, so I don't have too many concerns there.

### DIFF
--- a/src/compiler/transformers/decorators-to-static/component-decorator.ts
+++ b/src/compiler/transformers/decorators-to-static/component-decorator.ts
@@ -6,6 +6,27 @@ import { convertValueToLiteral, createStaticGetter, retrieveTsDecorators } from 
 import { getDeclarationParameters } from './decorator-utils';
 import { styleToStatic } from './style-to-static';
 
+/**
+ * Perform code generation to create new class members for a Stencil component
+ * which will drive the runtime functionality specified by various options
+ * passed to the `@Component` decorator.
+ *
+ * Inputs are validated (@see {@link validateComponent}) before code generation
+ * is performed.
+ *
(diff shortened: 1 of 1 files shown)


src/compiler/transformers/decorators-to-static/component-decorator.ts
```diff
@@ -6,6 +6,27 @@ import { convertValueToLiteral, createStaticGetter, retrieveTsDecorators } from
...
+ * Perform code generation to create new class members for a Stencil component
+ * which will drive the runtime functionality specified by various options
+ * passed to the `@Component` decorator.
+ *
+ * Inputs are validated (@see {@link validateComponent}) before code generation
+ * is performed.
+ *
+ * **Note**: in this function and in functions that it called the `newMembers`
+ * parameter is treated as a out parameter and mutated, with new class members
+ * added to it.
```
rwaskiewicz, line 19: Nit - this reads a little funky. Maybe something like:
```suggestion
 * **Note**: in this function and in functions that it calls, the `newMembers`
 * parameter is treated as an out parameter and mutated, with new class members
 * added to it.
```

rwaskiewicz (approved):

tanner-reits (approved):

PR #4737, opened 2023-08-29
chore(snc): bring type of string props on `Diagnostic` inline with use

alicewriteswrongs:
This interface has two properties on it, `absFilePath` and `relFilePath`, which are habitually instantiated as `null` throughout the codebase but were previously typed `string | undefined`. This changes the type on the interface declaration to be `string | undefined | null` to match the interface to the usage.

This technically constitutes a breaking change, since it is modifying the type of a public interface, but in practice it should not be a problem since the type is just being widened (and widened to what it actually will be at runtime in a lot of cases).

GitHub Issue Number: N/A

## What is the new behavior?

- 

- [x] Yes
- [ ] No

A public interface is being changed. However, as the type is being widened, it shouldn't be possible for this to generate any new type errors for users, and in fact this change makes the type declaration more in line with the actual runtime type. Still, we may want to put this in a minor instead of a patch.

### DIFF
--- a/src/compiler/config/test/load-config.spec.ts
+++ b/src/compiler/config/test/load-config.spec.ts
@@ -114,14 +114,14 @@ describe('load config', () => {
       const loadedConfig = await loadConfig({ configPath: noTsConfigPath });
       expect(loadedConfig.diagnostics).toHaveLength(1);
       expect<d.Diagnostic>(loadedConfig.diagnostics[0]).toEqual({
-        absFilePath: null,
+        absFilePath: undefined,
         header: 'Missing tsconfig.json',
         level: 'error',
         lines: [],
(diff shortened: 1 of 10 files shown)


tanner-reits (approved):
I'd say this is safe to go into a minor/patch release

rwaskiewicz (commented):
Playing devil's advocate here, what's the cost/level of effort to change the instantiations from `null` to `undefined`?

Is it that the risk of updating the value used at runtime is considered to be higher than that of changing the type of the property at the TS level?

alicewriteswrongs:
I don't think it's necessarily more risky, just a bigger change! I can try it out and see if anything comes up

alicewriteswrongs:
@rwaskiewicz I updated that so that the usage is always `string | undefined`

rwaskiewicz (approved):

PR #4738, opened 2023-08-29
refactor(compiler): create type for static getters

alicewriteswrongs:
This is a small refactor which turns a `Set<string>` into a `ReadonlyArray<string>` and pushes a bit more information into the type system about how 'static getters' are created during the compilation process.

If you're unfamiliar with this, when Stencil is compiling a component it finds any properties which are decorated with our custom decorators (e.g. `@Prop` and so on) and transforms the TypeScript `ClassDeclaration` node to 1) remove the syntax nodes for the decorated property and 2) add a `static get () { ... }` member to the class.

These static getters are then used in various ways later in the compilation process. Generally the values stuck in them are parsed into the `ComponentCompilerMeta`, then the `dist-custom-elements` output target removes them entirely, and lazy-based builds (like `www`) use them at runtime.

Right now, unfortunately, these are somewhat stringly typed! Before this change we did not have a type-level assurance that the names being used for these static getters were consistent. This change moves the needle on this a bit, although it's possible that the new type (`StencilStaticGetter`) could be used more widely.

STENCIL-856

- [ ] Yes
- [x] No

## Testing

This is essentially a type-only change, with a very small change in runtime logic. I think if the project builds and the tests are passing we should be good.

### DIFF
--- a/src/compiler/transformers/remove-static-meta-properties.ts
+++ b/src/compiler/transformers/remove-static-meta-properties.ts
@@ -1,13 +1,15 @@
+import { readOnlyArrayHasStringMember } from '@utils';
 import ts from 'typescript';
 
+import { StencilStaticGetter } from './decorators-to-static/decorators-constants';
 import { retrieveTsModifiers } from './transform-utils';
 
 /**
  * Creates a new collection of class members that belong to the provided class node and that do not exist in
- * {@link REMOVE_STATIC_GETTERS}
+ * {@link STATIC_GETTERS_TO_REMOVE}
  * @param classNode the class node in the syntax tree to inspect
  * @returns a new collection of class members belonging to the provided class node, less those found in
- * {@link REMOVE_STATIC_GETTERS}
+ * {@link STATIC_GETTERS_TO_REMOVE}
  */
 export const removeStaticMetaProperties = (classNode: ts.ClassDeclaration): ts.ClassElement[] => {
   if (classNode.members == null) {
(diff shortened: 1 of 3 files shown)


src/compiler/transformers/remove-static-meta-properties.ts
```diff
@@ -16,32 +18,33 @@ export const removeStaticMetaProperties = (classNode: ts.ClassDeclaration): ts.C
...
+  'states',
   'style',
-  'styles',
+  'styleMode',
   'styleUrl',
-  'watchers',
   'styleUrls',
-]);
+  'styles',
+  'watchers',
```
alicewriteswrongs, line 49: the diff looks bigger than it really is here because I alphabetized these while I was in here

src/compiler/transformers/transform-utils.ts
```diff
@@ -156,7 +156,10 @@ const objectToObjectLiteral = (obj: { [key: string]: any }, refs: WeakSet<any>):
  * @param returnExpression a TypeScript AST node to return from the getter
  * @returns an AST node representing a getter
  */
-export const createStaticGetter = (propName: string, returnExpression: ts.Expression): ts.GetAccessorDeclaration => {
+export const createStaticGetter = (
+  propName: StencilStaticGetter,
```
alicewriteswrongs, line 160: anyone know if there are other functions where we should change `string` to `StencilStaticGetter`?

alicewriteswrongs (commented):
notes

src/compiler/transformers/transform-utils.ts
tanner-reits: I'm sure there's more, but `getStaticGetter` could use it. It's only used in testing, but types anywhere are good. Only issue is it gets called [here](https://github.com/ionic-team/stencil/blob/main/src/compiler/transformers/test/parse-encapsulation.spec.ts#L15) with a "mode" property which doesn't look like it's in the array of valid names.

I couldn't find a call to `createStaticGetter` with a "mode" property name though, so that test file could just be wrong.

tanner-reits (approved):

src/compiler/transformers/transform-utils.ts
alicewriteswrongs: Yeah, I saw that usage and I wasn't sure if that function really does what it says it does...

rwaskiewicz (approved):

PR #4739, opened 2023-08-29
fix(compiler): resolve implicit enum types

### LINKED ISSUE #2243: StencilJS - Typescript: Explicit type definition of enum is required
**Stencil version:**

```
 @stencil/core@1.8.11
```

**I'm submitting a:**

[x] bug report
[ ] feature request

**Current behavior:**

```
[ ERROR ]  TypeScript: src/components.d.ts:32:24
           Cannot find name 'BreadcrumbStyle'.

     L31:  'breadcrumbItems': Array<BreadcrumbItem>;
     L32:  'breadcrumbStyle': BreadcrumbStyle;
     L33:  'i18n': BreadcrumbI18n;

[37:07.4]  build failed in 5.07 s
```

**Expected behavior:**

```
[33:42.9]  build finished in 5.44 s
```

**Steps to reproduce:**

1) Generate pure stencil library project
2) Create a component with the code snippet from the `Related code` section
3) Run `npm run build`

**Related code:**

**breadcrumb.tsx**

```tsx
import { Component, Host, h, Prop } from '@stencil/core';
import {
  BreadcrumbStyle
} from './breadcrumb.data';

@Component({
  tag: 'wk-breadcrumb',
  styleUrl: 'breadcrumb.css,
  shadow: true
})
export class Breadcrumb {

  @Prop( {reflect: true})
  breadcrumbStyle = BreadcrumbStyle.Default;

  render() {
    return (
      <Host>
        <slot></slot>
      </Host>
    );
  }

}

```

**breadcrumb.data.ts**

```ts
export enum BreadcrumbStyle {
  Default = 'default',
  Inverse = 'inverse'
}
```

**Other information:**

It works if I define a property with an explicit type like 
```
  @Prop( {reflect: true})
  breadcrumbStyle: BreadcrumbStyle = BreadcrumbStyle.Default;
```
but since Typescript allows you to define type by setting default value, compiler also should handle this case

tanner-reits:

Stencil will not resolve an implicit enum type on a `Prop()`:

```ts
export enum Mode {
  DEFAULT = 'default'
}

@Prop() mode = Mode.DEFAULT
```

would result in an `any` type used in the generated `components.d.ts` file(s).

GitHub Issue Number: Fixes #2243 

## What is the new behavior?

Stencil will now correctly infer the type for the enum so long as it is exported (either from a separate file, or the component file). 

- [ ] Yes
- [X] No

## Testing

**Automated tests**

Added a unit test for the case where an enum is exported from the component file that uses it. Testing the case where the type lives in a separate file is not possible with the current test setup and `transpileModule` methods.

**Manual tests**

Tested both cases for exported types in a component started. The generated `component.d.ts` files correctly imported and re-exported the referenced type.

## Other information

There are still some use cases where Stencil will not correctly resolve types (inferred or not). The most prominent instance is if a type is not exported. In this case, the generated `components.d.ts` file will still show the type as `any`. This is because we would need to recursively resolve the string representation of the complex type. This is something we could add in the future, but was out of scope of this work.

### DIFF
--- a/src/compiler/transformers/decorators-to-static/prop-decorator.ts
+++ b/src/compiler/transformers/decorators-to-static/prop-decorator.ts
@@ -174,7 +174,21 @@ const getComplexType = (
   return {
     original: nodeType ? nodeType.getText() : typeToString(typeChecker, type),
     resolved: resolveType(typeChecker, type),
-    references: getAttributeTypeInfo(node, node.getSourceFile(), typeChecker, program),
+    references: getAttributeTypeInfo(
+      // If the node did not explicity have a type set (i.e. `name: string`), then
+      // we can generate a type node via the type checker to resolve references for inferred types.
+      //
+      // This is only a concern with non-primitive types such as a user-defined enum.
+      //
+      // For instance, a @Prop() defined as:
+      // @Prop() mode = ComponentModes.DEFAULT;
+      // Should be able to correctly infer the type to be `ComponentModes` and
+      // resolve the reference to this type for use in generated component type declarations.
+      nodeType ? node : typeChecker.typeToTypeNode(type, undefined, undefined),
+      node.getSourceFile(),
+      typeChecker,
+      program,
+    ),
   };
 };
 
--- a/src/compiler/transformers/decorators-to-static/component-decorator.ts
+++ b/src/compiler/transformers/decorators-to-static/component-decorator.ts
@@ -109,7 +109,9 @@ const validateComponent = (
   if (!config._isTesting) {
     const nonTypeExports = typeChecker
       .getExportsOfModule(typeChecker.getSymbolAtLocation(cmpNode.getSourceFile()))
-      .filter((symbol) => (symbol.flags & (ts.SymbolFlags.Interface | ts.SymbolFlags.TypeAlias)) === 0)
+      .filter(
+        (symbol) => (symbol.flags & (ts.SymbolFlags.Interface | ts.SymbolFlags.TypeAlias | ts.SymbolFlags.Enum)) === 0,
+      )
       .filter((symbol) => symbol.name !== cmpNode.name.text);
 
     nonTypeExports.forEach((symbol) => {
--- a/src/compiler/transformers/test/parse-props.spec.ts
+++ b/src/compiler/transformers/test/parse-props.spec.ts
@@ -34,6 +34,41 @@ describe('parse props', () => {
     expect(t.cmp?.hasProp).toBe(true);
   });
 
+  it('should correctly parse a prop with an inferred enum type', () => {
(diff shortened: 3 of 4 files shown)


src/compiler/transformers/decorators-to-static/prop-decorator.ts
```diff
@@ -174,7 +174,21 @@ const getComplexType = (
...
+      // If the node did not explicity have a type set (i.e. `name: string`), then
+      // we can generate a type node via the type checker to resolve references for inferred types.
+      //
+      // This is only a concern with non-primitive types such as a user-defined enum.
+      //
+      // For instance, a @Prop() defined as:
+      // @Prop() mode = ComponentModes.DEFAULT;
+      // Should be able to correctly infer the type to be `ComponentModes` and
+      // resolve the reference to this type for use in generated component type declarations.
+      nodeType ? node : typeChecker.typeToTypeNode(type, undefined, undefined),
```
alicewriteswrongs, line 187: just making sure I understand, was this change necessary for support enums, or just something to take care of while you're in here?

alicewriteswrongs (commented):
All looks good to me - I was wondering do you think it might be possible to use some of the stuff I put in place in #4212 to handle the case where an enum is not exported? I think a decent string representation of the type should be present in the [type library](https://github.com/ionic-team/stencil/blob/main/src/compiler/transformers/type-library.ts) - if so you might be able to do something like

```ts
getTypeLibrary()[typeId]
```

if you already have the id that is set in the `complexType` for the prop. Maybe that would need to be added in the bit of code that actually generates `components.d.ts`?

src/compiler/transformers/decorators-to-static/prop-decorator.ts
tanner-reits: Yep, it was needed for the enums. In the case where the type is inferred there is no `TypeNode` in the prop declaration, so we can create one from the assignment value.

tanner-reits:
> All looks good to me - I was wondering do you think it might be possible to use some of the stuff I put in place in #4212 to handle the case where an enum is not exported? I think a decent string representation of the type should be present in the [type library](https://github.com/ionic-team/stencil/blob/main/src/compiler/transformers/type-library.ts) - if so you might be able to do something like
> 
> ```ts
> getTypeLibrary()[typeId]
> ```
> 
> if you already have the id that is set in the `complexType` for the prop. Maybe that would need to be added in the bit of code that actually generates `components.d.ts`?

@alicewriteswrongs I can try this out today! I'll do is separate from this work since it's kinda out-of-scope from the original issue I was solving, but would be good for us to support.

alicewriteswrongs (approved):
we talked and think it makes sense to go forward as-is!

PR #4741, opened 2023-08-30
refactor(compiler): use updateConstructor helper in native-constructor.ts

alicewriteswrongs:
We previously had some exceptionally similar code hanging around. This just reuses a helper called `updateConstructor` that was introduced a while back to cut down a bit on code duplication.

As part of that change, the `updateConstructor` function was moved from `convert-decorators.ts` to `transform-utils.ts`.

GitHub Issue Number: N/A

## What is the new behavior?

- 

- [ ] Yes
- [x] No

## Testing

I ran this through Framework as well as our test suite and didn't notice anything amiss.

### DIFF
--- a/src/compiler/transformers/transform-utils.ts
+++ b/src/compiler/transformers/transform-utils.ts
@@ -945,3 +945,97 @@ export const retrieveTsDecorators = (node: ts.Node): ReadonlyArray<ts.Decorator>
 export const retrieveTsModifiers = (node: ts.Node): ReadonlyArray<ts.Modifier> | undefined => {
   return ts.canHaveModifiers(node) ? ts.getModifiers(node) : undefined;
 };
+
+/**
+ * Helper util for updating the constructor on a class declaration AST node.
+ *
+ * @param classNode the class node whose constructor will be updated
+ * @param classMembers a list of class members for that class
+ * @param statements a list of statements which should be added to the
+ * constructor
+ * @returns a list of updated class elements
+ */
+export const updateConstructor = (
+  classNode: ts.ClassDeclaration,
+  classMembers: ts.ClassElement[],
+  statements: ts.Statement[],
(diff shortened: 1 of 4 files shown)


tanner-reits (approved):

src/compiler/transformers/transform-utils.ts
```diff
@@ -945,3 +945,97 @@ export const retrieveTsDecorators = (node: ts.Node): ReadonlyArray<ts.Decorator>
...
+    // into the array of class elements
+    if (needsSuper(classNode)) {
+      statements = [createConstructorBodyWithSuper(), ...statements];
+    }
+
+    // add the new constructor to the class members, putting it at the
+    // beginning
+    classMembers.unshift(
+      ts.factory.createConstructorDeclaration(undefined, [], ts.factory.createBlock(statements, true)),
+    );
```
rwaskiewicz, line 1004: Verifying my understanding - in moving this function from `convert-decorators` to here, is this the only block of code (other than comments) that changed? 

Comparing this against - https://github.com/ionic-team/stencil/pull/4741/files#diff-84d52fb230ffdf800b6b3debb4f554136efd13683b396c03d1dd8a129f81a602L464-L468

```diff
-    classMembers = [
-      ts.factory.createConstructorDeclaration(undefined, [], ts.factory.createBlock(statements, true)),
-      ...classMembers,
-    ];
+    classMembers.unshift(
+     ts.factory.createConstructorDeclaration(undefined, [], ts.factory.createBlock(statements, true)),
+    );
```

alicewriteswrongs: yep! I made that change in order to ensure that the `classMembers` array that's passed in gets modified, since that aligns with how the modification of the constructor for the native constructor transformers was being done

rwaskiewicz (approved):

PR #4746, opened 2023-08-31
feat(compiler): computed properties can be used with Stencil decorators

### LINKED ISSUE #4350: feat: computed @Prop name
### Prerequisites

Could the `@Prop` decorator support computed names similar to the computed property names feature of ES6?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#computed_property_names

### Describe the Use Case

Avoid code repetition when prop names are shared across multiple components in case of a design system library.

```tsx
@Prop() [name]: Type;
```

### Related Code

```tsx
enum Spacing {
    Margin = 'm',
    MarginX = 'mx',
    MarginY = 'my',
};

@Component({
  tag: 'button',
  styleUrl: 'button.css',
  shadow: true,
})
export class MyComponent {
  @Prop() label: string;
  @Prop() [Spacing.Margin]: SpacingSize;

  render() {
    const className = this[Spacing.Margin] && Spacing.Margin + '-' + this[Spacing.Margin];
    return (
      <button class={className}>{this.label}</button>
    );
  }
}
```

### Additional Information

Currently the related code does not compute the prop name while transpiling to React:
```ts
declare namespace LocalJSX {
    interface AdsButton {
        "[Spacing.Margin]"?: SpacingSize;
        "label"?: string;
    }
 }
```

sean-perkins:

GitHub Issue Number: Fixes #4350 

- Using computed property names is not valid for Stencil decorators. Trying something like:

```ts
enum MyEnum {
  first = 'first'
}

...

@Prop() [myEnum.first]: string;
```

Will result in the string `[myEnum.first]` being used as the prop name.

- Using computed values is not valid for Stencil decorator arguments. Trying something like:

```ts
enum MyEnum {
  first = 'first'
}

...

@Prop() first: string;

@Watch(MyEnum.first) 
firstChanged() { }
```

Will throw the following error:
> [ ERROR ]  invalid decorator argument: MyEnum.first at getDeclarationParameter

## What is the new behavior?

- Type checker looks up the symbol reference for a property access expression or computed property name to find the actual value.
- This allows any of the above examples to generate valid Stencil output

- [ ] Yes
- [x] No

## Testing

Manual tested in a new Stencil project. 
- Created a new type for `MyEnum` based on the code snippet shared above
- Updated the `@Prop() first` decorator to `@Prop() [MyEnum.first]`
- Added a `@Watch` decorator to `MyComponent.tsx` 
- Logged the result in the callback
- Used a local dev build of Stencil with the modified change
- Verified log statement logs when watch callback fires
- Verified no exceptions from the compiler are thrown

### DIFF
--- a/src/compiler/transformers/decorators-to-static/decorator-utils.ts
+++ b/src/compiler/transformers/decorators-to-static/decorator-utils.ts
@@ -1,29 +1,35 @@
 import ts from 'typescript';
 
 import { objectLiteralToObjectMap } from '../transform-utils';
-import { StencilDecorator } from './decorators-constants';
+import type { StencilDecorator } from './decorators-constants';
 
-export const getDeclarationParameters: GetDeclarationParameters = (decorator: ts.Decorator): any => {
+export const getDeclarationParameters: GetDeclarationParameters = (
+  decorator: ts.Decorator,
+  typeChecker: ts.TypeChecker,
+): any => {
   if (!ts.isCallExpression(decorator.expression)) {
     return [];
   }
-  return decorator.expression.arguments.map(getDeclarationParameter);
+  return decorator.expression.arguments.map((arg) => getDeclarationParameter(arg, typeChecker));
 };
 
-const getDeclarationParameter = (arg: ts.Expression): any => {
+const getDeclarationParameter = (arg: ts.Expression, typeChecker: ts.TypeChecker): any => {
   if (ts.isObjectLiteralExpression(arg)) {
     return objectLiteralToObjectMap(arg);
   } else if (ts.isStringLiteral(arg)) {
     return arg.text;
   } else if (ts.isPropertyAccessExpression(arg)) {
-    /**
-     * Enum members are property access expressions, so we can evaluate them
-     * to get the enum member value as a string.
-     *
-     * This enables developers to use enum members in decorators.
-     * e.g. @Watch(MyEnum.VALUE)
-     */
-    return arg.name.getText();
+    const symbol = typeChecker.getTypeAtLocation(arg);
+    if (symbol !== undefined && symbol.isLiteral()) {
+      /**
+       * Enum members are property access expressions, so we can evaluate them
+       * to get the enum member value as a string.
+       *
+       * This enables developers to use enum members in decorators.
+       * e.g. @Watch(MyEnum.VALUE)
+       */
+      return symbol.value;
+    }
   }
 
   throw new Error(`invalid decorator argument: ${arg.getText()}`);
@@ -45,7 +51,7 @@ export const isDecoratorNamed = (propName: StencilDecorator) => {
 };
 
 export interface GetDeclarationParameters {
-  <T>(decorator: ts.Decorator): [T];
-  <T, T1>(decorator: ts.Decorator): [T, T1];
-  <T, T1, T2>(decorator: ts.Decorator): [T, T1, T2];
+  <T>(decorator: ts.Decorator, typeChecker: ts.TypeChecker): [T];
+  <T, T1>(decorator: ts.Decorator, typeChecker: ts.TypeChecker): [T, T1];
+  <T, T1, T2>(decorator: ts.Decorator, typeChecker: ts.TypeChecker): [T, T1, T2];
 }
--- a/src/compiler/transformers/decorators-to-static/decorator-utils.ts
+++ b/src/compiler/transformers/decorators-to-static/decorator-utils.ts
@@ -22,11 +22,10 @@ const getDecoratorParameter = (arg: ts.Expression, typeChecker: ts.TypeChecker):
     const type = typeChecker.getTypeAtLocation(arg);
     if (type !== undefined && type.isLiteral()) {
       /**
-       * Enum members are property access expressions, so we can evaluate them
-       * to get the enum member value as a string.
-       *
-       * This enables developers to use enum members in decorators.
-       * e.g. @Watch(MyEnum.VALUE)
+       * Using enums or variables require us to resolve the value for
+       * the computed property/identifier via the TS type checker. As long
+       * as the type resolves to a literal, we can grab its value to be used
+       * as the `@Watch()` decorator argument.
        */
       return type.value;
     }
--- a/src/compiler/transformers/decorators-to-static/decorator-utils.ts
+++ b/src/compiler/transformers/decorators-to-static/decorator-utils.ts
@@ -19,8 +19,8 @@ const getDeclarationParameter = (arg: ts.Expression, typeChecker: ts.TypeChecker
   } else if (ts.isStringLiteral(arg)) {
     return arg.text;
   } else if (ts.isPropertyAccessExpression(arg)) {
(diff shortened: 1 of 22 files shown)


sean-perkins:
This is actually a bit more complicated than anticipated here. Currently this is using the key of the enum as the property name, not the actual assigned string. Need to investigate how to accomplish this/if this is possible.

alicewriteswrongs (changes requested):
I tried this out locally and I was able to do something along the lines of your example, with a component [like this](https://gist.github.com/alicewriteswrongs/0a9844d33b0a0d18e3d62a06ece0b0a0). I think this all looks good though! I just had one code question about a function name

src/compiler/transformers/decorators-to-static/decorator-utils.ts
```diff
@@ -1,20 +1,35 @@
...
+  } else if (ts.isPropertyAccessExpression(arg) || ts.isIdentifier(arg)) {
+    const type = typeChecker.getTypeAtLocation(arg);
+    if (type !== undefined && type.isLiteral()) {
+      /**
+       * Enum members are property access expressions, so we can evaluate them
+       * to get the enum member value as a string.
+       *
+       * This enables developers to use enum members in decorators.
+       * e.g. @Watch(MyEnum.VALUE)
+       */
```
rwaskiewicz, line 30: We may wanna tweak this comment a bit - I think the scope of this has changed quite a bit now (since we support all property expr + identifiers that are literals for decorators other than `@Watch`)

rwaskiewicz (changes requested):
A couple thoughts/asks.

Unsure who the current owner of this is, so I wrote this with @sean-perkins in mind here:

## PR Name 
I updated the name of the PR to use ‘feat’ instead of ‘fix’ to ensure this gets listed as a ‘feature’ in the auto-generated `CHANGELOG.md`. No action required here.
## Karma
Can we add a few Karma tests to this PR? Karma tests are helpful for us in this particular case in that they:
1. Help validate that “real world” Stencil components that use compile correctly
2. Help exercise the runtime to ensure that a component is properly rendered

I think it’d be useful for us to create three separate Karma test suites here, one for each of the following use cases:
1. Using a computed property in a `@Prop()` class member, validating that we can render the value stored in the computed property name (if there is one). This would in turn exercise property name resolution and that we can render it (over multiple render cycles). It might be nice to have two `@Prop()`s, one with a default value and one without. Not strictly required, but it might be worth looking into setting `attribute` and `reflect` to `true` in two additional tests to verify that those continue to act as expected.
2. Using a computed property in a `@Watch()` class member, validating that the function tied to the computed name gets called 
3. Using a computed property on both `@Watch()` and `@Prop()`, verifying the aforementioned works, that the correct property name gets piped into the third argument of the function decorated with `@Watch()`

Typically, we create a Karma test suite by copy/pasting an existing one. I’ve also created a tiny template in the following patch:
```diff --git a/test/karma/test-app/karma-template/cmp-root.tsx b/test/karma/test-app/karma-template/cmp-root.tsx
new file mode 100644
index 000000000..2030e7e21
--- /dev/null
+++ b/test/karma/test-app/karma-template/cmp-root.tsx
@@ -0,0 +1,13 @@
+import { Component, Prop, h } from '@stencil/core';
+
+@Component({
+  tag: 'karma-template',
+  shadow: true,
+})
+export class KarmaTemplate {
+  @Prop() myProp: string;
+
+  render() {
+    return <div>Hello World! Prop is: {this.myProp}</div>;
+  }
+}
diff --git a/test/karma/test-app/karma-template/index.html b/test/karma/test-app/karma-template/index.html
new file mode 100644
index 000000000..75cfddc5b
--- /dev/null
+++ b/test/karma/test-app/karma-template/index.html
@@ -0,0 +1,6 @@
+<!DOCTYPE html>
+<meta charset="utf8">
+<script src="./build/testapp.esm.js" type="module"></script>
+<script src="./build/testapp.js" nomodule></script>
+
+<karma-template my-prop="my-value"></karma-template>
diff --git a/test/karma/test-app/karma-template/karma.spec.ts b/test/karma/test-app/karma-template/karma.spec.ts
new file mode 100644
index 000000000..168084aca
--- /dev/null
+++ b/test/karma/test-app/karma-template/karma.spec.ts
@@ -0,0 +1,15 @@
+import { setupDomTests } from '../util';
+
+describe('karma-template', () => {
+  const { setupDom, tearDownDom } = setupDomTests(document);
+  let app: HTMLElement;
+
+  beforeEach(async () => {
+    app = await setupDom('/karma-template/index.html');
+  });
+  afterEach(tearDownDom);
+
+  it('has the correct @Prop() value', () => {
+    expect(app.querySelector('karma-template').shadowRoot.textContent).toBe('Hello World! Prop is: my-value');
+  });
+});
```

To get started, you can apply this patch (`git apply karma-template.patch`) - next, replace all instances of `karma-template` and `KarmaTemplate` with a more appropriate name. From the `test/karma` directory, you can run `npm start` to kick off a dev server and nav to `test-www/RENAMED_FILE_NAME` to see the component that will be rendered/tested. From there, the test file/component can be updated to be more representative of the functionality we want to test. That’s a lot, and I’m sure I missed something, so please let me know if I’ve missed anything!

[karma-template.patch](https://github.com/ionic-team/stencil/files/12562900/karma-template.patch)

src/compiler/transformers/decorators-to-static/decorator-utils.ts
```diff
@@ -1,20 +1,35 @@
...
 
-const getDeclarationParameter = (arg: ts.Expression): any => {
+const getDeclarationParameter = (arg: ts.Expression, typeChecker: ts.TypeChecker): any => {
   if (ts.isObjectLiteralExpression(arg)) {
     return objectLiteralToObjectMap(arg);
   } else if (ts.isStringLiteral(arg)) {
     return arg.text;
+  } else if (ts.isPropertyAccessExpression(arg)) {
+    const type = typeChecker.getTypeAtLocation(arg);
+    if (type !== undefined && type.isLiteral()) {
```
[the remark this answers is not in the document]
tanner-reits: I don't think we need to worry about this, at least right now, considering that a user can directly use any of these values for a class property and get the same result.

```tsx
@Prop() '1e0': string;
@Prop() '1': string;
@Prop() 1: string;
```

If this was something we'd wanna protect against, I say it falls outside the scope of this work and would be a separate enhancement.

alicewriteswrongs (approved):
Just tried this out locally and it looks like everything is working to me!

test/karma/test-app/computed-properties-prop-decorator/karma.spec.ts
```diff
@@ -1,6 +1,6 @@
 import { setupDomTests, waitForChanges } from '../util';
 
-describe('conditional-basic', function () {
+describe('computed-properties-peop-decorator', function () {
```
rwaskiewicz, line 3: Minor - 'peop'

rwaskiewicz (approved):
LGTM! One non-blocking typo

PR #4747, opened 2023-09-01
chore(spelling): spellcheck comments in TypeScript code using cspell

alicewriteswrongs:
This adds [cspell](https://cspell.org/) to do spell-checking and configures it to check comments within TypeScript files and Markdown. A wordlist, based on the one we use for the Stencil website, is added as well.

GitHub Issue Number: N/A

## What is the new behavior?

- 

- [ ] Yes
- [x] No


scripts/bundles/internal-platform-hydrate.ts
```diff
@@ -51,12 +51,12 @@ export async function internalHydrate(opts: BuildOptions) {
...
      *
-     * If you say `moduleSideEffects: false` you're basically saying 'dont assume
-     * that imported a module has a desired side effect' which allows an import that
+     * If you say `moduleSideEffects: false` you're basically saying "don't assume
+     * that imported a module has a desired side effect" which allows an import that
      * doesn't result in any symbols being pulled into the importing module to be
      * eliminated.
      *
-     * Another way to say this is that without this change the treeshaking
+     * Another way to say this is that without this change the tree-shaking
```
alicewriteswrongs, line 59: some of these things are sort of "style" questions - most usage around the web seems to be "tree shaking" or "tree-shaking" (eg <https://developer.mozilla.org/en-US/docs/Glossary/Tree_shaking>, <https://rollupjs.org/faqs/#what-is-tree-shaking>) but comparatively little usage of "treeshaking". I went with "tree-shaking" throughout, but with things like this I think generally it should be just like formatting with prettier, i.e. if we decide we want on variant to the exclusion of the others we can use `!tree-shaking` type rules in CSpell to forbid certain variants.

cspell.json
```diff
@@ -0,0 +1,17 @@
...
+  "dictionaryDefinitions": [
+    {
+      "name": "custom-words",
+      "path": "./cspell-wordlist.txt",
+      "addWords": true
+    }
+  ],
+
+  "ignoreRegExpList": [
+    "/`[a-zA-Z-_., /*']+`/g"
```
alicewriteswrongs, line 12: together with `"CStyleComment"` in the `"includeRegExpList"` this excludes things surrounding with backticks from spellchecking, letting us reference variable or function names which don't spellcheck in comments (you can see a few examples where I put backticks around some words that weren't previously backtick-ed, this is why)

tanner-reits (approved):
LGTM! Cool stuff!

rwaskiewicz (approved):

PR #4749, opened 2023-09-01
chore(snc): fix 5 SNC errors in logger-rollup.ts

alicewriteswrongs:
Just fixing a few SNC errors I found in here

GitHub Issue Number: N/A

## What is the new behavior?

- 

- [ ] Yes
- [x] No

### DIFF
--- a/src/utils/logger/logger-rollup.ts
+++ b/src/utils/logger/logger-rollup.ts
@@ -47,18 +47,18 @@ export const loadRollupDiagnostics = (
           try {
             const srcLines = splitLineBreaks(sourceText);
 
-            const errorLine: d.PrintLine = {
+            const errorLine = {
               lineIndex: loc.line - 1,
               lineNumber: loc.line,
               text: srcLines[loc.line - 1],
               errorCharStart: loc.column,
               errorLength: 0,
-            };
+            } satisfies d.PrintLine;
 
             diagnostic.lineNumber = errorLine.lineNumber;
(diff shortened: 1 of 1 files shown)


src/utils/logger/logger-rollup.ts
```diff
@@ -47,18 +47,18 @@ export const loadRollupDiagnostics = (
...
               errorCharStart: loc.column,
               errorLength: 0,
-            };
+            } satisfies d.PrintLine;
 
             diagnostic.lineNumber = errorLine.lineNumber;
             diagnostic.columnNumber = errorLine.errorCharStart;
 
-            const highlightLine = errorLine.text.slice(loc.column);
+            const highlightLine = errorLine.text?.slice(loc.column);
```
rwaskiewicz, line 61: Should we default this variable to something like an empty string? Although this fixes the SNC, wouldn't this still fail on the next line (and the line after)?
```
            for (let i = 0; i < highlightLine.length; i++) { // accessing length
              if (charBreak.has(highlightLine.charAt(i))) { // calling charAt()
```
```suggestion
            const highlightLine = errorLine.text?.slice(loc.column) ?? '';
```

alicewriteswrongs: ah true! I'll add a lil `??`

rwaskiewicz (approved):

tanner-reits (approved):

PR #4760, opened 2023-09-06
feat(runtime): watch native HTML attributes

tanner-reits:

Stencil does not allow developers to watch native HTML attributes via the `@Watch()` decorator. So, syntax like the following is not valid:

```tsx
@Watch('aria-label')
onAriaLabelChange() {
  // do something
}
```

In fact, the compiler will throw an error if the string provided to the decorator is not a Stencil class "member" (state or prop).

GitHub Issue Number: N/A

## What is the new behavior?

The `@Watch()` decorator can now be used with both Stencil class "member" (state or prop), as well as any native HTML attributes like aria attributes, `src`, etc.

- [ ] Yes
- [X] No

## Testing

**Automated Tests**

All existing tests continue to pass and a new e2e test was added to test the behavior with a native attribute.

**Manual Tests**

I tested this with both lazy and custom-elements builds. A build of this commit can be used to test in a component starter for lazy builds. 

## Other information

The Framework team has asked for us to support this for some time.

### DIFF
--- a/src/compiler/transformers/decorators-to-static/convert-decorators.ts
+++ b/src/compiler/transformers/decorators-to-static/convert-decorators.ts
@@ -119,7 +119,7 @@ const visitClassDeclaration = (
       filteredMethodsAndFields,
     );
     elementDecoratorsToStatic(diagnostics, decoratedMembers, typeChecker, filteredMethodsAndFields);
-    watchDecoratorsToStatic(config, diagnostics, decoratedMembers, watchable, filteredMethodsAndFields);
+    watchDecoratorsToStatic(decoratedMembers, filteredMethodsAndFields);
     listenDecoratorsToStatic(diagnostics, decoratedMembers, filteredMethodsAndFields);
   }
 
--- a/src/compiler/transformers/decorators-to-static/convert-decorators.ts
+++ b/src/compiler/transformers/decorators-to-static/convert-decorators.ts
@@ -103,11 +103,10 @@ const visitClassDeclaration = (
   componentDecoratorToStatic(config, typeChecker, diagnostics, classNode, filteredMethodsAndFields, componentDecorator);
 
   // stores a reference to fields that should be watched for changes
-  const watchable = new Set<string>();
   // parse member decorators (Prop, State, Listen, Event, Method, Element and Watch)
   if (decoratedMembers.length > 0) {
-    propDecoratorsToStatic(diagnostics, decoratedMembers, typeChecker, program, watchable, filteredMethodsAndFields);
-    stateDecoratorsToStatic(decoratedMembers, watchable, filteredMethodsAndFields);
+    propDecoratorsToStatic(diagnostics, decoratedMembers, typeChecker, program, filteredMethodsAndFields);
+    stateDecoratorsToStatic(decoratedMembers, filteredMethodsAndFields);
     eventDecoratorsToStatic(diagnostics, decoratedMembers, typeChecker, program, filteredMethodsAndFields);
     methodDecoratorsToStatic(
       config,
--- a/src/compiler/transformers/decorators-to-static/watch-decorator.ts
+++ b/src/compiler/transformers/decorators-to-static/watch-decorator.ts
@@ -1,4 +1,4 @@
-import { augmentDiagnosticWithNode, buildError, buildWarn, flatOne } from '@utils';
+import { flatOne } from '@utils';
 import ts from 'typescript';
 
 import type * as d from '../../../declarations';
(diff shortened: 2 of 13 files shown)


src/compiler/transformers/decorators-to-static/watch-decorator.ts
```diff
@@ -23,22 +15,12 @@ export const watchDecoratorsToStatic = (
...
   const methodName = method.name.getText();
   const decorators = retrieveTsDecorators(method) ?? [];
   return decorators.filter(isDecoratorNamed('Watch')).map((decorator) => {
     const [propName] = getDeclarationParameters<string>(decorator);
-    if (!watchable.has(propName)) {
-      const diagnostic = config.devMode ? buildWarn(diagnostics) : buildError(diagnostics);
-      diagnostic.messageText = `@Watch('${propName}') is trying to watch for changes in a property that does not exist.
-        Make sure only properties decorated with @State() or @Prop() are watched.`;
-      augmentDiagnosticWithNode(diagnostic, decorator);
-    }
```
tanner-reits, line 41: We do lose a compile-time check with this change. Only way around it would be to maintain a list of all the valid HTML attributes to check against which would be a nightmare.

src/declarations/stencil-private.ts
```diff
@@ -1386,6 +1386,9 @@ export type ComponentRuntimeMetaCompact = [
 
   /** listeners */
   ComponentRuntimeHostListener[]?,
+
+  /** watchers */
+  ComponentConstructorWatchers?,
```
tanner-reits, line 1391: Needed to add this so we can have a map of the watchers available when bootstrapping the lazy-build before the module is loaded. Otherwise, the class instance that gets registered with the custom element registry won't include any of the watched native attributes in the `observedAttributes` array.

src/runtime/proxy-component.ts
```diff
@@ -133,25 +133,60 @@ export const proxyComponent = (
             // if the propName exists on the prototype of `Cstr`, this update may be a result of Stencil using native
             // APIs to reflect props as attributes. Calls to `setAttribute(someElement, propName)` will result in
             // `propName` to be converted to a `DOMString`, which may not be what we want for other primitive props.
+            return;
+          } else if (propName == null) {
```
tanner-reits, line 137: Most of the stuff in this block comes from the existing [`setValue()`](https://github.com/ionic-team/stencil/blob/main/src/runtime/set-value.ts#L11C2-L11C2) function

src/runtime/proxy-component.ts
```diff
@@ -133,25 +133,60 @@ export const proxyComponent = (
...
-        });
+      // Create an array of attributes to observe
+      // This list in comprised of all strings used within a `@Watch()` decorator
+      // on a component as well as any Stencil-specific "members" (`@Prop()`s and `@State()`s).
+      // As such, there is no way to guarantee type-safety here that a user hasn't entered
+      // an invalid attribute.
+      Cstr.observedAttributes = Array.from(
+        new Set([
+          ...Object.keys(cmpMeta.$watchers$ ?? {}),
+          ...members
```
tanner-reits, line 177: If we don't include all "members" here (regardless if they have an associate watcher), then they components will not set their values and update appropriately.

tanner-reits (commented):
Just noting a few things.

Associated docs: https://github.com/ionic-team/stencil-site/pull/1224

src/compiler/transformers/decorators-to-static/convert-decorators.ts
alicewriteswrongs, line 1: github won't let me comment on the appropriate line directly, but does removing the `watchable` arg  from `watchDecoratorsToStatic` mean we can remove it from `propDecoratorsToStatic` and `stateDecoratorsToStatic`?

https://github.com/ionic-team/stencil/blob/b2440d6eb76d192ae7c440bc07be1b6a30db0d66/src/compiler/transformers/decorators-to-static/convert-decorators.ts#L106-L110

Looking through `parsePropDecorator` and `parseStateDecorator` it doesn't _look_ like they do anything with `watchable` other than add their prop names to it, so if we're no longer using it for `watchDecoratorsToStatic` it might be superfluous now

src/compiler/transformers/decorators-to-static/watch-decorator.ts
```diff
@@ -1,20 +1,12 @@
...
-  diagnostics: d.Diagnostic[],
-  decoratedProps: ts.ClassElement[],
-  watchable: Set<string>,
-  newMembers: ts.ClassElement[],
-) => {
-  const watchers = decoratedProps
-    .filter(ts.isMethodDeclaration)
-    .map((method) => parseWatchDecorator(config, diagnostics, watchable, method));
+export const watchDecoratorsToStatic = (decoratedProps: ts.ClassElement[], newMembers: ts.ClassElement[]) => {
+  const watchers = decoratedProps.filter(ts.isMethodDeclaration).map((method) => parseWatchDecorator(method));
```
alicewriteswrongs, line 9: super nit: this could be slightly more concise

```suggestion
  const watchers = decoratedProps.filter(ts.isMethodDeclaration).map(parseWatchDecorator);
```

alicewriteswrongs (changes requested):
I think this all generally makes sense to me, just had a few thoughts as I was reading through

src/compiler/transformers/decorators-to-static/convert-decorators.ts
tanner-reits: Ah, yes, I think you're right! Good catch!

alicewriteswrongs (approved):
lgtm!

src/compiler/transformers/decorators-to-static/watch-decorator.ts
sean-perkins: In Framework we maintain a list of attributes to inherit here: https://github.com/ionic-team/ionic-framework/blob/f82709595d5031a759e35b11dbcccb0ea5870ce0/core/src/utils/helpers.ts#L118-L175

But our reasoning is to not inherit all attributes that exist on the host. I'm not really sure why you'd restrict what attribute a developer can watch a value change on. That's based on the implementation details of the component, not a compiler constraint.

rwaskiewicz: @sean-perkins That's great to know!

Getting that context is what I'm really after here to both:
1. influence my understanding of this particular piece of code
2. add historical context to this PR - this helps prevent the issue where we don't capture this info anywhere in VCS/issue tracking and that information gets lost (which is sadly, lacking in older commits in Stencil/something I'd like us to prevent reverting to)

In that case, I think this code/logic fine as is 👍

rwaskiewicz (approved):

PR #4762, opened 2023-09-06
refactor(compiler): create a util function for name of a property decl

alicewriteswrongs:
Pull some code out of `convert-decorators.ts` into a helper function in our transformer utility module for getting the name of a property declaration as a string. If you had a class like this:

```ts
class {
  foo: "bar";
}
```

this utility would let you get the string `"foo"` given the TypeScript syntax tree node corresponding to the `foo: "bar";` line in the source.

GitHub Issue Number: N/A

## What is the new behavior?

- 

- [ ] Yes
- [x] No

### DIFF
--- a/src/compiler/transformers/decorators-to-static/convert-decorators.ts
+++ b/src/compiler/transformers/decorators-to-static/convert-decorators.ts
@@ -2,7 +2,12 @@ import { augmentDiagnosticWithNode, buildError } from '@utils';
 import ts from 'typescript';
 
 import type * as d from '../../../declarations';
-import { retrieveTsDecorators, retrieveTsModifiers, updateConstructor } from '../transform-utils';
+import {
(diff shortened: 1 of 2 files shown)


rwaskiewicz (commented):
The value in pulling this out isn't super clear to me here - is this something that we wanna test in isolation? Does it get used in an incoming ElementInternals PR?

The code itself makes sense, but I'm not sure as to the "why" here - can you help me with that/fill that in the PR summary?

alicewriteswrongs:
was going to put a note in slack and forgot - this is something I need for the ElementInternals stuff, just trying to keep that PR smaller (same for this: https://github.com/ionic-team/stencil/pull/4741)

rwaskiewicz (approved):
Thanks for the explanation! Makes sense!

tanner-reits (approved):

PR #4767, opened 2023-09-07
fix(compiler): restrict config extras slot fix flags

tanner-reits:

A user can currently set both the `experimentalSlotFixes` flag and any of the other scoped slot patch flags (like `slotChildNodesFix`). So a config may look like:

```ts
// stencil.config.ts

extras: {
  experimentalSlotFixes: true,
  slotChildNodesFix: true
}
```

This can be confusing since the `experimentalSlotFixes` flag will override any of the other patch flags.

GitHub Issue Number: N/A

## What is the new behavior?

The config type (and downstream validation logic) will only allow the individual patch flags to be set if `experimentalSlotFixes` is omitted or explicitly set `false`

```ts
// stencil.config.ts

extras: {
  experimentalSlotFixes: true,
  slotChildNodesFix: true // <-- valid
}
```

```ts
// stencil.config.ts

extras: {
  experimentalSlotFixes: false,
  slotChildNodesFix: true // <-- valid
}
```

```ts
// stencil.config.ts

extras: {
  slotChildNodesFix: true // <-- valid
}
```

```ts
// stencil.config.ts

extras: {
  experimentalSlotFixes: true,
  slotChildNodesFix: false // <-- invalid
}
```

If a user does build a project with one of the invalid config options, a warning will be logged at compile time stating that the individual flags will be overridden based on `experimentalSlotFixes`.

- [ ] Yes
- [X] No

## Testing

**Automated tests**
Automated tests for config validation were updated accordingly.

**Manual tests**
A build of this commit can be installed in a component starter. In the `stencil.config.ts`, verify that the following results in intellisense errors and a compile-time warning during `npm run build`:

```ts
extras: {
    experimentalSlotFixes: true,
    slotChildNodesFix: false,
}
```

### DIFF
--- a/src/compiler/config/validate-config.ts
+++ b/src/compiler/config/validate-config.ts
@@ -112,32 +112,31 @@ export const validateConfig = (
   // TODO(STENCIL-914): remove when `experimentalSlotFixes` is the default behavior
   // If the user set `experimentalSlotFixes` and any individual slot fix flags, we need to log a warning
   // to the user that we will "override" the individual flags
-  if (
-    validatedConfig.extras.experimentalSlotFixes === true &&
-    ((validatedConfig.extras as any).appendChildSlotFix === true ||
-      (validatedConfig.extras as any).cloneNodeFix === true ||
-      (validatedConfig.extras as any).slotChildNodesFix === true ||
-      (validatedConfig.extras as any).scopedSlotTextContentFix === true)
-  ) {
-    const warning = buildError(diagnostics);
-    warning.level = 'warn';
-    warning.messageText =
-      'The `experimentalSlotFixes` flag cannot be used in combination with other slot fix flags. These individual flags will be ignored. Please update your Stencil config accordingly.';
-  }
+  // if (
+  //   validatedConfig.extras.experimentalSlotFixes === true &&
+  //   ((validatedConfig.extras as any).appendChildSlotFix === true ||
+  //     (validatedConfig.extras as any).cloneNodeFix === true ||
+  //     (validatedConfig.extras as any).slotChildNodesFix === true ||
+  //     (validatedConfig.extras as any).scopedSlotTextContentFix === true)
+  // ) {
+  //   const warning = buildError(diagnostics);
+  //   warning.level = 'warn';
+  //   warning.messageText =
+  //     'The `experimentalSlotFixes` flag cannot be used in combination with other slot fix flags. These individual flags will be ignored. Please update your Stencil config accordingly.';
+  // }
 
   // TODO(STENCIL-914): remove `experimentalSlotFixes` when it's the default behavior
   validatedConfig.extras.experimentalSlotFixes = !!validatedConfig.extras.experimentalSlotFixes;
-  if (validatedConfig.extras.experimentalSlotFixes === false) {
+  if (validatedConfig.extras.experimentalSlotFixes === true) {
+    validatedConfig.extras.appendChildSlotFix = true;
+    validatedConfig.extras.cloneNodeFix = true;
+    validatedConfig.extras.slotChildNodesFix = true;
+    validatedConfig.extras.scopedSlotTextContentFix = true;
+  } else {
     validatedConfig.extras.appendChildSlotFix = !!validatedConfig.extras.appendChildSlotFix;
     validatedConfig.extras.cloneNodeFix = !!validatedConfig.extras.cloneNodeFix;
     validatedConfig.extras.slotChildNodesFix = !!validatedConfig.extras.slotChildNodesFix;
     validatedConfig.extras.scopedSlotTextContentFix = !!validatedConfig.extras.scopedSlotTextContentFix;
-  } else {
-    // Make sure the options don't exist on the config at all
-    delete (validatedConfig.extras as any).appendChildSlotFix;
-    delete (validatedConfig.extras as any).cloneNodeFix;
-    delete (validatedConfig.extras as any).slotChildNodesFix;
-    delete (validatedConfig.extras as any).scopedSlotTextContentFix;
   }
 
   validatedConfig.buildEs5 =
--- a/src/compiler/config/validate-config.ts
+++ b/src/compiler/config/validate-config.ts
@@ -1,7 +1,14 @@
 import { createNodeLogger, createNodeSys } from '@sys-api-node';
 import { buildError, isBoolean, isNumber, isString, sortBy } from '@utils';
 
-import { ConfigBundle, Diagnostic, LoadConfigInit, UnvalidatedConfig, ValidatedConfig } from '../../declarations';
+import {
+  ConfigBundle,
+  ConfigExtras,
+  Diagnostic,
+  LoadConfigInit,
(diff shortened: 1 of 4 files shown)


src/declarations/stencil-public-compiler.ts
```diff
@@ -367,19 +376,26 @@ export interface ConfigExtras {
...
+    | ConfigExtrasSlotFixes
+    | {
+        // TODO(STENCIL-914): remove `experimentalSlotFixes` when it's the default behavior
+        /**
+         * Enables all slot-related fixes such as {@link slotChildNodesFix}, and
+         * {@link scopedSlotTextContentFix}.
+         */
+        experimentalSlotFixes?: true;
+      }
+  );
```
tanner-reits, line 398: So, with this approach, even a config like the following would be invalid via the TS language server, but wouldn't throw a compiler warning:

```ts
extras: {
  experimentalSlotFixes: true,
  slotChildNodesFix: false // <-- doesn't like that this is specified
}
```

This may be confusing. We could, alternatively, have the types setup where if `experimentalSlotFixes: true` is set, then all other flags _could_ be specified, but would need to be `false`:

```ts
extras: {
  experimentalSlotFixes: true,
  slotChildNodesFix: false // <-- now this would be valid
}
```

Likewise, the logic for `experimentalSlotFixes: false`, would work as it does with the current implementation.

Making this change could revert the changes in [this](https://github.com/ionic-team/stencil/pull/4767/files#diff-bd7059cb44e3fa48f56fceff1b07ccba76c002943745485cf28d32b81449e024) file and reduce/remove the `any` casts in the [validation tests](https://github.com/ionic-team/stencil/pull/4767/files#diff-164622140dac3d1aa088d2b1c0f6b8de0630cf3b715c25b301c54208f5d3b5bf)

Lemme know what you guys think.

src/compiler/config/validate-config.ts
```diff
@@ -104,24 +104,40 @@ export const validateConfig = (
...
+    validatedConfig.extras.experimentalSlotFixes === true &&
+    ((validatedConfig.extras as any).appendChildSlotFix === true ||
+      (validatedConfig.extras as any).cloneNodeFix === true ||
+      (validatedConfig.extras as any).slotChildNodesFix === true ||
+      (validatedConfig.extras as any).scopedSlotTextContentFix === true)
+  ) {
+    const warning = buildError(diagnostics);
+    warning.level = 'warn';
+    warning.messageText =
+      'The `experimentalSlotFixes` flag cannot be used in combination with other slot fix flags. These individual flags will be ignored. Please update your Stencil config accordingly.';
```
tanner-reits, line 125: I'm no wordsmith so open to better phrasing here 😂

src/declarations/stencil-public-compiler.ts
rwaskiewicz: > This may be confusing. We could, alternatively, have the types setup where if experimentalSlotFixes: true is set, then all other flags could be specified, but would need to be false

I like this approach better, but I'm also happy with how it's working ATM - so I suppose I only lean towards this approach (requiring if the other flags are set, they have to be false) for removing type assertions

src/compiler/config/validate-config.ts
rwaskiewicz: Based on the outcome of https://github.com/ionic-team/stencil/pull/4767/files#r1318694085, this might be moot, but one thing we could do is collect a `Set` of flags that are erroneously set:

```ts
const flagsToCollect = new Set<string>();
if (validatedConfig.extras.experimentalSlotFixes === true) {
  ['appendChildSlotFix', 'slotChildNodesFix', 'cloneNodeFix', 'scopedSlotTextContentFix'].forEach((flag) => {
    if (validatedConfig.extras[flag] === true) {
      flagsToCollect.add(flag);
    }
  });
}
```

then print them out to help folks understand what a 'slot fix flag' "means":
```ts
    warning.messageText =
      'The `experimentalSlotFixes` flag cannot be used in combination when the following flags are set: ${[...flagsToCollect].join(',')}. These individual flags will be ignored. Please update your Stencil config accordingly.';
```

src/compiler/config/validate-config.ts
```diff
@@ -104,24 +104,43 @@ export const validateConfig = (
...
-  // TODO(STENCIL-914): remove `experimentalSlotFixes` when it's the default behavior
-  validatedConfig.extras.experimentalSlotFixes = !!validatedConfig.extras.experimentalSlotFixes;
 
-  // if the user has set `extras.experimentalSlotFixes` then we turn on all of
-  // the slot-related fixes automatically.
+  // TODO(STENCIL-914): remove when `experimentalSlotFixes` is the default behavior
+  // If the user set `experimentalSlotFixes` and any individual slot fix flags to `false`, we need to log a warning
+  // to the user that we will "override" the individual flags
+  if (validatedConfig.extras.experimentalSlotFixes === true) {
+    const possibleFlags: (keyof ValidatedConfig['extras'])[] = [
```
rwaskiewicz, line 116: I think this array is causing a couple new SNC violations.

Because `ValidatedConfig#Extras` is an optional type, accessing it as `ValidatedConfig['extras']` creates a type of `never` (since it can be `undefined`).

I think this should be fixable by using the `ConfigExtras` type directly:
```suggestion
    const possibleFlags: (keyof ConfigExtras)[] = [
```

rwaskiewicz (approved):
Approved with a handful of non-blocking comments

src/declarations/stencil-public-compiler.ts
```diff
@@ -358,28 +331,61 @@ export interface ConfigExtras {
...
 
+  /**
+   * Enables the tagNameTransform option of `defineCustomElements()`, so the component tagName
+   * can be customized at runtime. Defaults to `false`.
+   */
+  tagNameTransform?: boolean;
+}
+
+// TODO(STENCIL-914): delete this interface when `experimentalSlotFixes` is the default behavior
+type ConfigExtrasSlotFixes<ExpirimentalFixesEnabled extends boolean, IndividualFlags extends boolean> = {
```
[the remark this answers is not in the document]
tanner-reits: cspell not doing its job 🤦‍♂️

rwaskiewicz (approved):

src/declarations/stencil-public-compiler.ts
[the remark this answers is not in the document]
alicewriteswrongs: it's not set yet to check variable names and whatnot - we could turn that on for sure! but we'd have to fix a lot of things and / or add a lot of "words" like "splt" to the dictionary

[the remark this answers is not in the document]
tanner-reits: I just meant locally (I have the extension for VsCode)

alicewriteswrongs (approved):
I think this all looks good, just had one question basically about the wording

src/declarations/stencil-public-compiler.ts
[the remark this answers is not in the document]
rwaskiewicz: I wonder if adding it to the repo overrides your previous settings 🤔

PR #4770, opened 2023-09-08
chore(snc): resolve snc in test init

rwaskiewicz:

GitHub Issue Number: N/A

## What is the new behavior?

this commit resolves a handful of strict null checks violations in `testing.ts`. most violations involved widening the type of some variable to explicitly state it could be null. this ought to be a safe operation, as we were already assigning `null` to these variables already - the change only aligns the type system with what was originally happening at runtime.

an extraneous call to `getAppStyleUrl` was removed as well. there are no known side effects that occur as a result of that function call, which ought to make this refactor safe

- [ ] Yes
- [x] No

## Testing

Type checker remains happy - there's only one section that could _potentially_ change how Stencil runs, I'll call that out below

### DIFF
--- a/src/testing/testing.ts
+++ b/src/testing/testing.ts
@@ -24,14 +24,20 @@ export const createTesting = async (config: ValidatedConfig): Promise<Testing> =
   const { createCompiler } = require('../compiler/stencil.js');
   const compiler: Compiler = await createCompiler(config);
 
-  let devServer: DevServer;
-  let puppeteerBrowser: puppeteer.Browser;
+  let devServer: DevServer | null;
+  let puppeteerBrowser: puppeteer.Browser | null;
 
-  const run = async (opts: TestingRunOptions = {}) => {
+  /**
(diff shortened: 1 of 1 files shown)


src/testing/testing.ts
```diff
@@ -126,7 +132,7 @@ export const createTesting = async (config: ValidatedConfig): Promise<Testing> =
 
           const styleUrl = getAppStyleUrl(config, devServer.browserUrl);
           if (styleUrl) {
-            env.__STENCIL_APP_STYLE_URL__ = getAppStyleUrl(config, devServer.browserUrl);
+            env.__STENCIL_APP_STYLE_URL__ = styleUrl;
```
rwaskiewicz, line 135: I have no idea why we had an extraneous call to `getAppStyleUrl` here - I can't find any side effects as a result of calling it, so my best guess is that it was just an oversight at author time/at some point if there was a side effect.

alicewriteswrongs (approved):

Issue #4631, opened 2023-07-28
Test fails with ReferenceError: _{method}_decorators is not defined when using custom decorators


Aadhisivam:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

4.0.2

### Current Behavior

*.spec.tsx test fails when using custom decorators in the component.
---
![image](https://github.com/ionic-team/stencil/assets/76022798/0501deb1-bb41-4f56-b025-e3d055c4bebb)

### Expected Behavior

Expect test should not fail due to decorators.

### System Info

```shell
System: node 18.16.1
    Platform: darwin (22.5.0)
   CPU Model: Apple M1 Pro (8 cpus)
    Compiler: /stencil-demo/stencil-demo/node_modules/@stencil/core/compiler/stencil.js
     Stencil: 4.0.2 😈
  TypeScript: 5.0.4
      Rollup: 2.42.3
      Parse5: 7.1.2
      Sizzle: 2.42.3
      Terser: 5.19.1
```

### Steps to Reproduce

Add a custom decorator and use within component, test script will fail with reference error.

### Code Reproduction URL

https://github.com/Aadhisivam/stencil-demo/

rwaskiewicz:
Thanks! I'm going to label this to get it ingested into our internal issue tracking system!

rwaskiewicz:
It looks like this bug can be traced back to v2.19.0 - updated the title of the bug accordingly

Aadhisivam:
FYI, it worked previously in stencil **v3.0.1** with jest and jest-cli **v26.0.1.**

Sharing the dependencies of working tests, for your reference


```
"dependencies": {
    "@stencil/core": "3.0.1"
  },
  "devDependencies": {
    "@types/jest": "^25.2.3",
    "@types/node": "^16.18.11",
    "jest": "26.0.1",
    "jest-cli": "26.0.1",
    "puppeteer": "^20.7.3"
  },
```

@rwaskiewicz

rwaskiewicz:
Thanks @Aadhisivam! Upon closer inspection, it looks like this was introduced in v3.2.0.

[...]

Issue #3831, opened 2022-11-17
bug: custom decorators are not invoked for state properties


jgroth:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/.github/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

2.19.0

### Current Behavior

If a property is decorated with both `@State` and another decorator, the code for the other decorator will never be invoked. 

In the following component nothing will be logged to the console with Stencil `2.19.0`. It works perfectly in `2.18.1`. Probably introduced in #3614 

```tsx
const MyDecorator: () => PropertyDecorator = () => {
  return (target: Object, propertyKey: string) => {
    console.log({target, propertyKey});
  };
};

@Component({
  tag: 'my-component',
  shadow: true,
})
export class MyComponent {
  @State()
  @MyDecorator()
  decoratedProp: any;
}
```

After inspecting the compiled javascript files, the TypeScript `__decorate` helper function is missing for the component in `2.19.0`, and is present in `2.18.1`

### Expected Behavior

Decorators added to properties also decorated with `@State` should be invoked

### Steps to Reproduce

Create a new component with a property decorated with both `@State` and something else. The other decorator will not work

### Code Reproduction URL

https://github.com/jgroth/stencil-decorator-bug

jgroth:
Thanks for the quick reply! Any estimate on a timeline when it might be fixed? 
This is a quite serious breaking change for us and will prevent us from upgrading. We use this quite extensively to reduce the amount of boilerplate we have to write when connecting a component to a Redux state, e.g.
```ts
class MyComponent {
    @State()
    @SelectCurrentUser()
    private currentUser: User;
}
```

alicewriteswrongs:
@jgroth ah sorry that it broke a common pattern you all are using. Part of the sticky bit with this issue is that we need to do some discovery work to figure out exactly the right way to address this — this doesn't mean that it will take eons to address, but just that at present I don't have a clear roadmap of exactly how we'll fix it in my head, so I can't commit to a specific timeline at present.

However! That said, the ticket for doing this work has been pulled in to our next sprint, so we will start work on it relatively soon.

Dayvisson:
[...]

PR #4771, opened 2023-09-10
fix(test): ensure legacy decorators are used when using transpile

louis-bompart:
Currently, file compiled through `stencil test` do uses the new ECMA decorators, contrary to the output of `stencil build` which always uses legacy TS decorators.
This, combined with what I think might be an issue with how TypeScript handles class modifiers updates, causes the behavior described in #4631.

GitHub Issue Number: #4631 

### Rejected fix:
Fixing/bypassing the TypeScript update behavior leaves us using the new decorators for `stencil test`, but the legacy ones for `stencil build`. The difference in behavior between the test & the source is not a good thing, obviously.

### Proposed fix:
Ensure we do set the `experimentalDecorators` in the compiler options used by `stencil test`

- [ ] Yes
- [x] No

## Testing

 - `patch-package` the behavior in https://github.com/Aadhisivam/stencil-demo/, runs the tests ✅ 
 - Tested on github.com/coveo/ui-kit. ✅ (caveat, we're impacted by https://github.com/ionic-team/stencil/issues/3831 too)

_____

Fixes: #4631

### DIFF
--- a/src/compiler/config/transpile-options.ts
+++ b/src/compiler/config/transpile-options.ts
@@ -78,6 +78,9 @@ export const getTranspileConfig = (input: TranspileOptions): TranspileConfig => 
   };
 
   const tsCompilerOptions: CompilerOptions = {
+    // ensure we uses legacy decorators
+    experimentalDecorators: true,
+
     // best we always set this to true
     allowSyntheticDefaultImports: true,
 


rwaskiewicz:
Hey @louis-bompart 👋 

Thanks! 

Can you do me a favor and provide step-by-step instructions for testing this change? I'm having some issues trying to verify that this works as described against https://github.com/Aadhisivam/stencil-demo/, as the changes here don't appear to be affecting the failure output of the tests. Of course, I could always be doing something wrong, which is why I ask for step-by-step instructions here 🙂 Thanks!

louis-bompart:
Hey @rwaskiewicz my bad, I should've included more reproducible steps 🤦 
https://github.com/louis-bompart/stencil-demo I forked & pushed a repo to see the fix, clone, npm i, npm test, all tests should be green

rwaskiewicz:
Ah, thanks @louis-bompart! I'm gonna let CI run on the rest of the PR and see if that catches anything I missed.

rwaskiewicz (approved):
Thanks!

PR #4772, opened 2023-09-10
fix(decorator): preserve stencil decorated class members

louis-bompart:
If a class member is decorated by both a non-Stencil and a Stencil decorators, the non-stencil decorators is effectively removed from the transpiled output.

This is because Stencil-decorated members are removed from the classMember and instead solely handled in the constructor (+ some statics)

GitHub Issue Number: #3831

## What is the new behavior?
Pretty much https://github.com/ionic-team/stencil/issues/3831#issuecomment-1712954058
But, in essence, members are now always kept as class members, regardless of their decorators.
However, they are all stripped of their Stencil-decorator so that they can be handled by the special sauce of Stencil.

- [ ] Yes
- [x] No

## Testing

Tested as patch in both:
 - https://github.com/jgroth/stencil-decorator-bug
 - https://github.com/coveo/ui-kit
 
 
 ___
 
 Fixes: #3831

### DIFF
--- a/src/compiler/transformers/decorators-to-static/convert-decorators.ts
+++ b/src/compiler/transformers/decorators-to-static/convert-decorators.ts
@@ -131,7 +131,7 @@ const visitClassDeclaration = (
   // We call the `handleClassFields` method which handles transforming any
   // class fields, removing them from the class and adding statements to the
   // class' constructor which instantiate them there instead.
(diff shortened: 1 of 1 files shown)


louis-bompart:
Won't do. Stencil's plan is to migrate their custom decorator's sauce to Standard ECMAScript Decorators.
See: https://github.com/ionic-team/stencil/issues/3831#issuecomment-1718288529

Issue #4278, opened 2023-04-18
bug: Stencil Hydrate | value.split is not a function when using <svg>


mayerraphael:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

Builds based on #2938 

### Current Behavior

If i add an `<svg>` to the jsx, Stencil hydrate adds a class to this element on server-side rendering.

Upon Hydration, Stencil tries to parse the classList of this element, but an error is thrown in `parseClassList` function:
https://github.com/ionic-team/stencil/blob/main/src/runtime/vdom/set-accessor.ts#L157

```
index-1d3c5472.js:3001 TypeError: value.split is not a function
    at parseClassList (index-1d3c5472.js:1780:56)
    at setAccessor (index-1d3c5472.js:1640:32)
    at updateElement (index-1d3c5472.js:1794:17)
    at patch (index-1d3c5472.js:2193:17)
    at updateChildren (index-1d3c5472.js:2002:13)
    at patch (index-1d3c5472.js:2199:13)
    at updateChildren (index-1d3c5472.js:2002:13)
    at patch (index-1d3c5472.js:2199:13)
    at updateChildren (index-1d3c5472.js:2002:13)
```

The problem seems that `element.className` returns an `SvgAnimatedString` for svg-Elements instead of an expected string.

![image](https://user-images.githubusercontent.com/25869389/232785855-741c5ed5-d128-4f4a-8b3c-08ec4920f41f.png)

In the `updateElement` method we see that `oldVNodeAttrs` holds the `SvgAnimatedString` instead of a normal string.
![image](https://user-images.githubusercontent.com/25869389/232786579-46c61aab-3a42-4b3e-8bee-51c222019707.png)

### Expected Behavior

Updating the class in Stencil Hydrate on svg Elements should work.

### System Info

```shell
Stencil 2.23.2, don't know if v3 is also affected.
```

### Steps to Reproduce

Add an `<svg>` element to a component. Use Stencil Hydrate and render the components server-side. Error occurs on client side hydration.

```jsx
render() {
   return <svg></svg>;
}
```
### Code Reproduction URL

[...]

PR #4783, opened 2023-09-12
test(hydrate): add a regression test for hydrating SVG children

alicewriteswrongs:
This adds a regression test for the issue reported in https://github.com/ionic-team/stencil/issues/4278, which doesn't currently occur on `main` but does occur in the branch associated with https://github.com/ionic-team/stencil/pull/2938 (a PR containing a bunch of changes for fixing bugs with slots).

This regression test will give us a bit more safety as we go to add more slot behavior fixes in the future.

GitHub Issue Number: #4278 

## What is the new behavior?

- 

- [ ] Yes
- [x] No

## Testing

Cherry-pick the commit on this branch over the top of #2938 and then confirm that the relevant test fails (due to how the prerender karma tests work what you'll see is actually all tests in that suite failing).

In order to do so you'll need to:

```sh
npm run clean && npm i && npm run build && npm run test.karma.prod
```

(after doing the cherry-pick)

### DIFF
--- a/test/karma/test-app/prerender-test/karma.spec.ts
+++ b/test/karma/test-app/prerender-test/karma.spec.ts
@@ -1,5 +1,25 @@
 import { setupDomTests, waitForChanges } from '../util';
 
+/**
+ * This monkey-patches `window.console.error` in order to fail a test if that
+ * function is called within a test.
+ *
+ * @returns a teardown function which ondoes the monkey-patch
+ */
+function patchConsoleError() {
+  const orgConsoleError = window.console.error;
+
+  window.console.error = function (...args: any[]) {
+    orgConsoleError.apply(window, args);
+    window.console.info('console.error', args);
+    throw new Error('console.error was called, this is not allowed in a unit test run');
+  };
+
+  return () => {
+    window.console.error = orgConsoleError;
+  };
+}
+
 describe('prerender', () => {
   const { setupDom, tearDownDom } = setupDomTests(document);
   let app: HTMLElement;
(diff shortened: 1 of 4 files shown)


test/karma/test-app/prerender-test/karma.spec.ts
```diff
@@ -1,10 +1,30 @@
 import { setupDomTests, waitForChanges } from '../util';
 
+const orgConsoleError = window.console.error;
+
+/**
+ * This monkey-patches `window.console.error` in order to fail a test if that
+ * function is called within a test.
+ */
+function patchConsoleToFailOnError() {
+  window.console.error = function (...args: any[]) {
```
rwaskiewicz, line 10: Do we ever restore `window.console.error` anywhere after these tests run? Is that done implicitly by Jest in this scenario somehow (I don't recall if I'm honest).

rwaskiewicz (commented):
2 small clarifying questions

test/karma/test-app/prerender-test/karma.spec.ts
```diff
@@ -1,10 +1,30 @@
...
+    orgConsoleError.apply(this, args);
+    try {
+      throw new Error('console.error');
+    } catch (err) {
+      console.info('console.error', args, err);
+    }
+    fail('console.error was called, this is not allowed in a unit test run');
```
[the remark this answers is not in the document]
alicewriteswrongs: sure, that makes sense!

test/karma/test-app/prerender-test/karma.spec.ts
alicewriteswrongs: probs we should...I'll add some code for that

alicewriteswrongs: I refactored the mocking function a little bit to be test-specific by doing the mock and then returning a teardown function to undo it, that way I can make the 'fail on console.error' behavior specific to just the one test I'm adding.

test/karma/test-app/prerender-test/karma.spec.ts
```diff
@@ -61,6 +81,16 @@ describe('prerender', () => {
...
   });
+
+  it('should render an svg child', async () => {
+    const teardown = patchConsoleError();
+    app = await setupDom('/prerender/index.html');
+    await waitForChanges(500);
+
+    const testSvg = app.querySelector('test-svg');
+    expect(testSvg.className).toBe('hydrated');
+    teardown();
```
rwaskiewicz, line 92: Will this get execute if the test fails or `throw`s somewhere between instantiating it and here??

rwaskiewicz (commented):
One more clarifying question!

test/karma/test-app/prerender-test/karma.spec.ts
alicewriteswrongs: mmm probably not, in which case the test suite is failing _anyway_, but I can wrap the test in a `describe` so we can `afterAll`

rwaskiewicz (approved):

Issue #2284, opened 2020-03-18
Support formAssociated components (using ElementInternals)


adriengibrat:
**Stencil version:**

```
 @stencil/core@1.8.8
```

**I'm submitting a:**

[ ] bug report
[x] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**

Stencil component with a static `formAssociated` property does not behave like a vanilla customElement, accessing form related  `internals` methods throws  "DOMException: Failed to read the 'form' property from 'ElementInternals': The target element is not a form-associated custom element." in chrome.

**Expected behavior:**

The static property formAssociated defined on a component should be attached to proxy class defining the customElement or at least Stencil could provide an idiomatic way to allow to attach static properties. It would be awesome to enventually provide a shim for unsuporting browser... but this may deserve it's own issue ;)

**Steps to reproduce:**

A click on the below component (when nested in a form) should submit the form and not throw an error on chrome.
**Related code:**

```tsx
@Component({
  tag: 'associated-to-form',
  shadow: true
})
export class Test {
  static formAssociated: true;
  @Element()
  private host: HTMLElement;
  constructor() {
    this._internals = this.host.attachInternals?.();
  }
  @Listen('click')
  submit() {
    this._internals.form?.submit();
  }
}
```

**Other information:**

More about formAssociated elements use cases: https://web.dev/more-capable-form-controls/

Specifications of formAssociated elements: https://html.spec.whatwg.org/multipage/custom-elements.html#custom-elements-face-example

A blog post about form related component with stencil \o/
https://www.hjorthhansen.dev/shadow-dom-form-participation/ , but formAssociated example seems to be pseudo code (not working) :/

Info about browser support https://www.chromestatus.com/feature/4708990554472448

splitinfinities:
Hey all! I'm going to do my research and work on defining what this means to be able to ship. No timeline, but I added a request for comments. If anybody can provide the different use cases or their ideal interface - whether it's a decorator or a modification to a decorator - that would be fantastic!

johnjenkins:
Before working out a higher level API I guess we'd need to know whether stencil wants to ship something 'official' that doesn't work cross-browser?

[...]

PR #4784, opened 2023-09-12
feat(compiler): add support for form-associated custom elements

alicewriteswrongs:
This adds support for building form-associated custom elements in Stencil components, allowing Stencil components to participate in HTML forms in a rich manner. This is a popular request in the Stencil community (see #2284).

A minimal Stencil component that uses the new APIs to integrate with a form could look like this:

```tsx
import { Component, h, AttachInternals } from '@stencil/core';

@Component({
  tag: 'my-component',
  styleUrl: 'my-component.css',
  shadow: true,
  formAssociated: true
})
export class MyComponent {
  @AttachInternals() internals: ElementInternals;

  render() {
    return <div>Hello, World!</div>;
  }
}
```

A few things to note:

- A new option called `formAssociated` has been added to the [`ComponentOptions`](https://github.com/ionic-team/stencil/blob/06f6fad174c32b270ce239afab5002c23d30ccbc/src/declarations/stencil-public-runtime.ts#L10-L55) interface.
- A new `@AttachInternals()` decorator can be used to indicate a property on a Stencil component to which an [`ElementInternals`](https://developer.mozilla.org/en-US/docs/Web/API/ElementInternals) object will be bound at runtime.
- Using `@AttachInternals()` is supported both for lazy builds ([`www`](https://stenciljs.com/docs/www), [`dist`](https://stenciljs.com/docs/distribution)) as well as for [`dist-custom-elements`](https://stenciljs.com/docs/custom-elements).

The new behavior is implemented at compile-time, and so should result in only very minimal increases in code / bundle size.

> [!NOTE]
> Browser support for the APIs that this feature depends on is still relatively low ([~84% as of this writing](https://caniuse.com/?search=attachInternals)) and the Stencil team does not plan to officially support or incorporate any polyfills for the browser functionality. Accordingly, this is a 'use at your own risk' feature: it is up to you as a developer to ensure the browsers you need to support have shipped the necessary APIs.

Stencil does not support form-associated custom elements! Forms and Stencil components just aren't talking to each other as much as they could be.

## What is the new behavior?

The Stencil compiler now supports emitting form-associated custom elements in the lazy and 'native' output targets.

- [ ] Yes
- [x] No

## Testing

There are new tests which cover most of the functionality here.

There's an example Stencil project making use of these APIs here: https://github.com/alicewriteswrongs/stencil-element-internals-example


src/compiler/transformers/test/parse-form-associated.spec.ts
```diff
@@ -10,10 +10,10 @@ describe('parse form associated', function () {
     export class CmpA {
     }
     `);
-    expect(t.cmp.formAssociated).toBe(true);
+    expect(t.cmp!.formAssociated).toBe(true);
```
rwaskiewicz, line 13: I'm guessing based on the commit msg " if those aren't there we should get a runtime failure in the tests anyway" these were causing SNC errors?

src/compiler/transformers/static-to-meta/form-associated.ts
```diff
@@ -0,0 +1,60 @@
...
+ * @returns whether or not the given component is form-associated
+ */
+const parseFormAssociated = (staticMembers: ts.ClassElement[]): boolean => {
+  const isFormAssociated = getStaticValue(staticMembers, 'formAssociated');
+
+  if (typeof isFormAssociated === 'boolean' && isFormAssociated) {
+    return true;
+  } else {
+    return false;
+  }
```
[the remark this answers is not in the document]
alicewriteswrongs: yeah, I had the more explicit thing just for clarity but why not collapse I guess!

src/compiler/transformers/decorators-to-static/form-internals-decorator.ts
```diff
@@ -0,0 +1,60 @@
...
+  diagnostics: d.Diagnostic[],
+  decoratedMembers: ts.ClassElement[],
+  newMembers: ts.ClassElement[],
+) => {
+  const formInternalsMembers = decoratedMembers.filter(ts.isPropertyDeclaration).filter((prop) => {
+    return !!retrieveTsDecorators(prop)?.find(isDecoratorNamed('FormInternals'));
+  });
+
+  // no decorator fields, return!
+  // TODO print a warning if the component was form-associated but no decorator
```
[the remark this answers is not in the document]
alicewriteswrongs: well, I thought about it a bit and I think if you set `formAssociated: true` but don't use `@FormInternals()` there's nothing really wrong with that, it's just a bit useless to have the option on, so I'm not sure that 

I think maybe I should have clarified about the PR overall that I have a number of little things like this (some of which are noted with `TODO` and some which aren't) which I don't yet consider final, so probably the most useful feedback would be on architecture / overall approach, and when we think that's really ready to go then I'll do a full code cleanup

src/compiler/transformers/test/parse-form-associated.spec.ts
alicewriteswrongs: yup that's right, and since it's a test I'm not bothered at all by saying "go ahead and access this"

rwaskiewicz: Gotcha, totally makes sense in this context - thanks!

src/compiler/transformers/static-to-meta/form-associated.ts
```diff
@@ -0,0 +1,60 @@
...
+export const parseFormAssociatedProperties = (staticMembers: ts.ClassElement[]) => {
+  const formAssociated = parseFormAssociated(staticMembers);
+  const formInternalsProp = parseFormInternals(staticMembers);
+
+  if (formAssociated) {
+    return { formAssociated, formInternalsProp };
+  } else {
+    return {
+      formAssociated: false,
+      // @ts-ignore
```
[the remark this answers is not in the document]
alicewriteswrongs: it does, I added a return type based on picking the props off of the `ComponentCompilerMeta` interface

src/compiler/transformers/test/parse-form-associated.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+    @Component({
+      tag: 'cmp-a',
+    })
+    export class CmpA {
+    }
+    `);
+    expect(t.cmp.formAssociated).toBe(false);
+    expect(t.cmp.formInternalsProp).toBe(null);
+  });
+});
```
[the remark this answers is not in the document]
alicewriteswrongs: sure, why not!

src/compiler/transformers/test/transpile.ts
```diff
@@ -108,10 +108,23 @@ export function transpileModule(
     styleImportData: 'queryparams',
   };
 
+  const collection: d.CollectionCompilerMeta = {
```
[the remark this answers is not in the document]
alicewriteswrongs: I was running into an issue I think because of these lines:

https://github.com/ionic-team/stencil/blob/d0292e0c37d31b87a72a2b83434c8dab90aff120/src/compiler/transformers/static-to-meta/visitor.ts#L41-L48

basically if the `collection` argument isn't present then some values arent' set, and when I was debugging tests I was trying to figure something out about the collection - we don't need to actually keep this change around though

src/runtime/bootstrap-lazy.ts
```diff
@@ -157,6 +157,14 @@ export const bootstrapLazy = (lazyBundles: d.LazyBundlesRuntimeData, options: d.
         }
       }
 
+      // TODO is this even necessary at all?
```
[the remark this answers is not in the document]
alicewriteswrongs: well sort of! Turns out I need to do `HostElement.formAssociated = true` here instead

src/runtime/bootstrap-lazy.ts
```diff
@@ -157,6 +157,14 @@ export const bootstrapLazy = (lazyBundles: d.LazyBundlesRuntimeData, options: d.
         }
       }
 
+      // if the component is formAssociated we need to set that on the host
+      // element so that it will be ready for `attachInternals` to be called on
+      // it later on
+      if (BUILD.shadowDom && BUILD.formAssociated && cmpMeta.$flags$ & CMP_FLAGS.formAssociated) {
+        // any way around this typecast?
+        (HostElement as any).formAssociated = true;
```
alicewriteswrongs, line 165: I don't love this but this does seem to be necessary - setting `HostElement.formAssociated` is necessary for it to work correctly and unfortunately the typing does't want to let me set the property on there :/

alicewriteswrongs (commented):
just pointing out one thing. I've cleanup up a few TODO items and little things here and there, I think this is ready for a more thorough review

src/compiler/transformers/component-lazy/form-internals.ts
```diff
@@ -0,0 +1,47 @@
...
+ * code generated by this function depends on their being a variable in scope
+ * called {@link HOST_REF_ARG} with type {@link HTMLElement}.
+ *
+ * @param cmp metadata about the component of interest, gathered during compilation
+ * @returns a list of expression statements
+ */
+export function createLazyFormInternalsBinding(cmp: d.ComponentCompilerMeta): ts.ExpressionStatement[] {
+  if (cmp.formAssociated && cmp.formInternalsMemberName) {
+    return [
+      ts.factory.createExpressionStatement(
```
tanner-reits, line 22: Any value in a comment of what this expression looks like in generated TS? Factory AST stuff isn't the most friendly to read 😂

src/compiler/transformers/decorators-to-static/form-internals-decorator.ts
```diff
@@ -0,0 +1,58 @@
...
+    return !!retrieveTsDecorators(prop)?.find(isDecoratorNamed('FormInternals'));
+  });
+
+  // no decorator fields, return!
+  if (formInternalsMembers.length === 0) {
+    return;
+  }
+
+  // found too many!
+  if (formInternalsMembers.length > 1) {
```
tanner-reits, line 48: We need to `return` here?

src/compiler/transformers/component-native/form-internals.ts
```diff
@@ -0,0 +1,42 @@
...
+ * developer with the `@FormInternals` decorator.
+ *
+ * @param cmp metadata about the component of interest, gathered during
+ * compilation
+ * @returns an expression statement syntax tree node
+ */
+export function createNativeFormInternalsBinding(cmp: d.ComponentCompilerMeta): ts.ExpressionStatement[] {
+  if (cmp.formAssociated && cmp.formInternalsMemberName) {
+    return [
+      ts.factory.createExpressionStatement(
```
tanner-reits, line 20: Same as previous comment about comment for expression result

src/compiler/transformers/component-native/form-internals.ts
tanner-reits, line 1: Doesn't need to happen now, but would it make sense to abstract the bulk of this? Looks like the only difference between this and lazy-build is what we call `attachInternals()` on.

src/runtime/bootstrap-lazy.ts
tanner-reits: I don't think there is, unfortunately. We should figure out these types tho cuz it feels mega wonky

tanner-reits (approved):
All-in-all looks good imo. Noted a couple things, but no blockers!

rwaskiewicz (changes requested):
All in all looks good! Architecturally I think this looks great! 

I took this for a spin this morning, and ran into a few things that I wanted to run by you/the team. Some are commented inline, others are higher level and are below in the summary here:

## Unit testing

When unit testing a stencil component starter (`npm init stencil@latest component`) and updating the component to support element internals:
Screenshot 2023-09-15 at 9 36 14 AM

I get the following error running tests on the project (`npm t`):
```
  ● Console

    console.log
      undefined

      at MyComponent.getText (src/components/my-component/my-component.tsx:30:13)

    console.log
      undefined

      at MyComponent.getText (src/components/my-component/my-component.tsx:30:13)

  ● my-component › renders

    TypeError: hostRef.$hostElement$.attachInternals is not a function

       8 |     formAssociated: true
       9 |   }
    > 10 | })
         |   ^
      11 | export class MyComponent {
      12 |   @FormInternals() internals: ElementInternals;
      13 |

      at new MyComponent (src/components/my-component/my-component.tsx:10:44)
      at initializeComponent (node_modules/@stencil/core/internal/testing/index.js:581:5)

  ● my-component › renders with values

    TypeError: hostRef.$hostElement$.attachInternals is not a function

       8 |     formAssociated: true
       9 |   }
    > 10 | })
         |   ^
      11 | export class MyComponent {
      12 |   @FormInternals() internals: ElementInternals;
      13 |

      at new MyComponent (src/components/my-component/my-component.tsx:10:44)
      at initializeComponent (node_modules/@stencil/core/internal/testing/index.js:581:5)
```

Did you see this at all in your testing? Anything abundantly clear that I'm doing wrong here?

## HMR console errors

In testing this with the dev server, it looks like on every reload we get a `DOMException: NotSupportedError` as a result of trying to call `AttachInternals()` again. In the video below, I have the component that we generate from running `npm init stencil@latest component`, adapted to use `formAssociated` and `@FormInternals` (you should be able to see the changes I made to the source file in the left gutter of the editor there):

https://github.com/ionic-team/stencil/assets/1930213/076541a9-e6cb-4623-b025-cef87526f372

I also tried:
- Adding a second copy of this component to `index.html` - it appears we'll get an error for each component in the DOM
- I verified this happens when a component does/doesn't have a parent `<form>`

## Karma Testing

Could we add a Karma test that verifies that we can compile a component that uses form internals & capture some form data coming from a custom element?

src/compiler/transformers/decorators-to-static/component-decorator.ts
```diff
@@ -108,6 +113,21 @@ const validateComponent = (
...
+  });
+
+  if (
+    formInternalsMembers.length > 0 &&
+    (typeof componentOptions.shadow === 'boolean' || !componentOptions?.shadow?.formAssociated)
+  ) {
+    const err = buildError(diagnostics);
+    err.messageText = `In order to use the @FormInternals() decorator to access ElementInternals a
+    component must set shadow.formAssociated to true.`;
+    augmentDiagnosticWithNode(err, formInternalsMembers[0]);
```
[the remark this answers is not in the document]
alicewriteswrongs: yeah, I think it makes sense to pass the node for the `@Component` bit instead, I'll make that change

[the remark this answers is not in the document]
alicewriteswrongs: changed it to get something like this instead:

Screenshot 2023-09-15 at 11 04 10 AM

src/compiler/transformers/component-lazy/form-internals.ts
alicewriteswrongs: makes sense! especially the one here for the lazy build isn't so intuitive. Adding a comment here (and so the native component one) which explain what we're trying to generate

src/compiler/transformers/decorators-to-static/form-internals-decorator.ts
alicewriteswrongs: mm I guess we should!

src/compiler/transformers/component-native/form-internals.ts
alicewriteswrongs: I think with two instances of closely related code my instinct would be that adding an abstraction now is probably a YAGNI situation, I'd say if we need a third or fourth variant that might be where I'd look at adding it in

src/compiler/transformers/decorators-to-static/component-decorator.ts
```diff
@@ -108,6 +113,21 @@ const validateComponent = (
     return false;
   }
 
+  const formInternalsMembers = cmpNode.members.filter(ts.isPropertyDeclaration).filter((prop) => {
+    return !!retrieveTsDecorators(prop)?.find(isDecoratorNamed('FormInternals'));
+  });
+
+  if (
+    formInternalsMembers.length > 0 &&
+    (typeof componentOptions.shadow === 'boolean' || !componentOptions?.shadow?.formAssociated)
```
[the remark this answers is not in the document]
alicewriteswrongs: yup! sometimes I just spray the `?` everywhere haha

alicewriteswrongs:
on the higher-level things:

### testing

I think we'll need to probably add an element internals polyfill to mock-doc in order to support this stuff in unit tests. I'll look into adding this today.

With Karma tests I was planning to add a test suite for all this behavior but held off until the rest of the code was stabilized and whatnot, so likewise I'll dig into that today too.

### HMR stuff

I've run into that error during local dev myself, essentially it's because you can't `attachInternals` on an element more than once (and, as you noted, it doesn't matter if an element is in a form or not since the browser lets you `attachInternals` anywhere afaict). I'll look into what we can do to mitigate this so you don't get an error when hot-reloading stuff happens.

alicewriteswrongs:
@rwaskiewicz @tanner-reits I managed to figure out the Karma issue, basically I ended up just adding `$hostElement$` to a do-not-mangle list in the options we pass to Terser. Possibly there's a better way...

alicewriteswrongs:
Alright I also just added an "implementation" of `attachInternals` for mock-doc which returns a  proxy that will log "hey this isn't implemented" if you try to access any properties on it. Here's what it looks like if you try to access something on it in a spec test:


https://github.com/ionic-team/stencil/assets/6207644/31935ed3-14b5-4b8e-9170-3ce48ac52f2e

test/karma/test-app/form-associated/karma.spec.ts
```diff
@@ -0,0 +1,23 @@
...
+    const elm = app.querySelector('form-associated');
+    expect(elm).not.toBeNull();
+  });
+
+  // it('should link up to the surrounding form', async () => {
+  //   const formEl = app.querySelector("form");
+  //   await waitForChanges();
+  //   const formData = new FormData(formEl);
+  //   expect(formData.get("test-input")).toBe("my default value");
+  // });
```
christian-bromann, line 22: Is there a specific reason for having this code commented out?

alicewriteswrongs: just haven't gotten it working yet! I need to circle back once I've sorted out the HMR stuff

alicewriteswrongs:
Ok so I haven't yet gone through the whole diff with a fine-toothed comb to make sure that all testing code and temporary changes and whatnot are all removed, but as of right now I think all of the functionality stuff I needed to get in place is finished. So HMR should now be fully supported! 🎉 So while I don't think this is _quite_ ready for another close code review I do think it's ready for more functional testing

src/compiler/transformers/static-to-meta/component.ts
```diff
@@ -55,64 +57,49 @@ export const parseStaticComponentMeta = (
   const isCollectionDependency = moduleFile.isCollectionDependency;
   const encapsulation = parseStaticEncapsulation(staticMembers);
   const cmp: d.ComponentCompilerMeta = {
```
alicewriteswrongs, line 59: the diff here looks scarier than it is, I only actually changed two lines but the object wasn't alphabetized

one Q for reviewers: do we think this is safe? It's possible that some of the helper methods here have side-effects which would mean that we can't assume it's possible to safely re-order them

alicewriteswrongs (commented):
noting one thing

alicewriteswrongs:
Ok I believe this is now ready for code review. There are some issues with the Karma tests which I'm working on ironing out right now, but I think the code should still be ready for review

src/runtime/bootstrap-lazy.ts
rwaskiewicz: Agreed, we probably just need to remove the comment here

rwaskiewicz (changes requested):
Small number of asks - I haven't battle tested this yet, I'll do that once the dust has settled on Karma tests

src/compiler/transformers/component-lazy/attach-internals.ts
```diff
@@ -0,0 +1,151 @@
...
+import ts from 'typescript';
+
+import { HOST_REF_ARG } from './constants';
+
+/**
+ * Create a binding for an `ElementInternals` object compatible with a
+ * lazy-load ready Stencil component.
+ *
+ * In order to create a lazy-loaded form-associated component we need to access
+ * the underlying host element (via the "$hostElement$" prop on {@link d.hostRef})
```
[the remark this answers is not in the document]
alicewriteswrongs: hmm I wish something would throw an error if those `@link` bits didn't resolve correctly :/

src/compiler/transformers/static-to-meta/component.ts
alicewriteswrongs: I ended up backing that change out already as I was trying to debug the karma issues, pushing that code up now

alicewriteswrongs:
some documentation up here: https://github.com/ionic-team/stencil-site/pull/1247

src/mock-doc/node.ts
```diff
@@ -223,12 +223,25 @@ export class MockNodeList {
...
+type MockElementInternals = Record<keyof ElementInternals, null>;
+
 export class MockElement extends MockNode {
   namespaceURI: string | null;
   __attributeMap: MockAttributeMap | null | undefined;
   __shadowRoot: ShadowRoot | null | undefined;
   __style: MockCSSStyleDeclaration | null | undefined;
 
+  attachInternals(): MockElementInternals {
+    return new Proxy({} as unknown as MockElementInternals, {
```
tanner-reits, line 235: Clever

src/compiler/transformers/component-lazy/attach-internals.ts
tanner-reits, line 1: appreciate all the documentation here!

tanner-reits (commented):
Code-wise, looks solid! Haven't had a change to test it out yet, so will get around to that asap!

alicewriteswrongs:
alright at long last the build is fully passing here, so I believe this is all ready for a code review!

rwaskiewicz (approved):
Looks good! I had one non-blocking question, and found one bug that I think we can work on next week (and don't need to handle just this second).

Video of the bug in question below. In it, I have the output of `npm init stencil@latest component`  with this branch installed. When I flip the bit on `formAssociated`, I get a few console errors in the browser:

https://github.com/ionic-team/stencil/assets/1930213/dcd417ca-88a6-4476-b557-5986dc5e731b

test/karma/stencil.config.ts
```diff
@@ -7,6 +7,7 @@ export const config: Config = {
   namespace: 'TestApp',
   srcDir: 'test-app',
   tsconfig: 'tsconfig-stencil.json',
+  sourceMap: false,
```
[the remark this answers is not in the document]
alicewriteswrongs: I think so, although I think I just inadvertently left that change in after fixing the real issue - I flipped it back and just pushed that up

alicewriteswrongs:
@rwaskiewicz does that only happen with HMR or on first page load too?

rwaskiewicz:
@alicewriteswrongs it always happens in HMR when the value goes from `false` to `true`. That is, if I set the value to `false` before I start the dev server, start the dev server, then switch it to `true`, I get that error. I also get that error if I have the value set to `true`, start the dev server, then set the value to `false` then `true` again (it only happens on the `false` -> `true` transition)

tanner-reits (approved):
Looks good!

PR #4785, opened 2023-09-13
chore(snc): add 'extras' to ValidatedConfig

rwaskiewicz:

GitHub Issue Number: N/A

## What is the new behavior?

this commit updates the `ValidatedConfig` interface to include `extras`. doing so eliminates ~25 strictNullChecks violations from the codebase. 

- [ ] Yes
- [x] No

## Testing

The type checker should continue to pass.

I updated any cases of using optional chaining on `extras` in the codebase that I could find.

## Other information

I can see a world where there's an argument to _not_ make this required. LMK what you think!


alicewriteswrongs (approved):
25, nice!

PR #4794, opened 2023-09-14
chore(snc): replace config argument types

rwaskiewicz:

GitHub Issue Number: N/A

## What is the new behavior?

this commit updates a handful of function parameter types from the unvalidated `Config` type `ValidatedConfig`. this has 2 benefits:

1. it "naturally" fixes a handful of strictNullChecks violations in cases where the caller of the function updated already had a ref to the `ValidatedConfig`
2. it helps with future refactorings of `ValidatedConfig` where we add new keys to be required on the object, as it allows us to: 
    1. find new violations where a property is needed a bit easier in these converted functions 
    1. in theory, automatically dismisses strictNullChecks violations when a key is added to the type, since it can now be properly checked

- [ ] Yes
- [x] No

## Testing

Tests should continue to pass

## Other information

This is the amount of work I could (safely) do/test over ~15 minutes, so it's by no means comprehensive.

For each fn I updated, I:
- Checked if the caller(s) already had a `ValidatedConfig` in their fn definitions
  - If all callers did, I updated the fn signature of the function in question 
  - If all callers did not have a `ValidatedConfig`, then I repeated the process for each caller. In all cases, I found a `ValidatedConfig`. This only consisted cases of where we didn't propagate the type down to callers all the way

I'll probably do the same exercise tomorrow. I suspect there are a few more easy wins here

### DIFF
--- a/src/compiler/app-core/app-polyfills.ts
+++ b/src/compiler/app-core/app-polyfills.ts
@@ -2,7 +2,11 @@ import { join } from 'path';
 
 import type * as d from '../../declarations';
 
-export const getClientPolyfill = async (config: d.Config, compilerCtx: d.CompilerCtx, polyfillFile: string) => {
+export const getClientPolyfill = async (
+  config: d.ValidatedConfig,
+  compilerCtx: d.CompilerCtx,
+  polyfillFile: string,
+) => {
   const polyfillFilePath = join(
     config.sys.getCompilerExecutingPath(),
     '..',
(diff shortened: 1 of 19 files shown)


src/compiler/app-core/app-polyfills.ts
```diff
@@ -2,7 +2,11 @@ import { join } from 'path';
 
 import type * as d from '../../declarations';
 
-export const getClientPolyfill = async (config: d.Config, compilerCtx: d.CompilerCtx, polyfillFile: string) => {
+export const getClientPolyfill = async (
```
rwaskiewicz, line 5: _Just_ brought us over the line limit by a hair by going from `Config`->`ValidatedConfig`

src/compiler/build/compiler-ctx.ts
```diff
@@ -62,7 +62,7 @@ export class CompilerContext implements d.CompilerCtx {
   }
 }
 
-export const getModuleLegacy = (_config: d.Config, compilerCtx: d.CompilerCtx, sourceFilePath: string) => {
+export const getModuleLegacy = (compilerCtx: d.CompilerCtx, sourceFilePath: string) => {
```
rwaskiewicz, line 65: We don't seem to export this fn, so this felt safe

tanner-reits (approved):

alicewriteswrongs (approved):

PR #3888, opened 2022-12-08
chore(ci): downgrade ubuntu

rwaskiewicz:
## Pull request checklist

Please check if your PR fulfills the following requirements:
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [ ] Build (`npm run build`) was run locally and any changes were pushed
- [ ] Unit tests (`npm test`) were run locally and passed
- [ ] E2E Tests (`npm run test.karma.prod`) were run locally and passed
- [ ] Prettier (`npm run prettier`) was run locally and passed

## Pull request type

 

 

Please check the type of change your PR introduces:
- [ ] Bugfix
- [ ] Feature
- [ ] Refactoring (no functional changes, no api changes)
- [x] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

## What is the current behavior?

Today, CI started to fail when we perform our bundler tests:
```
08 12 2022 15:51:12.909:INFO [launcher]: Starting browser ChromeHeadless
08 12 2022 15:51:14.283:ERROR [launcher]: Cannot start ChromeHeadless
	
DevTools listening on ws://127.0.0.1:9222/devtools/browser/9c353de5-4e6d-445a-9e2c-aca2afe3b555
[1208/155113.030568:ERROR:gpu_init.cc(441)] Passthrough is not supported, GL is swiftshader
[1208/155113.179045:ERROR:gpu_process_host.cc(995)] GPU process exited unexpectedly: exit_code=159
[1208/155113.179070:WARNING:gpu_process_host.cc(1309)] The GPU process has crashed 1 time(s)
[1208/155113.186693:ERROR:gpu_init.cc(441)] Passthrough is not supported, GL is swiftshader
[1208/155113.301548:ERROR:gpu_process_host.cc(995)] GPU process exited unexpectedly: exit_code=159
[1208/155113.301567:WARNING:gpu_process_host.cc(1309)] The GPU process has crashed 2 time(s)
[1208/155113.308256:ERROR:gpu_init.cc(441)] Passthrough is not supported, GL is swiftshader
[1208/155113.419375:ERROR:gpu_process_host.cc(995)] GPU process exited unexpectedly: exit_code=159
[1208/155113.419394:WARNING:gpu_process_host.cc(1309)] The GPU process has crashed 3 time(s)
[1208/155113.421735:ERROR:gpu_init.cc(441)] Passthrough is not supported, GL is disabled
[...]

PR #4795, opened 2023-09-14
chore(repo): use latest ubuntu in tests

rwaskiewicz:

GitHub Issue Number: N/A

## What is the new behavior?

this commit finishes the revert of
https://github.com/ionic-team/stencil/pull/3888. the ubuntu versions in the changed github matrices can't be properly updated/detected by renovate, which is why they fell behind in the first place

STENCIL-755

- [ ] Yes
- [x] No

## Testing

All test workflows updated here continue to pass.

You can see new 'ubuntu-latest' checks occurring/passing here, and the old v20 ones obviously won't run

## Other information

I will probably wait to merge this until after Monday's release. Reason being is that I need to update the CI checks that we gate on, and I don't wanna cause a snowball of rebases right before the release.

I found this while culling the backlog yesterday, and thought it'd be a nice/quick win for something we could easily "just close out"

### DIFF
--- a/.github/workflows/test-analysis.yml
+++ b/.github/workflows/test-analysis.yml
@@ -12,7 +12,7 @@ jobs:
       fail-fast: false
       matrix:
         node: ['16', '18', '20']
-        os: ['ubuntu-20.04', 'windows-latest']
+        os: ['ubuntu-latest', 'windows-latest']
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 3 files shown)


tanner-reits (approved):
Looks like CI is doing what we want 👍

alicewriteswrongs (approved):

PR #4798, opened 2023-09-15
chore(): remove legacy meeting notes and ADRs

tanner-reits:
Removes the meeting notes in `/notes` and Architecture Decision Docs (ADRs) in `/docs/adr`. 

These docs don't really belong in the repo and don't seem to add any value to the team or external contributors (I have never once referenced these, nor knew they even existed). If we want to migrate any of these over to a design doc or "wiki" article, we can easily recover them from Git history. The only document I made a note about that might worth migrating is the Telemetry stuff, but I can do that when I get around to writing some of these wiki-type docs.


rwaskiewicz (changes requested):
One small ask - can you remove the `.adr.json` file in the root of the repo as well please?

tanner-reits:
> One small ask - can you remove the `.adr.json` file in the root of the repo as well please?

Didn't know that was a thing! Thanks for pointing it out. Done in af78331f7277c07ea5bfa2cec0c4c7db8ebf6fbe

rwaskiewicz (approved):

alicewriteswrongs (approved):

PR #4810, opened 2023-09-18
chore(deps): remove typescript from karma deps

alicewriteswrongs:

GitHub Issue Number: N/A

## What is the new behavior?

- 

- [ ] Yes
- [ ] No

### DIFF
--- a/test/karma/package-lock.json
+++ b/test/karma/package-lock.json
@@ -20,7 +20,6 @@
         "karma-typescript-es6-transform": "^5.5.4",
         "normalize.css": "^8.0.1",
         "rollup-plugin-node-polyfills": "^0.1.2",
-        "typescript": "~5.0.4",
         "webpack-cli": "^4.9.1",
         "workbox-build": "4.3.1"
       }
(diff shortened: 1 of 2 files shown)


test/karma/package-lock.json
```diff
@@ -39,47 +38,48 @@
       }
     },
     "node_modules/@babel/code-frame": {
-      "version": "7.21.4",
```
rwaskiewicz, line 42: By chance did an `npm i` get run in this directory? If I uninstall typescript on my machine locally, I don't see these changes.

rwaskiewicz: Could it be `karma-typescript` that's keeping it around? They list TS as a peer dep, which I believe should be getting installed by default (and is probably why TS 5.0.4 is still hanging around).

alicewriteswrongs: I think so yeah, so one solution could be to explicitly install typescript in `test/karma` and use the same version as in our root one

rwaskiewicz: In that case, it may be worth re-enabling renovate for this directory. I promised that PR yesterday and failed to deliver 😆 - thought's on us re-enabling that and see how that works out? That way, renovate should bump both TS versions (here and in the root of the project) at the same time

rwaskiewicz: https://github.com/ionic-team/stencil/pull/4816 created

alicewriteswrongs: I think it makes sense!

rwaskiewicz: https://github.com/ionic-team/stencil/pull/4820 created to keep things in line

Issue #4661, opened 2023-08-08
bug: dist/esm/polyfills is empty 


AnnaHoege:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

v4.0.4

### Current Behavior

When running `npm run build`, the resulting dist/esm/polyfills folder is empty every other time the command is run. This causes issues as the loader exports the index file of the polyfills folder.

### Expected Behavior

The polyfills folder should contain at least an index file on every build.

### System Info

```shell
System: node 18.17.0
      Platform: windows (10.0.19045)
      CPU Model: Intel(R) Core(TM) i7-10610U CPU @ 1.80GHz (8 cpus)
      Compiler: C:\Users\AHOEGE\Projects\stencil-bug-reproduction\bug-reproduction\node_modules\@stencil\core\compiler\stencil.js
      Build: 1691405536
      Stencil: 4.0.4
      TypeScript: 5.0.4
      Rollup: 2.42.3
      Parse5: 7.1.2
      Sizzle: 2.42.3
      Terser: 5.19.2
```

### Steps to Reproduce

1. Create a new project via `npm init stencil`
2. Run `npm i`
3. Run `npm build`
4. Run `npm build`

### Code Reproduction URL

https://github.com/AnnaHoege/stencil-bug-reproduction

FirstVertex:
See this intermittently on my Windows 11 machine.  `stencil build` completes and the dist folder is generated, there's a polyfills folder under esm/ but it is empty.  When I run build again it is fixed.

FirstVertex:
Please, make it stop.  All week long I have to build twice.  1 minute per build.  Building dozens of times per day as I figure out our Storybook integration.  So much wasted time.  Please, this bug needs a high priority and severity.

rwaskiewicz:
Can anyone running into this issue please try out this dev build for me and let me know if this works?
```
npm i @stencil/core@4.4.1-dev.1697228154.e0f8280
```
Thanks!

imagoiq:
Hi,

[...]

PR #4815, opened 2023-09-19
chore(snc): fix two snc errors related to `.typesDir`

alicewriteswrongs:
Just fixing two SNC errors.

GitHub Issue Number: N/A

## What is the new behavior?

Just dropping some SNC errors! 

- [ ] Yes
- [x] No

## Testing

This shouldn't break anything since I don't think the runtime behavior of anything will really be affected. Unit tests and whatnot should all be passing.


src/compiler/config/outputs/validate-dist.ts
```diff
@@ -27,8 +27,11 @@ import { validateCopy } from '../validate-copy';
  */
 export const validateDist = (config: d.ValidatedConfig, userOutputs: d.OutputTarget[]): d.OutputTarget[] => {
   const distOutputTargets = userOutputs.filter(isOutputTargetDist);
-  return distOutputTargets.reduce((outputs: d.OutputTarget[], o: d.OutputTargetDist) => {
-    const distOutputTarget = validateOutputTargetDist(config, o);
+
+  const outputs: d.OutputTarget[] = [];
+
+  for (const outputTarget of distOutputTargets) {
+    const distOutputTarget = validateOutputTargetDist(config, outputTarget);
```
alicewriteswrongs, line 34: while I was in here I thought hey, this `.reduce` call is kind of weird, let's just `for ... of` instead

src/utils/output-target.ts
```diff
@@ -37,7 +37,14 @@ export const relativeImport = (pathFrom: string, pathTo: string, ext?: string, a
...
 
-export const getComponentsDtsTypesFilePath = (outputTarget: d.OutputTargetDist | d.OutputTargetDistTypes) =>
+/**
+ * Helper to get an appropriate file path for `components.d.ts` for a `"dist"`
+ * or `"dist-types"` output target.
+ *
+ * @param outputTarget the output target of interest
+ * @returns a properly-formatted path
+ */
+export const getComponentsDtsTypesFilePath = (outputTarget: Required<d.OutputTargetDist> | d.OutputTargetDistTypes) =>
```
alicewriteswrongs, line 47: I added the `Required<>` wrapper here to make sure the `.typesDir` is defined, and this is legit since `getComponentsDtsTypesFilePath` is only called in the `validate-dist` file above where the argument satisfies that type

alicewriteswrongs (commented):
two notes

src/compiler/config/outputs/validate-dist.ts
```diff
@@ -41,7 +44,7 @@ export const validateDist = (config: d.ValidatedConfig, userOutputs: d.OutputTar
       systemDir: config.buildEs5 ? lazyDir : undefined,
       systemLoaderFile: config.buildEs5 ? join(lazyDir, namespace + '.js') : undefined,
       legacyLoaderFile: join(distOutputTarget.buildDir, namespace + '.js'),
-      polyfills: distOutputTarget.polyfills !== undefined ? !!distOutputTarget.polyfills : true,
+      polyfills: true,
```
alicewriteswrongs, line 47: Since we're actually ensuring (below) that `.polyfills` is always a boolean here this like ends up being equivalent to just always `true`...is this what we what? should the `polyfill` property on `DIST_LAZY` instead always be what the value on `DIST` is? that's what I think would make the most sense to me, but the tests as-written seem to require that it's always `true`

alicewriteswrongs (commented):
one other question

src/compiler/config/outputs/validate-dist.ts
tanner-reits: Wait, shouldn't this be `distOutputTarget.polyfills`?

src/compiler/config/outputs/validate-dist.ts
```diff
@@ -135,13 +138,13 @@ const validateOutputTargetDist = (config: d.ValidatedConfig, o: d.OutputTargetDi
     typesDir: o.typesDir || DEFAULT_TYPES_DIR,
     esmLoaderPath: o.esmLoaderPath || DEFAULT_ESM_LOADER_DIR,
     copy: validateCopy(o.copy ?? [], []),
-    polyfills: isBoolean(o.polyfills) ? o.polyfills : undefined,
+    polyfills: typeof o.polyfills === 'boolean' ? o.polyfills : false,
```
tanner-reits, line 141: Just curious, any reason to move away from `isBoolean()`?

tanner-reits (changes requested):

src/compiler/config/outputs/validate-dist.ts
alicewriteswrongs: just doesn't seem necessary to me, and I noticed recently that the `isBoolean` helper in node `utils` is marked as deprecated

src/compiler/config/outputs/validate-dist.ts
alicewriteswrongs: I'll go ahead and change it to that though since it really does seem like The Right Thing

rwaskiewicz:
This code is very much in the same area as #4661 - would it be okay if we held off on this until I got to the bottom of that issue (est early next week)?

src/compiler/config/outputs/validate-dist.ts
rwaskiewicz: I think this is fine, the only thing I'm confused by is:
> I noticed recently that the isBoolean helper in node utils is marked as deprecated

Doesn't `isBoolean` resolve to https://github.com/ionic-team/stencil/blob/cb3eac3a2b7795a5c8d3f9100812337d1a8d8a9d/src/utils/helpers.ts#L171?

src/compiler/config/test/validate-output-dist.spec.ts
```diff
@@ -41,7 +41,7 @@ describe('validateDistOutputTarget', () => {
         empty: false,
         isBrowserBuild: true,
         legacyLoaderFile: path.join(rootDir, 'my-dist', 'my-build', 'testing.js'),
-        polyfills: true,
+        polyfills: false,
```
rwaskiewicz, line 44: The changes from `undefined` to `false` here make sense to me, but the value changes from `true` to `false` less so. Can you help me understand why this is happening?

rwaskiewicz (changes requested):

src/compiler/config/outputs/validate-dist.ts
alicewriteswrongs: yeah it does, I just referenced the node utils deprecation because the reason for it was basically "why not just use `typeof` directly" which I think applies to our little helper too

src/compiler/config/test/validate-output-dist.spec.ts
alicewriteswrongs: [This comment gives some context](https://github.com/ionic-team/stencil/pull/4815#discussion_r1330579976), basically I changed it to ensure that the value on the generated output targets matches what it set on the `"dist"` output target (the logic was more confusing before, if `.polyfills` was `undefined` on the `dist` output target then it would end up always being `true` on the `dist-lazy` one, which doesn't seem right?)

src/compiler/config/outputs/validate-dist.ts
rwaskiewicz: I think the "deprecation" is my source of confusion - to clarify, it's not marked as deprecated in the code, right? Did we have a conversation about deprecating this as a team? TBH, I don't quite remember 🤔

alicewriteswrongs: sorry should have been more clear, I'm referring to this function: https://github.com/nodejs/node/blob/main/doc/api/util.md#utilisbooleanobject

src/compiler/config/test/validate-output-dist.spec.ts
tanner-reits: I agree it seems like they should be in sync

src/compiler/config/outputs/validate-dist.ts
tanner-reits: Hmm, don't really understand why they deprecated their util. It doesn't make any difference technically, but feels like we should be consistent by either:
1. Continuing to use our util **or**
2. Deprecating our util

I don't think Node deprecating theirs needs to influence us here, but I don't really care which way we go either. Not gonna consider this a blocker for me either.

tanner-reits (approved):
Left some opinions, but LGTM

src/compiler/config/outputs/validate-dist.ts
alicewriteswrongs: I think it's just a little bit more direct what's going on, like there's a keyword in the language which does what's needed here, so there's no reason to have a util function which doesn't do anything other than return the result of a comparison based on that keyword

that said, this seems controversial and I'm not at all committed to it so I will remove

rwaskiewicz:
Ugh, sorry @alicewriteswrongs - I thought I had submitted this like, last week 😭  It's a little lost in the GH UI, but https://github.com/ionic-team/stencil/pull/4815#discussion_r1338976196 is my review/question

src/compiler/config/test/validate-output-dist.spec.ts
alicewriteswrongs: mm yeah it's a bit complicated, I guess previously we actually had three different states we treated, basically

1. `dist.polyfills` is `true`, then `dist-lazy` should be true
2. `dist.polyfills` is `false`, then `dist-lazy` should be false
3. `dist.polyfills` is `undefined`, then `dist-lazy` should be `true`

I think the issue is that right now the `undefined` case isn't distinguished from the `false` case, I'll make the logic here check the original output target instead of the validated one (return value of `validateOutputTargetDist`)

tanner-reits (approved):
LGTM. Lots of back-and-forth for 2 checks 😂

rwaskiewicz (changes requested):
One last clarifying question!

alicewriteswrongs:
It won't let me reply with a normal comment (idk how to use github apparently haha) so RE:

> Is there a reason we can't use the validated `DIST` OT here? That way, since the return of `validateOutputTargetDist` is `Required<OutputTargetDist>` and with [this change in mind](https://github.com/ionic-team/stencil/pull/4815/files#diff-1677525851f8c0d25cf1838e531ac1778ed57f1bba3cf62a86923e451c47904bR141) where we force the return value to `false` if `polyfills` isn't a bool, this could be simplified to:

Making that change will actually change the behavior. If we use the `polyfills` property on the validated version then we won't distinguish between the case where `polyfills` is set to `false` on the unvalidated `dist` OT and when it is `undefined`. The existing behavior is that if it is `undefined` we default to `true` on the other OTs generated here, while if it is `false` then it will be `false` on the other OTs.

Tried to clarify that behavior here: https://github.com/ionic-team/stencil/pull/4815#discussion_r1349192580

It's a bit weird for sure, but lmk if you think that should change or if that's not clear

rwaskiewicz (approved):

PR #4833, opened 2023-09-21
revert(test): update dependency rollup-plugin-node-polyfills to ^0.2.0

rwaskiewicz:

This reverts commit 7e7122706e4313eb0c49e60b1901efd59505ac8d, #4824.

This commit was causing flakiness in our karma tests (or so it's believed).

When 7e7122706e4313eb0c49e60b1901efd59505ac8d was applied, we'd see some flakiness to the tune of the following in our karma tests:

> Firefox 117.0 (Windows 10) prerender server componentDidLoad Order FAILED
>	Error: console.error was called, this is not allowed in a unit test run thrown

While we did introduce this error recently, I believe this has more to do with the polyfills more than anything

GitHub Issue Number: N/A

## What is the new behavior?

It's a revert! Go back to the version that worked (for now) 

- [ ] Yes
- [ ] No

## Testing

Tested this locally. Let's exercise it on CI too

### DIFF
--- a/test/karma/package-lock.json
+++ b/test/karma/package-lock.json
@@ -19,7 +19,7 @@
         "karma-typescript": "^5.5.4",
         "karma-typescript-es6-transform": "^5.5.4",
         "normalize.css": "^8.0.1",
-        "rollup-plugin-node-polyfills": "^0.2.0",
+        "rollup-plugin-node-polyfills": "^0.1.2",
         "typescript": "~5.1.0",
         "webpack-cli": "^4.9.1",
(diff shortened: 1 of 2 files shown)


alicewriteswrongs (approved):

alicewriteswrongs:
I also have a draft PR (#4809) which removes the dependency entirely, if we do want to just do that

rwaskiewicz:
@alicewriteswrongs Ah, nice! Let's un-flake the karma tests (if that's even possible 🙃) and then I'm totally down for removing these in #4809

PR #4835, opened 2023-09-22
chore(ts): enable strictBindCallApply ts compiler flag

alicewriteswrongs:
This turns on the `strictBindCallApply` typescript compiler option and then does what I _think_ is something like the minimal set of changes to get the build to pass with that option on.

From [the TypeScript docs](https://www.typescriptlang.org/tsconfig#strictBindCallApply):

> When set, TypeScript will check that the built-in methods of functions `call`, `bind`, and `apply` are invoked with correct argument for the underlying function[...] Otherwise, these functions accept any arguments and will return `any`

I didn't quite realize that without this option doing a `.call` on a function was basically a big type hole! Not good!

- [ ] Yes
- [x] No

## Testing

Unit tests and whatnot are passing, but would be good to check also that framework builds without issues.


rwaskiewicz (approved):

tanner-reits (approved):
Sweet! Didn't know we could get type safety on these!

PR #4836, opened 2023-09-23
chore(release): ensure correct vermoji baked into distributable

rwaskiewicz:

we generate 2 vermoji during the release process. this is due to a small error in not forwarding the one generated during prerelease to the actual release process.

as an example, in stencil 4.3.0, the vermoji shown in `npx stencil info` is 🐫 , but the changelog shows 🎱 

GitHub Issue Number: N/A

## What is the new behavior?

this commit fixes a bug in our release scripts where the vermoji generated in the prerelease step of the process would not be properly propagated to the release step. when the entire process was manual, we'd do it all in one fell swooop, writing a tiny file to disk and reading that.

in lieu of that, we can pull the vermoji off of the head of the CHANGELOG.md! this ensures consistency between the two - running `stencil info` and looking at the changelog should show the same emoji.

note that this change intentionally does not get applied to the `getOptions` function for releases. since we still technically support manual releases, adding the change there for vermoji generation would negatively impact vermoji generation

- [ ] Yes
- [x] No

## Testing

Manually tested this to ensure that we parse this correctly.

I had some issues extending/mocking/overriding `node:fs` here for unit tests, hence the lack of them here

### DIFF
--- a/scripts/release.ts
+++ b/scripts/release.ts
@@ -7,6 +7,7 @@ import { promptRelease } from './release-prompts';
 import { runReleaseTasks } from './release-tasks';
 import { BuildOptions, getOptions } from './utils/options';
 import { getNewVersion } from './utils/release-utils';
+import { getLatestVermoji } from './utils/vermoji';
 
 /**
  * Runner for creating a release of Stencil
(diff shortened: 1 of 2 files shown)


tanner-reits (approved):
Works, but feels kinda unorthodox... then again, historically so has most of the Stencil release process so better than it being wrong 🤷‍♂️ As long as we don't change how the release notes generate, we should be fine

rwaskiewicz:
@tanner-reits agreed - I'm kind of at the end of my rope with the vermoji for production builds. On one hand, I love them for local development builds as a quick signifier that I'm using the build I just generated locally (and it's not the one I generated 10 minutes ago while debugging), as opposed to parsing a timestamp string or something similar. On the other, I've sunk more time than I'd like for these things. Maybe for v5 we consider retiring them 🤔

PR #4847, opened 2023-09-26
refactor(jest): move jest 27 and under support to separate directory

rwaskiewicz:

Stencil's Jest infrastructure assumes that the typings used by Stencil internally will match those
used by end-user's projects. This makes it rather difficult to upgrade Jest for both Stencil internally
and to support newer versions of Jest for users.

GitHub Issue Number: N/A

## What is the new behavior?

Existing Jest 27 code is moved to a new directory. In this directory, a new "internal" package
is created that declares its own support for `@types/jest`, breaking the coupling of type 
dependencies between Stencil's internal tests and the infrastructure that we build for end users.

Prior to this commit, the typings for Jest are declared in Stencil's root-level `package.json` and propagate to the `testing` submodule:

```
├── src/testing              <- Testing submodule
│   │    index.ts
│   ├── jest                 <- Files in this directory type check against @types/jest@27 in Stencil's root-level node_modules/
│   │   ├── jest-config.ts
│   │   ├── jest-environment.ts
│   │   ├── jest-preprocessor.ts
│   │   ├── jest-runner.ts
│   │   ├── jest-screenshot.ts
│   │   ├── jest-serializer.ts
│   │   ├── jest-setup-test-framework.ts
│   │   └── test/
└── package.json             <- Root-level package.json - A dependency on @types/jest@27 is declared here
```

Upon applying this commit, that `@types/jest` coupling is broken with the introduction of a new directory with a separate `package.json` that declares its own `@types/jest` dependency:
```
├── src/testing/jest/
│   └── jest-27-and-under
│       ├── jest-config.ts                 <- All 'jest-*.ts' files type check against @types/jest@27 from the local node_modules/
│       ├── jest-runner.ts
│       ├── node_modules/
│       ├── package.json                    <- A dependency on @types/jest@27 is defined here
│       ├── test
│       └── tsconfig.json
└── package.json                            <- Root-level package.json - A dependency on @types/jest@27 is declared here
``` 

- [ ] Yes
- [x] No

## Testing

This commit is largely a lift and shift operation of existing files. As such, I expect the existing unit tests to pass (especially those for the component starter).

I also manually tested this as follows:
1. Build the Project from the Stencil project root and generate a tarball:
```console
npm run clean && npm ci && npm run build && npm pack
```

2. Test in a component starter
```console
npm init stencil@latest component jest-shift-test \
&& cd $_ \
&& npm i [PATH_TO_TARBALL]
&& npm t
```

### Jest 28+ Support

This is the first piece of the puzzle (well, technically the second if we count #4666) to support newer versions of Jest. In subsequent PRs, we'll add support for detecting the Jest version, supporting installs of `@types/jest` for Stencil engineers, and of course, infra for Jest 28+ 

### For Reviewers

It may be beneficial to review this pull request commit-by-commit. You'll notice that each commit has more detail as to what was done in it to give a little better context. In short:
- Commit 1 - moves existing jest infrastructure to the new directory using `git mv` (to preserve history)
- Commit 2 - updates paths in import statements that are affected by the previous move
- Commit 3 - moves existing jest matchers to the new directory using `git mv` (to preserve history)
- Commit 4 - updates paths in import statements that are affected by the previous move
- Commit 5 - adds `package.json` that will declare `@types/jest`
- Commit 6 - adds `@types/jest` to commit 5's `package.json`
- Commit 7 - fmt the code

### Merging

This commit can be merged upon passing all checks + PR approvals. I'll squash it with a more concise version of all the info presented in this PR summary + the commits themselves

### Misc

JIRA: STENCIL-951


src/testing/jest/jest-27-and-under/package.json
```diff
@@ -0,0 +1,14 @@
+{
```
rwaskiewicz, line 1: See the commit body where I added this file for details behind some of the decisions. Happy to answer questions about this one

src/testing/jest/test/tsconfig.json
```diff
@@ -1,3 +0,0 @@
-{
-  "extends": "../../tsconfig.internal.json"
```
rwaskiewicz, line 2: This was `git mv`'ed, but then when I updated the value for 'extends' here, the GH diff algo thinks they're entirely different files

christian-bromann (commented):
Codewise LGTM 👍 

The only question I have reading this is whether this has an impact for users updating to the latest Stencil version. If the decoupling requires adding `@types/jest` to the user dependencies, we might want to have a migration plan. I don't know enough yet to be sure about this.

rwaskiewicz:
@christian-bromann 

> The only question I have reading this is whether this has an impact for users updating to the latest Stencil version. If the decoupling requires adding @types/jest to the user dependencies, we might want to have a migration plan. I don't know enough yet to be sure about this.

Good point! Stencil actually does this for folks already (assuming they're using Stencil testing infrastructure via `stencil test`). The quickest way to test this out is to spin up a Stencil component starter and remove `@types/jest`:
```
npm init stencil@latest component types-test
\ && cd $_ 
\ && npm i 
\ && npm uninstall @types/jest
```
Then try to run the tests, which gives us get the following error:
```
✗ npm t

> types-test@0.0.1 test
> stencil test --spec --e2e

[34:51.0]  @stencil/core
[34:51.2]  v4.3.0 🐫

[ ERROR ]  Please install supported versions of dev dependencies with either npm or yarn.
           npm install --save-dev @types/jest@27.0.3
```

alicewriteswrongs (approved):

src/testing/jest/jest-27-and-under/package.json
```diff
@@ -0,0 +1,14 @@
...
+  "name": "@stencil/jest-27-and-under",
+  "version": "1.0.0",
+  "description": "Internal package for supporting Jest 27 and under with Stencil",
+  "license": "MIT",
+  "author": "Ionic Team",
+  "homepage": "https://stenciljs.com/",
+  "volta": {
+    "extends": "../../../../package.json"
```
tanner-reits, line 9: Oh cool, didn't know you could extend Volta configs

tanner-reits (approved):
LGTM! Baby steps!

PR #4848, opened 2023-09-26
docs(vdom): comment some slot-related code

alicewriteswrongs:
Adding some documentation reflecting what I _think_ all these things are for!

GitHub Issue Number: N/A

## What is the new behavior?

- 

- [ ] Yes
- [x] No

### DIFF
--- a/src/runtime/runtime-constants.ts
+++ b/src/runtime/runtime-constants.ts
@@ -1,13 +1,20 @@
+/**
+ * Bit flags for recording various properties of VDom nodes
+ */
 export const enum VNODE_FLAGS {
+  /**
+   * Whether or not a vdom node is a slot reference
+   */
   isSlotReference = 1 << 0,
 
-  // slot element has fallback content
-  // still create an element that "mocks" the slot element
+  /**
+   * Whether or not a slot element has fallback content
+   */
   isSlotFallback = 1 << 1,
 
-  // slot element does not have fallback content
-  // create an html comment we'll use to always reference
-  // where actual slot content should sit next to
+  /**
+   * Whether or not an element is a host element
+   */
   isHost = 1 << 2,
 }
 
@@ -17,7 +24,9 @@ export const enum PROXY_FLAGS {
 }
 
 export const enum PLATFORM_FLAGS {
-  // designates a node in the DOM as being actively moved by the runtime
+  /**
+   * designates a node in the DOM as being actively moved by the runtime
+   */
   isTmpDisconnected = 1 << 0,
   appLoaded = 1 << 1,
   queueSync = 1 << 2,
@@ -25,6 +34,13 @@ export const enum PLATFORM_FLAGS {
   queueMask = appLoaded | queueSync,
 }
 
+/**
+ * A (subset) of node types which are relevant for the Stencil runtime. These
+ * values are based on the values which can possibly be returned by the
+ * `.nodeType` property of a DOM node. See here for details:
+ *
+ * {@link https://dom.spec.whatwg.org/#ref-for-dom-node-nodetype%E2%91%A0}
+ */
 export const enum NODE_TYPE {
   ElementNode = 1,
   TextNode = 3,
--- a/src/runtime/vdom/vdom-render.ts
+++ b/src/runtime/vdom/vdom-render.ts
@@ -177,7 +177,7 @@ const putBackInOriginalLocation = (parentElm: Node, recursive: boolean) => {
 
       // remove the old original location comment entirely
       // later on the patch function will know what to do
-      // and move this to the correct spot in need be
+      // and move this to the correct spot if need be
       childNode['s-ol'].remove();
       childNode['s-ol'] = undefined;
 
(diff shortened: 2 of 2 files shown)


src/runtime/vdom/vdom-render.ts
```diff
@@ -757,32 +780,42 @@ const relocateSlotContent = (elm: d.RenderNode) => {
...
+ *
+ * @param nodeToRelocate the node of interest
+ * @param slotName the slot name to check
+ * @returns whether the node is located in the slot or not
+ */
+const isNodeLocatedInSlot = (nodeToRelocate: d.RenderNode, slotName: string): boolean => {
   if (nodeToRelocate.nodeType === NODE_TYPE.ElementNode) {
-    if (nodeToRelocate.getAttribute('slot') === null && slotNameAttr === '') {
+    if (nodeToRelocate.getAttribute('slot') === null && slotName === '') {
+      // TODO why is this true here?
```
rwaskiewicz, line 801: Minor - unaddressed TODO. Can we either put this in the backlog or address this before it lands?

src/runtime/vdom/vdom-render.ts
```diff
@@ -737,18 +756,22 @@ const relocateSlotContent = (elm: d.RenderNode) => {
             }
 
             if (node['s-sr']) {
+              // is `node` itself a slot reference (as `childNode` is) ?
```
rwaskiewicz, line 759: Minor - unaddressed TODO(?). This reads a bit more as a question rather than an implementation not. Unsure if that was the intent or not here

src/runtime/vdom/vdom-render.ts
```diff
@@ -629,36 +629,46 @@ export const patch = (oldVNode: d.VNode, newVNode: d.VNode) => {
...
 
-        for (j = 0; j < ilen; j++) {
-          nodeType = childNodes[j].nodeType;
+        // we need to check all of its sibling nodes in order to see if
+        // `childNode` should be hidden
+        for (const siblingNode of childNodes) {
+          if (siblingNode === childNode) {
+            // same node, let's get outta here
+            continue;
+          }
```
rwaskiewicz, line 666: This looks like a change in actual functionality (please correct me if I'm wrong though!), can you help me understand why it's safe to do this/how we can verify that?

src/runtime/vdom/vdom-render.ts
```diff
@@ -683,50 +693,59 @@ const updateFallbackSlotVisibility = (elm: d.RenderNode) => {
...
-      slotNameAttr = childNode['s-sn'];
+      const slotName = childNode['s-sn'];
 
+      // iterate through all the nodes under the location where the host was
+      // originally rendered
       for (j = hostContentNodes.length - 1; j >= 0; j--) {
         node = hostContentNodes[j] as d.RenderNode;
 
+        // check that the node is not a content reference node or a node
+        // reference (???) and then check that the host name does not match
```
rwaskiewicz, line 728: Just calling out another '???'

rwaskiewicz (changes requested):

src/runtime/vdom/vdom-render.ts
alicewriteswrongs: I was describing the check here, but I don't think it adds _so_ much so I'll just remove this line

src/runtime/vdom/vdom-render.ts
alicewriteswrongs: yeah, this one is a bit confusing to me, it _looks_ like the code basically assumes that if a slot is unnamed then any node that we check will end up being marked as part of the slot, which is a bit confusing and maybe not right? I'll just explain that for now I think

src/runtime/vdom/vdom-render.ts
alicewriteswrongs: haha good catch, I think I had the refactoring madness on this one and while I remember thinking I needed to add this I don't remember now why I thought so, but this check definitely wasn't there before.

rwaskiewicz (approved):

src/runtime/runtime-constants.ts
```diff
@@ -1,13 +1,20 @@
...
   isSlotFallback = 1 << 1,
 
-  // slot element does not have fallback content
-  // create an html comment we'll use to always reference
-  // where actual slot content should sit next to
+  /**
+   * slot element does not have fallback content
+   * create an html comment we'll use to always reference
+   * where actual slot content should sit next to
+   */
```
tanner-reits, line 17: I know you didn't write this, but is this comment accurate to what this really does? Seems like this is more closely tied to the host node (`s-hn`). Not sure where slot fallback content fits into this and why that's mentioned in the comment.

src/runtime/vdom/vdom-render.ts
```diff
@@ -683,50 +688,59 @@ const updateFallbackSlotVisibility = (elm: d.RenderNode) => {
...
+ * Component-global information about nodes which are either currently being
+ * relocated or will be shortly.
+ */
 const relocateNodes: RelocateNodeData[] = [];
 
+/**
+ * Relocate the content of a slot
+ *
+ * @param elm a render node whose child nodes need to be relocated
+ */
```
tanner-reits, line 701: This method is one that might benefit from a rename. It's a bit confusing since the relocation doesn't actually happen here. We just figure out which nodes should be moved and where they'll be moved to (if anywhere). It kinda just generates a map from an original location to a new location.

src/runtime/runtime-constants.ts
alicewriteswrongs: yeah I just changed the formatting to make them visible, but I agree it doesn't make sense...I'll come up with something else

src/runtime/vdom/vdom-render.ts
alicewriteswrongs: right because the nodes get added to the `relocateNodes` data structure...I'm with you on the renaming, how about `markSlotContentForRelocation`? I'll add a node to the jsdoc clarifying that this doesn't actually move the nodes but just puts them into the 'queue' for that to happen

christian-bromann (approved):
I like the usage of `@link` in the comments
LGTM 👍

christian-bromann (approved):
I like the usage of `@link` in the comments
LGTM 👍

christian-bromann:
ups .. double approval due to flaky internet connection / page load

src/runtime/vdom/vdom-render.ts
```diff
@@ -683,50 +688,61 @@ const updateFallbackSlotVisibility = (elm: d.RenderNode) => {
...
 
-const relocateSlotContent = (elm: d.RenderNode) => {
+/**
+ * Mark the contents of a slot for renaming via adding references to them to
+ * the {@link relocateNodes} data structure. The actual work of relocating them
+ * will then be handled in {@link renderVdom}.
+ *
+ * @param elm a render node whose child nodes need to be relocated
+ */
+const markSlotContentForRenaming = (elm: d.RenderNode) => {
```
tanner-reits, line 704: Did you mean for this to be `markSlotContentForRelocation` instead of "renaming"?

alicewriteswrongs: haha sure did!

tanner-reits (approved):
Looks good! I'll update the wiki with the renamed method

PR #4852, opened 2023-09-27
feat(typescript): update to typescript 5.2

alicewriteswrongs:
This does the TS 5.2 upgrade. In order to get the project to build I didn't have to do much other than install the package.

I did have to adjust the `module` setting in `scripts/package.json`, to account for this change: <https://devblogs.microsoft.com/typescript/announcing-typescript-5-2/#module-and-moduleresolution-must-match-under-recent-node-js-settings>

Yucky, old, awful typescript 5.1!

## What is the new behavior?

TypeScript 5.2

- [ ] Yes
- [x] No

## Testing

After upgrading the package I built it locally and installed it into Framework and checked that Framework builds with no issues and the tests run.

### DIFF
--- a/package.json
+++ b/package.json
@@ -122,7 +122,7 @@
     "semver": "^7.3.7",
     "sizzle": "^2.3.6",
     "terser": "5.20.0",
-    "typescript": "~5.1.6",
+    "typescript": "~5.2.2",
     "webpack": "^5.75.0",
     "ws": "8.14.2"
   },
--- a/renovate.json5
+++ b/renovate.json5
@@ -60,9 +60,9 @@
       allowedVersions: '<21.0.0',
     },
     {
-      // TODO(STENCIL-710): Increment this value as a part of updating TypeScript
+      // Increment this value as a part of updating TypeScript
       matchPackageNames: ['typescript'],
-      allowedVersions: '<5.1.6',
+      allowedVersions: '<5.3.0',
     },
     {
       // disable Jest updates until the new testing architecture is in place
--- a/package-lock.json
+++ b/package-lock.json
@@ -77,7 +77,7 @@
         "semver": "^7.3.7",
         "sizzle": "^2.3.6",
         "terser": "5.20.0",
-        "typescript": "~5.1.6",
+        "typescript": "~5.2.2",
         "webpack": "^5.75.0",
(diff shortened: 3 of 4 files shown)


package.json
```diff
@@ -122,7 +122,7 @@
     "semver": "^7.3.7",
     "sizzle": "^2.3.6",
     "terser": "5.20.0",
-    "typescript": "~5.1.6",
+    "typescript": "^5.2.2",
```
rwaskiewicz, line 125: Can we re-install this so that we continue to use the tilde?

renovate.json5
```diff
@@ -60,9 +60,9 @@
       allowedVersions: '<21.0.0',
     },
     {
-      // TODO(STENCIL-710): Increment this value as a part of updating TypeScript
+      // Increment this value as a part of updating TypeScript
       matchPackageNames: ['typescript'],
-      allowedVersions: '<5.1.6',
+      allowedVersions: '<5.2.2',
```
rwaskiewicz, line 65: Since we don't know if there'll be a 5.2.3, what're your thoughts on updating this to something like:
```suggestion
      allowedVersions: '<5.3.0',
```

That way, anything that's 5.2.x can get managed by Renovate?

rwaskiewicz (changes requested):
Did a quick pass through - can you do me a favor and make sure this is marked as a `feat()` please?

renovate.json5
alicewriteswrongs: oh yeah that's probably a good idea, I put the exact version in since I thought we usually do all typescript version changes manually but I guess we can probably do patch releases safely

rwaskiewicz: Yeah, I think that compounds with the last time we did a bump (5.0->5.1) we were 2 minors behind, so we assumed there wasn't gonna be an additional patch release for 5.1

rwaskiewicz (approved):

rwaskiewicz:
@alicewriteswrongs can you open a PR to update the Doc Site's table for TS support?

alicewriteswrongs:
oops forgot to copy-paste it! it's up already: https://github.com/ionic-team/stencil-site/pull/1238

tanner-reits (approved):

PR #4861, opened 2023-09-28
chore(test): remove old jest todos

rwaskiewicz:

When I copied Jest infrastructure in #4847, I failed to address/remove TODOs in the code that no longer applied
 
GitHub Issue Number: N/A

## What is the new behavior?

remove a handful of TODOs from the Jest 27 and under section of the codebase. these TODOs were originally written when we planned to have to continue to maintain the internally used version of Jest and the one users use, keeping them in sync. now that these two versions of jest are independent, we can remove these TODOs (as the Jest 27 and under code is not expected to be changed for Jest 28).

- [ ] Yes
- [x] No

## Testing

I only removed comments, no testing applies (obviously CI should still pass)

### DIFF
--- a/src/testing/jest/jest-27-and-under/jest-config.ts
+++ b/src/testing/jest/jest-27-and-under/jest-config.ts
@@ -2,7 +2,6 @@ import type { Config } from '@jest/types';
 import type * as d from '@stencil/core/internal';
 import { isString } from '@utils';
 
-// TODO(STENCIL-306): Remove support for earlier versions of Jest
 /**
  * Helper function for retrieving legacy Jest options. These options have been provided as defaults to Stencil users
  * by Jest + yargs for all users using Jest versions 24 through 26 (inclusively). Between Jest v26 and v27, a few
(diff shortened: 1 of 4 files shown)


src/testing/jest/jest-27-and-under/jest-config.ts
```diff
@@ -146,7 +145,6 @@ export function buildJestConfig(config: d.ValidatedConfig): string {
     jestConfig.verbose = stencilConfigTesting.verbose;
   }
 
-  // TODO(STENCIL-307): Move away from Jasmine runner for Stencil tests, which involves a potentially breaking change
   jestConfig.testRunner = 'jest-jasmine2';
```
rwaskiewicz, line 148: Technically, Jest 27 supports jest-jasmine2 and jest-circus. At this time, we don't allow folks on Jest 27 and lower to configure this. We may need to change that in the future, but I believe the changes would be more than updating the assignment here to 'jest-circus' or something like:
```ts
jestConfig.testRunner = stencilConfigTesting.testRunner ?? 'jest-jasmine2';
```

alicewriteswrongs (approved):

tanner-reits (approved):
Any of these JIRA tickets need to be updated or removed?

rwaskiewicz:
@tanner-reits - They're still valid for Stencil's internally testing setup (and are gonna be completed when we upgrade our usage of Jest to v29). They're just no longer valid for these Jest 27-specific files (since the current plan is that we won't be modifying the Jest 27 code moving forward beyond bug fixes)

PR #4851, opened 2023-09-27
fix(jest): use correct minimum jest version

rwaskiewicz:
## What is the current behavior?

When a user attempts to use jest 24, they are greeted with an error:
```console
$ npm t

> new-test@0.0.1 test
> stencil test --spec --e2e

[51:38.0]  @stencil/core
[51:38.1]  v4.3.0 🐫

[ ERROR ]  Please install supported versions of dev dependencies with either npm or yarn.
           npm install --save-dev jest@27.0.3
```
GitHub Issue Number: N/A

## What is the new behavior?

this commit updates the check for the minimum allowed version of jest from v24.9.1 to v24.9.0. while working on the jest refactoring, it was discovered that v24.9.1 does not exist, making it impossible for anyone to use v24

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Testing

You can validate the bug exists today by trying to use Jest 24 in a new component starter project:
```bash
cd /tmp \
&& npm init stencil@latest component jest-24-test \
&& cd $_ \
&& npm i jest@24 jest-cli@24 @types/jest@24 \
&& npm t

> new-test@0.0.1 test
> stencil test --spec --e2e

[51:38.0]  @stencil/core
[51:38.1]  v4.3.0 🐫

[ ERROR ]  Please install supported versions of dev dependencies with either npm or yarn.
           npm install --save-dev jest@27.0.3
```

To verify this fix, build this branch:
```bash
npm ci \
&& npm run clean \
&& npm run build \
&& npm pack
```

Then spun up a new stencil component library with it:
```bash
cd /tmp \
&& npm init stencil@latest component jest-fix-test \
&& cd $_ \
&& npm i [PATH_TO_TARBALL] \
&& npm i jest@24 jest-cli@24 @types/jest@24 \
&& npm t
```

## Other information

this commit updates the component starter ci tests to run each version of
jest that we support

STENCIL-953

PR #4863, opened 2023-09-28
chore(ci): adjust component start test name to include jest version

alicewriteswrongs:
We added the Jest version to the matrix for the component starter test in #4851 which increases the number of tests being run for the component starter but the names of the tests don't signify which version of jest is being installed in each one.

Here we just adjust the name to include the node and jest versions directly, just so it's easier to see what's up if something is failing.

Screenshot 2023-09-28 at 10 03 18 AM

## What is the new behavior?

The names should be different!

Screenshot 2023-09-28 at 10 51 52 AM

- [ ] Yes
- [x] No

### DIFF
--- a/.github/workflows/test-component-starter.yml
+++ b/.github/workflows/test-component-starter.yml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   analysis_test:
-    name: (${{ matrix.os }}.${{ matrix.node }})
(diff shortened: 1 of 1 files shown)


rwaskiewicz (approved):
Ugh - sorry about that! LMK and I can update the gated checks

christian-bromann (approved):
👍

tanner-reits (approved):

PR #4864, opened 2023-09-28
test(vdom): setting readonly props is ignored

christian-bromann:
## What is the new behavior?
While going through `set-accessor.ts` I was wondering why we would try/catch the prop setting part. My assumption is for cases where we set readonly properties like e.g. `namespaceURI`. This actually is compliant with what the browser does:

Screenshot 2023-09-28 at 4 28 18 PM

- [ ] Yes
- [x] No

## Testing

This patch adds a test for this case.

## Other information

n/a


christian-bromann:
I will take a look at the failing tests tomorrow.

christian-bromann:
This is ready for review now.

rwaskiewicz (approved):

tanner-reits (approved):
LGTM!

PR #4867, opened 2023-09-29
test(vdom): adding tests for insertVdomAnnotations

christian-bromann:
There are no unit tests for `insertVdomAnnotations` and reading the code is hard to understand what is going on.

GitHub Issue Number: N/A

## What is the new behavior?
Added some unit tests to clarify the functionality.

- [ ] Yes
- [x] No

## Testing

I hope it is ok to use `toMatchSnapshot` since I barely seeing usage of it in the repo.

## Other information

Something that concerns me is this text node being added to the output:

```html
<section>
  <cmpa-test>
      <!---->o.0.1 <=== here
      <div>
        <!--t.0.1-->slotContent
      </div>
  </cmpa-test>
  <cmpb-test>
      <!---->o.0.2 <=== and here
      <div>
        <!--t.0.2-->slotContent
      </div>
  </cmpb-test>
</section>
```

I am not sure if these will be replaced when we hydrate the page. Anyone has some insights?


src/runtime/client-hydrate.ts
```diff
@@ -254,7 +254,7 @@ const clientHydrate = (
   }
 };
 
-export const initializeDocumentHydrate = (node: d.RenderNode, orgLocNodes: Map<string, any>) => {
+export const initializeDocumentHydrate = (node: d.RenderNode, orgLocNodes: d.PlatformRuntime['$orgLocNodes$']) => {
```
rwaskiewicz, line 257: I think this introduces a new strictNullCheck error, since `$orgLocNodes$` is an optional field on `PlatformRuntime`. I think in the context of this function, it makes sense for us to narrow it if we want to use the typings on `PlatformRuntime`:
```diff
- export const initializeDocumentHydrate = (node: d.RenderNode, orgLocNodes: d.PlatformRuntime['$orgLocNodes$']) => {
+ export const initializeDocumentHydrate = (node: d.RenderNode, orgLocNodes: Exclude<d.PlatformRuntime['$orgLocNodes$'], undefined>) => {
```

Note: If we use this, we'll probably want to format it with Prettier:
```suggestion
export const initializeDocumentHydrate = (
  node: d.RenderNode,
  orgLocNodes: Exclude<d.PlatformRuntime['$orgLocNodes$'], undefined>,
) => {
```

src/runtime/vdom/test/vdom-annotations.spec.tsx
```diff
@@ -0,0 +1,105 @@
+import { Component, forceUpdate, h, Prop } from '@stencil/core';
+import { newSpecPage } from '@stencil/core/testing';
+
+import { insertVdomAnnotations } from '../vdom-annotations';
+// import { createDocument } from '../../../mock-doc';
```
rwaskiewicz, line 5: Non-blocking - can we remove this?

alicewriteswrongs (approved):
lgtm!

PR #4868, opened 2023-09-29
chore(client): remove performance.mark and performance.measure polyfill

christian-bromann:
We polyfill `performance.mark` and `performance.measure` in case it is not supported. After taking a look into MDN it seems that support for these primitives is way advanced and pretty much given across all environments:

__`performance.mark`__
Screenshot 2023-09-29 at 12 54 10 PM

__`performance.measure`__
Screenshot 2023-09-29 at 12 54 00 PM

## What is the new behavior?
Removing the polyfill and reduce some bytes in bundle size.

- [ ] Yes
- [x] No

Looking at Stencils [browser support policy](https://stenciljs.com/docs/support-policy#browser-support) non of the browser we marked as supported miss these web platform features.

## Testing

n/a

## Other information

n/a

### DIFF
--- a/src/client/client-patch-browser.ts
+++ b/src/client/client-patch-browser.ts
@@ -14,19 +14,6 @@ export const patchBrowser = (): Promise<d.CustomElementsDefineOptions> => {
     patchCloneNodeFix((H as any).prototype);
   }
 
-  if (BUILD.profile && !performance.mark) {
-    // not all browsers support performance.mark/measure (Safari 10)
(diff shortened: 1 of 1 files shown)


rwaskiewicz (approved):
I think this is fine - we could have done this as early as Stencil v3 (and honestly probably just forgot). Since our polyfills/stubs never accepted any args nor returned anything, this should be fine.

Non-blocking - can you do me a favor and update this ticket (STENCIL-323) in JIRA?

alicewriteswrongs (approved):
🪓

PR #4870, opened 2023-09-29
docs(dev-server): document hmr utils

alicewriteswrongs:
Add some documentation to `src/dev-server/client/hmr-util.ts` and friends

- [ ] Yes
- [x] No

## Testing

This should be a comment-only change

### DIFF
--- a/src/dev-server/client/hmr-util.ts
+++ b/src/dev-server/client/hmr-util.ts
@@ -1,3 +1,5 @@
+import { NODE_TYPE } from '../../runtime/runtime-constants';
+
 export const getHmrHref = (versionId: string, fileName: string, testUrl: string) => {
   if (typeof testUrl === 'string' && testUrl.trim() !== '') {
     if (getUrlFileName(fileName) === getUrlFileName(testUrl)) {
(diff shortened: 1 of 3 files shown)


src/dev-server/client/hmr-util.ts
```diff
@@ -54,20 +56,54 @@ export const updateCssUrlValue = (versionId: string, fileName: string, oldCss: s
...
+/**
+ * Determine whether a given element is a `<link>` tag pointing to a stylesheet
+ *
+ * @param elm the element to check
+ * @returns whether or not the element is a link stylesheet
+ */
+export const isLinkStylesheet = (elm: Element): boolean =>
   elm.nodeName.toLowerCase() === 'link' &&
   (elm as HTMLLinkElement).href &&
   (elm as HTMLLinkElement).rel &&
```
rwaskiewicz, line 68: Couldn't these be `undefined`? Wouldn't that make it such that the return type of this function could be `boolean | undefined` as a result?

rwaskiewicz (commented):
One quick question - I think this might have contributed to the SNC increase, but I'm not entirely sure 🤔

src/dev-server/client/hmr-util.ts
tanner-reits: Could add `!!` to both those checks (which would also evaluate `false` if `href` is an empty string)

alicewriteswrongs: mm true, I'll throw a `!!` in there

rwaskiewicz (approved):

PR #4883, opened 2023-10-02
refactor(jest): detect jest version for testing

rwaskiewicz:

stencil will always assume that jest v27 is being used for testing.
this is a result of years  of using the same files for all jest infrastructure for end users.
while this pr does not do away with always picking jest v27, it gets us one step closer

GitHub Issue Number: N/A

## What is the new behavior?

add a new abstraction layer to stencil's jest testing configuration/running such that the version of jest is detected when jest is run. based on the version that is detected, defer to the correct piece of versioned infrastructure to setup and run tests.

in https://github.com/ionic-team/stencil/pull/4847, the directory structure of the testing submodule was updated to look something like:
```
├── src/testing/jest/
│   └── jest-27-and-under
│       ├── jest-config.ts                 
│       ├── jest-runner.ts
│       ├── node_modules/
│       ├── package.json                   
│       └── test
└── package.json                            
```

this commit introduces the following infrastructue to dispatch to jest-27-and-under at runtime:

```diff
  ├── src/testing/jest/
  │   └── jest-27-and-under
  │       ├── jest-config.ts                
  │       ├── jest-environment.ts
+ │       ├── jest-facade.ts                  <- Implements base class in `jest-adapter.ts`
  │       ├── jest-preprocessor.ts
  │       ├── jest-runner.ts
  │       ├── jest-screenshot.ts
  │       ├── jest-serializer.ts
  │       ├── jest-setup-test-framework.ts
  │       ├── node_modules/
  │       ├── package.json                    
  │       └── test
+ ├── jest-apis.ts                            <- Interactions with Jest prior to knowing the version
+ ├── jest-facade.ts                         <- Base class all `jest-facade` implementations should follow
+ ├── jest-stencil-connector.ts               <- Routes Jest logic through the correct `jest-facade` implementation
  └── package.json                            
```
Where:
- `jest-stencil-connector` - performs the detection of the current version of jest - dispatching to the correct infrastructure via `jest-facade` implementations
- `jest-facade` - a file containing a base class to define the API by which version-specific implementations can be loaded. a jest v27-specific implementation is included as well in `jest-27-and-under/`
- `jest-apis` - a file containing typings for cases where the jest typings are ambiguous, particularly for when we do not know what the current version of jest is (yet)

the end result of this commit should be nothing to the end user (for now). this work shall be used to help support jest v28+ in the near future (dispatching the the correct version)

- [ ] Yes
- [x] No

## Testing

Pull down this branch and build it. Make note of the location of the generated tarball:
```bash
npm run clean \                   
&& npm ci \
&& npm run build \
&& npm pack
```
Next, create a new stencil component library with that tarball:
```bash
cd /tmp \
&& npm init stencil@latest component jest-test \
&& cd $_ \
&& npm i [PATH_TO_TARBALL]
```
Stencil Testing should continue to pass for Jest [24,27]:
```bash
npm i jest@24 jest-cli@24 @types/jest@24 && npm t -- --no-cache && \
npm i jest@25 jest-cli@25 @types/jest@25 && npm t -- --no-cache && \
npm i jest@26 jest-cli@26 @types/jest@26 && npm t -- --no-cache && \
npm i jest@27 jest-cli@27 @types/jest@27 && npm t -- --no-cache
```
Next, Stencil should continue to warn if a testing module is not installed:
```
npm uninstall jest \
&& npm t -- --no-cache # expect error
```

### DIFF
--- a/src/testing/jest/jest-27-and-under/jest-config.ts
+++ b/src/testing/jest/jest-27-and-under/jest-config.ts
@@ -2,6 +2,8 @@ import type { Config } from '@jest/types';
 import type * as d from '@stencil/core/internal';
 import { isString } from '@utils';
 
+import { Jest27Stencil } from './jest-facade';
+
 /**
  * Helper function for retrieving legacy Jest options. These options have been provided as defaults to Stencil users
  * by Jest + yargs for all users using Jest versions 24 through 26 (inclusively). Between Jest v26 and v27, a few
@@ -145,7 +147,7 @@ export function buildJestConfig(config: d.ValidatedConfig): string {
     jestConfig.verbose = stencilConfigTesting.verbose;
   }
 
-  jestConfig.testRunner = 'jest-jasmine2';
+  jestConfig.testRunner = new Jest27Stencil().getDefaultJestRunner();
 
   return JSON.stringify(jestConfig);
 }
--- a/src/testing/jest/jest-27-and-under/jest-facade.ts
+++ b/src/testing/jest/jest-27-and-under/jest-facade.ts
@@ -1,0 +1,40 @@
+import { JestFacade } from '../jest-facade';
+import { createJestPuppeteerEnvironment as createJestPuppeteerEnvironment27 } from './jest-environment';
+import { jestPreprocessor as jestPreprocessor27 } from './jest-preprocessor';
+import { createTestRunner as createTestRunner27 } from './jest-runner';
+import { runJest as runJest27 } from './jest-runner';
+import { runJestScreenshot as runJestScreenshot27 } from './jest-screenshot';
+import { jestSetupTestFramework as jestSetupTestFramework27 } from './jest-setup-test-framework';
+
+/**
+ * `JestFacade` implementation for communicating between this directory's version of Jest and Stencil
+ */
+export class Jest27Stencil implements JestFacade {
+  getJestCliRunner() {
+    return runJest27;
+  }
+
+  getRunJestScreenshot() {
+    return runJestScreenshot27;
+  }
+
+  getDefaultJestRunner() {
+    return 'jest-jasmine2';
+  }
+
+  getCreateJestPuppeteerEnvironment() {
+    return createJestPuppeteerEnvironment27;
+  }
+
+  getJestPreprocessor() {
+    return jestPreprocessor27;
+  }
+
+  getCreateJestTestRunner() {
+    return createTestRunner27;
+  }
+
+  getJestSetupTestFramework() {
+    return jestSetupTestFramework27;
+  }
+}
--- a/src/testing/jest/jest-27-and-under/jest-runner.ts
+++ b/src/testing/jest/jest-27-and-under/jest-runner.ts
@@ -3,6 +3,7 @@ import type * as d from '@stencil/core/internal';
 
 import type { ConfigFlags } from '../../../cli/config-flags';
 import { setScreenshotEmulateData } from '../../puppeteer/puppeteer-emulate';
+import type { JestTestRunner } from '../jest-apis';
 import { buildJestArgv, getProjectListFromCLIArgs } from './jest-config';
 
 export async function runJest(config: d.ValidatedConfig, env: d.E2EProcessEnv) {
@@ -49,7 +50,7 @@ export async function runJest(config: d.ValidatedConfig, env: d.E2EProcessEnv) {
  * Creates a Stencil test runner
  * @returns the test runner
  */
-export function createTestRunner(): any {
+export function createTestRunner(): JestTestRunner {
   // The left hand side of the '??' is needed for Jest v27, the right hand side for Jest 26 and below
   const TestRunner = require('jest-runner').default ?? require('jest-runner');
 
--- a/src/testing/jest/jest-apis.ts
+++ b/src/testing/jest/jest-apis.ts
@@ -1,0 +1,48 @@
+/*!
+ * This file contains Jest API usages for situations where it is difficult to determine which API should be used.
+ *
+ * An example of this is determining the version of Jest, which is retrieved via the `getVersion` API.
+ * It's difficult at compile & runtime to determine:
+ * 1. If such an API exists
+ * 2. If it's typings are the same across all versions of Jest
+ * 3. If there are variants of this API, which one to use and when
+ *
+ * Short of probing the directory where a user keeps their modules (e.g. `node_modules/`), we need to make a "best
+ * guess" at things. This file is meant to only contain functions for these types of scenarios. It is expected that this
+ * file be added to sparingly.
+ */
+
+import { getVersion } from 'jest';
+
+// TODO(STENCIL-959): Improve this typing by narrowing it
+export type JestPuppeteerEnvironment = any;
+
+type Jest26CacheKeyOptions = { instrument: boolean; rootDir: string };
+type Jest26Config = { instrument: boolean; rootDir: string };
+type Jest27TransformOptions = { config: Jest26Config };
+export type JestPreprocessor = {
+  process(
+    sourceText: string,
+    sourcePath: string,
+    jestConfig: Jest26Config | Jest27TransformOptions,
+    transformOptions?: Jest26Config,
+  ): string;
+  getCacheKey(
+    sourceText: string,
+    sourcePath: string,
+    jestConfigStr: string | Jest27TransformOptions,
+    transformOptions?: Jest26CacheKeyOptions,
+  ): string;
+};
+
+// TODO(STENCIL-960): Improve this typing by narrowing it
+export type JestTestRunner = any;
+
+/**
+ * Get the current major version of Jest that Stencil reconciles
+ *
+ * @returns the major version of Jest.
+ */
+export const getJestMajorVersion = (): string => {
+  return getVersion();
+};
--- a/src/testing/jest/jest-stencil-connector.ts
+++ b/src/testing/jest/jest-stencil-connector.ts
@@ -1,0 +1,114 @@
+/*!
+ * This file acts as the connector/bridge between Stencil and Jest.
+ *
+ * It defines/caches a `JestFacade` implementation to dispatch Jest-related configuration calls to the correct section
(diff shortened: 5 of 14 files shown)


src/testing/jest/jest-27-and-under/jest-runner.ts
```diff
@@ -49,7 +50,7 @@ export async function runJest(config: d.ValidatedConfig, env: d.E2EProcessEnv) {
  * Creates a Stencil test runner
  * @returns the test runner
  */
-export function createTestRunner(): any {
```
rwaskiewicz, line 52: This was previously typed as `any` - `JestTestRunner` is just a type alias for `any` at this point, but I saved that piece of tech debt for another day - https://github.com/ionic-team/stencil/pull/4883/files#diff-4b0dc677b3f6e55c5f547f5585915852d874a4a1ff77a6211becc016251521fcR38-R39

src/testing/jest/jest-adapter.ts
```diff
@@ -0,0 +1,74 @@
+import { JestPreprocessor, JestPuppeteerEnvironment, JestTestRunner } from './jest-apis';
+
+/**
+ * Base class for Jest-version specific code implementations that interact with Stencil.
+ *
+ * It is expected that there exists a Jest version-specific implementation for this adapter in each version-specific
+ * directory Stencil supports.
+ */
+export abstract class JestAdapter {
+  // TODO(STENCIL-961): Fix build validation when types are pulled in from `@stencil/core/declarations`
```
rwaskiewicz, line 10: Importing from `@stencil/core` or another subdirectory, the `test.dist` scripts begin to fail. I'm kicking the can down the road a little here on that, in case a better design reveals itself as I work on v28+ implementatons

src/testing/jest/jest-stencil-connector.ts
```diff
@@ -0,0 +1,114 @@
...
+ * Retrieve the cached local variable containing a Jest adapter, based on the version of Jest detected.
+ * If no Jest adapter is cached, set it.
+ *
+ * @returns the cached Jest adapter.
+ */
+const getJestAdapter = (): JestAdapter => {
+  if (!JEST_ADAPTER) {
+    const version = getVersion();
+    if (version <= 27) {
+      JEST_ADAPTER = new Jest27StencilAdapter();
```
tanner-reits, line 42: I thought maybe this would be able to resolve the "adapter" based on consistent file structure and naming conventions (and a default export from the file). That too magical?

src/testing/jest/jest-27-and-under/jest-facade.ts
tanner-reits, line 1: Just curious, where did the term "facade" come from for these when we use "adapter" for the actual implementations and definitions?

src/testing/jest/jest-27-and-under/jest-config.ts
```diff
@@ -145,7 +146,7 @@ export function buildJestConfig(config: d.ValidatedConfig): string {
     jestConfig.verbose = stencilConfigTesting.verbose;
   }
 
-  jestConfig.testRunner = 'jest-jasmine2';
+  jestConfig.testRunner = getDefaultJestRunner();
```
tanner-reits, line 149: Should this pull directly from the the v27 adapter rather than the abstract one? I know it evaluates to the same thing, but that would keep the version-specific adapters from "knowing about" the more generic stuff

src/testing/jest/jest-apis.ts
```diff
@@ -0,0 +1,48 @@
...
+
+import { getVersion } from 'jest';
+
+// TODO(STENCIL-959): Improve this typing by narrowing it
+export type JestPuppeteerEnvironment = any;
+
+type Jest26CacheKeyOptions = { instrument: boolean; rootDir: string };
+type Jest26Config = { instrument: boolean; rootDir: string };
+type Jest27TransformOptions = { config: Jest26Config };
+export type JestPreprocessor = {
```
tanner-reits, line 23: Another spot where it'd be nice if the abstract layer didn't need to worry about version-specific APIs. Anything we can do with generics here? How are these types consumed?

src/testing/jest/jest-stencil-connector.ts
rwaskiewicz: The original prototype actually used something very similar to what you're describing:

```ts
class Jest27StencilAdapter extends JestFacade {
  static override async getRunner() {
    return await import('./experimental/jest-27-and-under/jest-runner');
  }
  static override async getScreenshot() {
    return await import('./experimental/jest-27-and-under/jest-screenshot');
  }
}
```

I found there were a couple shortcomings with this:
1. Type safety wasn't as "nice" as I wanted it to be (the return type of `getRunner` is a `Promise<>` that returns the shape of _whatever_ the export of `jest-runner` is). I found this to be kind of annoying to work with/harder to understand
2. It introduced `async/await` to the the whole call chain, which wasn't strictly necessary
3. Refactoring (say for instance, I were to hoist 'jest-27-and-under' out of that 'experimental' directory 😉) wasn't very pleasurable either.

By using the Facade pattern in this way, we can be super explicit about what adapters/implementations map to which versions of Jest, avoid some async/await overhead, and have some better (IMO) refactoring ability.

src/testing/jest/jest-27-and-under/jest-facade.ts
rwaskiewicz: "Facade" comes from our use of the Facade Pattern here - adapter isn't _quite_ right in terms of a classic Gang of Four pattern, but I renamed things so many times that it probably "just made sense" to me at the time. I'll rename these a bit for some better....well, to be better 😆

src/testing/jest/jest-27-and-under/jest-config.ts
rwaskiewicz: I'm not sure - my only concern here is that the general version has a cache of the adapter, whereas the version specific versions don't. Let me verify this won't run >1 for `npx stencil test` and I'll circle back

src/testing/jest/jest-apis.ts
rwaskiewicz: > Another spot where it'd be nice if the abstract layer didn't need to worry about version-specific APIs. 

Trust me, I wish we could 😆 The problem is that the generic version of our connectors/adapters need to return _something_. And those something's need to be compatible with Jest.

> Anything we can do with generics here? How are these types consumed?

Right now, they're only used by `jest-adapter.ts`. I think in the near future, we'll end up refactoring this to be less aware of these types. But the exact shape/form hasn't "presented itself" yet - I have a feeling once we get Jest 28 in place, that'll start to become more evident

src/testing/jest/jest-27-and-under/jest-facade.ts
rwaskiewicz: Okay, I think I made this a little more cohesive in 2ac845e4a829781fac4d6f8aed7d3bf824ee95b0 - LMK what you think

src/testing/jest/jest-27-and-under/jest-config.ts
rwaskiewicz: Mkay - we can do this 👍 Verified that this isn't called >1, so we don't end up instantiating a new Facade instance each time we test. Updated in 615402cfec0416af55214d8af7c4a8d0599ba37d

src/testing/jest/jest-27-and-under/jest-facade.ts
```diff
@@ -0,0 +1,40 @@
...
+import { jestPreprocessor as jestPreprocessor27 } from './jest-preprocessor';
+import { createTestRunner as createTestRunner27 } from './jest-runner';
+import { runJest as runJest27 } from './jest-runner';
+import { runJestScreenshot as runJestScreenshot27 } from './jest-screenshot';
+import { jestSetupTestFramework as jestSetupTestFramework27 } from './jest-setup-test-framework';
+
+/**
+ * `JestFacade` implementation for communicating between between this directory's version of Jest and Stencil
+ */
+export class Jest27Stencil extends JestFacade {
```
alicewriteswrongs, line 12: j/w, are there any advantages / disadvantages either way between the approaches of having an ABC and implementations that extend it (_a là_ `JEst27Stencil` here) vs an interface that the actual implemetations then `implement`? classes with the `abstract` keyword do actually have a runtime representation and can have concrete methods on them - right now it looks like the `JestFacade` doesn't have any concrete / non-abstract methods on it so would basically be `class JestFacade { }` after compilation.

Did you give any thought to using an interface to constrain the shape of the facade classes (like `Jest27Stencil` here) instead of an ABC? Maybe we will need to have some behavior on the base class in the future

anyway just sort of thinking through this as I read the PR

alicewriteswrongs (approved):
I was able to build this and then run tests in a new project with no problems, so everything seems in working order - just had one sort of structural question about using an ABC vs possible a class or object conforming to an interface. Looks good overall and no real concerns!

src/testing/jest/jest-27-and-under/jest-facade.ts
rwaskiewicz: I did give it some thought, and originally had gone the ABC route because I did have a few concrete implementations in the base class here - but those have obviously been removed since then (hence your question). It's relatively trivial to convert between the two. I say let's go with an interface for now unless we find we really need it later down the road. Done in 4327cfcca

alicewriteswrongs (approved):

tanner-reits (approved):
LGTM. I think this would be a good candidate for some sort of guide to follow when adding support (and removing support) for Jest versions. Just because we have to add the version-specific adapter code and remember to update the abstract layer to recognize and use the correct versions

rwaskiewicz:
@tanner-reits 

> ... I think this would be a good candidate for some sort of guide to follow when adding support (and removing support) for Jest versions. Just because we have to add the version-specific adapter code and remember to update the abstract layer to recognize and use the correct versions

That's my current plan - I'll write one up while I do the Jest 28 related stuff, then try to follow it verbatim for Jest 29 😆 🤞

alicewriteswrongs:
ditto on the upgrade guide - ideally just a README.md in the right spot so you can't miss it

PR #4887, opened 2023-10-03
chore(repo): ensure jest packages have installed deps

rwaskiewicz:

one consequence of moving to the design laid out in https://github.com/ionic-team/stencil/pull/4847 is that there are now `package.json` file(s) that:
- define jest dependencies that are not used by stencil's own internal testing
- do not have their dependencies installed when we run `npm ci` (or equivalent)

as a result, it's entirely possible for a stencil engineer to work on the project without having
the right dependencies!

GitHub Issue Number: N/A

## What is the new behavior?

this commit introduces a small shell script that iterates over all jest subdirectories (currently just one, `jest-27-and-under`) and runs `npm ci`. this ensures that these dependencies are installed/up to date when working with stencil

- [ ] Yes
- [x] No

## Testing

1. Pull this branch
2. Manually remove `src/testing/jest/jest-27-and-under/node_modules` if they exist
3. Run `npm ci` from the root of this project
4. See that `node_modules` have been recreated

### DIFF
--- a/.github/workflows/test-component-starter.yml
+++ b/.github/workflows/test-component-starter.yml
@@ -82,7 +82,7 @@ jobs:
         shell: bash
 
       - name: Install Stencil Eval
-        run: npm i ../stencil-eval.tgz
+        run: npm i --ignore-scripts ../stencil-eval.tgz
         working-directory: ./tmp-component-starter
         shell: bash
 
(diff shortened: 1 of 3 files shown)


rwaskiewicz:
Lint is expected to fail until https://github.com/ionic-team/stencil/pull/4886 lands

rwaskiewicz:
Pulling this back into draft, something's clearly not right 🤣

rwaskiewicz:
Closing this, it's not as elegant/usable as I'd like. I'll be back with....something else

PR #4890, opened 2023-10-03
chore(snc): resolve (most) jest strictNullChecks

rwaskiewicz:

there's a handful of strictNullChecks in the Jest v27 infra. before I start on the jest 28+ work, I'd like to sqaush these
GitHub Issue Number: N/A

## What is the new behavior?

resolve strictNullCheck violations in `src/testing/jest/jest-27-and-under` that only require updating the types on function/variable declarations, or a type assertion in the case of tests not understanding that a property will be added to an object

- [ ] Yes
- [x] No

## Testing

Unit tests should continue to pass.

The updates in this PR _shouldn't_ cause a functional change, but let me know if something smells fishy to you!

## Other information

I didn't squash all of the SNC violations for this directory - there are a handful of them in the matchers dir that require some additional investigation (which may require more work), so I'll put that in a separate PR

### DIFF
--- a/src/testing/jest/jest-27-and-under/jest-runner.ts
+++ b/src/testing/jest/jest-27-and-under/jest-runner.ts
@@ -60,7 +60,7 @@ export function createTestRunner(): any {
       // filter out only the tests the flags said we should run
       tests = tests.filter((t) => includeTestFile(t.path, env));
 
-      if (env.__STENCIL_SCREENSHOT__ === 'true') {
+      if (env.__STENCIL_SCREENSHOT__ === 'true' && env.__STENCIL_EMULATE_CONFIGS__) {
         // we're doing e2e screenshots, so let's loop through
         // each of the emulate configs for each test
 
@@ -105,8 +105,8 @@ export function includeTestFile(testPath: string, env: d.E2EProcessEnv) {
   return false;
 }
 
-export function getEmulateConfigs(testing: d.TestingConfig, flags: ConfigFlags) {
-  let emulateConfigs = testing.emulate.slice();
+export function getEmulateConfigs(testing: d.TestingConfig, flags: ConfigFlags): d.EmulateConfig[] {
+  let emulateConfigs = testing.emulate?.slice() ?? [];
 
   if (typeof flags.emulate === 'string') {
     const emulateFlag = flags.emulate.toLowerCase();
--- a/src/testing/jest/jest-27-and-under/jest-setup-test-framework.ts
+++ b/src/testing/jest/jest-27-and-under/jest-setup-test-framework.ts
@@ -93,7 +93,7 @@ export function jestSetupTestFramework() {
  *
  * @param node The mocked DOM node that will be removed from the DOM
  */
-export function removeDomNodes(node: MockNode) {
+export function removeDomNodes(node: MockNode | undefined | null) {
   if (node == null) {
     return;
   }
--- a/src/testing/jest/jest-27-and-under/test/jest-setup-test-framework.spec.ts
+++ b/src/testing/jest/jest-27-and-under/test/jest-setup-test-framework.spec.ts
@@ -15,7 +15,7 @@ describe('jest setup test framework', () => {
     });
 
     it('does nothing if there is no parent node', () => {
-      const parentNode: MockNode = undefined;
+      const parentNode: MockNode | undefined = undefined;
 
       removeDomNodes(parentNode);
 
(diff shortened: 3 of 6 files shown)


src/testing/jest/jest-27-and-under/jest-runner.ts
```diff
@@ -60,7 +60,7 @@ export function createTestRunner(): any {
       // filter out only the tests the flags said we should run
       tests = tests.filter((t) => includeTestFile(t.path, env));
 
-      if (env.__STENCIL_SCREENSHOT__ === 'true') {
```
rwaskiewicz, line 63: Add this check since `JSON.parse` would [fail here](https://github.com/ionic-team/stencil/pull/4890/files#diff-7cec921f2833685474ae4b5c745bb512842c238e18cd9c729f9db6de28d81f5bL69) if `__STENCIL_EMULATE_CONFIGS__` was `undefined | null` or a falsy string.  This check doesn't handle the case of malformed JSON, as I was only looking at SNC violation fixes here

src/testing/jest/jest-27-and-under/jest-runner.ts
```diff
@@ -105,8 +105,8 @@ export function includeTestFile(testPath: string, env: d.E2EProcessEnv) {
   return false;
 }
 
-export function getEmulateConfigs(testing: d.TestingConfig, flags: ConfigFlags) {
-  let emulateConfigs = testing.emulate.slice();
+export function getEmulateConfigs(testing: d.TestingConfig, flags: ConfigFlags): d.EmulateConfig[] {
+  let emulateConfigs = testing.emulate?.slice() ?? [];
```
rwaskiewicz, line 109: We technically backfill this property when we validate the testing config, but at this time, don't have a concept of a `ValidatedTestingConfig`

src/testing/jest/jest-27-and-under/test/jest-setup-test-framework.spec.ts
```diff
@@ -15,8 +15,9 @@ describe('jest setup test framework', () => {
     });
 
     it('does nothing if there is no parent node', () => {
-      const parentNode: MockNode = undefined;
+      const parentNode: MockNode | undefined = undefined;
 
+      // the intent of this test is to ensure
```
tanner-reits, line 20: Ensure what? TELL ME 😂

tanner-reits (changes requested):
Looks good, one itty bitty thing that got missed

src/testing/jest/jest-27-and-under/test/jest-setup-test-framework.spec.ts
rwaskiewicz: NEVER! I'M TAKING SECRET ONE WITH ME!

rwaskiewicz: I think this was just a backed out copy/paste that I didn't fully undo 🙃 

I'll give it the ol' delete-a-roo

src/testing/jest/jest-27-and-under/jest-setup-test-framework.ts
```diff
@@ -93,7 +93,7 @@ export function jestSetupTestFramework() {
  *
  * @param node The mocked DOM node that will be removed from the DOM
  */
-export function removeDomNodes(node: MockNode) {
+export function removeDomNodes(node: MockNode | undefined) {
```
alicewriteswrongs, line 96: should this be `| undefined | null` ? just wondering based on the `== null` below, which I know will coerce `undefined` correctly, but I _think_ if it's typed like this there would still be a SNC error at the call site if you did `removeDomNodes(null)` and the `== null` sort of implies that could happen?

alicewriteswrongs (changes requested):
just one Q

src/testing/jest/jest-27-and-under/jest-setup-test-framework.ts
rwaskiewicz: It probably should be - at the moment, it's not contributing to any SNC's (on this branch) at the moment, which is interesting 🤔 I'll update it!

rwaskiewicz: cf2c060ea

alicewriteswrongs (approved):

tanner-reits (approved):

PR #4891, opened 2023-10-03
fix(screenshot): alert user when toMatchScreenshot uses NaN

rwaskiewicz:

we have a few strictNullChecks violations for `toMatchScreenshot`, which are a result of a bug/not so-nice failure mode in the matcher.

GitHub Issue Number: N/A

## What is the new behavior?

ensures that `deviceScaleFactor` is properly narrowed to type `number` from `number | undefined`. prior to this commit, the property _could_ be set to `undefined`. this would result in multiplying by `undefined`, producing `NaN` in our calculation of mismatched pixels. this value would be used in a comparison operator, which would always return false.

that is, if `deviceScaleFactor` is `undefined`, we would always fail a test using this matcher. that logic doesn't actually change - instead, we throw a much more explicit error instead of silently failing

- [ ] Yes
- [x] No


src/testing/jest/jest-27-and-under/matchers/screenshot.ts
```diff
@@ -12,7 +12,15 @@ export function toMatchScreenshot(compare: d.ScreenshotDiff, opts: d.MatchScreen
   }
 
   if (typeof compare.mismatchedPixels !== 'number') {
-    throw new Error(`expect toMatchScreenshot() value is not a screenshot compare`);
+    throw new Error(
+      `expect toMatchScreenshot() value is not a valid screenshot compare object - 'mismatchedPixels' has type '${typeof compare.mismatchedPixels}', but should be a number`,
```
rwaskiewicz, line 16: This error message before wasn't the greatest - since I was already doing the same thing for `deviceScaleFactor` I decided to copy/paste/replace that error msg here too

tanner-reits (approved):

alicewriteswrongs (approved):

alicewriteswrongs:
The fix for this issue has been released as a part of today's [Stencil
v4.4.1 release](https://github.com/ionic-team/stencil/releases/tag/v4.4.1).

PR #4901, opened 2023-10-05
chore(compiler): cleanup www validation

tanner-reits:

The validation code is a bit messy and contains a few SNC violations

GitHub Issue Number: N/A

## What is the new behavior?

Code structure matches some other validation logic we do for things like our config and other output targets. Less SNC violations.

- [ ] Yes
- [X] No

## Testing

Tests continue to pass


src/compiler/config/outputs/validate-www.ts
```diff
@@ -34,105 +33,151 @@ export const validateWww = (config: d.ValidatedConfig, diagnostics: d.Diagnostic
...
-      outputs.push({
-        type: DIST_GLOBAL_STYLES,
-        file: join(buildDir, `${config.fsNamespace}.css`),
-      });
-
-      return outputs;
-    },
-    [],
-  );
+  const outputs: (
```
tanner-reits, line 36: Looks like a lot more changes here than really does. This is similar to what Alice did on recent PR to just get rid of the `reduce` call that can make this a bit less readable

src/compiler/config/outputs/validate-www.ts
```diff
@@ -34,105 +33,151 @@ export const validateWww = (config: d.ValidatedConfig, diagnostics: d.Diagnostic
...
 };
 
 const validateWwwOutputTarget = (
   config: d.ValidatedConfig,
   outputTarget: d.OutputTargetWww,
   diagnostics: d.Diagnostic[],
-) => {
-  if (!isString(outputTarget.baseUrl)) {
-    outputTarget.baseUrl = '/';
+): d.ValidatedOutputTargetWww => {
```
tanner-reits, line 90: Broke a lot of these checks into separate helpers since certain fields often relied on each other. Didn't document the functions since that seems more like documentation for the sake of documentation rather than documentation that would actually add extra context

src/compiler/config/outputs/validate-www.ts
```diff
@@ -34,105 +33,151 @@ export const validateWww = (config: d.ValidatedConfig, diagnostics: d.Diagnostic
...
+  if (!isAbsolute(indexHtml)) {
+    indexHtml = join(appDir, indexHtml);
   }
 
-  if (!isAbsolute(outputTarget.indexHtml)) {
-    outputTarget.indexHtml = join(outputTarget.appDir, outputTarget.indexHtml);
+  return indexHtml;
+};
+
+const validatePrerenderConfig = (
```
tanner-reits, line 154: Moved this over from the separate file where it was before. We only use this helper validating this target and we didn't have any tests written specifically for it

tanner-reits:
Tests should be fixed once the CI patch lands and this is rebased

src/compiler/config/outputs/validate-www.ts
alicewriteswrongs: I think it's pretty clear what `validateAppDir` does!

src/compiler/config/test/validate-service-worker.spec.ts
```diff
@@ -33,161 +30,108 @@ describe('validateServiceWorker', () => {
...
    *
-   * @param target the output target from which we want to pull the serviceWorker
+   * @param serviceWorker The value returned from `validateServiceWorker`
    * @returns a serviceWorker object or `void`, with a `void` return being
    * accompanied by a manually-triggered test failure.
    */
-  function getServiceWorker(target: OutputTargetWww) {
-    if (target.serviceWorker) {
-      return target.serviceWorker;
+  function getServiceWorker(serviceWorker: d.ValidatedOutputTargetWww['serviceWorker']) {
```
alicewriteswrongs, line 37: this function name now seems a bit out-of-date, maybe `ensureServiceWorkerIsDefined` ? or something along those lines

also do we have a generic null-guard function like this hanging around somewhere? I thought maybe we did...

src/compiler/config/validate-service-worker.ts
```diff
@@ -18,68 +18,70 @@ import type * as d from '../../declarations';
...
+ * or a {@link d.ServiceWorkerConfig} otherwise.
  */
-export const validateServiceWorker = (config: d.ValidatedConfig, outputTarget: d.OutputTargetWww): void => {
-  if (outputTarget.serviceWorker === false) {
-    return;
+export const validateServiceWorker = (
+  config: d.ValidatedConfig,
+  serviceWorker: d.OutputTargetWww['serviceWorker'],
+  appDir: string,
+): d.ValidatedOutputTargetWww['serviceWorker'] => {
```
alicewriteswrongs, line 31: is `d.ValidatedOutputTargetWww['serviceWorker']` a different type than `d.OutputTargetWww['serviceWorker']` ? afaik they'll both resolve to `ServiceWorkerConfig | null | false`

also after thinking on it a bit I'm wondering if we want to have types like `ValidatedOutputTargetWww` hanging around (i.e. in `src/declarations/stencil-public-compiler.ts`) since I think if we did this for all of the output targets and other things we validate it could be quite a proliferation of types.

One alternative might be to inline `Required<OutputTargetWww>` in `src/compiler/config/outputs/validate-www.ts`. Just thinking that even if `validateWww` returns an array holding a union that includes `ValidatedOutputTargetWww` the `outputTargets` prop on `ValidatedConfig` is still going to be just `OutputTarget[]` so although we have some greater type assurance _while validating_ from `ValidatedOutputTargetWww` we would essentially type-hole that when `.outputTargets` is accessed later on...

I guess the "real" way to address that would be to introduce 'validated' versions of _all_ the output targets and then change the type of `ValidatedConfig['outputTargets']` to be `ValidatedOutputTarget[]` which is a big lift!

Anywhoo - just thinking that through a bit, lmk what you think

alicewriteswrongs (changes requested):
two questions after reading

src/compiler/config/test/validate-service-worker.spec.ts
tanner-reits: Good call on the function, the comment was outdated too so I'll get that fixed up. I'm not sure about the null guard. Haven't seen anything like that, but I can swap it out if you know where that might be

src/compiler/config/validate-service-worker.ts
tanner-reits: I'm not tied to the validated target types by any means. I think they might make more sense down the road as we fix up types in all our validation logic since there _may_ be a case where the types between a validated and non-validated output target config are different other than being required... but idk. 

I know we **didn't** create a dedicated type for `dist` after fixing the SNCs in there, so happy to switch this back!

alicewriteswrongs (approved):

PR #4904, opened 2023-10-06
refactor(jest): make jest presets version-specific

rwaskiewicz:
## Note to Reviewers

Depends on #4883, please review but do not merge 

Our existing Jest preset is fixed. That is, it assumes that it will be static/unchanged regardless of the version of Jest that a user uses. This PR seeks to change that

GitHub Issue Number: N/A

## What is the new behavior?

this commit updates the existing infrastructure for jest presets to move
them under version-specific directories (although there's just one such
directory at this time).

in doing so, the `jest-preset.js` file that was previously housed under
the `scripts/` directory now has the same dispatch behavior as the other
javascript files in the same directory (`git mv` was used to preserve
history of the original contents). this allows the correct version of
`jest-preset` to be selected at runtime when a user invokes jest.

`jest-preset.ts` contains the original preset with the following
changes:
1. it no longer uses CommonJS exports to make loading the file at
   runtime easier and compile-time compliant. it is worth noting that
   the contents of this file, in their previous location, would simply
   be copied to the output directory. now, they are run through the
   typescript compiler and placed in `testing/index.js` after the
   compiler has been built
2. `path` is now imported with an `import` statement, rather than a
   `require` statement as a consequence of point 1.

the resulting directory structure looks something like:

```diff
  ├── scripts/bundles/helpers/jest/
  │   └── jest-preset.js                      <- Now has same dispatch behavior as other files in this directory
  ├── src/testing/jest/
  │   └── jest-27-and-under
  │       ├── jest-config.ts
  │       ├── jest-environment.ts
  │       ├── jest-facade.ts
  │       ├── jest-preprocessor.ts
+ │       ├── jest-preset.ts                  <- Contains preset previously in `scripts/`, now as a module
  │       ├── jest-runner.ts
  │       ├── jest-screenshot.ts
  │       ├── jest-serializer.ts
  │       ├── jest-setup-test-framework.ts
  │       ├── node_modules/
  │       ├── package.json
  │       └── test
  ├── jest-adapter.ts
  ├── jest-apis.ts
  ├── jest-stencil-connector.ts
  └── package.json
```

the jest adapter and it's related classes have been updated to return
the correct jest preset

- [ ] Yes
- [x] No

## Testing

Pull down this branch and build it. Make note of the location of the generated tarball:
```bash
npm run clean \                   
&& npm ci \
&& npm run build \
&& npm pack
```
Next, create a new stencil component library with that tarball:
```bash
cd /tmp \
&& npm init stencil@latest component jest-test \
&& cd $_ \
&& npm i [PATH_TO_TARBALL]
```
Stencil Testing should continue to pass for Jest [24,27]:
```bash
npm i jest@24 jest-cli@24 @types/jest@24 && npm t -- --no-cache && \
npm i jest@25 jest-cli@25 @types/jest@25 && npm t -- --no-cache && \
npm i jest@26 jest-cli@26 @types/jest@26 && npm t -- --no-cache && \
npm i jest@27 jest-cli@27 @types/jest@27 && npm t -- --no-cache
```

## Other information

I recommended walking the commits in this PR to review. They better illustrate the `git mv` step and the changes made. Although the summation of the diffs is something like `+74/-39`, the 'show me all changes in this PR view' makes t[his look far worse than it actually is.](https://github.com/ionic-team/stencil/pull/4904/files)


src/testing/jest/jest-27-and-under/jest-preset.ts
```diff
@@ -0,0 +1,41 @@
...
+const testingDir = __dirname;
+const rootDir = join(testingDir, '..');
+const internalDir = join(rootDir, 'internal');
+
+// NOTE: if you change this, also change compiler/transpile.ts. Search for 'mod_extensions_jest' to find other comments
+// like it.
+const moduleExtensions = ['ts', 'tsx', 'js', 'mjs', 'jsx'];
+const moduleExtensionRegexp = '(' + moduleExtensions.join('|') + ')';
+
+const preset: Config.InitialOptions = {
```
tanner-reits, line 17: Can we use the `JestConfig` type you made?

tanner-reits (approved):
One non-blocker. Haven't ran any of the test rework stuff locally yet, but I'll do that once I'm out of PR review hell

src/testing/jest/jest-27-and-under/jest-preset.ts
rwaskiewicz: The `JestConfig` type is actually going to be changing in the near-ish future when we add support for v28+ to be wider than just the typings of Jest 27. By using `Config.InitialOptions` here, we guarantee that this preset adheres to Jest 27's configuration shape, rather than one that (will) be a combination of 27 + 28 + 29

Issue #4916, opened 2023-10-09
feat: use `globalThis` as window fallback in client-window.ts


dwirz:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already include this feature request, without success.

### Describe the Feature Request

Use `globalThis` as `window` fallback in `client-window.ts` so that it gets possible to run it in an Node environment. 

### Describe the Use Case

In one of our client project we'd like to render stencil generated web components in a Next.js app directory application. With the help of [luwes/wesc](https://github.com/luwes/wesc) this is "possible" some stuff is still not working but the biggest change we need is that stencil is using `globalThis` so we can provide the necessary functions when the components are rendered server side in Node. (PR incoming)

rwaskiewicz:
Hey @dwirz 👋 

Thanks for the feature request + PR! 

While the code that you provided in the PR seems safe, we'd like a little more information here. From our side, we don't quite understand why this change is necessary based on the information provided. We'd love the have the ability to reproduce the issue that you're seeing so that we can verify the fix works as intended. This is important for the long term maintenance of the code, so that we can understand why a change was made and to avoid regressions.

Can you do me a favor and provide a minimal reproduction case for the error you're seeing? From there, we can move forward with your PR.


Thanks

dwirz:
Hi @rwaskiewicz 

I created a repo with a minimal reproduction description [here](https://github.com/smartive/stencil-nextjs-example).

Cheers

dwirz:
Hey @rwaskiewicz 

> Is it the case that this Stencil library is using the WeSC library to wrap the components to use React, so that they can be used in a React/NextJS environment? 

Yes, the [WeSC wrapper](https://github.com/smartive/stencil-nextjs-example/blob/main/packages/app/src/components/button.tsx#L18) component and [server import](https://github.com/smartive/stencil-nextjs-example/blob/main/packages/app/src/app/page.tsx#L1) makes it possible to render Web Components server side in a Node environment.

> Do the [Stencil React Wrappers](https://www.npmjs.com/package/@stencil/react-output-target) not fit your needs?

[...]

PR #4917, opened 2023-10-09
feat: use globalThis as fallback if window is not available

dwirz:
Fixes: #4916

## What is the new behavior?
It is possible to provide server side needed functions (e.g. `customElements`) via `globalThis`.

- [ ] Yes
- [x] No

### DIFF
--- a/src/client/client-window.ts
+++ b/src/client/client-window.ts
@@ -2,7 +2,7 @@ import { BUILD } from '@app-data';
 
 import type * as d from '../declarations';
 
-export const win = typeof window !== 'undefined' ? window : ({} as Window);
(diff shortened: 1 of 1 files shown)


rwaskiewicz:
Hey @dwirz!

We would like to discuss this change further in https://github.com/ionic-team/stencil/issues/4916#issuecomment-1756154248 in order to proceed with this PR

EriCreator (approved):
nice catch!

rwaskiewicz:
Thanks for the PR!

We've decided not to move forward with this implementation, which I've explained the reasons for in https://github.com/ionic-team/stencil/issues/4916#issuecomment-1792948113. As a result, I'm going to close this out.

PR #4935, opened 2023-10-16
chore(): propagate our patched `join` helper

alicewriteswrongs:
This propagates usage of our `join` helper in `@utils` throughout the codebase in places where it may possibly be relevant.

GitHub Issue Number: N/A

## What is the new behavior?

- 

- [ ] Yes
- [ ] No

### DIFF
--- a/src/compiler/output-targets/copy/hashed-copy.ts
+++ b/src/compiler/output-targets/copy/hashed-copy.ts
@@ -1,5 +1,5 @@
-import { dirname, extname } from 'path';
 import { join } from '@utils';
+import { dirname, extname } from 'path';
 
 import type * as d from '../../../declarations';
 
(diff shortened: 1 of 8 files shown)


tanner-reits (approved):

rwaskiewicz (commented):
Can you explain how you came up with this list? I started to do this a bit in connection with #4932, but got a list of >100 files that were using `'path'`. I wanna make sure we're thorough here (and I'm OK if we split PRs up to do it)

alicewriteswrongs:
I just opened up the typedef for the node.js `path` module and started looking through the references to `join` and did a few that looked straightforward and like they wouldn't cause problems - I didn't make any effort to address every call site in this PR or anything like that

rwaskiewicz:
@alicewriteswrongs Gotcha - let's circle back around to this next sprint and get a plan together for removing these. With the changes in this branch applied, I still find quite a few usages of `join` being imported from path:
Screenshot 2023-10-20 at 9 08 13 AM

PR #4936, opened 2023-10-16
refactor(jest): resolve v27 typings correctly

rwaskiewicz:

See next section

GitHub Issue Number: N/A

## What is the new behavior?

add `jest@27` as a dev dependency (as it is not a part of the production build) to the `jest-27-and-under` directory/package. previously, `@jest/types` imports in this directory would resolve to the root level `node_modules/`, which is not what we want when we start to upgrade jest internally for stencil

- [ ] Yes
- [x] No

## Testing

The build continues to pass (`tsc`), and our smoke tests for the component starter continue to pass.

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -95,7 +95,7 @@
     },
     {
       "matchFileNames": ["src/testing/jest/jest-27-and-under/package.json"],
-      matchPackageNames: ['@types/jest'],
+      matchPackageNames: ['@types/jest', 'jest'],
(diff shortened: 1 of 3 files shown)


tanner-reits (approved):

rwaskiewicz:
@tanner-reits @alicewriteswrongs I realized I needed to update `renovate.json5` here as well - can you PTAL?

alicewriteswrongs (approved):

PR #4938, opened 2023-10-17
fix(compiler): consistently generate additional type files

### LINKED ISSUE #4716: bug: The types folder will not be packaged
### Prerequisites

### Stencil Version

4.0.2

### Current Behavior

image

Because of the `title` field within the component, the packaging reports `warn`, which causes the `types` folder to not be packaged correctly! In case I have to use `title`, I would like to ask how to solve this problem? Or how to turn off `warn` completely?

### Expected Behavior

Why is it not packed? What went wrong?

### System Info

_No response_

### Steps to Reproduce

`pnpm build` is executed multiple times, occasionally packaged, occasionally not!

### Code Reproduction URL

https://github.com/ayangweb/stencil-test

### Additional Information

_No response_

HansClaasen:

The if statement doesn't really make sense to me as I would assume you would return early in case no types have changed. However this is causing an interesting side effect where the filesystem in memory determines to delete the directory because the directory and the file in it is not being written. A second execution actually re-creates the file.

It it has hard to determine the origin of this change. Likely this wasn't causing any harm until the memory FS logic started to clear out directories that had no files written to them.

This has potentially a small performance impact since types are now always generated which they would have been before too except the first time. However if we are not writing the file to memory, the directory will be cleaned up by the memfs system when we commit changes in [`writeBuild`](https://github.com/ionic-team/stencil/blob/b91f94a122dc9add876f0277f99afc92c96e66dd/src/compiler/build/write-build.ts#L24).

Fixes #4716

GitHub Issue Number: #4716

## What is the new behavior?

With this patch, we always generate additional type files for the component, including for the first type generation.

- [ ] Yes
- [x] No

## Testing

No tests added to verify this new behavior given there aren't any tests for part of the code base already. Given that we don't really know the expected behavior for a lot of these parts I am not sure about the value of adding unit tests for the current state. Happy to do so if you think it does make sense.

### DIFF
--- a/src/compiler/transpile/run-program.ts
+++ b/src/compiler/transpile/run-program.ts
@@ -109,10 +109,6 @@ export const runTsProgram = async (
 
   // create the components.d.ts file and write to disk
   const hasTypesChanged = await generateAppTypes(config, compilerCtx, buildCtx, 'src');
-  if (hasTypesChanged) {
-    return true;
-  }
-
   if (typesOutputTarget.length > 0) {
     // copy src dts files that do not get emitted by the compiler
     // but we still want to ship them in the dist directory
@@ -149,7 +145,7 @@ export const runTsProgram = async (
     buildCtx.diagnostics.push(...tsSemantic);
   }
 
-  return false;
+  return hasTypesChanged;
 };
 
 /**


tanner-reits (changes requested):
@HansClaasen I think we're good to make the change to avoid the early return, but we'll want to keep the result of `generateAppTypes` around and use it as the return value [here](https://github.com/ionic-team/stencil/blob/ccd3f6ac5cfd26a23603ea86e7682926551256c8/src/compiler/transpile/run-program.ts#L148) for the context where `runTsProgram()` gets executed

HansClaasen:
@tanner-reits thanks for your review. I made necessary changes, please let me know if they align with what you had in mind. Thank you!

rwaskiewicz (approved):

tanner-reits (approved):

PR #4940, opened 2023-10-17
fix(runtime): relocate slot content from non-shadow to shadow components w/ slot name change

tanner-reits:

It is possible for content to not be relocated correctly when it is relocated from a non-shadow to a shadow component if a slot is slotted into another slot and the names are different. A basic example may look something like:

```html
// index.html
<host-cmp>
  <b slot="main-content">main slotted content</b>
  <b slot="secondary-content">secondary slotted content</b>
</host-cmp>

```

```tsx
// host.tsx
<div>
  <parent-cmp>
    <slot name="main-content" slot="main-content" />
    <div slot="slotted-content-element">
      <child-cmp>
        <slot name="secondary-content" />
      </child-cmp>
    </div>
  </parent-cmp>
</div>

// parent.tsx
<div>
  <slot name="main-content" />
  <slot name="slotted-content-element" />
</div>

// child.tsx
<div>
  <p>content before</p>
  <slot />
  <p>content after</p>
</div>
```

In the above example, the "secondary slotted content" text won't render at all since we don't correctly remove the `slot` attribute for it to render in the default slot in it's final location.

Fixes: #4525 

## What is the new behavior?

When we move the node under the shadow component, we check the element's `slot` attribute against the slot ref node to make sure the two slot names match.

- [ ] Yes
- [X] No

## Testing

**Manual Testing**

Install a build of this branch into the reproduction case for the issue (#4525)

**Automated Testing**

New e2e test added for reproduction case. All other unit & e2e tests continue to pass

## Other information

Only available when the `experimentalSlotFixes` config extras flag is enabled


src/runtime/vdom/vdom-render.ts
```diff
@@ -1004,6 +1008,29 @@ render() {
...
+              // In this case, we need to update the relocated node's slot attribute to match
+              // the slot name it is being relocated into.
+              //
+              // There is a very niche use case where we may be relocating a text node. For now,
+              // we ignore anything that is not an element node since non-element nodes cannot have
+              // attributes to specify the slot. We'll deal with this if it becomes a problem... but super edge-case-y
+              if (
+                BUILD.experimentalSlotFixes &&
+                nodeToRelocate.nodeType === NODE_TYPE.ElementNode &&
+                slotRefNode.parentElement?.shadowRoot != null &&
```
tanner-reits, line 1024: Couple things here for context:
1. I decided to just ignore non-element nodes for now (i.e. comment nodes and text nodes). Per our discussion, the edge-case where these types of nodes is relevant is highly unlikely, so seems okay to hold on that until we have a need to fix it.
2. We check the `shadowRoot` here because we only need to deal with these slot `slot` (yes, the double "slot" there is intentional) values if we're moving the content into a shadow component since the native shadow DOM will handle things from there.

tanner-reits (commented):
Just noting one thing. Also, the added test was pulled directly from the associated issue

test/karma/test-app/non-shadow-to-shadow-slot-relocation/karma.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  it('should correctly render slotted content', () => {
+    expect(host).toBeDefined();
+
+    const mainDiv = host.firstElementChild.firstElementChild;
+    expect(mainDiv).toBeDefined();
+    expect(mainDiv.children.length).toBe(2);
+
+    // Main content
+    expect(mainDiv.children[0].tagName).toBe('B');
```
rwaskiewicz, line 23: Non-blocking: For some of these assertions, it might be nice to pull some of the repeated accessors (e.g. `mainDiv.children[1]`) into a variable for the sake of 'cleaning' the code up. But how far we take that is pretty subjective - up to you how much you wanna do (or not)

rwaskiewicz (approved):
Two non-blocking asks

PR #4946, opened 2023-10-19
fix(runtime): apply textnodes to shadow DOM instead of light DOM

### LINKED ISSUE #4231: bug:
### Prerequisites

### Stencil Version

2.21.0

### Current Behavior

Get build error when using top-level await. 

"Module format cjs does not support top-level await." 

### Expected Behavior

Solution should build successfully. 

### System Info

```shell
System: node 14.17.1

    Compiler: \@stencil\core\compiler\stencil.js
     Stencil: 2.21.0
  TypeScript: 4.9.4
      Rollup: 2.42.3
      Parse5: 7.1.2
      Sizzle: 2.42.3
      Terser: 5.16.1
```

### Steps to Reproduce

in ts file: 
`export class Hello{
  static async SayHello() {
    return Promise.resolve('Hello World');
  }
}

export const hello= await Profile.SayHello();
`

in another ts file import variable hello and use. Then build npm run build. You will receive the error. 

### Code Reproduction URL

see sample code above. 

### Additional Information

_No response_

HansClaasen:
By default the `vdomRender` build flag is `false`. The Stencil parser detects any usage of a `h` function, this flag will be switched. In the component provided by author of the bug there hasn't been any vDOM to be parsed, therefor there was no usage of the function `h`.

Now, in `callRender` if we end up not having to render any vDOM we used to just attach the string (can also be a boolean or number) as text to the host element. This however doesn't work when a shadow DOM is registered for the component. In this case the text content is added to the light dom which is not being rendered.

Screenshot 2023-10-18 at 4 52 28 PM

## What is the new behavior?

To fix this we check if the component has a shadow DOM if attach the text node to that node.

Screenshot 2023-10-18 at 4 53 09 PM

- [ ] Yes
- [x] No

## Testing

Added a unit test but not sure if shadow DOM cases can be tested this way.

## Other information

n/a


rwaskiewicz (commented):
Hey @HansClaasen 👋 

Can you do me a favor and double check the issue referenced in the PR summary? The one linked references using `async/await` at the lop level. Perhaps a typo (switched digits)?

HansClaasen:
Ah, I guess I discovered this bug then when investigating for the issue. I will remove the issue reference then. When I checked out the [example repo](https://github.com/runr02/toplevelawaitissue.git) I didn't run into any CJS issues.

rwaskiewicz (approved):
LGTM!

src/runtime/update-component.ts
```diff
@@ -266,7 +266,12 @@ const callRender = (hostRef: d.HostRef, instance: any, elm: HTMLElement, isIniti
           renderVdom(hostRef, instance, isInitialLoad);
         }
       } else {
-        elm.textContent = instance;
+        const shadowRoot = elm.shadowRoot;
+        if (BUILD.shadowDom) {
```
tanner-reits, line 270: Should this be checking `hostRef.$cmpMeta$.flags & CMP_FLAGS.shadowDomEncapsulation` instead? I don't think it's guaranteed that this component has a shadow root just because the build flag is set.

tanner-reits (changes requested):
One thing to double-check

src/runtime/update-component.ts
HansClaasen: Good catch @tanner-reits 👌 updated the if statement.

tanner-reits (approved):

PR #4951, opened 2023-10-19
chore(snc): resolve dist-lazy violations

rwaskiewicz:

We have ~1400 SNC failures

GitHub Issue Number: N/A

## What is the new behavior?

resolve a handful of dist-lazy strictNullChecks violations where `undefined` values were not properly being filtered out of the arrays being passed to `generate-lazy-module`.

in all code paths, this code would be failing if `undefined` were in these arrays in practice, as the values are used in `path.join` calls (which error in the event of `undefined` being provided as an arg)

- [ ] Yes
- [ ] No

## Testing

I'm relying on existing tests/the assumption that I'm correct in that all code paths would result in an `Error` being thrown if `undefined` actually made it in to these data structs


tanner-reits (approved):
Looks good. Could also use our `isString` helper in `@utils`

alicewriteswrongs (approved):
I like it 👍

PR #4952, opened 2023-10-19
chore(jest): add directions for adding new jest support

rwaskiewicz:

GitHub Issue Number: N/A

## What is the new behavior?

create a README file that instructs folks how to add support for a new version of jest

- [ ] Yes
- [x] No

## Testing

n/a - readme only (although these steps are being followed/created as we add support for newer versions of jest, so I suppose technically we're testing it)
## Other information

I wanted to push this up & get it out of my Jest 28+ branch. I'll mark it ready for review once Jest 28 is ready for review.

### DIFF
--- a/src/testing/jest/README.md
+++ b/src/testing/jest/README.md
@@ -1,0 +1,182 @@
+# Jest Support in Stencil
+
+This directory contains support for the Jest testing library.
+
+## Adding Support for a New Version of Jest
+
+The steps for adding support for a new version of Jest can be found below.
+These steps are intended to be followed in order, and be a part of a single pull request.
+**It is HIGHLY recommended that you create a commit for each step in the process.**
+This allows for 'well known' commits that map directly back to these steps to be created, speeding up the review process.
+
+1. In this directory, there are several version-specific directories to support Jest.
+   For example, `jest-27-and-under/` is for Jest v27 and below, `jest-28/` is for Jest v28, etc.
+   To start, create a copy of the entire directory contents for **the latest version** of Jest in this directory, ensuring the version number is in the directory name.
+   
+   If we wanted to add support for Jest v29 and had the following existing directory structure:
+   ```
+   └── src/testing/jest/
+       ├── jest-27-and-under
+       └── jest-28
+   ```
+   Adding support for Jest v29 would begin with making a copy of `jest-28/` (the latest supported version) and naming it `jest-29/`.
+   The end result of this step are two identical copies of the copied directory, with one bearing the name of the new version of Jest:
+   ```
+   └── src/testing/jest/
+       ├── jest-27-and-under
+       ├── jest-28
+       └── jest-29            <- Newly copied version of jest-28, just with a different name
+   ``` 
+
+1. **Commit your changes to help your reviewers out - this makes it easier to review changes that are coming down the line.**
+   Last reminder, please do this after every step.
+1. Delete the `package-lock.json` file that was copied into your new directory.
+   We'll regenerate this later after a few modifications have been made to `package.json` in the next step.
+1. Update the name and description of the package in the newly created directory's `pacakge.json` to reflect the version of Jest it supports.
+    ```diff
+    diff --git a/src/testing/jest/jest-29/package.json b/src/testing/jest/jest-29/package.json
+    index 92b5ee419..bd1614fe0 100644
+    --- a/src/testing/jest/jest-29/package.json
+    +++ b/src/testing/jest/jest-29/package.json
+    @@ -1,5 +1,5 @@
+    {
+    -  "name": "@stencil/jest-28",
+    +  "name": "@stencil/jest-29",
+       "version": "1.0.0",
+    -  "description": "Internal package for supporting Jest 28 with Stencil",
+    +  "description": "Internal package for supporting Jest 29 with Stencil",
+       "license": "MIT",
+    
+    ```
+1. Update any Jest related dependencies to use the targeted version of Jest.
+   It is up to you to determine/verify the packages needed here.
+   ```bash
+   $ cd src/testing/jest/jest-29
+   $ npm i -D jest@29 @types/jest@29
+   ```
+   Note: Running `npm install` will regenerate the `package-lock.json` file you deleted in a previous step.
+   Note: All dependencies can simply be dev-dependencies, since they're only used for type checking.
+1. Update `renovate.json` to not bump Jest beyond the version that this directory is responsible for:
+    ```diff
+    index 4861b2477..7cf75e16d 100644
+    --- a/renovate.json5
+    +++ b/renovate.json5
+    @@ -98,6 +98,11 @@
+           matchPackageNames: ['@types/jest', 'jest'],
+           allowedVersions: '<=28'
+         },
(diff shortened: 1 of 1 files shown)


src/testing/jest/README.md
```diff
@@ -0,0 +1,210 @@
...
+    strategy:
+    fail-fast: false
+    matrix:
+    -        jest: ['24', '25', '26', '27']
+    +        jest: ['24', '25', '26', '27', '28']
+             node: ['16', '18', '20']
+             os: ['ubuntu-latest', 'windows-latest']
+      runs-on: ${{ matrix.os }}
+   ```
+1. Rename the `JestFacade` implementation in the new directory's `jest-facade.ts` file.
```
tanner-reits, line 123: This step could be a little clearer that you need to rename the class in `src/testing/jest/jest-<version>/jest-facade.ts` and also its references in `src/testing/jest/jest-<version>/jest-config.ts`

src/testing/jest/README.md
```diff
@@ -0,0 +1,210 @@
...
+        * `JestFacade` implementation for communicating between this directory's version of Jest and Stencil
+        */
+    -  export class Jest27Stencil implements JestFacade {
+    +  export class Jest28Stencil implements JestFacade {
+        getJestCliRunner() {
+        return runJest;
+      }
+    
+    ```
+1. Update the `stencil-jest-connector.ts` to call the new `JestFacade` implementation:
```
tanner-reits, line 159: Unless this gets renamed in one of your other branches, I believe this should be `jest-stencil-connector.ts`

src/testing/jest/README.md
```diff
@@ -0,0 +1,210 @@
...
+   To start, create a copy of the entire directory contents for the latest version of Jest in this directory, ensuring the version number is in the directory name.
+   Note that `jest-27-and-under/` is a special case of supporting more than one version of Jest.
+   Going forward, there should be one directory for each version of Jest that is supported.
+   
+   If we wanted to add support for Jest v28 and had the following directory structure:
+   ```
+   └── src/testing/jest/
+       └── jest-27-and-under
+   ```
+   Adding support for Jest v28 would begin with making a copy of `jest-27-and-under/` (the latest supported version) and naming it `jest-28/`.
```
tanner-reits, line 23: Can we somehow make the "latest supported version" callout a bit more prominent? I think we all have enough intuition to use the latest version's adapter in practice, but better safe than sorry. Only calling this out so we don't end up cloning the `jest-27-and-under` directory as the starting point each time since that has some special cases in files like the preprocessor that have to make additional version checks.

tanner-reits (changes requested):
Couple small things

src/testing/jest/README.md
rwaskiewicz: I'm not sure I follow this - wouldn't [this line](https://github.com/ionic-team/stencil/pull/4952/files#diff-2c531f2d33dab017a4385abe8baffc2b3f78067802baf95f257f4f4b5deee42eR125) cover it? 

> It is recommended that you use your editor's renaming/refactoring utility to do this step:

Maybe it's different for each editor, but I think IntelliJ (what I use/my basis for this statement) does that. I don't think we have any usages of any Facade implementation in our `tests/` either, so I'm not sure how to best act on this one 🤔

src/testing/jest/README.md
rwaskiewicz: 👍 I tried to do this in https://github.com/ionic-team/stencil/pull/4952/commits/ac9c04aa148b8f70357cba91065077c9aff4dbda - I added Jest 28 into the example as if it already existed and the example goes through adding Jest 29. Hopefully that and some bolding of certain areas brings out the intent more

src/testing/jest/README.md
rwaskiewicz: D'Oh - https://github.com/ionic-team/stencil/pull/4952/commits/948af85113167a95b889c85eeb09bc2708e708cf

src/testing/jest/README.md
tanner-reits: Oh yeah, I guess I didn't realize that the editors would rename the symbol in other files that referenced it with an import

src/testing/jest/README.md
tanner-reits: Yep, think its pretty clear now. Changing the examples to be a migration from Jest 28 to 29 was a good move 👍

tanner-reits (approved):

src/testing/jest/README.md
```diff
@@ -0,0 +1,211 @@
...
+1. Update any Jest related dependencies to use the targeted version of Jest.
+   Note: It is up to you to determine/verify the packages needed here.
+   ```bash
+   $ cd src/testing/jest/jest-29
+   $ npm i -D jest@29 @types/jest@29
+   ```
+   Note that all dependencies can simply be dev-dependencies, since they're only used for type checking.
+1. Update the name of the package in the newly created directory to reflect the version of Jest it supports.
+   Do so by replacing the one instance of the package name in `package.json` and two instances in `package-lock-json`.
+   Following the example above, `@stencil/jest-28` would be replaced with `@stencil/jest-29`.
```
alicewriteswrongs, line 42: would it make sense to update the package name before installing the new dependencies? just thinking then the `package-lock.json` would be updated already. also do we maybe want to delete that (the `package-lock.json`) when starting a new adapter?

src/testing/jest/README.md
alicewriteswrongs: fwiw any editor that uses the typescript language server should be able to do this (I can do that in neovim with it)

src/testing/jest/README.md
```diff
@@ -0,0 +1,211 @@
...
+    +    jest: { minVersion: '24.9.0', recommendedVersion: '29', maxVersion: '29.0.0' },
+    +    'jest-cli': { minVersion: '24.9.0', recommendedVersion: '29', maxVersion: '29.0.0' },
+         puppeteer: { minVersion: '10.0.0', recommendedVersion: '20' },
+         'puppeteer-core': { minVersion: '10.0.0', recommendedVersion: '20' },
+         'workbox-build': { minVersion: '4.3.1', recommendedVersion: '4.3.1' },
+    ```
+1. It's time to get the project to compile.
+   Build Stencil and resolve any typing errors that have arisen from adding new versions of Jest.
+1. Once Stencil compiles, generate a version that can be installed in a Stencil component starter project (either a tarball or a dev build) and install it into the component starter.
+   Attempting to run the tests, working through any initial errors that you may run into.
```
alicewriteswrongs, line 204: I think maybe

```suggestion
   Attempt to run the tests, working through any initial errors that you may run into.
```

alicewriteswrongs (changes requested):
a few questions / comments

src/testing/jest/README.md
rwaskiewicz: Yeah, that's probably the better thing to do - I tested it out locally and everything worked as expected. Doc updated in https://github.com/ionic-team/stencil/pull/4952/commits/09129f4ebd816f0ec5dee7a416cd5d090231ee5f

alicewriteswrongs (approved):

PR #4966, opened 2023-10-23
fix(hydrate): prevent dead code elimination of patch dom implementation

sean-perkins:

In certain environments, such as Angular Universal (with Angular 16 or newer), the `baseURI` patch code will be eliminated from the final bundle. Javascript optimizers will treat the code as deadcode and eliminate it. 

This causes issues in Ionic Framework: https://github.com/ionic-team/ionic-framework/issues/28077

GitHub Issue Number: N/A

## What is the new behavior?

- `baseURI` patched implementation is assigned, so that it is not eliminated from the bundle.
- Ionic Framework (v7) works with Angular Universal with Angular 16 or newer

- [ ] Yes
- [x] No

## Testing

- Ryan provided a dev-build of this change for Stencil (`4.4.1-dev.1697216096.18bf460`)
- Ionic Framework was updated with the dev-build and a new dev-build was created/posted here: https://github.com/ionic-team/ionic-framework/issues/28077#issuecomment-1761950254
- Installed the dev-build in the user's test application and confirmed the application renders
   - Additionally verified that the `baseURI` code ends up in the final bundle

## Other information

This guidance was provided directly by the Angular team.

### DIFF
--- a/src/hydrate/runner/patch-dom-implementation.ts
+++ b/src/hydrate/runner/patch-dom-implementation.ts
@@ -38,7 +38,8 @@ export function patchDomImplementation(doc: any, opts: d.HydrateFactoryOptions) 
   }
 
   try {
-    doc.baseURI;
+    // Assigning the baseURI prevents JavaScript optimizers from treating this as dead code
+    win.__stencil_baseURI = doc.baseURI;
   } catch (e) {
     Object.defineProperty(doc, 'baseURI', {
       get() {


sean-perkins:
Let me know if we have any concerns to this approach. Happy to run through the Angular Universal testing I did locally. I also have additional comments to the issue in the linked framework issue.

src/hydrate/runner/patch-dom-implementation.ts
```diff
@@ -38,7 +38,8 @@ export function patchDomImplementation(doc: any, opts: d.HydrateFactoryOptions)
   }
 
   try {
-    doc.baseURI;
+    // Assigning the baseURI prevents JavaScript optimizers from treating this as dead code
+    win.__stencil_baseURI = doc.baseURI;
```
rwaskiewicz, line 42: Do we need to assign to `win` here to avoid dead code elimination? I.E. would something like:
```suggestion
    const __stencil_baseURI = doc.baseURI;
```
(with maybe an eslint disable rule pragma above it for unused variables) prevent us from elimination of this code without adding something to `global`?

rwaskiewicz (changes requested):
@sean-perkins Thanks!

Can you expound on 

> Installed the dev-build in the user's test application and confirmed the application renders
>    Additionally verified that the baseURI code ends up in the final bundle

and the steps you took to do that? Ideally, if we were to be going through `git blame` in the future and wanted to figure out how this was tested, it'd be ideal if there were steps that someone on the team could use to reproduce the error/fix

src/hydrate/runner/patch-dom-implementation.ts
sean-perkins: Test reproduction steps would follow the feedback here: https://github.com/ionic-team/ionic-framework/issues/28077#issuecomment-1738132757

Only with the modifications of:
1. Install the provided Ionic dev-build
2. Run against Angular 16 (the issue reporters originally reproduction app)
3. Validate that:
```ts
try {
  win.__stencil_baseURI = e2.baseURI;
} catch {
  Object.defineProperty(e2, "baseURI", {
    get() {
      const t3 = e2.querySelector("base[href]");
      return t3
        ? new URL(t3.getAttribute("href"), r.location.href).href
        : r.location.href;
    },
  });
}
```

is in the `patchDomImplementation` function in the compiled bundle. 

I do not know if a local assignment of an unused constant would get treated as a side-effect. The Angular team was not certain to which underlying package update between Angular 15 → Angular 16 had an impact on the analysis and removal of dead code. The recommendation from them was assigning to `win` after looking at Stencil's implementation, but if there is some performance impact to assigning to mock window we can discuss that.

rwaskiewicz: Okay, that all sounds reasonable to me - can you help me understand where I'd find that code snippet? So far, I've:
1. Downloaded the reproduction case linked above
2. Installed your dev build
3. Verified we're using Angular 16
4. Run `npm run build`, but I'm not sure where I should be looking for `patchDomImplementation`. I'm sure I'm missing something, but I'm not sure what exactly.

For context, I put together a branch of Stencil with the suggested change to try it out, created a dev build of Stencil and then used _that_ to create a dev build of Framework. I wanted to try this out myself, but first want to verify that that code snippet you provided with the existing dev build (so I could tell if/what my build changed)

SebastianKohler: Hope I'm not stepping on any toes pitching in ... (in that case, apologies).

Run the command `npm run build:ssr` to build the SSR app. Open the generated `dist/server/main.js`, optionally format it, and look for `function patchDomImplementation` in it.

I can find the code snippet, with one difference. I have `r.__stencil_baseURI=e2.baseURI` instead of `win.__stencil_baseURI=e2.baseURI`

sean-perkins: Apologies for the delay testing this, I created a new build with your dev-build for using a const: `7.5.6-dev.11700071366.1395b143` (Framework dev-build). 

This does not resolve the problem. Bundler still optimizes that bit of code for removal.

rwaskiewicz (approved):

Issue #4955, opened 2023-10-20
bug: onClickCapture not working?


kraftwer1:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

4.5.0

### Current Behavior

This is not working, "click" won't appear in the console:

    onClickCapture={() => console.log('click')}

### Expected Behavior

"click" appears in the console.

### Steps to Reproduce

In the stencil-component-starter repo, add this line:

    render() {
      return <div onClickCapture={() => console.log('click')}>Hello, World! I'm {this.getText()}</div>;
    }

### Code Reproduction URL

https://github.com/ionic-team/stencil-component-starter/

### Additional Information

I know I can add use `@Listen("click", { capture: true })` and do some filtering, but I'd prefer the simpler, declarative JSX variant. If there's a reason that `onClickCapture` does not work, maybe it should be removed from the `stencil-public-runtime.d.ts`.

Note that I've tested using the latest 4.5.0, but it also happens on lower versions.

tanner-reits:
Hey @kraftwer1 👋 

Thanks for reporting this issue! I was able to confirm there's a problem and should have a fix up soon. Someone from the team will post again here once it makes it into an upcoming release!

PR #4968, opened 2023-10-23
fix(runtime): support "capture" style events

tanner-reits:

Our element interfaces include capture type events (like `onClickCapture`, `onInputCapture`, etc.), but our runtime does not correctly bind the event listeners for these types. So, currently, nothing happens when using these attributes since they were bound like this:

```ts
el.addEventListener('clickCapture', <handler>, false)`
```

but should be bound like this:

```ts
el.addEventListener('click', <handler>, true)`
```

Fixes: #4955 

## What is the new behavior?

The runtime has been updated to correctly bind the listeners for capture events based on the "Capture" suffix in the event attribute

- [ ] Yes
- [X] No

## Testing

**Manual testing**
1. Add an `onClickCapture` to any element in a Stencil component starter that just logs out the event. Notice nothing gets logged clicking the element.
2. Build and install this branch into the starter
3. Event gets logged correctly when clicking the element

**Automated testing**
- All existing tests continue to pass.
- Added a unit test for `setAccessor` to ensure the call to bind the event listeners is getting the correct args
- Added a e2e test to test the actual event handlers for capture events

### DIFF
--- a/src/runtime/vdom/set-accessor.ts
+++ b/src/runtime/vdom/set-accessor.ts
@@ -113,12 +113,9 @@ export const setAccessor = (
         // Need to account for "capture" events.
         // If the event name ends with "Capture", we'll update the name to remove
         // the "Capture" suffix and make sure the event listener is setup to handle the capture event.
-        let capture = false;
-        if (memberName.endsWith(CAPTURE_EVENT_SUFFIX)) {
-          capture = true;
-          // Make sure we only replace the last instance of "Capture"
-          memberName = memberName.replace(new RegExp(CAPTURE_EVENT_SUFFIX + '$'), '');
-        }
+        const capture = memberName.endsWith(CAPTURE_EVENT_SUFFIX);
+        // Make sure we only replace the last instance of "Capture"
+        memberName = memberName.replace(CAPTURE_EVENT_REGEX, '');
 
         if (oldValue) {
           plt.rel(elm, memberName, oldValue, capture);
@@ -183,3 +180,4 @@ export const setAccessor = (
 const parseClassListRegex = /\s/;
 const parseClassList = (value: string | undefined | null): string[] => (!value ? [] : value.split(parseClassListRegex));
 const CAPTURE_EVENT_SUFFIX = 'Capture';
+const CAPTURE_EVENT_REGEX = new RegExp(CAPTURE_EVENT_SUFFIX + '$');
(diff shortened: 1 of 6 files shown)


src/runtime/vdom/set-accessor.ts
```diff
@@ -109,11 +109,23 @@ export const setAccessor = (
...
-      }
-      if (newValue) {
-        plt.ael(elm, memberName, newValue, false);
+      if (oldValue || newValue) {
+        // Need to account for "capture" events.
+        // If the event name ends with "Capture", we'll update the name to remove
+        // the "Capture" suffix and make sure the event listener is setup to handle the capture event.
+        let capture = false;
+        if (memberName.endsWith(CAPTURE_EVENT_SUFFIX)) {
+          capture = true;
```
rwaskiewicz, line 118: Minor, couldn't we avoid the double assignment here of `capture` by using the conditional directly since we're always calling it?
```suggestion
        const capture = memberName.endsWith(CAPTURE_EVENT_SUFFIX);
        if (capture) {
```

src/runtime/vdom/set-accessor.ts
```diff
@@ -170,3 +182,4 @@ export const setAccessor = (
 };
 const parseClassListRegex = /\s/;
 const parseClassList = (value: string | undefined | null): string[] => (!value ? [] : value.split(parseClassListRegex));
+const CAPTURE_EVENT_SUFFIX = 'Capture';
```
rwaskiewicz, line 185: Could we avoid the [re-instantiation of this regex above](https://github.com/ionic-team/stencil/pull/4968/files#diff-7f05c2f1645c35f29f698766956966cc2b5d915b37004316adfb635648ce9a62R120) on every call to by storing the regex as well?
```suggestion
const CAPTURE_EVENT_SUFFIX = 'Capture';
const CAPTURE_EVENT_REGEX = new RegExp(`${CAPTURE_EVENT_SUFFIX}$`);
```

rwaskiewicz (changes requested):
Two minor questions/asks

src/runtime/vdom/set-accessor.ts
tanner-reits: Made that change, also just removed the conditional that would follow so we always just make the call to `replace`, just won't do anything if there's nothing to replace (obviously).

rwaskiewicz (approved):

alicewriteswrongs (approved):
tested this out locally and it works great!

PR #4969, opened 2023-10-23
fix(compiler): correctly generate CSS rules using `::slotted` outside shadow DOM

tanner-reits:

Using a `::slotted` pseudo-selector outside a shadow DOM component can result in styles not working as expected. The issue is with how we transform the CSS to account for our pseudo-slot behavior (relocation).

GitHub Issue Number: #3977 

## What is the new behavior?

The CSS in transformed into the correct rule(s) to account for nested elements. Also updated the logic so we replace _all_ instances of the pre-transformed string in the CSS text, rather than just the first.

- [ ] Yes
- [X] No

## Testing

**Automated tests**
Updated the existing tests to account for the change in logic (which is what they should have been in the first place)

**Manual testing**
Install a build of this branch into the linked issue reproduction case and verify the button styles for the "scoped" button are applied as expected to match the other buttons.

### DIFF
--- a/src/utils/regular-expression.ts
+++ b/src/utils/regular-expression.ts
@@ -1,0 +1,9 @@
+/**
+ * Utility function that will escape all regular expression special characters in a string.
+ *
+ * @param text The string potentially containing special characters.
+ * @returns The string with all special characters escaped.
+ */
+export const escapeRegExpSpecialCharacters = (text: string) => {
+  return text.replace(/[-[\]{}()*+?.,\\^$|#\s]/g, '\\$&');
+};
--- a/src/utils/regular-expression.ts
+++ b/src/utils/regular-expression.ts
@@ -4,6 +4,6 @@
  * @param text The string potentially containing special characters.
  * @returns The string with all special characters escaped.
  */
-export const escapeRegExpSpecialCharacters = (text: string) => {
-  return text.replace(/[-[\]{}()*+?.,\\^$|#\s]/g, '\\$&');
+export const escapeRegExpSpecialCharacters = (text: string): string => {
+  return text.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
 };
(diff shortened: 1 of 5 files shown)


src/utils/regular-expression.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Utility function that will escape all regular expression special characters in a string.
+ *
+ * @param text The string potentially containing special characters.
+ * @returns The string with all special characters escaped.
+ */
+export const escapeRegExpSpecialCharacters = (text: string) => {
```
tanner-reits, line 7: Made this a util since I could see this coming up again. If something like this already exists, or we do something similar in other places, I can update accordingly

src/utils/shadow-css.ts
```diff
@@ -471,15 +473,32 @@ const scopeCssText = (
     cssText = scopeSelectors(cssText, scopeId, hostScopeId, slotScopeId, commentOriginalSelector);
   }
 
-  cssText = cssText.replace(/-shadowcsshost-no-combinator/g, `.${hostScopeId}`);
+  cssText = replaceShadowCssHost(cssText, hostScopeId);
```
tanner-reits, line 476: Tried to reorder this code so we didn't have to worry about doing this replacement in multiple places, but that caused other problems, so figured this was the solution that would cause the least headaches for now

tanner-reits (commented):
This doesn't resolve all the problems in the linked issue, just the styling stuff. Once I finish patching `textContent` and `innerText`, I'll have that PR be the one to close out the issue

rwaskiewicz (changes requested):

src/utils/regular-expression.ts
alicewriteswrongs: I don't know of anything but this looks like it will be handy in the future!

alicewriteswrongs (approved):
tried it out and all looks good!

rwaskiewicz (approved):

PR #4983, opened 2023-10-25
fix(www): fix an inconsistency between www builds

alicewriteswrongs:
This fixes a bug where subsequent `www` builds would not always produce working output. In particular, prior to this change a 'clean' `www` build (i.e one done after deleting all previous output) would produce a working build but subsequent builds would not.

This had to do with an optimization that Stencil attempts to do during the `www` build where it will attempt to inline the JS code for the main entry point into the `index.html` that it produces.

As part of doing this it needs to rewrite some relative paths for any module specifiers that are part of that entry point, since the HTML file is found at `www/index.html` while the JS entry point will be typically found at `www/build/project-name.js`. Thus if that entry point has an import like `import foo from './bar.js'` that will need to be rewritten to `import foo from '/build/bar.js'` in order for the import to resolve once its inlined into the HTML.

This inlining optimization was previously broken because the path rewrite code did not account for the presence of more than one instance of a given module specifier in the soon-to-be-inlined code. However, by default we have both `import` and `export` declarations in that code, so the inlined code would end up non-functional.

To give an example, the problematic excerpt of the inlined code would look typically like this:

```js
import {
  p as e,
  b as t
} from "/build/p-71aeae62.js";
export {
  s as setNonce
}
from "./p-71aeae62.js";
```

so the first `import` declaration is being correctly rewritten to begin with `/build/` but the `export` declaration is not. The code should look like this:

```js
import {
  p as e,
  b as t
} from "/build/p-71aeae62.js";
export {
  s as setNonce
}
from "/build/p-71aeae62.js";
```

This commit fixes the issue by transforming all instances of a given module specifier in the text, and additionally adds a regression test that tests for this specific case.

You can reproduce the issue by doing the following:

1. Create a new stencil project:
   ```sh
   cd /tmp
   npm init stencil@latest component test-www-double-build
   cd test-www-double-build
   npm i
   ```
2. Edit the `stencil.config.ts` to only do the `www` build (just for simplicity)
3. build the project:
   ```sh
   npx stencil build
   ```
4. Observe that `build/index.html`. It should look something like this:
   ```html
   <!doctype html>
   <html dir="ltr" lang="en">
      <head>
         <meta charset="utf-8">
         <meta name="viewport" content="width=device-width, initial-scale=1.0, minimum-scale=1.0, maximum-scale=5.0">
         <script type="module" src="/build/test-dev-build-reload.esm.js" data-stencil data-resources-url="/build/" data-stencil-namespace="test-dev-build-reload"></script> <script nomodule="" src="/build/test-dev-build-reload.js" data-stencil>. 
         </script> 
      </head>
      <body>
         <my-component first="Stencil" last="'Don't call me a framework' JS"></my-component>
      </body>
   </html>
   ```
5. Try this build output in a browser. The easiest way is to use a simple webserver to serve the `www` directory locally, for instance:
   ```sh
   npx http-server www
   ```
   This should work with no errors in the console and the `<my-component>` tag should work properly.
6. Rebuild the project _without deleting or modifying anything in `www/`_ and check out `build/index.html` again. It should now have inlined JS in the `<script>` tag, like so:
   ```html
   <!doctype html>
   <html dir="ltr" lang="en">
      <head>
         <meta charset="utf-8">
         <meta name="viewport" content="width=device-width, initial-scale=1.0, minimum-scale=1.0, maximum-scale=5.0">
         <script type="module" data-stencil data-resources-url="/build/" data-stencil-namespace="test-dev-build-reload">
         import{p as e,b as t}from"/build/p-71aeae62.js";export{s as setNonce}from"./p-71aeae62.js";const o=()=>{const s=import.meta.url;const t={};if(s!==""){t.resourcesUrl=new URL(".",s).href}return e(t)};o().then((e=>t([["p-a9942e82",[[1,"my-component",{first:[1],middle:[1],last:[1]}]]]],e)));
            //# sourceMappingURL=test-dev-build-reload.esm.js.map
         </script> <script nomodule="" src="/build/test-dev-build-reload.js" data-stencil></script> 
      </head>
      <body>
         <my-component first="Stencil" last="'Don't call me a framework' JS"></my-component>
      </body>
   </html>
   ```
7. Now try out this build output in the browser (as above). It should fail because the `./p-71aeae62.js` module specifier can't be resolved (it should be `/build/p-71aeae62.js` instead).

## What is the new behavior?

If you build and install this in the same project as above you should be able to rebuild without producing a non-working build.

- [ ] Yes
- [x] No

## Testing

See above. I also expanded the unit tests for this bit of code a bit.

### DIFF
--- a/src/compiler/html/inline-esm-import.ts
+++ b/src/compiler/html/inline-esm-import.ts
@@ -7,12 +7,30 @@ import { generateHashedCopy } from '../output-targets/copy/hashed-copy';
 import { getAbsoluteBuildDir } from './html-utils';
 import { injectModulePreloads } from './inject-module-preloads';
 
+/**
+ * Attempt to optimize an ESM import of the main entry point for a `www` build
+ * by inlining the imported script within the supplied HTML document, if
+ * possible.
+ *
+ * This will only do this for a `<script>` with type `"module"` where the
+ * `"src"` attr matches the main entry point for the build. If such a
+ * `<script>` is found the imported file will be resolved and edited in order
+ * to allow it to be properly inlined. If there's no such `<script>` _or_ if
+ * the file referenced by the `<script>` can't be resolved then no action
+ * will be taken.
+ *
+ * @param config the current user-supplied Stencil config
+ * @param compilerCtx a compiler context
+ * @param doc the document in which to search for scripts to inline
+ * @param outputTarget the output target for the www build we're optimizing
+ * @returns whether or not a script was found and inlined
+ */
 export const optimizeEsmImport = async (
   config: d.ValidatedConfig,
   compilerCtx: d.CompilerCtx,
   doc: Document,
   outputTarget: d.OutputTargetWww,
-) => {
+): Promise<boolean> => {
   const resourcesUrl = getAbsoluteBuildDir(outputTarget);
   const entryFilename = `${config.fsNamespace}.esm.js`;
   const expectedSrc = join(resourcesUrl, entryFilename);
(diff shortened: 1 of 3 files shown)


src/compiler/html/inline-esm-import.ts
```diff
@@ -56,50 +74,85 @@ export const optimizeEsmImport = async (
...
+ * @param newDir the directory which should be prepended to all module
+ * specifiers in the code
+ * @returns a manifest containing transformed code and a list of transformed
+ * module specifiers
+ */
 export const updateImportPaths = (code: string, newDir: string) => {
-  const orgImportPaths: string[] = [];
-  const tsSourceFile = ts.createSourceFile('module.ts', code, ts.ScriptTarget.Latest);
-  ts.transform(tsSourceFile, [readImportPaths(orgImportPaths)]);
+  const orgModulePaths = readModulePaths(code);
```
alicewriteswrongs, line 89: I made a change here because I don't think it's actually necessary to use `ts.transform` here, we were doing so purely to gather information about the source code, not to actually run a transformer and then emit code from it. So instead we can just create a `ts.SourceFile` and then iterate through the `statements` on it

src/compiler/html/inline-esm-import.ts
```diff
@@ -56,50 +74,85 @@ export const optimizeEsmImport = async (
...
+/**
+ * Small type coercion / guard helper that returns whether or not a
+ * {@link ts.Statement} is an import / export declaration
+ *
+ * @param stmt the statement of interest
+ * @returns whether this is an import or export declaration or neither
+ */
+function isImportOrExportDecl(stmt: ts.Statement): stmt is ts.ImportDeclaration | ts.ExportDeclaration {
+  return ts.isExportDeclaration(stmt) || ts.isImportDeclaration(stmt);
+}
```
alicewriteswrongs, line 155: this is I think necessary because I don't think TS is smart enough (yet!) to know that `ts.isExportDeclaration(foo) || ts.isImportDeclaration(foo)` evaluating to `true` means it must be one of those two types - instead when I inlined the usage of these two functions it was inferring a type of `ts.Statement` :/

src/compiler/html/test/update-esm-import-paths.spec.ts
```diff
@@ -15,33 +15,51 @@ describe('updateImportPaths', () => {
...
     const input = `import 'leave-me-be';`;
     const o = updateImportPaths(input, newAbsDir);
     expect(o.code).toBe(`import 'leave-me-be';`);
   });
+
+  it('should update both an import and an export statement', () => {
+    const input = `import './p-010101.js'; export { foo } from './p-010101.js'`;
+    const o = updateImportPaths(input, newAbsDir);
+    expect(o.code).toBe(`import '/build/p-010101.js'; export { foo } from '/build/p-010101.js'`);
+  });
```
alicewriteswrongs, line 64: This is a regression test for the bug, I confirmed that it fails on `main`

alicewriteswrongs (commented):
some notes

src/compiler/html/inline-esm-import.ts
```diff
@@ -56,50 +74,85 @@ export const optimizeEsmImport = async (
...
 
-  orgImportPaths.forEach((orgImportPath) => {
-    const newPath = replacePathDir(orgImportPath, newDir);
+  for (const orgImportPath of orgModulePaths) {
+    const newPath = updateImportPathDir(orgImportPath, newDir);
     if (newPath) {
-      code = code.replace(`"${orgImportPath}"`, `"${newPath}"`);
-      code = code.replace(`'${orgImportPath}'`, `'${newPath}'`);
+      code = code.replace(new RegExp(`"${orgImportPath}"`, 'g'), `"${newPath}"`);
+      code = code.replace(new RegExp(`'${orgImportPath}'`, 'g'), `'${newPath}'`);
```
tanner-reits, line 95: Do we need to keep the same quotation style that the file originally had? Otherwise we could instantiate a single regex to match the import with either quotation style

tanner-reits (approved):
One possible change, but not a blocker

src/compiler/html/inline-esm-import.ts
alicewriteswrongs: yeah I suppose we don't - I kept the same quotation-style-preserving logic that was already present, but I really just did so because that was the precedent and not for any real reason

src/compiler/html/inline-esm-import.ts
```diff
@@ -56,50 +74,85 @@ export const optimizeEsmImport = async (
...
-        .filter((s) => s.moduleSpecifier != null)
-        .filter((s) => ts.isStringLiteral(s.moduleSpecifier) && s.moduleSpecifier.text);
+/**
+ * Gather all module specifiers used in the `import` and `export` declarations
+ * in a bit of JS code
+ *
+ * @param code the code to transform
+ * @returns a list of the module specifiers present in the code
+ */
+function readModulePaths(code: string) {
```
rwaskiewicz, line 129: can we throw a return type on this function please?

src/compiler/html/inline-esm-import.ts
rwaskiewicz: Agreed - I can't think of/see a reason to use `transform` here 🤔

src/compiler/html/inline-esm-import.ts
rwaskiewicz: Yeah, I don't think we need it, but I'm not opposed to keeping it either. The only thing I can think if is that different quote styles can be used to allow _another_ quote style to be used in a string:
```ts
const foo = "We are 'the' champions";
```
which I don't _think_ would be allowed in module paths (otherwise that'd be _very_ confusing for users!)

rwaskiewicz (approved):
LGTM - I had one tiny ask that's not blocking

src/compiler/html/inline-esm-import.ts
alicewriteswrongs: I thought about doing the lean-more-into-TypeScript refactor to actually emit the code but I think that's a _lot_ more overhead actually

src/compiler/html/inline-esm-import.ts
alicewriteswrongs: ok maybe out of an abundance of caution I'll just leave it as-is for now

PR #4984, opened 2023-10-25
docs(output-targets): document generateHashedCopy function

alicewriteswrongs:
This just adds some JSDoc to this function, which it was sorely lacking.

GitHub Issue Number: N/A

## What is the new behavior?

- 

- [ ] Yes
- [x] No

### DIFF
--- a/src/compiler/output-targets/copy/hashed-copy.ts
+++ b/src/compiler/output-targets/copy/hashed-copy.ts
@@ -2,7 +2,42 @@ import { dirname, extname, join } from 'path';
 
 import type * as d from '../../../declarations';
 
-export const generateHashedCopy = async (config: d.ValidatedConfig, compilerCtx: d.CompilerCtx, path: string) => {
+/**
+ * Create a copy of a given file in its current directory but with a filename
+ * derived by hashing the file contents. This allows us to refer to a file by
+ * name and ensure that a given filename always refers to the same file
+ * contents.
+ *
+ * So for instance if you had a directory:
+ *
+ * ```
+ * build
+ * └── index.js
+ * ```
+ *
+ * and the contents of `build/index.js` hashed to `1234abcd` then write a
+ * file to `build/p-1234abcd.js` with the contents of `build/index.js`,
+ * giving:
+ *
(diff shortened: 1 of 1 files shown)


src/compiler/output-targets/copy/hashed-copy.ts
```diff
@@ -2,7 +2,38 @@ import { dirname, extname, join } from 'path';
...
+ * Create a copy of a given file in its current directory but with a filename
+ * derived by hashing the file contents. This allows us to refer to a file by
+ * name and ensure that a given filename always refers to the same file
+ * contents.
+ *
+ * So for instance if you had a directory:
+ *
+ * build
+ * └── index.js
+ *
```
rwaskiewicz, line 15:
```suggestion
 * ```
 * build
 * └── index.js
 * ```
```

src/compiler/output-targets/copy/hashed-copy.ts
```diff
@@ -2,7 +2,38 @@ import { dirname, extname, join } from 'path';
...
+ * and the contents of `build/index.js` hashed to `1234abcd` then write a
+ * file to `build/p-1234abcd.js` with the contents of `build/index.js`,
+ * giving:
+ *
+ * build
+ * ├── index.js
+ * └── p-1234abcd.js
+ *
```
rwaskiewicz, line 23:
```suggestion
 * ```
 * build
 * ├── index.js
 * └── p-1234abcd.js
 * ```
```

rwaskiewicz (approved):
One non-blocking ask - can we throw those directory examples in code blocks to make sure they get rendered with monospace font by IDEs?

src/compiler/output-targets/copy/hashed-copy.ts
alicewriteswrongs: oh yeah good idea!

alicewriteswrongs: just use the terminal, everything is monospaced :D

tanner-reits (approved):

PR #4992, opened 2023-10-26
fix(www): ensure that files necessary for www build are on disk

### LINKED ISSUE #4948: bug: Stencil first prod build is broken after dev build
**Update 2023-10-25:** The first Stencil prod build after a dev build always seem to differ from the second (or third...) build. Actually, has nothing to do with Ionic. In the first build, the whole `patchBrowser = () => ...` stuff is present (non-minified), whereas a second or third build creates a different kind of release file. But if you include Ionic, the whole app doesn't even load, because of stuff like this: `import './ionic-global-8c59e7e6.js';`.

Still, I don't think the first prod build should differ from subsequent builds at all. Quite confusing and error-prone.

### Prerequisites

### Stencil Version

4.5.0

### Current Behavior

```
"@ionic/core": "7.5.1",
"@stencil/core": "4.5.0"
```

A project that uses Ionic and Stencil creates a **broken build** when running `stencil build` for the first time after `stencil build --dev --watch --serve`. I always have to build twice to get the proper output.

The output of the first run contains a JS file that includes stuff that simply doesn't exist:

`index.html`:
```html
<script type="module" src="/build/p-630a4e0e.js" data-stencil data-resources-url="/build/"
        data-stencil-namespace="app"></script>
```

`/build/p-630a4e0e.js` (excerpt):
```js
import { B as BUILD, c as consoleDevInfo, H, d as doc, N as NAMESPACE, p as promiseResolve, b as bootstrapLazy } from './index-5be4bbac.js';
export { s as setNonce } from './index-5be4bbac.js';
import { g as globalScripts } from './app-globals-f8ab0fed.js';
import './ionic-global-8c59e7e6.js';
import './utils-23c30c62.js';
import './animation-4811d121.js';
import './index-150d1a25.js';
import './helpers-a0b2a380.js';
import './index-914e555f.js';
import './ios.transition-1739a486.js';
import './md.transition-538de680.js';
import './cubic-bezier-67bf1f69.js';
import './index-89e0f6ee.js';
import './config-a7320371.js';
import './theme-7dce6eff.js';
import './index-d0e75c8c.js';
import './index-bf616dc7.js';
import './hardware-back-button-90902d55.js';
import './overlays-1cb3af53.js';
import './framework-delegate-0227fbeb.js';

/*
 Stencil Client Patch Browser v4.5.0 | MIT Licensed | https://stenciljs.com
 */
const patchBrowser = () => {
    // NOTE!! This fn cannot use async/await!
    if (BUILD.isDev && !BUILD.isTesting) {
        consoleDevInfo('Running in development mode.');
    }
//...
```

Most of these imported files are not present in the actual build output. If `stencil build` is run a second time, a different JS file is created:

```html
<script type="module" src="/build/p-3040ee01.js" data-stencil data-resources-url="/build/"
        data-stencil-namespace="app"></script>
```

Which looks roughly like this `/build/p-3040ee01.js` (first few lines): 
```js
import{p as e,b as o}from"./p-ef0d46fc.js";export{s as setNonce}from"./p-ef0d46fc.js";import{g as n}from"./p-3bebd51e.js";import"./p-f38c8989.js";import"./p-fd141259.js";const t=()=>{const o=import.meta.url;const n={};if(o!==""){n.resourcesUrl=new URL(".",o).href}return e(n)};t().then((e=>{n();return o(JSON.parse('[["p-f5d63535",[[33,
//...
```

If you deploy the broken build, the app cannot start, because it fails to load non-existing .js files.

I disabled the cache in `stencil.config.ts`, but it doesn't make a difference. I provided a sample Github repo below, but it is essentially the Stencil starter with `ionic-pwa` and updated all dependencies to the most recent versions.

### Expected Behavior

Build the proper prod build the first time when I run `stencil build`.

### System Info

```shell
System: node 20.7.0
    Platform: darwin (22.5.0)
   CPU Model: Apple M1 Max (10 cpus)
    Compiler: /Users/cm/Code/ionic-bugtest/node_modules/@stencil/core/compiler/stencil.js
       Build: 1697469010
     Stencil: 4.5.0 📢
  TypeScript: 5.2.2
      Rollup: 2.42.3
      Parse5: 7.1.2
      Sizzle: 2.42.3
      Terser: 5.21.0
```

### Steps to Reproduce

1. Take any Stencil + Ionic project, run the default npm start command, which is `stencil build --dev --watch --serve`
2. Cancel the watcher, run the default npm run build, which is `stencil build` to create a prod build
3. Try to load the resulting app in your browser or simply inspect the files in `www/`

### Code Reproduction URL

https://github.com/cmaas/ionic-bugtest

### Additional Information

- This has been an issue for years and I remember that it was pointed out before, but can't find it anymore.
- At least one other person reported the same problem in the Discord.

alicewriteswrongs:
This fixes an issue where the `www` build and, in particular, some code that attempts to inline the entry point for the `www` build into the `index.html` file that the build outputs, was depending on a file which, in certain circumstances, would not yet be built. This produced inconsistency across builds, where the `www` build would pick up and inline the entry point from the _previous_ build (if the build wasn't deleted) instead of from the current build.

fixes #4948

See the linked issue. Basically, with the `www` output target at present if you do

```sh
stencil build --dev --watch --serve
```

in a Stencil project and _then_ you run a build (without deleting `www/`) like

```sh
stencil build
```

You'll get a non-working build. Not good!

Why does this happen? We have this function which is called as part of the `www` build:

https://github.com/ionic-team/stencil/blob/5c4ea3d5354596e5cfb544f91bde7f61de1888b4/src/compiler/html/inline-esm-import.ts#L10-L105

This will basically look for the entry point for the `www` build at `www/build/project-name.js` and, if it finds it, it will inline that code in the `index.html` file, reducing the number of JS files that have to be fetch on page-load by one.

However! Without this change we have a race condition where the lazy build (which writes the file we're looking for to disk) is not written until after the `www` build executes. This means that:

- on the first build the entry point is not found, so nothing is inlined in the HTML. this produces a working build, but the optimization (inlining) is missed.
- on the second build, if `www/` is not deleted then the entry point from the _first_ build is found and inlined _or_ if the file is larger than a cutoff size then a hashed copy of it will be made and the `<script>` tag in `index.html` will be changed to point to the hashed copy.
    - in either case the result is that `index.html` will include the entry point _for the previous build_ rather than for the _current_ build and this can produce broken behavior, in particular because this might import _other_ files that are no longer on the disk (and also because, you know, if you run a build you expect that the build emits code based on the current source code, not the source code the last time you ran a build)

Further, if both builds were 1) non-HMR / dev-server builds (i.e. the result of `stencil build`) then, provided there were no code changes between build one and build two, a working, inlined build will be the result. However! If _either_ the first build was an HMR build (i.e. from `stencil build --dev --watch --serve`) _or_ there was a code change so the hashed filenames will change then the file that is picked up will refer to things that don't exist. Not good!

## What is the new behavior?

Now the `www` build is run separately from the other output targets after they all finish, to ensure that the files from the lazy build are written before the `www` build tries to read them.

- [ ] Yes
- [x] No

## Testing

You can reproduce the issue by doing the following:

1. Create a new stencil project:
   ```sh
   cd /tmp
   npm init stencil@latest component test-www-previous-build
   cd test-www-previous-build
   npm i
   ```
2. Edit the `stencil.config.ts` to only do the `www` build (just for simplicity)
3. start the project in watch mode:
   ```sh
   npx stencil build --dev --watch --serve
   ```
   This should work as normal
4. Now do a normal build:
   ```sh
   npx stencil build
   ```
   this should exit without error
5. Check the build output in the browser:
   ```sh
   npx http-server www
   ```
   This should not work, giving you some console errors about not resolving certain resources and whatnot.

If you build, pack, and install this branch the issue should be fixed.

### DIFF
--- a/src/compiler/html/inline-esm-import.ts
+++ b/src/compiler/html/inline-esm-import.ts
@@ -59,7 +59,11 @@ export const optimizeEsmImport = async (
       if (results.orgImportPaths.length > 0) {
         // insert inline script
         script.removeAttribute('src');
-        script.innerHTML = results.code;
+        // the `'\n'` is added here to avoid possible issues with HTML parsers
+        // since the inlined JS will often end in a sourcemap-related `//`
+        // comment a newline ensures that in the rendered HTML string the
+        // closing script tag will be on its own line
+        script.innerHTML = results.code + '\n';
       }
     } else {
       const hashedFile = await generateHashedCopy(config, compilerCtx, entryPath);
--- a/src/compiler/output-targets/index.ts
+++ b/src/compiler/output-targets/index.ts
@@ -33,8 +33,13 @@ export const generateOutputTargets = async (
     outputHydrateScript(config, compilerCtx, buildCtx),
     outputLazyLoader(config, compilerCtx),
     outputLazy(config, compilerCtx, buildCtx),
-    outputWww(config, compilerCtx, buildCtx),
   ]);
+
+  // the www output target depends on the output of the lazy output target
+  // since it attempts to inline the lazy build entry point into `index.html`
+  // so we want to ensure that the lazy OT has already completed and written
+  // all of its files before the www OT runs.
+  await outputWww(config, compilerCtx, buildCtx);
 
   // must run after all the other outputs
   // since it validates files were created


rwaskiewicz (approved):
Interestingly, https://github.com/ionic-team/stencil/pull/2970/files makes the same change (sans some comment differences). I tried to get #2970 working (since we hadn't touched it in a while) and couldn't - nice!

We can probably close #2970 out if this works out :-)

alicewriteswrongs:
I think I broke a karma test, investigating that this morning

rwaskiewicz:
@alicewriteswrongs FWIW, I think that's the problem I ran into myself. I don't have much to offer beyond that 😭 but LMK if you wanna sync up - I was hacking on this when the team was in Madison earlier this year, and never root caused the issue

src/compiler/html/inline-esm-import.ts
```diff
@@ -59,7 +59,8 @@ export const optimizeEsmImport = async (
       if (results.orgImportPaths.length > 0) {
         // insert inline script
         script.removeAttribute('src');
-        script.innerHTML = results.code;
+        // the `'\n'` is necessary here to avoid CI failures on browserstack
+        script.innerHTML = results.code + '\n';
```
rwaskiewicz, line 63: in case we ever remove BS, can we state that this prevents a potential HTML parsing issue with inlined JS l source map comments?

rwaskiewicz (approved):

tanner-reits (approved):

PR #4993, opened 2023-10-26
fix(runtime): slot name forwarding & attribute reset

tanner-reits:

There is a possible issue with slot relocation when moving nodes into nested components where slot names change along the path of relocation. Currently, Stencil will not execute the logic to change slot attributes on relocated elements unless they are being relocated into a Shadow element. This was causing issues with elements receiving the `hidden` attribute when relocated because we weren't resolving the slot reference node based on the correct slot name.

GitHub Issue Number: N/A

## What is the new behavior?

We now perform the same slot name forwarding logic we do when moving from a non-shadow to shadow component. The `putBackInOriginalLocation` method was also updated to reset the `slot` attribute on re-relocated nodes back to their original value so relocation can work correctly on subsequent re-renders.

- [ ] Yes
- [X] No

## Testing

All current e2e and unit tests continue to pass. Added a e2e test for the specific slot-name forwarding case this was located in.

## Other information

This behavior is only available with the `experimentalSlotFixes` extras config flag active.

### DIFF
--- a/src/runtime/vdom/vdom-render.ts
+++ b/src/runtime/vdom/vdom-render.ts
@@ -187,6 +187,13 @@ const putBackInOriginalLocation = (parentElm: Node, recursive: boolean) => {
       // Reset so we can correctly move the node around again.
       childNode['s-sh'] = undefined;
 
+      // When putting an element node back in it's original location,
+      // we need to reset the `slot` attribute back to the value it originally had
+      // so we can correctly relocate it again in the future
+      if (childNode.nodeType === NODE_TYPE.ElementNode && !childNode['s-sr']) {
+        childNode.setAttribute('slot', childNode['s-sn'] ?? '');
+      }
+
       checkSlotRelocate = true;
     }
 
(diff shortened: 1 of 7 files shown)


src/runtime/vdom/vdom-render.ts
```diff
@@ -666,23 +673,29 @@ const updateFallbackSlotVisibility = (elm: d.RenderNode) => {
...
-            // any element or text node (with content)
-            // should hide the default fallback slot node
-            if (
-              siblingNode.nodeType === NODE_TYPE.ElementNode ||
-              (siblingNode.nodeType === NODE_TYPE.TextNode && siblingNode.textContent.trim() !== '')
-            ) {
-              childNode.hidden = true;
-              break;
+          // Don't check the node against itself
+          if (siblingNode !== childNode) {
```
tanner-reits, line 677: All I did here was wrap the existing logic in a check so we aren't hiding slot fallback content when comparing the exact same node

tanner-reits (commented):
Noting one thing

src/runtime/vdom/vdom-render.ts
```diff
@@ -187,6 +187,13 @@ const putBackInOriginalLocation = (parentElm: Node, recursive: boolean) => {
       // Reset so we can correctly move the node around again.
       childNode['s-sh'] = undefined;
 
+      // When putting an element node back in it's original location,
```
rwaskiewicz, line 190:
```suggestion
      // When putting an element node back in its original location,
```

src/runtime/vdom/vdom-render.ts
```diff
@@ -187,6 +187,13 @@ const putBackInOriginalLocation = (parentElm: Node, recursive: boolean) => {
...
+      // we need to reset the `slot` attribute back to the value it originally had
+      // so we can correctly relocate it again in the future
+      if (childNode.nodeType === NODE_TYPE.ElementNode && !childNode['s-sr']) {
```
rwaskiewicz, line 193: For my own edification, why do we check that `childNode['s-sr']` is falsy?

tanner-reits: Actually, good catch. We don't need to since a slot ref node can only be a text node or comment node, so they'd never pass the first check anyway

rwaskiewicz: Wooo!

rwaskiewicz (approved):

alicewriteswrongs (approved):

PR #4994, opened 2023-10-26
fix(runtime): slotted content order with sibling elements

tanner-reits:

It is possible for slot content that gets relocated to a nested component to appear in the wrong order with the nested component's other non-slotted elements. This was due to logic we had for keeping relocated content in the correct order _within_ a slot executing when it shouldn't and changing our relocation node. 

Fixes: #2997 

## What is the new behavior?

- The culprit code has been wrapped in a check to only execute in certain situations that make sense (i.e. when the slot has other relocated elements)
- Documented the code responsible for determining where to insert the relocated node for posterity
- Cleaned up some of the code to make things more readable

- [ ] Yes
- [X] No

## Testing

All existing unit and e2e tests continue to pass. Added an e2e test for the reproduction case from the linked issue

## Other information

This is only available with the `experimentalSlotFixes` extra config flag is active

### DIFF
--- a/src/runtime/vdom/vdom-render.ts
+++ b/src/runtime/vdom/vdom-render.ts
@@ -989,7 +989,7 @@ render() {
           //
           // If there is no node immediately following the slot reference node, then we will just
           // end up appending the node as the last child of the parent.
-          let insertBeforeNode = slotRefNode.nextSibling as d.RenderNode;
+          let insertBeforeNode = slotRefNode.nextSibling as d.RenderNode | null;
 
           // If the node we're currently planning on inserting the new node before is an element,
           // we need to do some additional checks to make sure we're inserting the node in the correct order.
@@ -1001,10 +1001,11 @@ render() {
             !BUILD.experimentalSlotFixes ||
             (insertBeforeNode && insertBeforeNode.nodeType === NODE_TYPE.ElementNode)
           ) {
-            let orgLocationNode = nodeToRelocate['s-ol'];
-            let refNode: d.RenderNode;
-            while ((orgLocationNode = orgLocationNode.previousSibling as d.RenderNode)) {
-              refNode = orgLocationNode['s-nr'];
+            let orgLocationNode = nodeToRelocate['s-ol']?.previousSibling as d.RenderNode | null;
+
+            while (orgLocationNode) {
+              let refNode = orgLocationNode['s-nr'] ?? null;
+
               if (refNode && refNode['s-sn'] === nodeToRelocate['s-sn'] && parentNodeRef === refNode.parentNode) {
                 refNode = refNode.nextSibling as any;
                 if (!refNode || !refNode['s-nr']) {
@@ -1012,6 +1013,8 @@ render() {
                   break;
                 }
               }
+
+              orgLocationNode = orgLocationNode.previousSibling as d.RenderNode | null;
             }
           }
 
--- a/src/runtime/vdom/vdom-render.ts
+++ b/src/runtime/vdom/vdom-render.ts
@@ -995,7 +995,12 @@ render() {
           // we need to do some additional checks to make sure we're inserting the node in the correct order.
           // The use case here would be that we have multiple nodes being relocated to the same slot. So, we want
           // to make sure they get inserted into their new how in the same order they were declared in their original location.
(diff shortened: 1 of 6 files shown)


tanner-reits:
Tried to break this down by commit to help with the noise of some cleanup and documentation stuff. Changes to actual logic are in https://github.com/ionic-team/stencil/pull/4994/commits/67c340a5c13fcfa30783b9a4539402f1e6fa3555 & https://github.com/ionic-team/stencil/pull/4994/commits/33de03069028c5c931be385a7580f659746aa2b9

src/runtime/vdom/vdom-render.ts
```diff
@@ -985,25 +976,41 @@ render() {
...
-                break;
+          const parentNodeRef = slotRefNode.parentNode;
+          // When determining where to insert content, the most simple case would be
+          // to relocate the node immediately following the slot reference node. We do this
+          // by getting a reference to the node immediately following the slot reference node
+          // since we will use `insertBefore` to manipulate the DOM.
+          //
+          // If there is no node immediately following the slot reference node, then we will just
+          // end up appending the node as the last child of the parent.
+          let insertBeforeNode = slotRefNode.nextSibling as d.RenderNode;
```
alicewriteswrongs, line 992: should this cast be to `d.RenderNode | null` ? I think if there's no sibling (which if I understand the comment correctly and the truthiness check below is possible) then I think `.nextSibling` would be `null`

src/runtime/vdom/vdom-render.ts
```diff
@@ -985,25 +976,41 @@ render() {
...
+          // The use case here would be that we have multiple nodes being relocated to the same slot. So, we want
+          // to make sure they get inserted into their new how in the same order they were declared in their original location.
+          //
+          // TODO(STENCIL-914): Remove `experimentalSlotFixes` check
+          if (
+            !BUILD.experimentalSlotFixes ||
+            (insertBeforeNode && insertBeforeNode.nodeType === NODE_TYPE.ElementNode)
+          ) {
+            let orgLocationNode = nodeToRelocate['s-ol'];
+            let refNode: d.RenderNode;
```
alicewriteswrongs, line 1005: why not move this inside of the block for the `while` loop? It doesn't look like its referenced outside of that scope

alicewriteswrongs (changes requested):
two questions!

src/runtime/vdom/vdom-render.ts
tanner-reits: Good catch!

src/runtime/vdom/vdom-render.ts
```diff
@@ -985,26 +976,45 @@ render() {
...
-                break;
+          const parentNodeRef = slotRefNode.parentNode;
+          // When determining where to insert content, the most simple case would be
+          // to relocate the node immediately following the slot reference node. We do this
+          // by getting a reference to the node immediately following the slot reference node
+          // since we will use `insertBefore` to manipulate the DOM.
+          //
+          // If there is no node immediately following the slot reference node, then we will just
+          // end up appending the node as the last child of the parent.
+          let insertBeforeNode = slotRefNode.nextSibling as d.RenderNode | null;
```
rwaskiewicz, line 992: Do we need this type assertion? I'm not sure what purpose it's serving here ATM

rwaskiewicz (approved):
one non-blocking question

src/runtime/vdom/vdom-render.ts
rwaskiewicz: I saw https://github.com/ionic-team/stencil/pull/4994#discussion_r1373518603 after I approved this, but I'm not sure I understand (but also, my brain is actively turning to goo)

tanner-reits: We need to cast it since we may reassign this value later and the types won't match. `nextSibling` is of type `ChildNode | null | undefined`, but in the while loop we reassign a node that could be `RenderNode | null | undefined`. I tried to update our `RenderNode` interface to type `nextSibling` and `previousSibling` as `RenderNode | null`, but that broke types in some other places, so didn't wanna fight with that right now.

alicewriteswrongs (approved):

PR #4996, opened 2023-10-26
wip / experiment: use codemod to convert join and relative usage from 'path' module

alicewriteswrongs:

GitHub Issue Number: N/A

## What is the new behavior?

- 

- [ ] Yes
- [ ] No


alicewriteswrongs:
since #4997 we no longer need this!

PR #4998, opened 2023-10-26
chore(utils): add path utils for 'resolve' and 'normalize'

alicewriteswrongs:
This adds two new wrapped / patched path utils to our `@utils/path` module.

As part of addressing our ongoing woes with `path` we want to make sure we have all the utils we'll need here.

## What is the new behavior?

- 

- [ ] Yes
- [x] No

### DIFF
--- a/src/utils/path.ts
+++ b/src/utils/path.ts
@@ -218,3 +218,31 @@ export function relative(from: string, to: string): string {
 export function join(...paths: string[]): string {
   return normalizePath(path.join(...paths), false);
 }
+
+/**
+ * A wrapped version of node.js' {@link path.resolve} which adds our custom
+ * normalization logic. This resolves a path to a given (relative or absolute)
+ * path.
+ *
+ * @throws the underlying node function will throw if any argument is not a
+ * string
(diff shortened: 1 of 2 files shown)


src/utils/path.ts
```diff
@@ -218,3 +218,31 @@ export function relative(from: string, to: string): string {
...
+ * normalization logic. This resolves a path to a given (relative or absolute)
+ * path.
+ *
+ * @throws the underlying node function will throw if any argument is not a
+ * string
+ * @param paths a path or path fragments to resolve
+ * @returns a resolved path!
+ */
+export function resolve(...paths: string[]): string {
+  return normalizePath(path.resolve(...paths));
```
rwaskiewicz, line 233: Do these calls need the `false` argument that the others (`relative`, `join`) have?

alicewriteswrongs: haha p sure I was adding that as you wrote this comment - they do, I just added it

tanner-reits (approved):

rwaskiewicz (approved):

PR #5009, opened 2023-10-31
refactor(sys): move logger impl to `sys/node/logger`

alicewriteswrongs:
Previously this lived in `src/compiler/sys/logger` but it was only referenced in `src/sys/node/node-logger.ts`. This moves it to live under `src/sys/node/logger`, consolidated the 'sys' related stuff in that directory.

I looked at this as part of building `sys/node` with esbuild, I just realized that it's a bit funny that this code lives under `src/compiler` for no real reason.

GitHub Issue Number: N/A

## What is the new behavior?

- 

- [ ] Yes
- [ ] No


src/sys/node/logger/index.ts
alicewriteswrongs, line 1: I called the directory just `logger` because 1) this is already in `sys/node` so calling is `node-logger` feels kind of redundant and 2) the impl for the node logger internals isn't part of our public API (just the `createNodeLogger` function itself) so I was thinking this sets us up to do a rename from `createNodeLogger` to just `createLogger` at some point in the future (that will have to be a breaking change though, and we could also rename all `sys/node` stuff to just `sys/` at that point)

alicewriteswrongs: also the contents of this file are almost identical to what was in `src/compiler/sys/node-logger.ts`, the only differences are re-pointing some imports

alicewriteswrongs: I also didn't really look at actually refactoring any of this code, but if we're only ever going to support a node.js runtime for the compiler there's probably some unnecessary abstraction here - maybe something to look at in the future

alicewriteswrongs (commented):
a few notes

src/sys/node/logger/index.ts
```diff
@@ -0,0 +1,100 @@
+import { Logger } from '@stencil/core/declarations';
```
rwaskiewicz, line 1: As far as I can tell, with the exception of imports, this file is a copy of `src/sys/node/node-logger.ts` (which we delete in this PR). Looking at this file's history (`git log -p -- src/sys/node/logger/index.ts`), it looks like this is a new file rather than a move:
```
diff --git a/src/sys/node/logger/index.ts b/src/sys/node/logger/index.ts
new file mode 100644
index 000000000..dce445409
--- /dev/null
+++ b/src/sys/node/logger/index.ts
```

Can you do me a favor a use `git mv` to move/rename the old file instead? That way, we can preserve git history between the two.

rwaskiewicz (changes requested):

src/sys/node/logger/index.ts
alicewriteswrongs: good idea - I made that update, so it's now properly noted as a move rather than a delete + creation

src/sys/node/logger/test/terminal-logger.spec.ts
```diff
@@ -1,8 +1,8 @@
 import { bgRed, blue, bold, cyan, dim, gray, green, magenta, red, yellow } from 'ansi-colors';
 
 import { LOG_LEVELS, LogLevel } from '../../../../declarations';
-import { createNodeLoggerSys } from '../../../../sys/node/node-logger';
 import { setupConsoleMocker } from '../../../../testing/testing-utils';
+import { createNodeLoggerSys } from '..';
```
tanner-reits, line 5: Probably a nitpick, can this point directly to the file rather than the directory (imports like this feel wrong)?

tanner-reits (approved):

rwaskiewicz (approved):

PR #5010, opened 2023-10-31
chore(esbuild): build sys-node bundle with esbuild

alicewriteswrongs:
This adds a new esbuild-based bundle for the sys-node bundle, which includes the `CompilerSystem` implementation for node.js (currently our only full implementation).

GitHub Issue Number: N/A

## What is the new behavior?

- 

- [ ] Yes
- [x] No

### DIFF
--- a/scripts/esbuild/cli.ts
+++ b/scripts/esbuild/cli.ts
@@ -50,9 +50,10 @@ export async function buildCli(opts: BuildOptions) {
   };
 
   // CommonJS build options
-  const cjsContext: ESBuildOptions = {
+  const cjsOptions: ESBuildOptions = {
     ...cliEsbuildOptions,
     outfile: join(outputDir, cjsFilename),
+    platform: 'node',
     format: 'cjs',
     banner: {
       js: getBanner(opts, `Stencil CLI (CommonJS)`, true),
@@ -78,5 +79,5 @@ export async function buildCli(opts: BuildOptions) {
     types: dtsFilename,
   });
 
-  return runBuilds([esmOptions, cjsContext], opts);
+  return runBuilds([esmOptions, cjsOptions], opts);
 }
--- a/scripts/esbuild/util.ts
+++ b/scripts/esbuild/util.ts
@@ -21,6 +21,7 @@ export function getEsbuildAliases(): Record<string, string> {
     '@stencil/core/dev-server': './dev-server/index.js',
     '@stencil/core/mock-doc': './mock-doc/index.cjs',
     '@stencil/core/internal/testing': './internal/testing/index.js',
+    '@sys-api-node': './sys/node/index.js',
 
     // mapping node.js module names to `sys/node` "cache"
     //
@@ -28,6 +29,8 @@ export function getEsbuildAliases(): Record<string, string> {
     // of the Stencil distributable but also have our separate bundles
     // reference the same file
     prompts: './sys/node/prompts.js',
+    glob: './sys/node/glob.js',
+    'graceful-fs': './sys/node/graceful-fs.js',
   };
 }
 
@@ -43,31 +46,17 @@ const externalNodeModules = [
   '@jest/core',
   '@jest/reporters',
   '@microsoft/typescript-etw',
-  'assert',
-  'buffer',
-  'child_process',
-  'console',
-  'constants',
   'expect',
   'fsevents',
-  'inspector',
   'jest',
   'jest-cli',
   'jest-config',
   'jest-message-id',
   'jest-pnp-resolver',
   'jest-runner',
-  'net',
   'puppeteer',
   'puppeteer-core',
-  'readline',
-  'stream',
-  'tty',
-  'url',
-  'util',
-  'vm',
   'yargs',
-  'zlib',
 ];
 
 /**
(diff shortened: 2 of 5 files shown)


scripts/esbuild/cli.ts
```diff
@@ -50,9 +50,10 @@ export async function buildCli(opts: BuildOptions) {
   };
 
   // CommonJS build options
-  const cjsContext: ESBuildOptions = {
+  const cjsOptions: ESBuildOptions = {
```
alicewriteswrongs, line 53: realized this needed to be renamed

scripts/esbuild/util.ts
```diff
@@ -43,31 +46,16 @@ const externalNodeModules = [
...
   'puppeteer',
   'puppeteer-core',
-  'readline',
-  'stream',
-  'tty',
-  'url',
-  'util',
-  'vm',
-  'yargs',
-  'zlib',
```
alicewriteswrongs, line 70: I neglected to read and heed my own comment on this variable! Basically if you run esbuild with `platform` set to `node` then it will automatically 1) externalize all node.js built-in modules and 2) it will automatically mark imports from those modules as side-effect free. This is important because it will allow dead-code-elimination to remove them. So! We actually don't want to have things like `crypto`, `buffer`, `util`, etc in here.

src/sys/node/worker.ts
```diff
@@ -1,10 +1,8 @@
-import '@stencil/core/compiler';
-
+import * as coreCompiler from '@stencil/core/compiler';
```
alicewriteswrongs, line 1: the only code change I needed to make - this appears to be fine in both the esbuild-based build and the normal build

alicewriteswrongs (commented):

tanner-reits (approved):
Whoops, forgot to hit the button on this earlier

src/sys/node/worker.ts
rwaskiewicz: Are there any ramifications from moving this to a compile-time reference from a runtime-based one? How can we test this out?

rwaskiewicz (changes requested):
I built this locally, and it looks like there's a new output file, `713.node-fetch.js` in the output here. Any idea what's causing that?

Screenshot 2023-11-06 at 2 20 11 PM

src/sys/node/worker.ts
alicewriteswrongs: tbh I'm not sure, I think the `prerender` task may use the node worker but I'm not certain -- I've basically confirmed that common functionality works as-expected (i.e. running a dev server, building, testing, etc)

scripts/esbuild/util.ts
```diff
@@ -43,31 +46,16 @@ const externalNodeModules = [
...
-  'net',
   'puppeteer',
   'puppeteer-core',
-  'readline',
-  'stream',
-  'tty',
-  'url',
-  'util',
-  'vm',
-  'yargs',
```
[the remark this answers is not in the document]
alicewriteswrongs: oh yeah good catch, we should keep that in the external list

alicewriteswrongs:
For that `713.node-fetch.js` file, I'm not sure _why_ it's being written but I'm seeing that file getting written after doing `git clean -fdX` and a clean `npm install` on `main`, following by an `npm build` -- are you seeing that too?

rwaskiewicz:
> For that 713.node-fetch.js file, I'm not sure why it's being written but I'm seeing that file getting written after doing git clean -fdX and a clean npm install on main, following by an npm build -- are you seeing that too?

Huh - I didn't have that previously, and now I do. 🤷 
Disregard then!

src/sys/node/worker.ts
rwaskiewicz: I'm not either 😆 Can you do me a favor and spend a little time digging into that? I think it'd be beneficial for us to build some of that knowledge while we're in here

alicewriteswrongs: After doing a bunch of reading and documenting for #5034 I think that a basic way to test this is to just confirm that work which is done on the worker is, you know, done!

One example is here:

https://github.com/ionic-team/stencil/blob/6e89bcc7aa88dcc0b073390d8729bc76c22af974/src/compiler/optimize/optimize-module.ts#L101-L115

this `prepareModule` function called on `sys.worker` will, assuming you don't have `maxConcurrentWorkers` set to 1 or something, dispatch to a worker on another thread. If you have `minifyJs: true` in your Stencil config then the actual minification happens in another thread because of this - however, there is error recovery where if the worker doesn't work it will (I think!) output unminified JS instead. So I think a good check if that if you run a build then your minification settings are actually respected and the output actually changes if toggle `minifyJs`

alicewriteswrongs: So more concretely, in a sample project you could

1. build and pack this branch and then install it
2. run a non-minified build (put `minifyJs: false` in `stencil.config.ts`)
3. check-in `dist/` with `git add -f dist/` to so you can see a diff later
4. run a minified build

and then just confirm that files that were unminified are now minified

alicewriteswrongs: I've run this locally and it all seems to be working correctly

rwaskiewicz (approved):
LGTM! Thanks for the testing steps!

PR #5012, opened 2023-11-01
chore(esbuild): build mock-doc with esbuild

alicewriteswrongs:
This adds support for building the mock-doc top-level-module using esbuild.

GitHub Issue Number: N/A

## What is the new behavior?

- 

- [ ] Yes
- [x] No

## Testing

CI passing with both the Rollup- and Esbuild-based builds is a good signal that things are alright here. I also build and packed the project locally and then installed it in a sample project, ensuring that I could then run a basic `stencil test --e2e --spec` without issues.

### DIFF
--- a/scripts/esbuild/mock-doc.ts
+++ b/scripts/esbuild/mock-doc.ts
@@ -1,0 +1,93 @@
+import type { BuildOptions as ESBuildOptions } from 'esbuild';
+import fs from 'fs-extra';
+import { join } from 'path';
+
+import { bundleParse5 } from '../bundles/plugins/parse5-plugin';
+import { writeSizzleBundle } from '../bundles/plugins/sizzle-plugin';
+import { getBanner } from '../utils/banner';
+import { bundleDts } from '../utils/bundle-dts';
+import { BuildOptions, createReplaceData } from '../utils/options';
+import { writePkgJson } from '../utils/write-pkg-json';
+import { getBaseEsbuildOptions, getEsbuildAliases, runBuilds } from './util';
+
+/**
+ * Use esbuild to bundle the `mock-doc` submodule
+ *
+ * @param opts build options
+ * @returns a promise for this bundle's build output
+ */
+export async function buildMockDoc(opts: BuildOptions) {
+  const inputDir = join(opts.buildDir, 'mock-doc');
+  const srcDir = join(opts.srcDir, 'mock-doc');
+  const outputDir = opts.output.mockDocDir;
+
+  // bundle d.ts
+  await bundleMockDocDts(opts, inputDir, outputDir);
+
+  writePkgJson(opts, outputDir, {
+    name: '@stencil/core/mock-doc',
+    description: 'Mock window, document and DOM outside of a browser environment.',
+    main: 'index.cjs',
+    module: 'index.js',
+    types: 'index.d.ts',
+    sideEffects: false,
+  });
+
+  // we need to call `createReplaceData` here not because we plan to use the
+  // replace data in this bundle but because the function has some side-effects
+  // that we need here. in particular, it sets the version of `parse5` on
+  // `opts` and the `bundleParse5` function has an implicit dependency on this
+  // value being already set.
+  createReplaceData(opts);
+
+  const mockDocAliases = getEsbuildAliases();
+
+  const sizzlePath = await writeSizzleBundle(opts);
+  mockDocAliases['sizzle'] = sizzlePath;
+
+  const [, parse5Path] = await bundleParse5(opts);
+  mockDocAliases['parse5'] = parse5Path;
+
+  const mockDocBuildOptions: ESBuildOptions = {
+    ...getBaseEsbuildOptions(),
+    entryPoints: [join(srcDir, 'index.ts')],
+    bundle: true,
+    alias: mockDocAliases,
+    logLevel: 'info',
+    target: 'node16',
+  };
+
+  const esmOptions: ESBuildOptions = {
+    ...mockDocBuildOptions,
+    format: 'esm',
+    outfile: join(outputDir, 'index.js'),
+    banner: { js: getBanner(opts, `Stencil Mock Doc`, true) },
+  };
+
+  const cjsOptions: ESBuildOptions = {
+    ...mockDocBuildOptions,
+    format: 'cjs',
+    outfile: join(outputDir, 'index.cjs'),
+    banner: { js: getBanner(opts, `Stencil Mock Doc (CommonJS)`, true) },
+  };
+
+  return runBuilds([esmOptions, cjsOptions], opts);
+}
+
+async function bundleMockDocDts(opts: BuildOptions, inputDir: string, outputDir: string) {
+  const bundled = await bundleDts(
+    opts,
+    join(inputDir, 'index.ts'),
+    {
+      // we want to suppress the `dts-bundle-generator` banner here because we do
+      // our own later on
+      noBanner: true,
+      // we also don't want the types which are inlined into our bundled file to
+      // be re-exported, which will change the 'surface' of the module
+      exportReferencedTypes: false,
+    },
+    false,
+  );
+
+  await fs.writeFile(join(outputDir, 'index.d.ts'), bundled);
+}
--- a/scripts/esbuild/build.ts
+++ b/scripts/esbuild/build.ts
@@ -1,6 +1,7 @@
 import { getOptions } from '../utils/options';
 import { buildCli } from './cli';
 import { buildCompiler } from './compiler';
+import { buildMockDoc } from './mock-doc';
 import { buildScreenshot } from './screenshot';
 import { buildSysNode } from './sys-node';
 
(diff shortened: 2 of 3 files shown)


scripts/esbuild/mock-doc.ts
```diff
@@ -0,0 +1,73 @@
...
+  const cjsOptions: ESBuildOptions = {
+    ...mockDocBuildOptions,
+    format: 'cjs',
+    outfile: join(outputDir, 'index.cjs'),
+  };
+
+  return runBuilds([esmOptions, cjsOptions], opts);
+}
+
+async function bundleMockDocDts(opts: BuildOptions, inputDir: string, outputDir: string) {
```
tanner-reits, line 57: This function seems quite different from the Rollup equivalent. How different are the written `index.d.ts` files? Also, was `index.d.ts-bundled.d.ts` file useless (IIUC, that isn't getting created anymore)?

tanner-reits (commented):
One thing I'd personally just like a little clarification on

scripts/esbuild/mock-doc.ts
alicewriteswrongs: I did a bunch of manual checking and I believe that the files are the same, i.e. they end up exporting the same types. The function in the Rollup bundle ([here](https://github.com/ionic-team/stencil/blob/5e3605779589105d6a3da73fcfc2bbe5ceeb5def/scripts/bundles/mock-doc.ts#L79)) is basically a super manual way to bundle up and 'statically link' (sort of) a `.d.ts` file to make it standalone and portable, so it actually reads the files from disk, concatenates things, etc.

Here in the esbuild version I wanted to see if we could achieve the same result using our `bundleDts` function which leverages [dts-bundle-generator](https://github.com/timocov/dts-bundle-generator). Instead of being a somewhat 'homemade' solution this is a pretty reliable and tested library.

The reason for not writing a `-bundled.d.ts` version here (by passing `false` for the `useCache` parameter) is because I think for this code we always just want to generate a fresh new `index.d.ts` since it's not very expensive to do so and we can just avoid a whole potentially class of bugs that way

tanner-reits: Okay, that all makes sense. In that case, I think this is g2g!

tanner-reits (approved):

scripts/esbuild/mock-doc.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  const esmOptions: ESBuildOptions = {
+    ...mockDocBuildOptions,
+    format: 'esm',
+    outfile: join(outputDir, 'index.js'),
+  };
+
+  const cjsOptions: ESBuildOptions = {
+    ...mockDocBuildOptions,
+    format: 'cjs',
```
rwaskiewicz, line 50: Do we no longer need the [`CJS_INTRO` and `CJS_OUTRO` portions](https://github.com/ionic-team/stencil/blob/9d2d0a0612caafb88c559907c6dec1a39e760615/scripts/bundles/mock-doc.ts#L41-L42) of the build?

scripts/esbuild/mock-doc.ts
```diff
@@ -0,0 +1,73 @@
...
+    outfile: join(outputDir, 'index.cjs'),
```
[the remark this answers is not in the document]
rwaskiewicz: Follow up - it looks like [`banner` is also the acceptable way to add `strict`](https://github.com/evanw/esbuild/issues/422#issuecomment-739740602) - should we do that as well? [Here's where we use this option today](https://github.com/ionic-team/stencil/blob/9d2d0a0612caafb88c559907c6dec1a39e760615/scripts/bundles/mock-doc.ts#L43)

scripts/esbuild/mock-doc.ts
```diff
@@ -0,0 +1,73 @@
...
+    module: 'index.js',
+    types: 'index.d.ts',
+    sideEffects: false,
+  });
+
+  const mockDocAliases = getEsbuildAliases();
+
+  const sizzlePath = await writeSizzleBundle(opts);
+  mockDocAliases['sizzle'] = sizzlePath;
+
```
rwaskiewicz, line 32: Should the `parse5` bundle be aliased here as well? It's in [our current configuration](https://github.com/ionic-team/stencil/blob/9d2d0a0612caafb88c559907c6dec1a39e760615/scripts/bundles/mock-doc.ts#L52), and I'm not sure if/why we should omit it here

rwaskiewicz (changes requested):

scripts/esbuild/mock-doc.ts
```diff
@@ -0,0 +1,73 @@
...
+    bundle: true,
+    alias: mockDocAliases,
+    logLevel: 'info',
+    target: 'node16',
+  };
+
+  const esmOptions: ESBuildOptions = {
+    ...mockDocBuildOptions,
+    format: 'esm',
+    outfile: join(outputDir, 'index.js'),
```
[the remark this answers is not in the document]
alicewriteswrongs: oh yup we should!

scripts/esbuild/mock-doc.ts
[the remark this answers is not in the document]
alicewriteswrongs: yeah I think we'll need to put that in if we want it, I'll throw it in the banner (it looks like that was more or less what we were doing before anyway)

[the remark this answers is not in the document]
alicewriteswrongs: actually I don't think this is necessary! it looks like because this `cjs` output is being built from ES Modules (our `.ts` files) it can automatically determine that it's safe to put the whole output file in strict mode. So if I `cat mock-doc/index.cjs | head -n 4` after building I get the following:

```js
/*!
 Stencil Mock Doc (CommonJS) v4.7.2-dev.1700164570.7d29ef4 | MIT Licensed | https://stenciljs.com
 */
"use strict";
```

scripts/esbuild/mock-doc.ts
alicewriteswrongs: I don't believe its necessary

scripts/esbuild/mock-doc.ts
alicewriteswrongs: ah yeah that's just an oversight, we do want to include it

alicewriteswrongs: added that in

alicewriteswrongs:
One thing I just realized, I need the code I have in #5027 for the `replacePlugin` over here as well, so I'll hold off on merging this until after #5027 lands and I can rebase over it and pull that in

scripts/esbuild/mock-doc.ts
rwaskiewicz: Can you expound on why?

rwaskiewicz (changes requested):
The build here is currently broken:
```
 /home/runner/work/stencil/stencil/build/sys/node/worker.js → sys/node/worker.js...
created sys/node/worker.js in 15ms
/home/runner/work/stencil/stencil/scripts/build/bundles/plugins/parse5-plugin.js:54
    const fileName = `parse5-${opts.parse5Version.replace(/\./g, '_')}-bundle-cache${opts.isProd ? '.min' : ''}.js`;
                                                  ^

TypeError: Cannot read properties of undefined (reading 'replace')
    at bundleParse5 (/home/runner/work/stencil/stencil/scripts/build/bundles/plugins/parse5-plugin.js:54:51)
    at buildMockDoc (/home/runner/work/stencil/stencil/scripts/build/esbuild/mock-doc.js:32:67)
    at async Promise.all (index 2)
    at async main (/home/runner/work/stencil/stencil/scripts/build/esbuild/build.js:14:5)
```

alicewriteswrongs:
Fixed the build issue - the `bundleParse5` function has an implicit dependency on a side-effect of the `createReplaceData` function so I added a call to that function to the mock-doc bundle

scripts/esbuild/mock-doc.ts
alicewriteswrongs: as far as I can tell the `CJS_INTRO` and `CJS_OUTRO` snippets are here to allow the cjs bundle of mock doc to be used outside of a node.js / commonjs context. In the Rollup build the whole bundle is wrapped with

```js
var mockDoc = (function(exports) {
'use strict';
```

and then

```js
if (typeof module !== "undefined" && module.exports) {
  module.exports = exports;
}
return exports;
})({});
```

so basically the whole thing gets wrapped in an IIFE which will override the `exports` identifier within the scope of the IIFE and then if `typeof module !== "undefined"` (i.e. if we're in a node environment) only then it will set `module.exports` to be `exports`. I don't think this is necessary anymore since we don't support in-browser compilation. Also afaict Esbuild doesn't output cjs code using the [`exports` shortcut](https://nodejs.org/api/modules.html#exports-shortcut) but instead always references `module.exports`, so I think that including this intro and outro as-is would actually just cause `module.exports` to be set to `{}`.

Instead Esbuild, in one place, sets `module.exports` like this:

```js
module.exports = __toCommonJS(mock_doc_exports);
```

the `__toCommonJs` is a helper emitted by Esbuild that produces a commonJS module from an object with getters on it - you can check it out locally to see how it works (kinda big to paste here haha)

but anyway, all that is to say that I don't think the IIFE wrapping does anything we need anymore. I've also been able to build mock-doc using Esbuild and then run tests in a sample project without issues.

rwaskiewicz (approved):

Issue #4999, opened 2023-10-27
bug: component.d.ts generation gives TS errors


josip-kovacevic-valcon:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

4.6.0

### Current Behavior

We have simple `setup` pipeline which generates everything project needs. One of the files is `components.d.ts` file which is auto-generated by stencil. We git-ignored this file and also we made slim config file only for this file to be generated and on v4.5.0 and below it works nicely, but on v4.6.0 it gives error (look at picture).

Script used for dts generation: `"build.dts": "stencil build --config stencil-dts.config.ts"`
`stencil-dts.config.ts`
```ts
import { Config } from '@stencil/core';

export const config: Config = {
  namespace: 'vnp-web-components',
  sourceMap: false,
  transformAliasedImportPaths: true,
  outputTargets: [
    {
      type: 'dist-custom-elements',
    },
  ],
};

```

![image](https://github.com/ionic-team/stencil/assets/80685558/9c296b26-c175-42a6-bcd1-e1f9a5cb9e42)

### Expected Behavior

It shouldn't give errors. If there is simpler way only to generate `component.d.ts`, please provide it. I couldn't find anything in documentation nor I found similar issue.

### System Info

```shell
System: node 20.9.0
Platform: linux (6.2.0-35-generic)
CPU Model: 12th Gen Intel(R) Core(TM) i7-1265U (12 cpus)
Compiler: /home/josip-kovacevic/public-www/vnp-web-components/node_modules/@stencil/core/compiler/stencil.js
Build: 1698073283
Stencil: 4.6.0 💥
TypeScript: 5.2.2
Rollup: 2.42.3
Parse5: 7.1.2
Sizzle: 2.42.3
Terser: 5.22.0
```

### Steps to Reproduce

Make clean project setup using stencil cli. Try to use `build` or `start` script. It will give you TS errors since initially `components.d.ts` file does not exist.

### Code Reproduction URL

-

verena-ifx:
same error for me, when trying to build via github-actions. Works fine up to v4.50., but throws the same error when trying with a higher version

[...]

PR #5013, opened 2023-11-01
fix(compiler): ignore TS diagnostics on builds where typedef file changes

tanner-reits:

Building a Stencil project that uses the component starter template can result in build errors if the `src/components.d.ts` file doesn't already exist. This is because the TS program used for the build generates the semantic error diagnostics before we generate the typedef file. This bug was introduced by #4938 since we no longer early-abort before the TS diagnostics are pushed into our own diagnostics. 

Fixes: #4999 

## What is the new behavior?

This is a temporary fix to prevent the build errors without needing to early abort and revert all the changes from #4938. If the typedef file was changed, we ignore the TS diagnostics.

- [ ] Yes
- [X] No

## Testing

Tested in a component starter by deleting the `src/components.d.ts` file before running a build.

### DIFF
--- a/src/compiler/transpile/run-program.ts
+++ b/src/compiler/transpile/run-program.ts
@@ -132,7 +132,10 @@ export const runTsProgram = async (
     await Promise.all(srcRootDtsFiles);
   }
 
-  if (config.validateTypes) {
+  // TODO(STENCIL-540): remove `hasTypesChanged` check and figure out how to generate types before
+  // executing the TS build program so we don't get semantic diagnostic errors about referencing the
+  // auto-generated `components.d.ts` file.
+  if (config.validateTypes && !hasTypesChanged) {
     const tsSemantic = loadTypeScriptDiagnostics(tsBuilder.getSemanticDiagnostics());
     if (config.devMode) {
       tsSemantic.forEach((semanticDiagnostic) => {


rwaskiewicz (approved):

src/compiler/transpile/run-program.ts
```diff
@@ -132,7 +132,10 @@ export const runTsProgram = async (
     await Promise.all(srcRootDtsFiles);
   }
 
-  if (config.validateTypes) {
+  // TODO(STENCIL-540): remove `hasTypesChanged` check and figure out how to generate types before
+  // executing the TS build program so we don't get semantic diagnostic errors about referencing the
+  // auto-generated `components.d.ts` file.
+  if (config.validateTypes && !hasTypesChanged) {
```
alicewriteswrongs, line 138: It looks like there's an `emitOnlyDtsFiles` option for the `.emit` callback that we call here:

https://github.com/ionic-team/stencil/blob/4fd0ecd17e72f6892c96b8256a0206f6e583be13/src/compiler/transpile/run-program.ts#L91-L92

possibly I'm misunderstanding the comment here - but if we ran a dts-only emit _first_ and then the whole program would that prevent the error? Probably that would be slower so not necessarily saying we want to, more just wondering if that's indeed how it works here

tanner-reits: I think there'd still be an issue since we don't generate the `components.d.ts` file until after the TS program has ran/emitted the files. What we really need to do is generate the `components.d.ts` file before we run the TS program so that the file exists before the build. We can do this, just need to restructure things a bit more since we don't have the components metadata on the `BuildCtx` until the build callback, at which point the TS program has already ran and generated the diagnostics. This was just intended as a quick fix to remedy the regression until we can take the step for the desired, long term solution

alicewriteswrongs: ahh I think I sort of forgot that the `components.d.ts` is so artisanal, makes sense!

alicewriteswrongs (approved):

PR #5026, opened 2023-11-06
chore(snc): resolve SNC for stencil-types.ts

rwaskiewicz:

We have a handful of SNC's in `stencil-types.ts`

GitHub Issue Number: N/A

## What is the new behavior?

resolves two strictNullChecks violations that were caused
by improperly handling a variable that was of type `string|undefined`
and assuming that it was of type `string`.

- [ ] Yes
- [x] No

## Testing

I ran the unit tests with coverage, and verified the [new/existing branches](https://github.com/ionic-team/stencil/pull/5026/files#diff-9520ec3dc9b12aac1a3735d5bc8a2808c1cf32a7c560ab81874b9e6aa10b9f37R56-R62) were covered. I ran the tests in the debugger with breakpoints in each conditional body, to verify they were actually called when I expected them to be.

I also repeated the same steps above without the code changes applied (but with the new test) to verify that the behavior for `undefined` being returned by our little helper method functionally did "the same thing".

### DIFF
--- a/src/compiler/types/stencil-types.ts
+++ b/src/compiler/types/stencil-types.ts
@@ -70,9 +70,9 @@ export const updateTypeIdentifierNames = (
  * @param sourceFilePath the component source file path to resolve against
  * @returns the path of the type import
  */
-const getTypeImportPath = (importResolvedFile: string | undefined, sourceFilePath: string): string => {
-  const isPathRelative = importResolvedFile && importResolvedFile.startsWith('.');
-  if (isPathRelative) {
+const getTypeImportPath = (importResolvedFile: string | undefined, sourceFilePath: string): string | undefined => {
+  if (importResolvedFile && importResolvedFile.startsWith('.')) {
+    // the path to the type reference is relative, resolve it relative to the provided source path
     importResolvedFile = resolve(dirname(sourceFilePath), importResolvedFile);
   }
 
--- a/src/compiler/types/stencil-types.ts
+++ b/src/compiler/types/stencil-types.ts
@@ -53,6 +53,10 @@ export const updateTypeIdentifierNames = (
   for (const typeReference of Object.values(typeReferences)) {
     const importResolvedFile = getTypeImportPath(typeReference.path, sourceFilePath);
 
+    if (typeof importResolvedFile !== 'string') {
+      continue;
+    }
+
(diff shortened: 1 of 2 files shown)


src/compiler/types/tests/stencil-types.spec.ts
```diff
@@ -71,6 +71,24 @@ describe('stencil-types', () => {
         expect(actualTypeName).toBe(expectedTypeName);
       });
 
+      it('returns the provided type for imports without the resolved file', () => {
```
rwaskiewicz, line 74: this test explicitly tests [this code](https://github.com/ionic-team/stencil/pull/5026/files#diff-9520ec3dc9b12aac1a3735d5bc8a2808c1cf32a7c560ab81874b9e6aa10b9f37R60-R62), which was previously existing (I know, right?). 

It's [this existing test](https://github.com/ionic-team/stencil/pull/5026/files#diff-f3b145b0c322c69d725a877f98b882fbe836e81787ee2043fa42d182841f82b2L57-L72) that covers the new conditional, [found here](https://github.com/ionic-team/stencil/pull/5026/files#diff-9520ec3dc9b12aac1a3735d5bc8a2808c1cf32a7c560ab81874b9e6aa10b9f37R56-R59)

src/compiler/types/stencil-types.ts
```diff
@@ -70,9 +74,9 @@ export const updateTypeIdentifierNames = (
  * @param sourceFilePath the component source file path to resolve against
  * @returns the path of the type import
  */
-const getTypeImportPath = (importResolvedFile: string | undefined, sourceFilePath: string): string => {
-  const isPathRelative = importResolvedFile && importResolvedFile.startsWith('.');
-  if (isPathRelative) {
+const getTypeImportPath = (importResolvedFile: string | undefined, sourceFilePath: string): string | undefined => {
```
rwaskiewicz, line 77: This fn could always return `undefined`, especially if `importResolvedFile` was falsy - this makes that explicit (and something we end up dealing with right after its [called here](https://github.com/ionic-team/stencil/pull/5026/files#diff-9520ec3dc9b12aac1a3735d5bc8a2808c1cf32a7c560ab81874b9e6aa10b9f37R56-R59))

alicewriteswrongs (approved):

PR #5028, opened 2023-11-07
fix(runtime): add display style to slot-fb elements

tanner-reits:

Stencil wraps slot fallback content in a `slot-fb` element when we patch slot behavior in non-shadow components. This can cause issues with styles like `display: flex` that target slotted elements. 

Fixes: #2937 

## What is the new behavior?

This adds a global style to set `display: contents` on all `slot-fb` elements. This is the same style that gets set on the `slot` elements in the native shadow DOM.

- [ ] Yes
- [X] No

## Testing

**Manual testing**

Confirmed the fix in the reproduction in the linked issue

**Automated testing**

- Added e2e tests to verify that the new styles are correctly applied
- Restored the existing slot fallback tests that were disabled

### DIFF
--- a/src/runtime/bootstrap-lazy.ts
+++ b/src/runtime/bootstrap-lazy.ts
@@ -15,7 +15,7 @@ import {
 import { hmrStart } from './hmr-component';
 import { createTime, installDevTools } from './profile';
 import { proxyComponent } from './proxy-component';
-import { HYDRATED_CSS, HYDRATED_STYLE_ID, PLATFORM_FLAGS, PROXY_FLAGS } from './runtime-constants';
+import { HYDRATED_CSS, HYDRATED_STYLE_ID, PLATFORM_FLAGS, PROXY_FLAGS, SLOT_FB_CSS } from './runtime-constants';
 import { convertScopedToShadow, registerStyle } from './styles';
 import { appDidLoad } from './update-component';
 export { setNonce } from '@platform';
@@ -185,7 +185,7 @@ export const bootstrapLazy = (lazyBundles: d.LazyBundlesRuntimeData, options: d.
     });
   });
 
-  dataStyles.innerHTML = 'slot-fb:not([hidden]){display:contents}';
+  dataStyles.innerHTML = SLOT_FB_CSS;
   dataStyles.setAttribute('data-styles', '');
   head.insertBefore(dataStyles, metaCharset ? metaCharset.nextSibling : head.firstChild);
 
--- a/src/runtime/bootstrap-lazy.ts
+++ b/src/runtime/bootstrap-lazy.ts
@@ -185,16 +185,18 @@ export const bootstrapLazy = (lazyBundles: d.LazyBundlesRuntimeData, options: d.
     });
   });
 
-  if (BUILD.invisiblePrehydration && (BUILD.hydratedClass || BUILD.hydratedAttribute)) {
-    visibilityStyle.innerHTML = cmpTags + HYDRATED_CSS;
-    visibilityStyle.setAttribute('data-styles', '');
+  visibilityStyle.innerHTML = 'slot-fb:not([hidden]){display:contents}';
+  visibilityStyle.setAttribute('data-styles', '');
+  head.insertBefore(visibilityStyle, metaCharset ? metaCharset.nextSibling : head.firstChild);
 
-    // Apply CSP nonce to the style tag if it exists
-    const nonce = plt.$nonce$ ?? queryNonceMetaTagContent(doc);
-    if (nonce != null) {
-      visibilityStyle.setAttribute('nonce', nonce);
-    }
-    head.insertBefore(visibilityStyle, metaCharset ? metaCharset.nextSibling : head.firstChild);
+  if (BUILD.invisiblePrehydration && (BUILD.hydratedClass || BUILD.hydratedAttribute)) {
+    visibilityStyle.innerHTML += cmpTags + HYDRATED_CSS;
+  }
+
+  // Apply CSP nonce to the style tag if it exists
+  const nonce = plt.$nonce$ ?? queryNonceMetaTagContent(doc);
+  if (nonce != null) {
+    visibilityStyle.setAttribute('nonce', nonce);
   }
 
   // Process deferred connectedCallbacks now all components have been registered
--- a/src/runtime/bootstrap-lazy.ts
+++ b/src/runtime/bootstrap-lazy.ts
@@ -32,7 +32,7 @@ export const bootstrapLazy = (lazyBundles: d.LazyBundlesRuntimeData, options: d.
   const customElements = win.customElements;
   const head = doc.head;
   const metaCharset = /*@__PURE__*/ head.querySelector('meta[charset]');
-  const visibilityStyle = /*@__PURE__*/ doc.createElement('style');
+  const dataStyles = /*@__PURE__*/ doc.createElement('style');
   const deferredConnectedCallbacks: { connectedCallback: () => void }[] = [];
   const styles = /*@__PURE__*/ doc.querySelectorAll(`[${HYDRATED_STYLE_ID}]`);
   let appLoadFallback: any;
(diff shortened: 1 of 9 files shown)


src/runtime/bootstrap-lazy.ts
```diff
@@ -32,7 +32,7 @@ export const bootstrapLazy = (lazyBundles: d.LazyBundlesRuntimeData, options: d.
   const customElements = win.customElements;
   const head = doc.head;
   const metaCharset = /*@__PURE__*/ head.querySelector('meta[charset]');
-  const visibilityStyle = /*@__PURE__*/ doc.createElement('style');
+  const dataStyles = /*@__PURE__*/ doc.createElement('style');
```
tanner-reits, line 35: Renamed this since it wasn't really only responsible for visibility styles anymore

src/runtime/bootstrap-lazy.ts
```diff
@@ -185,16 +185,18 @@ export const bootstrapLazy = (lazyBundles: d.LazyBundlesRuntimeData, options: d.
...
-    // Apply CSP nonce to the style tag if it exists
-    const nonce = plt.$nonce$ ?? queryNonceMetaTagContent(doc);
-    if (nonce != null) {
-      visibilityStyle.setAttribute('nonce', nonce);
-    }
-    head.insertBefore(visibilityStyle, metaCharset ? metaCharset.nextSibling : head.firstChild);
+  // Apply CSP nonce to the style tag if it exists
+  const nonce = plt.$nonce$ ?? queryNonceMetaTagContent(doc);
+  if (nonce != null) {
+    dataStyles.setAttribute('nonce', nonce);
```
tanner-reits, line 199: This shouldn't have been contingent on the `invisiblePrehydration` stuff in the first place

src/runtime/runtime-constants.ts
```diff
@@ -60,4 +60,6 @@ export const HYDRATED_STYLE_ID = 'sty-id';
 export const HYDRATE_CHILD_ID = 'c-id';
 export const HYDRATED_CSS = '{visibility:hidden}.hydrated{visibility:inherit}';
 
+export const SLOT_FB_CSS = 'slot-fb{display:contents}slot-fb[hidden]{display:none}';
```
tanner-reits, line 63: We have to have two style rules here rather than using `:not([hidden])` since we support some browsers that don't support `:not()` yet

test/karma/test-app/slot-fallback/karma.spec.ts
```diff
@@ -1,7 +1,6 @@
 import { setupDomTests, waitForChanges } from '../util';
 
-// TODO(STENCIL-18) Restore this test and fix the underlying issue.
-xdescribe('slot-fallback', () => {
+describe('slot-fallback', () => {
```
tanner-reits, line 3: Went ahead and restored these tests while I was in here since I needed to add some new ones anyway.

tanner-reits (commented):
Noting a few things

alicewriteswrongs (approved):
looks good to me!

src/runtime/runtime-constants.ts
rwaskiewicz: Can you do me a favor and add a (code) comment here with that information? That might not be immediately obvious if someone were reading the source file as-is.

src/runtime/bootstrap-lazy.ts
```diff
@@ -185,16 +185,18 @@ export const bootstrapLazy = (lazyBundles: d.LazyBundlesRuntimeData, options: d.
     });
   });
 
+  dataStyles.innerHTML = SLOT_FB_CSS;
+  dataStyles.setAttribute('data-styles', '');
+  head.insertBefore(dataStyles, metaCharset ? metaCharset.nextSibling : head.firstChild);
```
rwaskiewicz, line 190: For my own understanding, why do we hoist this out of [this conditional](https://github.com/ionic-team/stencil/pull/5028/files#diff-391b4dd275b4968d6629df297b178c24ebb07597f843b3f1ce5a3d6f4de82e16R192)?

src/runtime/styles.ts
```diff
@@ -67,6 +67,8 @@ export const addStyle = (styleContainerNode: any, cmpMeta: d.ComponentRuntimeMet
           styleContainerNode.insertBefore(styleElm, styleContainerNode.querySelector('link'));
         }
 
+        styleElm.innerHTML += SLOT_FB_CSS;
```
rwaskiewicz, line 70: For my own understanding, why do we add this here?

test/karma/test-app/invisible-prehydration-false/karma.spec.ts
```diff
@@ -5,13 +5,13 @@ describe('invisible-prehydration-false', () => {
...
-  it('the style element will not be placed in the head', async () => {
+  it('the style element will be placed in the head', async () => {
     tearDownStylesScripts();
     await setupDom('/invisible-prehydration-false/index.html', 1000);
 
     // Is the component hydrated?
     expect(document.querySelector('prehydrated-styles').innerHTML).toEqual('<div>prehydrated-styles</div>');
 
-    expect(document.head.querySelectorAll('style[data-styles]').length).toEqual(0);
+    expect(document.head.querySelectorAll('style[data-styles]').length).toEqual(1);
```
rwaskiewicz, line 15: Previously in these tests, checking for `style[data-styles]` was the only way to differentiate between invisible-prehydration being on/off. Can we update these tests to evaluate the content of that selector as well?

rwaskiewicz (changes requested):

src/runtime/bootstrap-lazy.ts
tanner-reits: Because the styles for `slot-fb` should not be contingently created and applied based on the `invisiblePrehydration` config option. So, we always create and insert the style element, but will continue to only inject the `invisiblePrehydration` styles based on that config value.

Also added a quick comment in the code to help clarify that a bit.

src/runtime/styles.ts
tanner-reits: This is how we style the `slot-fb` element for CE builds. This appends the `slot-fb` styles to the style element in the DOM. Added a code comment as well

test/karma/test-app/invisible-prehydration-false/karma.spec.ts
tanner-reits: Oh, meant to mention this on my initial comment review. I tried to do this, but at least one of the tests would fail when checking the content. But, if you ran the failing test(s) on their own, they would pass. So, there's something weird with multiple tests running that are related. These tests use a component and script from a different build, so maybe something with that?

But I also tried updating the tests so they each had their own component, but ran into basically the same issue so the elements in the DOM `head` may not be getting created correctly between tests.

Let me know if you have any ideas and I can try them out.

src/runtime/bootstrap-lazy.ts
```diff
@@ -185,16 +185,20 @@ export const bootstrapLazy = (lazyBundles: d.LazyBundlesRuntimeData, options: d.
     });
   });
 
+  // These style should always be constructed and inserted into the DOM
+  // We'll conditionally add some more styles later on
+  dataStyles.innerHTML = SLOT_FB_CSS;
```
rwaskiewicz, line 190: Double checking - why should we always add this style to DOM, even if a component doesn't have any slots? I think that's still lost on me, even with the added comment

src/runtime/bootstrap-lazy.ts
rwaskiewicz: Can you expound on that for me? `invisiblePrehydration` is `true` by default, which meant we were setting the `innerHTML` and `data-styles` by default for most folks only if `hydratedClass` or `hydratedAttribute` were set (or both). In fact, before `invisibleHydration` was added (a3d2928d), this section looked like the following in ac49d7c3:
```ts
  if (BUILD.hydratedClass || BUILD.hydratedAttribute) {
    visibilityStyle.innerHTML = cmpTags + HYDRATED_CSS;
    visibilityStyle.setAttribute('data-styles', '');
    head.insertBefore(visibilityStyle, metaCharset ? metaCharset.nextSibling : head.firstChild);
  }
```

src/runtime/styles.ts
rwaskiewicz: This probably ties directly into my question above (GH won't let me link it since I haven't submitted the review yet):

> Double checking - why should we always add this style to DOM, even if a component doesn't have any slots? I think that's still lost on me, even with the added comment

How do we know this is a `slot-fb` element?

tanner-reits: You're right, we shouldn't. For some reason didn't realize we actually know if we need to apply this for each component. Updated to wrap this in a conditional based on `CMP_FLAGS.hasSlotRelocation`

src/runtime/bootstrap-lazy.ts
tanner-reits: We shouldn't. Didn't realize we know specifically if any components are using slots outside the Shadow DOM. So, I added a boolean that will be set `true` if any components in the lazy build do that and that is used to conditionally add the `slot-fb` styles to the stylesheet in the DOM head.

src/runtime/bootstrap-lazy.ts
tanner-reits: That sill happens. So now, if you have a component using a slot outside the Shadow DOM (i.e. `scoped` or no encapsulation) and have the conditions for `invisiblePrehydration`, you get styles like the following:

```css
my-component {
  visibility:hidden
}
.hydrated {
  visibility:inherit
}
slot-fb {
  display: contents
}
slot-fb[hidden] {
  display: none
}
```

rwaskiewicz (approved):

PR #5032, opened 2023-11-08
test(path): add additional tests to join

rwaskiewicz:

I was questioning some divergent behavior between our `path.join` wrapper and the underlying Node call with respect to trailing slashes. This led to adding additional tests/docs

GitHub Issue Number: N/A

## What is the new behavior?

add additional tests to stencil's wrapped version of `path.join`, adding additional documentation to how the wrapped function works. specifically, add tests to verify the following behavior:
1. how trailing slashes are treated by the function
2. how 1+ string arguments of '/' are handled
3. how relative paths going 'up' one directory level are handled such that only a file is resolved

- [ ] Yes
- [ ] No

## Testing

No functional changes have occurred - only adding unit tests. I did track the execution of the tests in a debugger and walked through `normalizePath` to verify they were hitting the code points that I expected them to.
## Other information

the first case of trailing slashes is documented in JSDoc as well to signify a divergence from `path.join`. this behavior was the impetus for this change - I didn't understand what I was seeing at first when debugging some Stencil 4.x related path related bugs. this commit is spun out from that work.

related to: #5029

### DIFF
--- a/src/utils/path.ts
+++ b/src/utils/path.ts
@@ -210,12 +210,22 @@ export function relative(from: string, to: string): string {
  * normalization logic. This joins all the arguments (path fragments) into a
  * single path.
  *
+ * The calculation of the returned path follows that of Node's logic, with one exception - any trailing slashes will
+ * be removed from the calculated path.
+ *
  * @throws the underlying node function will throw if any argument is not a
  * string
  * @param paths the paths to join together
  * @returns a joined path!
  */
 export function join(...paths: string[]): string {
+  /**
+   * When normalizing, we should _not_ attempt to relativize the path returned by the native Node `join` method. When
+   * calculating the path from each of the string-based parts, Node does not prepend './' to any calculated path.
+   *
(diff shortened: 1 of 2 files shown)


tanner-reits (approved):

src/utils/path.ts
```diff
@@ -210,12 +210,22 @@ export function relative(from: string, to: string): string {
...
  * string
  * @param paths the paths to join together
  * @returns a joined path!
  */
 export function join(...paths: string[]): string {
+  /**
+   * When normalizing, we should _not_ attempt to relativize the path returned by the native Node `join` method. When
+   * calculating the path from each of the string-based parts, Node does not prepend './' to any calculated path.
+   *
+   * Note that our algorithm does differ from that of Node's, as described in this function's JSDoc regarding trailing
```
alicewriteswrongs, line 226:
```suggestion
   * Note that our algorithm does differ from Node's, as described in this function's JSDoc regarding trailing
```

alicewriteswrongs (approved):
looks good! just noticed one typo

PR #5034, opened 2023-11-08
refactor(worker): document worker code, improve types

alicewriteswrongs:
This documents a bunch of code related to our node.js worker setup.

Decided to do this while working on #5010. I didn't understand this code much before! Now I understand it a bit better.

## What is the new behavior?

This mainly adds JSDoc but also does a little bit of typing improvements.

- [ ] Yes
- [x] No

### DIFF
--- a/src/declarations/stencil-private.ts
+++ b/src/declarations/stencil-private.ts
@@ -1,4 +1,5 @@
 import { result } from '@utils';
+import type { Serializable as CPSerializable } from 'child_process';
 
 import type { InMemoryFileSystem } from '../compiler/sys/in-memory-fs';
 import type {
@@ -2321,6 +2322,13 @@ export interface VNodeProdData {
   [key: string]: any;
 }
 
+/**
+ * An abstraction to bundle up four methods which _may_ be handled by
+ * dispatching work to workers running in other OS threads or may be called
+ * synchronously. Environment and `CompilerSystem` related setup code will
+ * determine which one, but in either case the call sites for these methods can
+ * dispatch to this shared interface.
+ */
 export interface CompilerWorkerContext {
   optimizeCss(inputOpts: OptimizeCssInput): Promise<OptimizeCssOutput>;
   prepareModule(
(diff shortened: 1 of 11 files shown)


src/declarations/stencil-private.ts
```diff
@@ -2352,7 +2396,17 @@ export interface CompilerWorkerTask {
   retries?: number;
 }
 
-export type WorkerMsgHandler = (msgToWorker: MsgToWorker) => Promise<any>;
+/**
+ * A handler for IPC messages from the main thread to a worker thread. This
+ * involves dispatches an action specified by a {@link MsgToWorker} object to a
```
rwaskiewicz, line 2401: Non-blocking nit
```suggestion
 * involves dispatching an action specified by a {@link MsgToWorker} object to a
```

src/declarations/stencil-public-compiler.ts
```diff
@@ -1255,10 +1259,30 @@ export interface ResolveModuleIdResults {
...
+   * Handle a particular method
+   *
+   * @param name of the method to be passed to a worker
+   * @returns a Promise wrapping the results
+   */
+  handler(name: WorkerContextMethod): (...args: any[]) => Promise<any>;
+  /**
+   * description
+   *
+   */
```
rwaskiewicz, line 1281: Whoops - looks like a placeholder made its way in here

src/sys/node/node-worker-main.ts
```diff
@@ -4,16 +4,31 @@ import { EventEmitter } from 'events';
...
 export class NodeWorkerMain extends EventEmitter {
+  /**
+   * A handle for the OS process that is running our worker code
+   */
   childProcess: cp.ChildProcess;
   tasks = new Map<number, d.CompilerWorkerTask>();
   exitCode: number = null;
   processQueue = true;
-  sendQueue: d.MsgToWorker[] = [];
+  sendQueue: d.MsgToWorker<any>[] = [];
```
rwaskiewicz, line 19: For my own edification, why does the generic argument here have to be `any`?

src/sys/node/node-worker-thread.ts
```diff
@@ -1,14 +1,21 @@
...
   const sendHandle = (err: NodeJS.ErrnoException) => {
     if (err && err.code === 'ERR_IPC_CHANNEL_CLOSED') {
-      prcs.exit(0);
+      process.exit(0);
     }
   };
 
   const errorHandler = (stencilMsgId: number, err: any) => {
-    const errMsgBackToMain: d.MsgFromWorker = {
+    const errMsgBackToMain: d.MsgFromWorker<any> = {
```
rwaskiewicz, line 18: Similar to my question above regarding using `any` here. Feel free to ignore if the answer is the same

src/sys/node/node-worker-thread.ts
```diff
@@ -22,30 +29,30 @@ export const initNodeWorkerThread = (prcs: NodeJS.Process, msgHandler: d.WorkerM
...
       try {
         // run the handler to get the data
-        const msgFromWorker: d.MsgFromWorker = {
+        const msgFromWorker = {
           stencilId: msgToWorker.stencilId,
           stencilRtnValue: await msgHandler(msgToWorker),
-          stencilRtnError: null,
-        };
+          stencilRtnError: null as string | null,
+        } satisfies d.MsgFromWorker<T>;
```
rwaskiewicz, line 44: I may be missing something here, but couldn't we avoid the type assertion on `stencilRtnError` by explicitly typing `msgFromWorker` like we currently do on `main`?

src/declarations/stencil-public-compiler.ts
```diff
@@ -1255,10 +1259,30 @@ export interface ResolveModuleIdResults {
...
   send(...args: any[]): Promise<any>;
-  handler(name: string): (...args: any[]) => Promise<any>;
+  /**
+   * Handle a particular method
+   *
+   * @param name of the method to be passed to a worker
+   * @returns a Promise wrapping the results
+   */
+  handler(name: WorkerContextMethod): (...args: any[]) => Promise<any>;
```
rwaskiewicz, line 1277: I'm not sure we can narrow this type without it technically constituting a breaking change (with respect to typings) - we expose this through our Node API on `CompilerSystem`.

rwaskiewicz (changes requested):
One additional ask - can you update the semantic commit type to `refactor` since we did more than add JSDoc here? I'd like to signal to folks bisecting/scanning the codebase that this does more than add JSDoc to the codebase

src/sys/node/node-worker-main.ts
alicewriteswrongs: was just the easiest way to get it to pass - I might have to pass something like `"optimizeCss" | "prepareModule" | ..." instead which is just clunky

src/declarations/stencil-public-compiler.ts
alicewriteswrongs: mmmm right...I'll put that in another PR that we can merge in the future or something

I'd be surprised if this caused too many issues for people though, so if we're open to shipping a small breaking change in a 4.8 release we could leave it in here

src/sys/node/node-worker-thread.ts
alicewriteswrongs: yeah maybe it's just easier to leave it that way, I'll put it back

src/declarations/stencil-public-compiler.ts
alicewriteswrongs: whoops, thought I got them all! 😅

src/declarations/stencil-public-compiler.ts
rwaskiewicz: > I'd be surprised if this caused too many issues for people though, so if we're open to shipping a small breaking change in a 4.8 release we could leave it in here

Yeah, I doubt it'd be _too_ widespread. However, out of abundance of caution, can we slate that change for v5 and add a TODO so we don't forget?

alicewriteswrongs: Sure, I'll put in a TODO

alicewriteswrongs: Just added `STENCIL-1005`

rwaskiewicz (approved):

tanner-reits (approved):

PR #5051, opened 2023-11-09
wip()

alicewriteswrongs:

GitHub Issue Number: N/A

## What is the new behavior?

- 

- [ ] Yes
- [ ] No


alicewriteswrongs:
For possibly future reference, this was a quick hour or so hack to just see about running our current Karma tests directly in Github actions, using the Karma runners for headless chrome, firefox, etc. Long story short it might work! But we're not going to do so right now.

PR #5031, opened 2023-11-07
fix(test): don't fail build when jest typings can't be resolved

rwaskiewicz:
## What is the current behavior?

prior to this fix, building a stencil project with jest 29 installed could lead to an error such as:
```
TypeScript: node_modules/@stencil/core/testing/jest/jest-stencil-connector.d.ts:81:31
       Could not find a declaration file for module '@jest/types/build/Config'.
       '/PROJECT_DIRECTORY/node_modules/@jest/types/build/Config.js' implicitly
       has an 'any' type.If the '@jest/types' package actually exposes this module, try adding a new declaration
       (.d.ts) file containing `declare module '@jest/types/build/Config';`
```

GitHub Issue Number: #5030 

## What is the new behavior?

patch `getJestPreset` in the jest-stencil connector to return `any` instead of an inferred type for the function. 

Explicitly setting this return type to `any` is a temporary measure to unblock adopters. A TODO for STENCIL-1003 has been added to resolve this fully.

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Testing

1. Pull down the repo that is associated with #5030 
2. Run `yarn` to install dependencies
3. Run `yarn build` to build the project, observe the error
4. Add the following dev build to the project: `yarn add @stencil/core@4.7.1-dev.1699392001.733589`
5. Run `yarn build` to build the project, observe no error occurs

## Other information

STENCIL-1002: bug: Stencil 4.7.x & Jest 29.5.7 TypeScript error
fixes: #5030

Issue #5030, opened 2023-11-07
bug: Stencil 4.7.x & Jest 29.5.7 TypeScript error


marksy:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

4.7.0

### Current Behavior

When running `stencil build` with StencilJS `4.7.0` and Jest `29.5.7`, a TypeScript error occurs and the build fails:

```bash
TypeScript: node_modules/@stencil/core/testing/jest/jest-stencil-connector.d.ts:81:31
       Could not find a declaration file for module '@jest/types/build/Config'.
       '/Users/mark.chambers/Sites/timely/TimelyApp/Timely.UI/node_modules/@jest/types/build/Config.js' implicitly
       has an 'any' type.If the '@jest/types' package actually exposes this module, try adding a new declaration
       (.d.ts) file containing `declare module '@jest/types/build/Config';`

 L80:  coverageProvider: "v8" | "babel";
 L81:  coverageReporters: import("@jest/types/build/Config").CoverageReporters;
 L82:  coverageThreshold: {
```

### Expected Behavior

The build should succeed without type errors.

### System Info

```shell
* MacOS 14.1
* M2 Max
* Yarn `3.6.3`
* TypeScript `^5.1.6`
```

### Steps to Reproduce

From the Code Reproduction URL:

1. run `yarn` to install the dependencies
2. run `yarn start` (which runs `stencil build --dev --watch --serve --port 2003`) to start the build

### Code Reproduction URL

https://github.com/marksy/stencil-jest-example

### Additional Information

Posted the original question on Discord under the stencil-help section:
https://discord.com/channels/520266681499779082/1170864699361996830

rwaskiewicz:
@marksy Can you give the following dev build a try?

```
yarn add @stencil/core@4.7.1-dev.1699392001.7335898
```

marksy:
> @marksy Can you give the following dev build a try?
> 
> ```
> yarn add @stencil/core@4.7.1-dev.1699392001.7335898
> ```

Yeah that fixed the issue - thanks for the quick turnaround on this 👍🏼

[...]

PR #5053, opened 2023-11-09
refactor(jest): narrow getJestPreset typings

rwaskiewicz:

We have a small piece of technical debt introduced in #5031 where we use an `any` type for the Jest infrastructure. This PR fixes it!

GitHub Issue Number: N/A

## What is the new behavior?

this commit is a follow up to #5031 (5df16e6). it narrows the return type of the `JestFacade#getJestPreset` function from `any` to an alias to Jest's `Config.InitialOptions`.

prior to #5031, the return type of `getJestPreset` would be inferred by the TypeScript compiler, and a dynamic import to an internal Jest type declaration file would be generated for the typing of the function:
```ts
// a snippet of what was generated in `jest-stencil-connector.d.ts
export declare const getJestPreset: () => Partial<{
    coverageReporters: import("@jest/types/build/Config").CoverageReporters;
    // other fields omitted
}>;
```
however, stencil cannot/should not make assumptions about the location of this file. by placing an alias, the dyanmic import is removed from the output `.d.ts` file (having the same effect as #5031).

pr #5031 eliminated this by using `any` as an explicit return type, which would generate:
```ts
export declare const getJestPreset: () => any;
```

with this commit, we now generate:
```ts
export declare const getJestPreset: () => JestPresetConfig;
```
which no longer inlines dynamic imports for typings.

- [ ] Yes
- [x] No

## Testing

1. Pull down the repo that is associated with #5030 
2. Run `yarn` to install dependencies
3. Run `yarn build` to build the project, observe the error
4. Add the following dev build to the project: `yarn add @stencil/core@4.7.1-dev.1699565413.1847940`
5. Run `yarn build` to build the project, observe no error occurs
## Other information

STENCIL-1003 Dynamic Import Type Resolution Fails for Jest 28, 29

### DIFF
--- a/src/testing/jest/jest-apis.ts
+++ b/src/testing/jest/jest-apis.ts
@@ -27,7 +27,16 @@ export type JestPreprocessor = {
 // TODO(STENCIL-960): Improve this typing by narrowing it
 export type JestTestRunner = any;
 
-export type JestConfig = Config.InitialOptions;
+/**
+ * This type serves as an alias for the type representing the initial configuration for Jest.
+ * This alias serves two purposes:
+ * 1. It allows Stencil to have a single source of truth for the return type(s) on {@link JestFacade} (and its
+ *    implementations)
+ * 2. It prevents TypeScript from expanding Jest typings in the generated `.d.ts` file. This is necessary as TypeScript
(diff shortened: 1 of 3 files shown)


tanner-reits (approved):
Tested out locally and works great 👍

alicewriteswrongs (approved):

PR #5054, opened 2023-11-10
refactor(build): remove whitespace edit from parse5Plugin

alicewriteswrongs:
This removes a strange bit of code from our `parse5Plugin`, which bundles `parse5` for inclusion in the compiler and mock-doc bundles.

That code is here:

https://github.com/ionic-team/stencil/blob/bb0b1d46f63175dc09d0a23445be4d4a0d891a01/scripts/bundles/plugins/parse5-plugin.ts#L41-L54

Essentially what this does is any place in the bundled code where there are four space characters it convert it into two space characters instead.

This produces the weird behavior where when we're generating files like `components.d.ts` somewhat 'by hand' we have code that looks like this:

https://github.com/ionic-team/stencil/blob/bb0b1d46f63175dc09d0a23445be4d4a0d891a01/src/compiler/types/generate-app-types.ts#L133-L135

One would think that this would produce lines indented by 8 spaces, but actually when this file is written to disk it looks like this:

https://github.com/ionic-team/stencil/blob/ce06708e9666c69c162eb6f100a7d925b3e14344/test/todo-app/src/components.d.ts#L68-L72

In other words it's now indented 4 spaces instead of 8! What!

This is because the template literal is rewritten by the rollup plugin from 8 spaces to four. 

This isn't good! It creates a situation where (contrary to any reasonable explanation) the behavior of Stencil doesn't obviously follow from the source code.

So this changes things to that the re-writing behavior is removed and the source code is adjusted in order to keep the same output as before (so that it has, for the example cited above, 4 spaces in the source code and 4 spaces of indentation in the `components.d.ts` written to disk).

Making this change required edits to a few functions that generate the typedef strings and corresponding edits to a lot of unit tests.

We have this weird source code / behavior mis-match!

GitHub Issue Number: I started doing this when working on #5027 where this thing was producing different behavior when the compiler is built with rollup or with esbuild.

## What is the new behavior?

Now the indentation for `components.d.ts` comes from the source code and not from a weird build-time confusing things.

- [ ] Yes
- [x] No

## Testing

When making the change I first adjusted the indentation in `generate-app-types.ts` and friends until I stopped seeing any changes in the `components.d.ts` files written to disk when running `npm run test.analysis`. Then I adjusted the unit tests to pass with the new code.

### DIFF
--- a/scripts/bundles/plugins/parse5-plugin.ts
+++ b/scripts/bundles/plugins/parse5-plugin.ts
@@ -2,8 +2,7 @@ import rollupCommonjs from '@rollup/plugin-commonjs';
 import rollupResolve from '@rollup/plugin-node-resolve';
 import fs from 'fs-extra';
 import { join } from 'path';
-import type { NormalizedOutputOptions, OutputBundle } from 'rollup';
-import { OutputChunk, Plugin, rollup } from 'rollup';
+import { Plugin, rollup } from 'rollup';
 
 import type { BuildOptions } from '../../utils/options';
 import { aliasPlugin } from './alias-plugin';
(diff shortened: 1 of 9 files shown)


rwaskiewicz (approved):

tanner-reits (approved):
But think of the optimizations were missing out on 😂

PR #5063, opened 2023-11-13
chore(snc): fix snc error in src/testing/test-transpile.ts

alicewriteswrongs:
Just fixing some snc errors in this file

There's a few errors here!

## What is the new behavior?

There are now fewer errors! Amazing!

- [ ] Yes
- [x] No

## Testing

I think if CI is passing and whatnot we should be good.

### DIFF
--- a/src/testing/test-transpile.ts
+++ b/src/testing/test-transpile.ts
@@ -35,6 +35,6 @@ export function transpile(input: string, opts: TranspileOptions = {}): Transpile
  * @param stencilTranspilePaths a value to 'parse'
  * @returns a boolean
  */
-function parseStencilTranspilePaths(stencilTranspilePaths: string): boolean {
+function parseStencilTranspilePaths(stencilTranspilePaths: string | undefined): boolean {
   return stencilTranspilePaths === 'true' ? true : false;
 }


src/testing/test-transpile.ts
```diff
@@ -10,12 +10,12 @@ export function transpile(input: string, opts: TranspileOptions = {}): Transpile
     coreImportPath: isString(opts.coreImportPath) ? opts.coreImportPath : '@stencil/core/internal/testing',
     currentDirectory: opts.currentDirectory || process.cwd(),
     module: 'cjs', // always use commonjs since we're in a node environment
-    proxy: null,
+    proxy: undefined,
     sourceMap: 'inline',
-    style: null,
+    style: undefined,
```
alicewriteswrongs, line 15: This is typed thusly:

https://github.com/ionic-team/stencil/blob/fa074ad852bda39ccd0a1c5f983cf16735593e82/src/declarations/stencil-public-compiler.ts#L2702-L2706

src/testing/test-transpile.ts
```diff
@@ -10,12 +10,12 @@ export function transpile(input: string, opts: TranspileOptions = {}): Transpile
     coreImportPath: isString(opts.coreImportPath) ? opts.coreImportPath : '@stencil/core/internal/testing',
     currentDirectory: opts.currentDirectory || process.cwd(),
     module: 'cjs', // always use commonjs since we're in a node environment
-    proxy: null,
+    proxy: undefined,
```
alicewriteswrongs, line 13: this too!

https://github.com/ionic-team/stencil/blob/fa074ad852bda39ccd0a1c5f983cf16735593e82/src/declarations/stencil-public-compiler.ts#L2696-L2701

alicewriteswrongs (commented):
notes!

src/testing/test-transpile.ts
```diff
@@ -15,7 +15,9 @@ export function transpile(input: string, opts: TranspileOptions = {}): Transpile
     style: null,
     styleImportData: 'queryparams',
     target: 'es2015', // default to es2015
-    transformAliasedImportPaths: parseStencilTranspilePaths(process.env.__STENCIL_TRANSPILE_PATHS__),
+    transformAliasedImportPaths: parseStencilTranspilePaths(
+      process.env.__STENCIL_TRANSPILE_PATHS__ as unknown as string,
```
tanner-reits, line 19: Does this need to be cast as `unknown` first? I think you can get away with just `as string` since the type is already `string | undefined`. Alternatively could just be:

```suggestion
      process.env.__STENCIL_TRANSPILE_PATHS__!,
```

Think they all work fine tho so nbd.

tanner-reits (approved):

tanner-reits (approved):

src/testing/test-transpile.ts
```diff
@@ -15,7 +15,9 @@ export function transpile(input: string, opts: TranspileOptions = {}): Transpile
     style: null,
     styleImportData: 'queryparams',
     target: 'es2015', // default to es2015
-    transformAliasedImportPaths: parseStencilTranspilePaths(process.env.__STENCIL_TRANSPILE_PATHS__),
+    transformAliasedImportPaths: parseStencilTranspilePaths(
+      process.env.__STENCIL_TRANSPILE_PATHS__ as string,
+    ),
```
rwaskiewicz, line 20: I'm not sure we (safely) can assert this to be a string. From my understanding, we rely on `parseStencilTranspilePaths` to do the transformation for the value assigned to `transformAliasedImportPaths` - I would think `parseStencilTranspilePaths` should accept any of the types of `__STENCIL_TRANSPILE_PATHS__`. Currently, it's typed as:
https://github.com/ionic-team/stencil/blob/45718df9b8f50072c91f58c99163f56be068ff26/src/testing/test-transpile.ts#L38-L40
where the function signature assumes that we accept a `string`. Wouldn't it be safer (in that we don't require a type assertion) if we change the function signature itself?

```diff
- function parseStencilTranspilePaths(stencilTranspilePaths: string): boolean {
+ function parseStencilTranspilePaths(stencilTranspilePaths: string | undefined): boolean {
  return stencilTranspilePaths === 'true' ? true : false;
}
```

rwaskiewicz (changes requested):

src/testing/test-transpile.ts
alicewriteswrongs: sure, makes sense

rwaskiewicz (approved):

Issue #4745, opened 2023-08-31
bug: popover is not a known attribute of [element]


gfellerph:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

3.4.2

### Current Behavior

When adding an element with the popover attribute `<div popover></div>`, there is a typescript error on the console.

```shell
Type '{ popover: true; }' is not assignable to type
'HTMLAttributes<HTMLDivElement>'.Property 'popover' does not exist
on type 'HTMLAttributes<HTMLDivElement>'.
     <Host data-version={version}>
       <div popover></div>
       <div class="tabs-wrapper">
rebuild failed, watching for changes... in 106 ms
```

### Expected Behavior

The standard popover attribute is a known attribute.

### System Info

```shell
System: node 16.20.0
    Platform: darwin (22.5.0)
   CPU Model: Apple M1 Pro (8 cpus)
    Compiler: /Users/.../design-system/node_modules/.pnpm/@stencil+core@3.4.2/node_modules/@stencil/core/compiler/stencil.js
       Build: 1690208000
     Stencil: 3.4.2 ☀️
  TypeScript: 5.0.4
      Rollup: 2.42.3
      Parse5: 7.1.2
      Sizzle: 2.42.3
      Terser: 5.18.1
```

### Steps to Reproduce

Write a component that uses `popover` on any element.

### Code Reproduction URL

https://github.com/gfellerph/stencil-popover-repro

rwaskiewicz:
Thanks @gfellerph! I've confirmed that this is an issue in the typings - we'll keep an eye on `popover`  (which is still marked as [experimental/in technical preview in Firefox and Safari](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/popover#browser_compatibility)) and add it accordingly. Thanks again!

alicewriteswrongs:
The fix for this issue has been released as a part of today's [Stencil v4.8.0 release](https://github.com/ionic-team/stencil/releases/tag/v4.8.0)!

PR #5064, opened 2023-11-13
feat(declarations): add popover attributes to JSX declarations

tanner-reits:

Stencil's JSX declarations do not support `popover` or associated popover attributes

Closes: #4745 

## What is the new behavior?

- Adds the `popover` attribute to the base `HTMLAttributes`
- Adds `popoverTargetAction`, `popoverTargetElement`, and `popoverTarget` to the declarations for button and input

- [ ] Yes
- [X] No

## Testing

**Automated testing**

Can't run automated tests because `popover` is not supported in all browsers we run e2e tests against via BrowserStack

**Manual testing**

Installed a build of this branch in a started project and was able to build a demo using popovers and triggers.

## Other information

`popover` is not yet supported in all browsers. It will be up to the developer to decide when they can leverage these APIs

### DIFF
--- a/src/declarations/stencil-public-runtime.ts
+++ b/src/declarations/stencil-public-runtime.ts
@@ -901,8 +901,8 @@ export namespace JSXBase {
     value?: string | string[] | number;
 
     // popover
-    popoverTargetAction?: 'hide' | 'show' | 'toggle';
-    popoverTargetElement?: Element;
+    popoverTargetAction?: string;
+    popoverTargetElement?: Element | null;
     popoverTarget?: string;
   }
 
@@ -1081,8 +1081,8 @@ export namespace JSXBase {
     width?: number | string;
 
     // popover
-    popoverTargetAction?: 'hide' | 'show' | 'toggle';
-    popoverTargetElement?: Element;
+    popoverTargetAction?: string;
+    popoverTargetElement?: Element | null;
     popoverTarget?: string;
   }
 
@@ -1374,7 +1374,7 @@ export namespace JSXBase {
     tabIndex?: number;
     tabindex?: number | string;
     title?: string;
-    popover?: 'auto' | 'manual';
+    popover?: string | null;
 
     // Unknown
     inputMode?: string;
--- a/src/declarations/stencil-public-runtime.ts
+++ b/src/declarations/stencil-public-runtime.ts
@@ -899,6 +899,11 @@ export namespace JSXBase {
     name?: string;
     type?: string;
     value?: string | string[] | number;
+
+    // popover
+    popoverTargetAction?: 'hide' | 'show' | 'toggle';
+    popoverTargetElement?: Element;
+    popoverTarget?: string;
   }
 
   export interface CanvasHTMLAttributes<T> extends HTMLAttributes<T> {
@@ -1074,6 +1079,11 @@ export namespace JSXBase {
     webkitdirectory?: boolean;
     webkitEntries?: any;
     width?: number | string;
+
+    // popover
+    popoverTargetAction?: 'hide' | 'show' | 'toggle';
+    popoverTargetElement?: Element;
+    popoverTarget?: string;
   }
 
   export interface KeygenHTMLAttributes<T> extends HTMLAttributes<T> {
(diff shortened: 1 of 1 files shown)


tanner-reits:
I tried writing an e2e test just to verify that everything is working correctly inside a Stencil component, but, AFAICT, there isn't a way to detect when the "popover" element is actually displayed or not. The element always exists in the DOM, but when displayed it lives in the [top-layer](https://developer.chrome.com/blog/what-is-the-top-layer/) which I couldn't find a way to know when an element is in that.

src/declarations/stencil-public-runtime.ts
```diff
@@ -1364,6 +1374,7 @@ export namespace JSXBase {
     tabIndex?: number;
     tabindex?: number | string;
     title?: string;
+    popover?: 'auto' | 'manual';
```
tanner-reits, line 1377: So, this deviates a bit from the reproduction case in the linked issue and the [HTML spec](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/popover) for the global `popover` attribute. According to the spec, the `popover` attribute will default to "auto" if you do something like this:

```html
<div popover>...</div>
```

But, if we update our declaration here to be:
```ts
popover?: true | 'auto' | 'manual';
```
and write the same HTML, we actually end up getting the behavior for the "manual" value because it does not reflect the attribute correctly.

Not sure if this is something we've encountered before or can get around, but wanted to call it out since with this definition you'll be forced to explicitly set "auto" or "manual".

tanner-reits (commented):
Noting one thing

src/declarations/stencil-public-runtime.ts
alicewriteswrongs: should we maybe leave a jsx comment here to that effect?

rwaskiewicz: +1 for a comment here.

Looking at `lib.dom.ts`, it gets typed as:
```
    popover: string | null;
```
I assume they use `string` here rather than `'auto' | 'manual'` to:
1. allow for additional values to be added to [the spec ](https://html.spec.whatwg.org/multipage/popover.html#the-popover-attribute)
2. adhere to this part of the spec: "The [popover](https://html.spec.whatwg.org/multipage/popover.html#attr-popover) attribute's [invalid value default](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#invalid-value-default) is the [manual](https://html.spec.whatwg.org/multipage/popover.html#attr-popover-manual-state) state..."

In the case of `null`, I think that's to handle the latter part of the same sentence in the spec (emphasis mine):
> The [popover](https://html.spec.whatwg.org/multipage/popover.html#attr-popover) attribute's [invalid value default](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#invalid-value-default) is the [manual](https://html.spec.whatwg.org/multipage/popover.html#attr-popover-manual-state) state and i**ts [missing value default](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#missing-value-default) is the no popover state.**

Thoughts on aligning this with `lib.dom.ts`' typings?

rwaskiewicz (changes requested):

src/declarations/stencil-public-runtime.ts
tanner-reits: We can do that. Thoughts on typing it as `'auto' | 'manual' | string | null`? I know that's different from the spec a bit, but offers a bit of intellisense for devs. But, I also get if we don't wanna do that in case they remove one of those values. Any preference?

rwaskiewicz: I think that's fine - I doubt one of those will be removed, so the only thing I wanna watch out for is making sure `string` there doesn't widen "over" the `auto | manual` (basically ensure that the intellisense works for those value).

tanner-reits: `string` did widen over the other types so I just ended up matching the `lib.dom.ts` types.

rwaskiewicz (approved):

alicewriteswrongs (approved):

PR #5066, opened 2023-11-13
chore(snc): fix snc errors in src/testing/testing-utils.ts

alicewriteswrongs:
Just adding some checks that values are the right type before we use them to fix a few SNC errors.

- [ ] Yes
- [x] No

## Testing

I think if CI is all green then we should be good!

### DIFF
--- a/src/testing/testing-utils.ts
+++ b/src/testing/testing-utils.ts
@@ -1,5 +1,5 @@
 import type * as d from '@stencil/core/internal';
-import { isOutputTargetDistLazy, isOutputTargetWww } from '@utils';
+import { isOutputTargetDistLazy, isOutputTargetWww, isString } from '@utils';
 import { join, relative } from 'path';
 
 import { InMemoryFileSystem } from '../compiler/sys/in-memory-fs';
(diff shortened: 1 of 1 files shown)


tanner-reits (approved):

rwaskiewicz (approved):
I think this is fine - although `getAppUrl` is not under test (can we add something to the backlog for that please?). It's worth noting `getAppUrl` gets called 2x:
https://github.com/ionic-team/stencil/blob/45718df9b8f50072c91f58c99163f56be068ff26/src/testing/testing-utils.ts#L68
https://github.com/ionic-team/stencil/blob/45718df9b8f50072c91f58c99163f56be068ff26/src/testing/testing-utils.ts#L74
 
But _those_ callers are always called in the context in a large `try`/`catch` block:
https://github.com/ionic-team/stencil/blob/45718df9b8f50072c91f58c99163f56be068ff26/src/testing/testing.ts#L42-L139

That `try`/`catch` just logs the output of any thrown value - I think this will suppress those errors. Although TBH, I have no clue how often they occur in practice today.

alicewriteswrongs:
@rwaskiewicz created STENCIL-1010 for that

PR #5067, opened 2023-11-13
chore(snc): add devMode to ValidatedConfig

alicewriteswrongs:
This adds the devMode property to the ValidatedConfig type, and makes a small refactor to our config validation code to respect that change.

Currently we have I think one SNC error because this property isn't on the `ValidatedConfig` yet.

## What is the new behavior?

This addresses that issue by adding the property to the type, moving some logic up in `validate-config.ts`, and tweaking a mocking function.

- [ ] Yes
- [x] No

## Testing

If tests are passing I think we're probably pretty good here


rwaskiewicz (commented):
I'm not opposed to this change, but it doesn't look like it drops the SNC count (which I assumed it would based on the PR description).

@alicewriteswrongs is there a particular SNC this should fix? Maybe https://github.com/ionic-team/stencil/blob/9d2d0a0612caafb88c559907c6dec1a39e760615/src/compiler/bundle/worker-plugin.ts#L187?

I wonder if we 'traded' one for another which gave us a net zero SNC 🤔

alicewriteswrongs:
oh hmm, I'll try to track that down

I do know of at least one that should have been resolved by this, here:

https://github.com/ionic-team/stencil/blob/9f913a93af2e09f3f129ea09b393d03b597fb425/src/compiler/bundle/worker-plugin.ts#L187-L187

Screenshot 2023-11-14 at 2 32 03 PM

alicewriteswrongs:
found the new one, it was an issue in `validate-config.ts`, I changed up the logic there a little bit, lmk what you think!

rwaskiewicz (approved):
👍 Makes sense to me!

tanner-reits (approved):

PR #5069, opened 2023-11-13
chore(ci): install jest dependencies script

rwaskiewicz:

until now, all jest-related types in `src/testing/jest/jest-*` have resolved to stencil's root-level `node_modules/jest*` directories (jest-27) in ci. this hasn't been an issue, as many of the types were effectively typed as `any` (or something nearly equally as wide). as we begin to narrow the typings in `src/testing/jest`, we will need these typings to be available to us (otherwise ci will fail).

GitHub Issue Number: N/A

## What is the new behavior?

add a shell script for installing jest-related dependencies for the `src/testing/jest/jest-*` subdirectories. this allows a developer to install all dependencies in one command, rather than chaining N number of directory changes and `npm ci` commands for every N directories.

this script has been wired up to the `get-core-dependencies` github action as well. this ensures that all jest dependencies are available during ci. 

this route was taken over an npm `postinstall` script, as a such a script will run when installing a packed tarball (and fail), breaking the existing `npm pack && npm install` workflow for testing stencil.

- [ ] Yes
- [x] No

### Verify this script can run anywhere
Start from the root of the repo with this branch checked out and run:
```bash
./src/testing/jest/install-dependencies.sh \
&& cd src \
&& ./testing/jest/install-dependencies.sh \
&& cd ./testing \
&& ./jest/install-dependencies.sh \
&& cd jest \
&& ./install-dependencies.sh
```

### Verify this Runs in CI
From [the logs](https://github.com/ionic-team/stencil/actions/runs/6855129064/job/18639601742?pr=5069)
```
 Installing dependencies in jest-27-and-under
npm WARN deprecated w3c-hr-time@1.0.2: Use your platform's native performance.now() and performance.timeOrigin.

added 342 packages, and audited 343 packages in 2s

30 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities

Installing dependencies in jest-28

added 289 packages, and audited 290 packages in 2s

32 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities

Installing dependencies in jest-29

added 291 packages, and audited 292 packages in 2s

32 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities
```
## Other information

Technically, I didn't add the `strictNullChecks` here. They've always been there, just hidden by the fact that we weren't resolving to the correct versions of Jest.

STENCIL-994 Improve Local Dev Experience for Jest Work

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,7 +36,8 @@ Please see our [Contributor Code of Conduct](https://github.com/ionic-team/stenc
    [Install it](https://docs.volta.sh/guide/getting-started) before proceeding.
    1. There's no need to install a specific version of npm or Node right now, it shall be done automatically for you in
       the next step
-5. Run `npm install`
+5. Run `npm ci`
+6. (Optional) If you are working on [Jest support](./src/testing/jest), [see the installation steps in that directory's README](./src/testing/jest/README.md#installing-dependencies).
 
 
 ### Updates
--- a/src/testing/jest/install-dependencies.sh
+++ b/src/testing/jest/install-dependencies.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+
+set -e -u -o pipefail
+
+# Get the directory where this script is located - doing so will allow us to run the script from anywhere in the
+# project. Since we retrieve the directory this script lives in via a subshell, we'll need to `cd` to in explicitly in
+# a separate command.
+SCRIPT_DIR=$(cd "$(dirname "$0")" && pwd)
+cd "$SCRIPT_DIR"
+
+# Flag to check if any jest-[SOME_NUMBER] directories were found and processed
+found=false
+
+# Loop through directories start with 'jest-', e.g. `jest-27-and-under`, `jest-28`, etc.
+for dir in jest-*; do
+    if [[ -d "$dir" ]]; then
+        found=true
+
+        cd "$dir"
+        echo "Installing dependencies in $dir"
+        npm ci
+        # print a newline to separate sequential installs
+        echo ""
+
+        # go back to where we started
+        cd -
+    fi
+done
+
+# If no directories were found and processed, print an error and exit
+if [[ $found == false ]]; then
+    echo "Error: No jest directories were found"
+    exit 1
+fi
--- a/.github/workflows/actions/get-core-dependencies/action.yml
+++ b/.github/workflows/actions/get-core-dependencies/action.yml
@@ -13,5 +13,8 @@ runs:
         cache: 'npm'
 
     - name: Install Dependencies
-      run: npm ci
+      run: |
+        npm ci \
+        && ./src/testing/jest/install-dependencies.sh
+
       shell: bash
--- a/src/testing/jest/README.md
+++ b/src/testing/jest/README.md
@@ -2,6 +2,13 @@
 
 This directory contains support for the Jest testing library.
 
+## Installing Dependencies
+
+To be able to work on the sub-projects in this directory, Jest dependencies are required to be installed explicitly.
+To do so, one may either:
+1. (Recommended) Run the `install-dependencies.sh` script found in this directory to install _all_ Jest dependencies
+1. Run `npm ci` in the Jest version-specific directory you plan on working in
+
 ## Adding Support for a New Version of Jest
 
 The steps for adding support for a new version of Jest can be found below.


src/testing/jest/install-dependencies.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+
+set -e
+
+# Get the directory where this script is located
+SCRIPT_DIR=$(cd "$(dirname "$0")" && pwd)
+
+# Change to the directory where the script is located so that it can be called from anywhere
+cd "$SCRIPT_DIR"
```
alicewriteswrongs, line 9: sometimes bash confuses me, but didn't we already `cd` to `$0` on line 6?

src/testing/jest/install-dependencies.sh
```diff
@@ -0,0 +1,43 @@
...
+        else
+            echo "Error: Cannot enter directory $dir"
+            exit 1
+        fi
+
+        # go back to where we started
+        cd - > /dev/null || {
+            echo "Error: Unable to leave $dir"
+            exit 1
+        }
```
alicewriteswrongs, line 35: with `set -e` won't we exit for a non-zero exit code here if the `cd` fails anyhow?

CONTRIBUTING.md
```diff
@@ -36,7 +36,8 @@ Please see our [Contributor Code of Conduct](https://github.com/ionic-team/stenc
    [Install it](https://docs.volta.sh/guide/getting-started) before proceeding.
    1. There's no need to install a specific version of npm or Node right now, it shall be done automatically for you in
       the next step
-5. Run `npm install`
+5. Run `npm ci`
+6. (Optional) If you are working on [Jest support](./src/testing/jest), [see the installation steps in that directory's README](./src/testing/jest/README.md#installing-dependencies).
```
alicewriteswrongs, line 40: would it make sense for us to put this in a `postinstall` script in our root `package.json` so the jest deps are just sort of always there for us?

alicewriteswrongs (changes requested):
a few comments and questions

CONTRIBUTING.md
rwaskiewicz: I don't think so, as it breaks the `npm pack` workflow of testing a Stencil tarball.
When installing a tarball of an npm package (so one where we generate via `npm ci && npm run build && npm pack`) then go to `npm install [PATH_TO_TARBALL]` in a starter project (or wherever), the post install scripts will attempt to run in the project we're installing the tarball into :-(

src/testing/jest/install-dependencies.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+
+set -e
```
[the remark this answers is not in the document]
rwaskiewicz: Yeah, let me add that (2343b6b9c) - I forgot to mention that I _did_ run this through shellcheck's online checker (and nothing came up in my IDE, which has a shellcheck plugin).

> we don't need to do this right now, but if we're going to start having more shellscripts around we might want to think about adding [shellcheck](https://github.com/koalaman/shellcheck) to CI (or a better tool if there's one around)

Agreed, definitely worth considering if/when we start to amass more shell scripts here

src/testing/jest/install-dependencies.sh
rwaskiewicz: In line 6, we're inside of a subshell, so that `cd` doesn't get persisted to our current directory state. Added a comment in bfd16e4a3 to clarify

src/testing/jest/install-dependencies.sh
rwaskiewicz: Oh, good point (I missed this comment when I resubmitted for review). Let me double check on that. We may be able to get rid of this + another `npm ci` related failure check

src/testing/jest/install-dependencies.sh
alicewriteswrongs: right, so it's like

1. dirname gives us the relative path from the calling location to the script's directory
2. cd-ing to that then puts us in that directory
3. doing `pwd` gives us the full path to the script dir

makes sense when you think it through haha

CONTRIBUTING.md
alicewriteswrongs: ohh right I forgot that they would run then too

hmmmmmm wondering if maybe we want a `bootstrap` command at some point 🤔 

something to worry about later!

alicewriteswrongs (approved):

tanner-reits (approved):
LGTM

PR #5073, opened 2023-11-14
chore(snc): fix some boolean-related snc errors

alicewriteswrongs:
This fixes a bunch of errors caused by passing `boolean | undefined` where a property or parameter is actually typed `boolean`. This isn't likely to actually cause too many issues at runtime since `undefined` is falsy, but it does cause `strictNullChecks` errors that we'd like to get rid of.

To fix the issues we can just do a `!!` coercion of the falsy `undefined` value to `false`.

GitHub Issue Number: N/A

## What is the new behavior?

- 

- [ ] Yes
- [ ] No

### DIFF
--- a/src/compiler/transformers/static-to-meta/component.ts
+++ b/src/compiler/transformers/static-to-meta/component.ts
@@ -65,7 +65,7 @@ export const parseStaticComponentMeta = (
     componentClassName: cmpNode.name ? cmpNode.name.text : '',
     elementRef: parseStaticElementRef(staticMembers),
     encapsulation,
-    shadowDelegatesFocus: parseStaticShadowDelegatesFocus(encapsulation, staticMembers),
+    shadowDelegatesFocus: !!parseStaticShadowDelegatesFocus(encapsulation, staticMembers),
     properties: parseStaticProps(staticMembers),
     virtualProperties: parseVirtualProps(docs),
     states: parseStaticStates(staticMembers),
--- a/src/cli/task-serve.ts
+++ b/src/cli/task-serve.ts
@@ -6,7 +6,7 @@ export const taskServe = async (config: ValidatedConfig) => {
   config.suppressLogs = true;
 
   config.flags.serve = true;
-  config.devServer.openBrowser = config.flags.open;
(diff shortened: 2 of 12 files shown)


src/compiler/transformers/static-to-meta/component.ts
```diff
@@ -65,7 +65,7 @@ export const parseStaticComponentMeta = (
     componentClassName: cmpNode.name ? cmpNode.name.text : '',
     elementRef: parseStaticElementRef(staticMembers),
     encapsulation,
-    shadowDelegatesFocus: parseStaticShadowDelegatesFocus(encapsulation, staticMembers),
+    shadowDelegatesFocus: !!parseStaticShadowDelegatesFocus(encapsulation, staticMembers),
```
alicewriteswrongs, line 68: this necessitates an edit to a spec file (`src/compiler/transformers/test/parse-encapsulation.spec.ts`) but I believe this is safe to do because the only reference to this field is here:

https://github.com/ionic-team/stencil/blob/1c359016e09b66aeb82c704a934d1355cceb4cb0/src/utils/format-component-runtime-meta.ts#L26-L28

we're checking that the value is truthy here, so coercing `null` to `false` won't have any effect on runtime behavior.

_however_ the `parseStaticShadowDelegatesFocus` function does specifically return `null` if the encapsulation mode isn't shadow:

https://github.com/ionic-team/stencil/blob/1c359016e09b66aeb82c704a934d1355cceb4cb0/src/compiler/transformers/static-to-meta/encapsulation.ts#L29-L48

however given that `shadowDelegatesFocus` property here is typed `boolean` and we're not doing anything with that `null` return value here I think coercing here is the right thing to do (although we could also make the return type for `parseStaticShadowDelegatesFocus` always be `boolean`, which would have the same result overall)

alicewriteswrongs (commented):
one note

src/compiler/transformers/static-to-meta/component.ts
rwaskiewicz: I think that makes sense!

rwaskiewicz (approved):

tanner-reits (approved):

PR #5078, opened 2023-11-16
refactor(build): stop removing process.binding in compiler

alicewriteswrongs:
This removes some code which purportedly rewrote the compiler bundle to remove a call to `process.binding` because it's deprecated.

The only `process.binding` call that _would_ otherwise be included in the bundle comes from the
[`builtin-modules`](https://github.com/sindresorhus/builtin-modules/) package which is included because it's a dependency of Rollup's node-resolve plugin and `eslint-plugin-jsdoc`

`process.binding` is an internal API which has to do with how node.js interfaces with native code. `builtin-modules` uses it here as a fallback way to get a list of builtin modules:

https://github.com/sindresorhus/builtin-modules/blob/0d234f3ad3965c57d8cc5b9ca198c505fe032452/index.js#L9-L11

The motivation for removing this rewrite from the compiler bundle is that the code of interest changed in this commit:

https://github.com/sindresorhus/builtin-modules/commit/0a73da3bb4926eeae34636746f193a9aa5bedbca

We upgraded to a version of `builtin-modules` which includes this change in https://github.com/ionic-team/stencil/pull/4494 when the resolved version of `builtin-modules` went from 3.2.0 to 3.3.0. This meant that our edit was no longer actually removing the offending line, but we didn't experience any problems leaving it in, so we concluded that its presence isn't justified.

GitHub Issue Number: N/A

## What is the new behavior?

- 

- [ ] Yes
- [x] No

### DIFF
--- a/scripts/bundles/compiler.ts
+++ b/scripts/bundles/compiler.ts
@@ -2,9 +2,8 @@ import rollupCommonjs from '@rollup/plugin-commonjs';
 import rollupJson from '@rollup/plugin-json';
 import rollupNodeResolve from '@rollup/plugin-node-resolve';
 import fs from 'fs-extra';
-import MagicString from 'magic-string';
 import { join } from 'path';
-import type { OutputChunk, RollupOptions, RollupWarning, TransformResult } from 'rollup';
+import type { OutputChunk, RollupOptions, RollupWarning } from 'rollup';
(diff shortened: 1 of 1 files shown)


christian-bromann (approved):
LGTM 👍

tanner-reits (approved):
🪓

PR #5079, opened 2023-11-17
chore(build): build the dev-server module with esbuild

christian-bromann:

This patch adds an esbuild step for compiling the `dev-server` package. Some note worthy comments:

- I've found some required build steps, e.g. for creating `open-in-editor-api.js`, `ws.js`, `visualstudio.vbs` and `xdg-open` in sys-node bundle and migrated them over. They only seem to be connected to `dev-server` features.
- because of that I moved the `buildSysNode` step as pre-requisite of other the CLI and dev-server build scripts as I discovered race conditions (e.g. `@sys-api-node` -> `./sys/node/index.js` not being able to get resolved by `dev-server.js`
- there were some esbuild plugins necessary to transpile the source code in a similar way as we have done with rollup, rollup was a bit better compiling down from ESM to CJS

GitHub Issue Number: N/A

## What is the new behavior?

- The dev server works the same way, just gets compiled by esbuild than rollup.

- [ ] Yes
- [x] No

## Testing

There are already some tests in place verifying that files are created successfully. I will do some further manual testing tomorrow.

## Other information

n/a

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5638,6 +5638,24 @@
         "sourcemap-codec": "^1.4.8"
       }
     },
+    "node_modules/esbuild-plugin-replace": {
+      "version": "1.4.0",
+      "resolved": "https://registry.npmjs.org/esbuild-plugin-replace/-/esbuild-plugin-replace-1.4.0.tgz",
+      "integrity": "sha512-lP3ZAyzyRa5JXoOd59lJbRKNObtK8pJ/RO7o6vdjwLi71GfbL32NR22ZuS7/cLZkr10/L1lutoLma8E4DLngYg==",
+      "dev": true,
+      "dependencies": {
+        "magic-string": "^0.25.7"
+      }
+    },
+    "node_modules/esbuild-plugin-replace/node_modules/magic-string": {
+      "version": "0.25.9",
+      "resolved": "https://registry.npmjs.org/magic-string/-/magic-string-0.25.9.tgz",
+      "integrity": "sha512-RmF0AsMzgt25qzqqLc1+MbHmhdx0ojF2Fvs4XnOqz2ZOBXzzkEwc/dJQZCYHAn7v1jbVOjAZfK8msRn4BxO4VQ==",
+      "dev": true,
+      "dependencies": {
+        "sourcemap-codec": "^1.4.8"
+      }
+    },
     "node_modules/escalade": {
       "version": "3.1.1",
       "resolved": "https://registry.npmjs.org/escalade/-/escalade-3.1.1.tgz",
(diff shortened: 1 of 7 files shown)


scripts/esbuild/build.ts
```diff
@@ -10,7 +11,9 @@ async function main() {
     isWatch: !!process.argv.includes('--watch'),
   });
 
-  await Promise.all([buildCli(opts), buildSysNode(opts)]);
+  // build the sys node files first as other components depend on it
+  await buildSysNode(opts);
+  await Promise.all([buildCli(opts), buildDevServer(opts)]);
```
alicewriteswrongs, line 16: what exactly in the dev server build depends on the sys-node build already being done? j/w because I'd really like to avoid having the builds depend on each other at build-time if possible, currently the approach I've been taking with the esbuild bundles so far is sort of that they all mark each other as external and then can basically independently re-build

they also currently support watch mode, which if we have `await buildSysNode` before the `Promise.all` will just only build sys-node in watch mode (since that single `Promise` will never resolve).

a general point I've found in going through our rollup-based builds to port them over is that there is _a lot_ of weird stuff in our builds, so it's also totally possible if you found something like that that we may want to remove the weirdness from both the esbuild and the rollup build at the same time

christian-bromann: You were totally right! I was depending on a file that should have been marked as "external". I resolved the issue and put all builds back to be run in parallel. I totally agree these builds should not depend on each other, that will only result in more complexity.

christian-bromann:
There are some build artifacts we now have twice in the bundles directory and esbuild task:

https://github.com/ionic-team/stencil/blob/1b18f2d938b26d20630950f40a3eae58728c4948/scripts/bundles/sys-node.ts#L110-L125

I added a comment to make sure we clean this up later on. I can't remove them just yet as we run the same build tests (`npm run test.dist`) for both.

scripts/esbuild/dev-server.ts
```diff
@@ -0,0 +1,304 @@
...
+import fs from 'fs-extra';
+import ts from 'typescript';
+
+import { createContentTypeData } from '../bundles/plugins/content-types-plugin';
+import { bundleExternal, sysNodeBundleCacheDir } from '../bundles/sys-node';
+import { getBanner } from '../utils/banner';
+import { type BuildOptions, createReplaceData } from '../utils/options';
+import { writePkgJson } from '../utils/write-pkg-json';
+import { getBaseEsbuildOptions, getEsbuildAliases, runBuilds } from './util';
+const connectorName = 'connector.html';
```
alicewriteswrongs, line 15: style nit: usually for constants I'll go for `const VARIABLE_NAME = ...` type of style

alicewriteswrongs (changes requested):
just noted a few things, but testing this out it all works great!

christian-bromann:
@alicewriteswrongs good catch, updated the PR

alicewriteswrongs (approved):
looks great!

package-lock.json
```diff
@@ -5807,6 +5807,24 @@
         "sourcemap-codec": "^1.4.8"
       }
     },
+    "node_modules/esbuild-plugin-replace": {
```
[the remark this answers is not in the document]
christian-bromann: Good catch, forgot to update my `package-lock.json`

rwaskiewicz (approved):

PR #5083, opened 2023-11-17
refactor(build): add esbuild-plugin-replace to CLI bundle

alicewriteswrongs:
This adds `esbuild-plugin-replace` plugin to the ESbuild-based CLI bundle. This was inadvertently omitted from #4187, an oversight which wasn't noticed until working on #5027.

The Rollup-based CLI bundle uses the `replacePlugin` here:

https://github.com/ionic-team/stencil/blob/7c86019e5910935e8731e10bd7156d25a85e997c/scripts/bundles/cli.ts#L78-L78

This injects some build-time constants into the code to support things like printing the current version and whatnot.

GitHub Issue Number: N/A

## What is the new behavior?

- 

- [ ] Yes
- [ ] No


christian-bromann (approved):
Nice 👌 I've done the same in the `dev-server` PR as well!

rwaskiewicz (commented):
I'm not sure this is strictly needed - I know in the PR summary the reasoning of this is build parity:

> This is being added out of an abundance of caution. I checked and the output does not seem to be different than it, but I want to make sure we have build setup parity as much as possible anyhow.

AFAICT, if I compile `cli/` using rollup but without the `replacePlugin` (effectively just commenting out [this line of code](https://github.com/ionic-team/stencil/blob/2dc43e57aee58ee8fbc04b910fdc7b3f44eaefd0/scripts/bundles/cli.ts#L78)), I'm not seeing any of [those replacement keys](https://github.com/ionic-team/stencil/blob/9a3e81cb773aedc65557768280d961c12780cf9f/scripts/utils/options.ts#L129-L142) in the build output (of `cli/` at the root level of the repo).  

If we want to move forward with this, can you please create a ticket in JIRA for us to audit usages of our plugins after we've migrated to ESBuild?

alicewriteswrongs:
We don't specifically need it...I've got it as part of #5027 which is where it will really count so I think we can close this

PR #5104, opened 2023-11-27
fix(runtime): call form-associated lifecycle callbacks w/ `this`

### LINKED ISSUE #5106: bug: can't access `this` in form-associated callbacks
### Prerequisites

### Stencil Version

4.8.0

### Current Behavior

In #4939 we added support for form-associated custom element lifecycle callbacks like `formAssociatedCallback`, however the way they're currently being called doesn't resolve the correct `this` value to ensure that properties on a Stencil component (e.g. ones decorated with `@State`) can be resolved properly.

So a little component like this one won't work correctly 😢:

```jsx

@Component({
  tag: 'my-component',
  styleUrl: 'my-component.css',
  shadow: true,
  formAssociated: true,
})
export class MyComponent {
  @State() showWords: boolean = false;

  formAssociatedCallback(form: any) {
    console.log('formAssociated called');
    console.log(this.showWords);
  }

  render() {
    return <div>Hello, World!</div>;
  }
}
```

### Expected Behavior

If I try to access `this` in a method as in the example it should work correctly.

### System Info

_No response_

### Steps to Reproduce

In any of the form-associated callbacks (e.g. `formAssociatedCallback`) try to access something on `this`, like a `@State` property. You should get an error that it's not found (cannot read property 'so and so' of undefined). See the attached repro case.

### Code Reproduction URL

https://github.com/alicewriteswrongs/stencil-fa-cb-repro

### Additional Information

_No response_

alicewriteswrongs:
This fixes an issue where form-associated lifecycle callbacks were not being called with the correct `this` value, making it impossible to set something like a `@State` value from within a callback.

The right `this` value isn't resolved for the form-associated callbacks.

GitHub Issue Number: fixes #5106

## What is the new behavior?

It's resolved the same way it is for other similar lifecycle callbacks.

- [ ] Yes
- [x] No

## Testing

I added some code to our current karma test for this functionality which functions as a regression test, i.e. it will cause a test to fail without this fix.

### DIFF
--- a/src/runtime/proxy-component.ts
+++ b/src/runtime/proxy-component.ts
@@ -36,10 +36,13 @@ export const proxyComponent = (
           const elm = BUILD.lazyLoad ? hostRef.$hostElement$ : this;
           const instance: d.ComponentInterface = BUILD.lazyLoad ? hostRef.$lazyInstance$ : elm;
           if (!instance) {
-            hostRef.$onReadyPromise$.then((instance: d.ComponentInterface) => instance[cbName]?.(...args));
+            hostRef.$onReadyPromise$.then((instance: d.ComponentInterface) => {
+              const cb = instance[cbName];
+              typeof cb === 'function' && cb.call(instance, ...args);
+            });
           } else {
             const cb = instance[cbName];
-            typeof cb === 'function' && cb(...args);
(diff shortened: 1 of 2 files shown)


christian-bromann (approved):
👍

PR #5105, opened 2023-11-27
fix(runtime): patch `textContent` whenever shadow DOM is not used

tanner-reits:

Stencil does not patch a component's `textContent` method for our pseudo-slot behavior in all cases. Patching this method should occur whenever a component does not use the Shadow DOM.

Fixes: #3977 

## What is the new behavior?

`textContent` gets patched for both scoped encapsulation and when not using any encapsulation strategy.

- [ ] Yes
- [X] No

## Testing

**Manual testing**
Tested in the linked issue's reproduction case

**Automated testing**
Added e2e tests to verify the patch behavior and that it happens for the appropriate encapsulation strategies.

## Other information

The patch implementation may change in the future. STENCIL-1019 was added as a potential task for a future major version to evaluate our implementation compared to the spec.

### DIFF
--- a/src/runtime/bootstrap-custom-element.ts
+++ b/src/runtime/bootstrap-custom-element.ts
@@ -44,7 +44,7 @@ export const proxyCustomElement = (Cstr: any, compactMeta: d.ComponentRuntimeMet
   }
 
   // TODO(STENCIL-914): this check and `else` block can go away and be replaced by just a
-  // check for shadow encapsulation once we default our pseudo-slot behavior
+  // check for shadow encapsulation/shim once we default our pseudo-slot behavior
   if (
     BUILD.experimentalSlotFixes &&
     (cmpMeta.$flags$ & CMP_FLAGS.needsShadowDomShim || !(cmpMeta.$flags$ & CMP_FLAGS.shadowDomEncapsulation))
@@ -60,8 +60,8 @@ export const proxyCustomElement = (Cstr: any, compactMeta: d.ComponentRuntimeMet
     if (BUILD.appendChildSlotFix) {
       patchSlotAppendChild(Cstr.prototype);
     }
-    if (BUILD.scopedSlotTextContentFix) {
-      patchTextContent(Cstr.prototype, cmpMeta);
+    if (BUILD.scopedSlotTextContentFix && BUILD.scoped && cmpMeta.$flags$ & CMP_FLAGS.scopedCssEncapsulation) {
+      patchTextContent(Cstr.prototype);
     }
   }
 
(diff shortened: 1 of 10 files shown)


src/runtime/bootstrap-custom-element.ts
```diff
@@ -57,8 +60,8 @@ export const proxyCustomElement = (Cstr: any, compactMeta: d.ComponentRuntimeMet
     if (BUILD.appendChildSlotFix) {
       patchSlotAppendChild(Cstr.prototype);
     }
-    if (BUILD.scopedSlotTextContentFix) {
-      patchTextContent(Cstr.prototype, cmpMeta);
+    if (BUILD.scopedSlotTextContentFix && BUILD.scoped && cmpMeta.$flags$ & CMP_FLAGS.scopedCssEncapsulation) {
```
tanner-reits, line 63: This check is the same that was in the `patchTextContent()` method. Just moved it to this level so we don't have to make changes in the future when we remove these `extras` flags.

src/runtime/bootstrap-lazy.ts
```diff
@@ -152,8 +155,8 @@ export const bootstrapLazy = (lazyBundles: d.LazyBundlesRuntimeData, options: d.
         if (BUILD.appendChildSlotFix) {
           patchSlotAppendChild(HostElement.prototype);
         }
-        if (BUILD.scopedSlotTextContentFix) {
-          patchTextContent(HostElement.prototype, cmpMeta);
+        if (BUILD.scopedSlotTextContentFix && BUILD.scoped && cmpMeta.$flags$ & CMP_FLAGS.scopedCssEncapsulation) {
```
tanner-reits, line 158: This check is the same that was in the `patchTextContent()` method. Just moved it to this level so we don't have to make changes in the future when we remove these `extras` flags.

src/runtime/dom-extras.ts
```diff
@@ -204,58 +204,56 @@ export const patchSlotInsertAdjacentElement = (HostElementPrototype: HTMLElement
...
+// TODO(STENCIL-1019): Evaluate if this patch should be modified
 /**
  * Patches the text content of an unnamed slotted node inside a scoped component
  * @param hostElementPrototype the `Element` to be patched
- * @param cmpMeta component runtime metadata used to determine if the component should be patched or not
  */
-export const patchTextContent = (hostElementPrototype: HTMLElement, cmpMeta: d.ComponentRuntimeMeta): void => {
-  if (BUILD.scoped && cmpMeta.$flags$ & CMP_FLAGS.scopedCssEncapsulation) {
-    const descriptor = Object.getOwnPropertyDescriptor(Node.prototype, 'textContent');
+export const patchTextContent = (hostElementPrototype: HTMLElement): void => {
```
tanner-reits, line 212: Looks like a lot of changes here, but it should all just be formatting from removing the top-level `if` condition

PR #5109, opened 2023-11-29
fix(compiler): make sure typesDir exist before writing to it

### LINKED ISSUE #5091: bug: Stencil occasionally does not generate dist/types directory
### Prerequisites

### Stencil Version

4.7.2

### Current Behavior

When some special conditions apply, Stencil sometimes fails to create the `dist/types` directory. This does **not** lead to a build error, which means that a broken version is generated if types are required in a downstream project.

Note: The missing types directory is the only noticable change, all other files looked the same.

### Expected Behavior

The types directory should always be generated.

### System Info

```shell
System: node 20.9.0
    Platform: windows (10.0.19045)
   CPU Model: Intel(R) Core(TM) i7-9850H CPU @ 2.60GHz (12 cpus)
    Compiler: O:\ws\temp\stencil-type-issue\node_modules\@stencil\core\compiler\stencil.js
       Build: 1699889895
     Stencil: 4.7.2
  TypeScript: 5.2.2
      Rollup: 2.42.3
      Parse5: 7.1.2
      Sizzle: 2.42.3
      Terser: 5.24.0

This has also been tested in different Linux environments (Ubuntu+Alpine).
```

### Steps to Reproduce

- Install `@stencil/sass` and use it as a plugin in your `stencil.config.ts`
- Make sure to use a `.scss` file for your component
- Add a property and mark it as internal using this comment above: the property `/** @internal */`
- Build the project multiple times and check whether the `types` directory is present in the `dist` folder (make sure to delete it before re-building, as Stencil does _not_ clear previous build results). For more convencience, a Python script has been added to the repository which performs the build and check automatically: `python loop.py 10` (runs the build 10 times and terminates if the types folder is missing).

### Code Reproduction URL

https://github.com/niklas2810/stencil-type-issue

### Additional Information

I know that this error seems rather cryptic, hopefully the reproduction repo can help with that. If you need more information, please let me know!

Best regards
Niklas

christian-bromann:
fixes #5091 / STENCIL-1024

Given the following Stencil config:

```ts
  namespace: 'stencil-type-issue',
  outputTargets: [
    {
      type: 'dist',
      esmLoaderPath: '../loader',
    },
    {
      type: 'dist-custom-elements',
    },
  ],
  plugins: [
    sass()
  ],
  testing: {
    browserHeadless: "new",
  },
};
```

running a Stencil build causes the types directory `/dist/types` not being generated.

My investigations were as follows:
- having `{ type: 'dist-custom-elements' }` entry in `outputTargets` seems to cause the type generation to run twice with the same configurations
- this causes [`generateTypes`](https://github.com/ionic-team/stencil/blob/2b0727efdf6062445655725ad6923fb2cad4eaab/src/compiler/types/generate-types.ts#L15) to be called twice as well as the command to write [`/dist/types/components.d.ts`](https://github.com/ionic-team/stencil/blob/5d8387f0a1c9f860151d133598dabb22e2b1f911/src/compiler/types/generate-app-types.ts#L46-L48) with the option [`immediateWrite: true`](https://github.com/ionic-team/stencil/blob/5d8387f0a1c9f860151d133598dabb22e2b1f911/src/compiler/types/generate-app-types.ts#L47)
- adding a "sleep" right after has shown that actually nothing is being written to file even though `fs.writeFile` is called in [`/src/sys/node/node-sys.ts#L612-L614`](https://github.com/ionic-team/stencil/blob/4959295c24ec3effcc8d63a8305dffd6e07a617d/src/sys/node/node-sys.ts#L612-L614)

GitHub Issue Number: #5091

## What is the new behavior?

As seen in the change-set it turns out that we didn't handle the error provided by the `sys.writeFile`. Essentially the call to create the file failed without handling its error. Further investigations have shown that the directory for the file didn't exist even though we were calling [`ensureDir`](https://github.com/ionic-team/stencil/blob/db01eb840602f4c023e488aaf266795cee2dec57/src/compiler/sys/in-memory-fs.ts#L735) before. The race condition originates from the question how fast other output items were able to generate their types (I believe).

My suggestion to resolve this issue is to use the `recursive` flag for creating the directory when we are not working within memory. I am afraid calling `mkdir` multiple times, especially on `graceful-fs` rather than the Node.js package must have caused weird side effects in the event loop. During the investigations I literally was able to resolve it by adding a simple `console.log`.

An alternative solution would be to remove the `immediateWrite: true` option and have it write to the filesystem when everything file operation is being flushed which is likely a point in time the directory has been created already. However since  we are calling `generateAppTypes` also in [`runTsProgram`](https://github.com/ionic-team/stencil/blob/2a75b6501f4f76dad0d8fa8304af57be1c04eef1/src/compiler/transpile/run-program.ts#L111) I assume it is important to keep this option so rewrites are happening on watch mode!?

- [ ] Yes
- [x] No

## Testing

See Jira ticket, apply the patch provided and reproduce the issue.

To test this change:

- checkout the branch
- build the project `npm run build`
- set your local Stencil dev environment as dependency in the reproduction project, e.g. `"@stencil/core": "file:/Users/christian.bromann/Sites/Ionic/projects/stencil"`
- run `./loop.sh` and see that the `types` dir is created every time

### DIFF
--- a/src/declarations/stencil-public-compiler.ts
+++ b/src/declarations/stencil-public-compiler.ts
@@ -1562,7 +1562,7 @@ export interface CompilerSystemRemoveFileResults {
 
 export interface CompilerSystemWriteFileResults {
   path: string;
-  error: Error;
+  error: any;
 }
 
 export interface Credentials {
--- a/src/declarations/stencil-public-compiler.ts
+++ b/src/declarations/stencil-public-compiler.ts
@@ -1562,7 +1562,7 @@ export interface CompilerSystemRemoveFileResults {
 
 export interface CompilerSystemWriteFileResults {
   path: string;
-  error: any;
+  error: Error;
 }
 
 export interface Credentials {
--- a/src/compiler/sys/in-memory-fs.ts
+++ b/src/compiler/sys/in-memory-fs.ts
@@ -66,9 +66,18 @@ export type FsItems = Map<string, FsItem>;
  * Options supported by write methods on the in-memory filesystem.
  */
 export interface FsWriteOptions {
+  /**
+   * only use the in-memory cache and do not write the file to disk
+   */
   inMemoryOnly?: boolean;
   clearFileCache?: boolean;
+  /**
+   * flush the write to disk immediately, skipping the in-memory cache
+   */
   immediateWrite?: boolean;
+  /**
+   * specify that the cache should be used
+   */
   useCache?: boolean;
   /**
    * An optional tag for the current output target for which this file is being
--- a/src/compiler/sys/in-memory-fs.ts
+++ b/src/compiler/sys/in-memory-fs.ts
@@ -733,7 +733,10 @@ export const createInMemoryFs = (sys: d.CompilerSystem) => {
 
         if (results.changedContent) {
           await ensureDir(filePath, false);
-          await sys.writeFile(filePath, item.fileText);
+          const { error } = await sys.writeFile(filePath, item.fileText);
+          if (error) {
+            throw error;
+          }
         }
       }
     } else {
--- a/src/compiler/sys/in-memory-fs.ts
+++ b/src/compiler/sys/in-memory-fs.ts
@@ -827,6 +827,15 @@ export const createInMemoryFs = (sys: d.CompilerSystem) => {
    * only change the in-memory cache
    */
   const ensureDir = async (path: string, inMemoryOnly: boolean) => {
+    /**
+     * in case we write to disk immediately, there is no need to split
+     * all directories into separate calls, we can just use the recursive flag
+     */
+    if (!inMemoryOnly) {
+      await sys.createDir(dirname(path), { recursive: true });
+      return;
+    }
+
     const allDirs: string[] = [];
 
     while (true) {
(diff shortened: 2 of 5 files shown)


tanner-reits (approved):
Tested this out and it works! Ran the script a few times and never had an instance where it didn't generate the types.

I'm not sure what the consequences of removing `immediateWrite` would be. The call to `generateAppTypes` in `runTsProgram` is only responsible for generating the `src/components.d.ts` file. Not sure if that matters 🤷‍♂️

src/declarations/stencil-public-compiler.ts
```diff
@@ -1557,7 +1557,7 @@ export interface CompilerSystemRemoveFileResults {
 
 export interface CompilerSystemWriteFileResults {
   path: string;
-  error: any;
+  error: Error;
```
rwaskiewicz, line 1560: Can we revert this + add a TODO(STENCIL-NUM) comment here? Technically, I think this would be a breaking change to the public Node API since we're narrowing this type.

christian-bromann: Good catch, I reverted the change.

rwaskiewicz (approved):

christian-bromann:

Issue #5113, opened 2023-11-29
bug: `E2EPage.find()` throws error when using piercing selector targeting a non-existing element


jcfranco:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

4.8.0

### Current Behavior

Calling `find` with a piercing selector where the target does not exist throws an error instead of returning `null` as the doc states.

### Expected Behavior

For `find()` to return `null` for shadow DOM elements that are not found.

### System Info

```shell
[36m      System:[39m node 18.16.0
[36m    Platform:[39m darwin (23.1.0)
[36m   CPU Model:[39m Apple M1 Pro (10 cpus)
[36m    Compiler:[39m /Users/.../stencil-component-starter/node_modules/@stencil/core/compiler/stencil.js
[36m       Build:[39m 1701098446
[36m     Stencil:[39m 4.8.0 🌞
[36m  TypeScript:[39m 5.2.2
[36m      Rollup:[39m 2.42.3
[36m      Parse5:[39m 7.1.2
[36m      Sizzle:[39m 2.42.3
[36m      Terser:[39m 5.24.0
```

### Steps to Reproduce

See repro case readme.

### Code Reproduction URL

https://github.com/jcfranco/stencil-4.8.0-breaks-piercing-selector-for-non-existent-elements

### Additional Information

4.7.2 does not throw an error and works as expected.

tanner-reits:
@jcfranco Thanks for letting us know! I was able to confirm the regression from v4.7.2. We'll work on getting this resolved ASAP!

tanner-reits:
The fix for this was included as a part of today's [v4.8.1 release](https://github.com/ionic-team/stencil/releases/tag/v4.8.1)!

jcfranco:
@tanner-reits Thanks for fixing!

PR #5114, opened 2023-11-30
fix(testing): re-add Puppeteer `asElement()` calls

tanner-reits:

Fixes a regression caused by the removal of a few `asElement()` calls in 91cfa2c9adb50f27c823653511a4913d320398b1. Turns out these are necessary for the case where the handle is `null`. To avoid type errors, we cast the returned value to be of type `ElementHandle<Element>` rather than `ElementHandle<Node`.

Fixes: #5113

## What is the new behavior?

No more errors 🙂

- [ ] Yes
- [X] No

## Testing

1. Open the repro case in the linked issue
2. Run `npm run test` and observe a failure on the second call to `page.find` in the e2e test
3. Install a build of this branch
4. Run `npm run test` and observe no more errors and tests pass!

### DIFF
--- a/src/testing/puppeteer/puppeteer-element.ts
+++ b/src/testing/puppeteer/puppeteer-element.ts
@@ -593,7 +593,7 @@ async function findWithCssSelector(
       return null;
     }
 
-    elmHandle = shadowHandle;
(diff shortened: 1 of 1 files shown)


christian-bromann (approved):
👍

PR #5120, opened 2023-12-01
fix(runtime): relocate slotted content when slot parent element tag changes

tanner-reits:

Stencil can re-render incorrectly when content is slotted in from the "implementation layer" (i.e. from `index.html`) and a `slot`s parent element tag changes (e.g. from `p` to `span`). This is only an issue when using slots outside the Shadow DOM. 

The issue is caused by our vDOM algorithm relocating slotted content back to its _original_ location when it determines that a child node has changed. When this happens, we call `putBackInOriginalLocation()` which will relocate a node back to where it was original defined. But, this can result in the node no being re-relocated if it was put back into a location inside another Stencil element that has already gone through its render cycle. In this case, the node will stay in its original location rather than being relocated through the DOM elements again.

Fixes: #4284 #3913 

## What is the new behavior?

Instead of using `putBackInOriginalLocation()` when we identify changed nodes, this PR introduces a new method, `relocateToHostRoot()`. This method will move all nodes that were slotted into the changed node to the root of the element currently being rendered. So, when we enter the `renderVdom()` method for the component, it will be able to relocate any slotted content back to its new location.

- [ ] Yes
- [X] No

## Testing

**Manual testing**
Two test cases can be reproduced using the repros from the linked issues and a build of this branch. These cases are also covered by e2e test.

A third test case can be created similar to the following (testing same behavior as first two, just more complex):
1. Create a new Stencil component starter
2. Enable `extras.experimentalSlotFixes` in the Stencil config
3. Create the following three components:
```tsx
@Component({
  tag: 'super-cmp',
  shadow: false,
  scoped: false,
})
export class SuperCmp implements ComponentInterface {
  @Prop() tag: any = 'p';

  render(): HTMLElement {
    return (
      <Host>
        <slot name="tmp"></slot>
        <my-cmp tag={this.tag}>
          <p>YOOO</p>
          <slot></slot>
        </my-cmp>
      </Host>
    );
  }
}
```

```tsx
@Component({
  tag: 'my-cmp',
  shadow: false,
  scoped: false,
})
export class MyCmp implements ComponentInterface {
  @Prop() tag: any = 'p';

  render(): HTMLElement {
    return (
      <Host>
        <base-cmp tag={this.tag}>
          <span>HOWDY</span>
          <slot></slot>
        </base-cmp>
      </Host>
    );
  }
}
```

```tsx
@Component({
  tag: 'base-cmp',
})
export class BaseCmp implements ComponentInterface {
  @Prop() tag: any = 'p';

  render() {
    return (
      <Host>
        <this.tag>
          <slot />
        </this.tag>
      </Host>
    );
  }
}
```
4. Update the `src/index.html` to have the following `body`
```html
<super-cmp>
  <p slot="tmp">Test</p>
  Hello
</super-cmp>
<button type="button">Click</button>

<script>
    document.querySelector('super-cmp').setAttribute('tag', 'span');
  });
</script>
```
5. The components will render correctly initially, but "Hello" will be in the incorrect location after clicking the button (should be in the `span` tag in the `base-cmp` component.
6. Install a build of this branch into the starter.
7. Run the project. The "hello" test will no relocate to the `span` tag after clicking the button. All "statically slotted" content will also render in the expected order ("yooo", "howdy", "hello")

**Automated testing**
All existing unit & e2e tests continue to pass. Also added new e2e tests for the reproduction cases in the linked issues.

## Other information

Only available with the `experimentalSlotFixes` extra flag enabled.

### DIFF
--- a/src/runtime/vdom/vdom-render.ts
+++ b/src/runtime/vdom/vdom-render.ts
@@ -172,8 +172,8 @@ const createElm = (oldParentVNode: d.VNode, newParentVNode: d.VNode, childIndex:
 
 /**
  * Relocates all child nodes of an element that were a part of a previous slot relocation
- * to the root of the Stencil component currently being rendered. This happens when a the parent
- * element of a slot reference node is dynamically changes and triggers a re-render. We cannot use
+ * to the root of the Stencil component currently being rendered. This happens when a parent
+ * element of a slot reference node dynamically changes and triggers a re-render. We cannot use
  * `putBackInOriginalLocation()` because that may relocate nodes to elements that will not be re-rendered
  * and so they will not be relocated again.
  *
--- a/src/runtime/vdom/vdom-render.ts
+++ b/src/runtime/vdom/vdom-render.ts
@@ -155,13 +155,38 @@ const createElm = (oldParentVNode: d.VNode, newParentVNode: d.VNode, childIndex:
       oldVNode = oldParentVNode && oldParentVNode.$children$ && oldParentVNode.$children$[childIndex];
       if (oldVNode && oldVNode.$tag$ === newVNode.$tag$ && oldParentVNode.$elm$) {
         // we've got an old slot vnode and the wrapper is being replaced
-        // so let's move the old slot content back to it's original location
-        putBackInOriginalLocation(oldParentVNode.$elm$, false);
+        // so let's move the old slot content to the root of the element currently being rendered
+        relocateToHostRoot(oldParentVNode.$elm$);
       }
     }
   }
 
   return elm;
+};
+
+const relocateToHostRoot = (parentElm: Element) => {
+  plt.$flags$ |= PLATFORM_FLAGS.isTmpDisconnected;
+
+  const host = parentElm.closest(hostTagName.toLowerCase());
+  for (const childNode of Array.from(parentElm.childNodes) as d.RenderNode[]) {
+    // Only relocate nodes that were slotted in
+    if (childNode['s-sh'] != null) {
+      host.insertBefore(childNode, null);
+      // Reset so we can correctly move the node around again.
+      childNode['s-sh'] = undefined;
+
+      // When putting an element node back in its original location,
+      // we need to reset the `slot` attribute back to the value it originally had
+      // so we can correctly relocate it again in the future
+      if (childNode.nodeType === NODE_TYPE.ElementNode && !!childNode['s-sn']) {
+        childNode.setAttribute('slot', childNode['s-sn']);
+      }
+
+      checkSlotRelocate = true;
+    }
+  }
+
+  plt.$flags$ &= ~PLATFORM_FLAGS.isTmpDisconnected;
 };
 
 const putBackInOriginalLocation = (parentElm: Node, recursive: boolean) => {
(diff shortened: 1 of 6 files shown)


src/runtime/vdom/vdom-render.ts
```diff
@@ -154,27 +154,64 @@ const createElm = (oldParentVNode: d.VNode, newParentVNode: d.VNode, childIndex:
...
 const putBackInOriginalLocation = (parentElm: Node, recursive: boolean) => {
   plt.$flags$ |= PLATFORM_FLAGS.isTmpDisconnected;
 
   const oldSlotChildNodes = parentElm.childNodes;
   for (let i = oldSlotChildNodes.length - 1; i >= 0; i--) {
     const childNode = oldSlotChildNodes[i] as any;
     if (childNode['s-hn'] !== hostTagName && childNode['s-ol']) {
-      // // this child node in the old element is from another component
-      // // remove this node from the old slot's parent
-      // childNode.remove();
```
tanner-reits, line 176: This chunk has been commented out for a while. I've tried uncommenting it a few times in different issues and haven't noticed any change in behavior so figured we just get rid of it to avoid future confusion

src/runtime/vdom/vdom-render.ts
```diff
@@ -1027,7 +1064,7 @@ render() {
             // has a different next sibling or parent relocated
 
             if (nodeToRelocate !== insertBeforeNode) {
-              if (!nodeToRelocate['s-hn'] && nodeToRelocate['s-ol']) {
+              if (!BUILD.experimentalSlotFixes && !nodeToRelocate['s-hn'] && nodeToRelocate['s-ol']) {
```
tanner-reits, line 1067: Idea is that this whole block will go away once the slot fix flag is defaulted. Setting `s-hn` for elements from `index.html` actually causes some weird behavior when putting nodes back since it won't relocate a node to the same host node (which we actually want in some scenarios). 

I didn't see any negative consequences from removing this while playing around with things, and all our automated tests still pass so feel like we're okay to do this.

tanner-reits (commented):
Couple notes

src/runtime/vdom/vdom-render.ts
rwaskiewicz: I'm in favor of this 👍

src/runtime/vdom/vdom-render.ts
```diff
@@ -154,27 +154,66 @@ const createElm = (oldParentVNode: d.VNode, newParentVNode: d.VNode, childIndex:
...
     }
   }
 
   return elm;
 };
 
+/**
+ * Relocates all child nodes of an element that were a part of a previous slot relocation
+ * to the root of the Stencil component currently being rendered. This happens when a the parent
+ * element of a slot reference node is dynamically changes and triggers a re-render. We cannot use
```
rwaskiewicz, line 176: Non-blocking nit
```suggestion
 * element of a slot reference node dynamically changes and triggers a re-render. We cannot use
```

rwaskiewicz (approved):
2 non-blocking nits

christian-bromann:
@tanner-reits I was able to verify the expected behavior after adding `experimentalSlotFixes: true`, thanks! 👍

christian-bromann:

PR #5121, opened 2023-12-01
chore(mock-doc): replace Sizzle with JQuery Selector Engine

christian-bromann:

We have been using Sizzle for a long time which is about to be discontinued and got merged into the jQuery project.

STENCIL-1039: Implement New Sizzle Infra

## What is the new behavior?

Replace Sizzle with a custom JQuery build that only bundles the selector engine.

- [ ] Yes
- [x] No

## Testing

I believe the unit tests are a good indicator if everything is working as expected.

## Other information

Instead of checking in a custom jQuery build we can also just run the script every time we build the project. I don't believe though there are a lot of changes to the selector engine happening, therefor we keep a better DX to just check the custom build in and only re-run the script if we want to do an update.

__Note:__
- I experience some heap stack issues in the build, investigate the root cause and work on a fix


christian-bromann:
I almost gave up on this as I faced a `RegExpCompiler Allocation failed - process out of memory` error when loading the file. Luckily I found the set of changes necessary to make the code work. The PR introduces a new script that we can run whenever we want to update the selector engine:

```sh
npm run build.updateSelectorEngine
```

The script will create a fresh jQuery bundle that contains the selector engine. It will also to some minor tweaks to the code to mitigate the RegExpCompiler Allocation error and writes the result to `src/mock-doc/third-party/jquery.ts` where it is imported from `src/mock-doc/selector.ts`. No need a rollup plugin anymore as we check in the file directly given that we don't expect any changes that often.

christian-bromann:
Still seeing an issue with `createWindowFromHtml` not being defined which doesn't seem to be connected to Sizzle at all 👀

christian-bromann:
Resolved the remaining issue, this is ready for review now.

alicewriteswrongs (changes requested):
code looks good, and I built locally and installed in Framework and was able to run the spec tests with no problems, so no concerns there.

just had a few comments and questions!

scripts/updateSelectorEngine.ts
```diff
@@ -0,0 +1,107 @@
...
+  console.log(`\nJQuery Selector engine updated to version ${jqueryPkgJSON.version}`);
+  console.log(`at ${thirdPartyDir} 🎉`);
+}
+
+function runCommand(cmd: string, cwd: string) {
+  return new Promise((resolve, reject) => {
+    console.log(`> ${cmd}`);
+    const child = cp.spawn(cmd, { cwd, shell: true });
+    child.stdout.pipe(process.stdout);
+    child.stderr.pipe(process.stderr);
```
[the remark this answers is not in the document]
christian-bromann: You are right, these two line can be removed as `stdio: 'pipe'` is the default setting which does exactly that.

src/cli/task-info.ts
```diff
@@ -25,7 +25,7 @@ export const taskInfo = (coreCompiler: CoreCompiler, sys: CompilerSystem, logger
   console.log(`${logger.cyan('  TypeScript:')} ${versions.typescript}`);
   console.log(`${logger.cyan('      Rollup:')} ${versions.rollup}`);
   console.log(`${logger.cyan('      Parse5:')} ${versions.parse5}`);
-  console.log(`${logger.cyan('      Sizzle:')} ${versions.sizzle}`);
+  console.log(`${logger.cyan('      Sizzle:')} ${versions.jquery}`);
```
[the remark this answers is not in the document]
christian-bromann: I will update to 
```suggestion
  console.log(`${logger.cyan('      jQuery:')} ${versions.jquery}`);
```
at this point there is no sizzle in this anymore.

alicewriteswrongs (approved):

package.json
```diff
@@ -104,6 +104,7 @@
     "jest": "^27.4.5",
     "jest-cli": "^27.4.5",
     "jest-environment-node": "^27.4.4",
+    "jquery": "https://github.com/jquery/jquery/tarball/c98597eaf5e144ee5e549cb41984687cd1033068",
```
tanner-reits, line 107: For my knowledge, can you explain what this is pulling and why we're pulling direct from Github?

scripts/updateSelectorEngine.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  const originalContent = await fs.readFile(path.resolve(jqueryDepDir, 'dist', 'jquery.js'), 'utf8');
+
+  /**
+   * This is a hack to make the jQuery selector engine work with the mock-doc package.
+   * Using the original jQuery bundle would cause a "RegExpCompiler Allocation failed - process out of memory"
+   * error due to the way the jQuery object is constructed. The following tweaks to the jQuery bundle
+   * resolve the issue:
+   */
+  const fixedJQuery = originalContent
```
tanner-reits, line 49: Changes like this give me hesitation, but as long as the plan is to never really change which version of jQuery we pull in, then I think we'll be okay.

tanner-reits (approved):

scripts/updateSelectorEngine.ts
christian-bromann: This is mostly to avoid the memory allocation issue I faced as there are tons of custom created Regular Expressions in there. This is reported in Chromium so chances are that eventually we can remove this and Node.js can handle this better. Given our Q1 roadmap it is more likely though that we nudge users to test on a different infrastructure other than our `mock-doc`, e.g. in a real headless browser, which also will remove the need for this dependency in general.

PR #5135, opened 2023-12-06
fix(runtime): hide slotted content with no destination in scoped components

tanner-reits:

Two current behaviors this solves:
1. If a `slot` is dynamically rendered in a component using `scoped` or no encapsulation and is not rendered initially, the content that would go into this slot is marked as `hidden`. But, if the slot was rendered on a subsequent render, we would not remove the `hidden` attribute, so the content would still be invisible in the DOM.
2. Content projected through to a component would not be hidden if it did not have a destination `slot` like how a component with a Shadow Root would hide any content in the light DOM not mapped to a slot. 

Fixes: #2877 

## What is the new behavior?

Solutions in order of the issues listed above:
1. We now track an "initially hidden" state for an element. This value is used to "reset" the `hidden` attribute when content is successfully relocated to a "slot" location. This accounts for if a user manually set the `hidden` attribute on an element so it would remain hidden after relocation. There _may_ be issues with this value getting programmatically changed throughout a component's lifetime and reset to an incorrect value. We can address this in the future as needed.
2. Leveraging the initial hidden state from #1, we now hide any content that is projected through to a `scoped` component that does not end up in a slot. If an appropriate "slot" is dynamically rendered later, the content will then be relocated and displayed accordingly. 

- [ ] Yes
- [X] No

## Testing

**Manual testing**
1. Spin up a new Stencil component library
2. Enable `experimentalSlotFixes` in the `extras` object of the Stencil config
3. Update the project's `my-component.tsx` and `index.html` files as follows:
```tsx
// my-component.tsx

@Component({
  tag: 'my-component',
  shadow: false,
  scoped: true,
})
export class MyComponent {
  @Prop() enabled = false;

  render() {
    return (
      <Host>
        <p>Test</p>
        {this.enabled && (
          <div data-txt="slot wrapper">
            <slot>
              <span>fallback default slot</span>
            </slot>
          </div>
        )}
      </Host>
    );
  }
}
```

```html
// index.html

<body>
    <my-component>
      <div>
        <span>outside default slot</span>
      </div>
    </my-component>

    <button type="button">Click</button>
</body>

<script>
      document.querySelector('my-component').setAttribute('enabled', true);
    });
</script>
```
4. Run the project and confirm that the `div` projected through from the `index.html` file _does_ appear in the DOM despite the `slot` not being rendered.
5. Install a build of this branch
6. Re-run and observe the `div` is not visible until after clicking the button (which dynamically renders the slot).

**Automated testing**
Existing unit & e2e tests continue to pass. Added e2e tests for the reproduction case in the associated issue.

## Other information

Only available with the `experimentalSlotFixes` extra flag enabled in a project's Stencil config.

### DIFF
--- a/src/runtime/vdom/vdom-render.ts
+++ b/src/runtime/vdom/vdom-render.ts
@@ -135,7 +135,7 @@ const createElm = (oldParentVNode: d.VNode, newParentVNode: d.VNode, childIndex:
     }
   }
 
-  // This needs to always happen so we can hide nodes that are projected through
+  // This needs to always happen so we can hide nodes that are projected
   // to another component but don't end up in a slot
   elm['s-hn'] = hostTagName;
   if (BUILD.slotRelocation) {
--- a/src/runtime/vdom/vdom-render.ts
+++ b/src/runtime/vdom/vdom-render.ts
@@ -155,13 +155,38 @@ const createElm = (oldParentVNode: d.VNode, newParentVNode: d.VNode, childIndex:
       oldVNode = oldParentVNode && oldParentVNode.$children$ && oldParentVNode.$children$[childIndex];
       if (oldVNode && oldVNode.$tag$ === newVNode.$tag$ && oldParentVNode.$elm$) {
         // we've got an old slot vnode and the wrapper is being replaced
-        // so let's move the old slot content back to it's original location
-        putBackInOriginalLocation(oldParentVNode.$elm$, false);
+        // so let's move the old slot content to the root of the element currently being rendered
+        relocateToHostRoot(oldParentVNode.$elm$);
       }
     }
   }
 
   return elm;
+};
+
+const relocateToHostRoot = (parentElm: Element) => {
+  plt.$flags$ |= PLATFORM_FLAGS.isTmpDisconnected;
+
+  const host = parentElm.closest(hostTagName.toLowerCase());
+  for (const childNode of Array.from(parentElm.childNodes) as d.RenderNode[]) {
+    // Only relocate nodes that were slotted in
+    if (childNode['s-sh'] != null) {
+      host.insertBefore(childNode, null);
+      // Reset so we can correctly move the node around again.
+      childNode['s-sh'] = undefined;
+
+      // When putting an element node back in its original location,
+      // we need to reset the `slot` attribute back to the value it originally had
+      // so we can correctly relocate it again in the future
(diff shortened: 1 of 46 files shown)


src/runtime/vdom/vdom-render.ts
```diff
@@ -1096,11 +1097,25 @@ render() {
...
+              }
             }
           }
         } else {
           // this node doesn't have a slot home to go to, so let's hide it
           if (nodeToRelocate.nodeType === NODE_TYPE.ElementNode) {
+            // Store the initial value of `hidden` so we can reset it later when
+            // moving nodes around.
+            if (isInitialLoad) {
+              nodeToRelocate['s-ih'] = nodeToRelocate.hidden ?? false;
```
tanner-reits, line 1116: It _might_ be possible to run into a situation where an element is programmatically hidden and we end up storing the incorrect state since this only captures the state on initial render. I figured that can be something we deal with as needed since the functionality is currently broken anyway so this is a step in the right direction

tanner-reits (commented):
Called out one thing

src/runtime/vdom/vdom-render.ts
```diff
@@ -135,9 +135,10 @@ const createElm = (oldParentVNode: d.VNode, newParentVNode: d.VNode, childIndex:
     }
   }
 
+  // This needs to always happen so we can hide nodes that are projected through
+  // to another component but don't end up in a slot
```
christian-bromann, line 139: nit, non-blocking suggestion
```suggestion
  // This needs to always happen so we can hide nodes that are projected
  // to another component but don't end up in a slot
```

rwaskiewicz:
@tanner-reits I have a quick clarifying questions regarding the manual testing steps:

1. The component has the tag `my-component`, but the query selector in `index.html` references a "pui-example" - I'm guessing we can replace "pui-example" for "my-component" here?
2. Does this fix the issue for all encapsulation models? In the manual testing example, we're using `shadow: true`, which differs from the karma tests in this PR

tanner-reits:
@rwaskiewicz Sorry bout that. 

1. The query in the steps should be 'my-component', that was a copy-pasta error when renaming things from the reproduction repo I was using locally. 
2. This change only modifies `scoped`. Which looks like that's the encapsulation mode I have in the testing steps so should be good there

rwaskiewicz:
@tanner-reits my bad on question 2 - IDK what I was looking at 😓

christian-bromann:

PR #5144, opened 2023-12-08
fix(compiler): make style getter behavior consistent between default and native components

christian-bromann:
Currently our nightly build is broken due changes introduced in #5131 (see https://github.com/ionic-team/ionic-framework/actions/runs/7124408050). Investigating this issue was difficult as it happens to be random behavior. However I found that the static style property wasn't applied correctly in the failing cases which gave me a hint to the location of this fix.

## What is the new behavior?
I made changes in the order of if statements in [`getSingleStyle`](https://github.com/ionic-team/stencil/blob/735d45afdda420420f6d3992662cb63ded2c937e/src/compiler/transformers/add-static-style.ts#L103)  and [`getMultipleModeStyle`](https://github.com/ionic-team/stencil/blob/735d45afdda420420f6d3992662cb63ded2c937e/src/compiler/transformers/add-static-style.ts#L69) 
 and have recognised a file with similar implementation features: `/src/compiler/transformers/component-native/native-static-style.ts`. Afaik this is done if `transformOpts.componentExport` is set to `customelement`.

It turns out that after aligning the order of if statements in `addSingleStyleGetter` with how it is in `getSingleStyle` the component was rendered correctly in a consistent fashion.

- [ ] Yes
- [ ] No

## Testing

Added some tests but you can reproduce the issue by:

- create a stencil project `npm init stencil`
- change the stencil dependency to be nightly in `package.json`: `"@stencil/core": "nightly"`
- `npm install`
- add another output target:
   ```
    {
      type: 'dist-custom-elements',
      dir: 'components',
    },
   ```
- go in `components/my-component.js` and verify that `MyComponentStyle` is not defined

No to verify the fix pack the build of this branch and link it to that project. After rebuilding the component you should see a `myComponentCss` defined.

### DIFF
--- a/src/compiler/transformers/component-native/native-static-style.ts
+++ b/src/compiler/transformers/component-native/native-static-style.ts
@@ -33,13 +33,6 @@ const addMultipleModeStyleGetter = (
       const styleLiteral = createStyleLiteral(cmp, style);
       const propStr = ts.factory.createPropertyAssignment(style.modeName, styleLiteral);
       styleModes.push(propStr);
-    } else if (typeof style.styleIdentifier === 'string') {
-      // direct import already written in the source code
-      // import myTagIosStyle from './import-path.css';
-      // static get style() { return { "ios": myTagIosStyle }; }
-      const styleIdentifier = ts.factory.createIdentifier(style.styleIdentifier);
-      const propIdentifier = ts.factory.createPropertyAssignment(style.modeName, styleIdentifier);
-      styleModes.push(propIdentifier);
     } else if (Array.isArray(style.externalStyles) && style.externalStyles.length > 0) {
       // import generated from @Component() styleUrls option
       // import myTagIosStyle from './import-path.css';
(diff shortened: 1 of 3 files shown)


src/compiler/transformers/component-native/native-static-style.ts
```diff
@@ -33,20 +33,20 @@ const addMultipleModeStyleGetter = (
       const styleLiteral = createStyleLiteral(cmp, style);
```
rwaskiewicz, line 33: GH won't allow me to comment on a line higher than this one - can we add a comment before this conditional starts (above the `if` statement) that the order of the conditional must match `add-static-style.ts#getSingleStyle`?

src/compiler/transformers/component-native/native-static-style.ts
```diff
@@ -65,18 +65,18 @@ const addSingleStyleGetter = (
     // static get style() { return "string"; }
```
rwaskiewicz, line 65: GH won't allow me to comment on a line higher than this one - can we add a comment before this conditional starts (above the `if` statement) that the order of the conditional must match `add-static-style.ts#getSingleStyle`?

src/compiler/transformers/component-native/native-static-style.ts
alicewriteswrongs: on this point, I was thinking it might make sense to circle back after the bugfix is out and factor out this logic into a helper function since it looks essentially the same in these two spots

alicewriteswrongs (approved):
I tried out the fix locally and it looks good!

tanner-reits (approved):

christian-bromann:
@rwaskiewicz added comments, can I go ahead and merge?

christian-bromann:
Got a 👍 through our internal communication platform. Going ahead and merge.

rwaskiewicz (approved):

PR #5146, opened 2023-12-08
fix(runtime): update `textContent` patch to mimic Shadow Root

tanner-reits:

Stencil's current `textContent` patch for `scoped` elements only deals with a component's default slot. However, the team has recently agreed that our `scoped` component behavior should mimic a `shadow` component as closely as possible.

Fixes: #3977 

## What is the new behavior?

The `textContent` patch now behaves like the native implementation for an element with a Shadow Root. The getter will return the combined `textContent` of all child nodes that are located in slot references. And the setter will replace all content within slot reference nodes with a single text node in the default slot reference (if one exists).

- [ ] Yes
- [X] No

## Testing

**Manual testing**
1. Create new component starter
2. Enable `extras.experimentalSlotFixes` in Stencil config
3. Update `my-component` to the following:
```tsx
@Component({
  tag: 'my-component',
  styleUrl: 'my-component.css',
  scoped: true,
})
export class MyComponent {
  render(): HTMLElement {
    return (
      <Host>
        <button>
          <div class="content-wrapper">
            <span class="prefix">
              <slot name="prefix" />
            </span>
            <slot />
            <slot name="suffix" />
          </div>
        </button>
      </Host>
    );
  }
}
```
4. Update `index.html` to have the following `body`:
```html
          <my-component>
            <span slot="prefix">Prefix</span>
            <p>Button</p>
            <span slot="suffix">Suffix</span>
          </my-component>
```
5. Run `npm start`
6. Execute the `textContent` getter/setter on the element from browser dev tools. Notice that it will only ever change the "Button" node in the default slot
7. Install a build of this branch
8. Execute the `textContent` getter/setter on the element from browser dev tools. Notice that it will now return the value of all slotted content and will replace all nodes when using the setter

**Automated testing**
Update a few existing tests to account for whitespace differences in return value. Added new e2e tests to verify behavior for `scoped` components.

## Other information

Only available with the `experimentalSlotFixes` extra flag enabled in a project's Stencil config.

This **does not** include a patch for `innerText`. That method is trickier to patch to correctly get whitespace and line break formatting.

### DIFF
--- a/src/runtime/bootstrap-custom-element.ts
+++ b/src/runtime/bootstrap-custom-element.ts
@@ -44,7 +44,7 @@ export const proxyCustomElement = (Cstr: any, compactMeta: d.ComponentRuntimeMet
   }
 
   // TODO(STENCIL-914): this check and `else` block can go away and be replaced by just the `scoped` check
-  if (BUILD.experimentalSlotFixes && cmpMeta.$flags$ & CMP_FLAGS.scopedCssEncapsulation) {
+  if (BUILD.experimentalSlotFixes && BUILD.scoped && cmpMeta.$flags$ & CMP_FLAGS.scopedCssEncapsulation) {
     patchPseudoShadowDom(Cstr.prototype, cmpMeta);
   } else {
     if (BUILD.slotChildNodesFix) {
--- a/src/runtime/bootstrap-custom-element.ts
+++ b/src/runtime/bootstrap-custom-element.ts
@@ -43,9 +43,8 @@ export const proxyCustomElement = (Cstr: any, compactMeta: d.ComponentRuntimeMet
     cmpMeta.$flags$ |= CMP_FLAGS.needsShadowDomShim;
   }
 
-  // TODO(STENCIL-914): this check and `else` block can go away and be replaced by just `BUILD.scoped` once we
-  // default our pseudo-slot behavior
-  if (BUILD.experimentalSlotFixes && BUILD.scoped) {
+  // TODO(STENCIL-914): this check and `else` block can go away and be replaced by just the `scoped` check
+  if (BUILD.experimentalSlotFixes && cmpMeta.$flags$ & CMP_FLAGS.scopedCssEncapsulation) {
     patchPseudoShadowDom(Cstr.prototype, cmpMeta);
   } else {
     if (BUILD.slotChildNodesFix) {
@@ -57,8 +56,8 @@ export const proxyCustomElement = (Cstr: any, compactMeta: d.ComponentRuntimeMet
     if (BUILD.appendChildSlotFix) {
       patchSlotAppendChild(Cstr.prototype);
     }
-    if (BUILD.scopedSlotTextContentFix) {
-      patchTextContent(Cstr.prototype, cmpMeta);
+    if (BUILD.scopedSlotTextContentFix && cmpMeta.$flags$ & CMP_FLAGS.scopedCssEncapsulation) {
+      patchTextContent(Cstr.prototype);
     }
   }
 
--- a/src/runtime/bootstrap-lazy.ts
+++ b/src/runtime/bootstrap-lazy.ts
@@ -146,9 +146,8 @@ export const bootstrapLazy = (lazyBundles: d.LazyBundlesRuntimeData, options: d.
         }
       };
 
-      // TODO(STENCIL-914): this check and `else` block can go away and be replaced by just `BUILD.scoped` once we
-      // default our pseudo-slot behavior
-      if (BUILD.experimentalSlotFixes && BUILD.scoped) {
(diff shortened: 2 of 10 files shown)


src/runtime/bootstrap-custom-element.ts
```diff
@@ -57,8 +56,8 @@ export const proxyCustomElement = (Cstr: any, compactMeta: d.ComponentRuntimeMet
     if (BUILD.appendChildSlotFix) {
       patchSlotAppendChild(Cstr.prototype);
     }
-    if (BUILD.scopedSlotTextContentFix) {
-      patchTextContent(Cstr.prototype, cmpMeta);
+    if (BUILD.scopedSlotTextContentFix && cmpMeta.$flags$ & CMP_FLAGS.scopedCssEncapsulation) {
```
tanner-reits, line 59: This additional condition is the same one that was a top-level check in the `patchTextContent` method. Just moved it here so we don't need to change things later when we make the `experimentalSlotFixes` the default behavior and remove these other `extras` options.

src/runtime/bootstrap-lazy.ts
```diff
@@ -160,8 +159,8 @@ export const bootstrapLazy = (lazyBundles: d.LazyBundlesRuntimeData, options: d.
         if (BUILD.appendChildSlotFix) {
           patchSlotAppendChild(HostElement.prototype);
         }
-        if (BUILD.scopedSlotTextContentFix) {
-          patchTextContent(HostElement.prototype, cmpMeta);
+        if (BUILD.scopedSlotTextContentFix && cmpMeta.$flags$ & CMP_FLAGS.scopedCssEncapsulation) {
```
tanner-reits, line 162: Same as previous comment

src/runtime/dom-extras.ts
```diff
@@ -207,14 +207,70 @@ export const patchSlotInsertAdjacentElement = (HostElementPrototype: HTMLElement
...
+          let slotContent = node.nextSibling as d.RenderNode | null;
+          while (slotContent && slotContent['s-sn'] === node['s-sn']) {
+            const tmp = slotContent;
+            slotContent = slotContent.nextSibling as d.RenderNode | null;
+            tmp.remove();
+          }
+
+          // If this is a default slot, add the text node in the slot location.
+          // Otherwise, destroy the slot reference node
+          if (node['s-sn'] === '') {
```
tanner-reits, line 263: I decided to make it so that the text node just won't be put into the DOM at all if there isn't a default slot reference node. Reason being you can't use `hidden` on text nodes so you'd have to style it to actually hide it, but it would still show up in `textContent` based on the spec.

tanner-reits (commented):
Few notes

src/runtime/bootstrap-custom-element.ts
```diff
@@ -43,9 +43,8 @@ export const proxyCustomElement = (Cstr: any, compactMeta: d.ComponentRuntimeMet
     cmpMeta.$flags$ |= CMP_FLAGS.needsShadowDomShim;
   }
 
-  // TODO(STENCIL-914): this check and `else` block can go away and be replaced by just `BUILD.scoped` once we
-  // default our pseudo-slot behavior
-  if (BUILD.experimentalSlotFixes && BUILD.scoped) {
+  // TODO(STENCIL-914): this check and `else` block can go away and be replaced by just the `scoped` check
+  if (BUILD.experimentalSlotFixes && cmpMeta.$flags$ & CMP_FLAGS.scopedCssEncapsulation) {
```
rwaskiewicz, line 47: For my own understanding, why did we remove the `BUILD.scoped` check here? Wouldn't we want this to get removed from the build output if there aren't any scoped components in a project?

test/karma/test-app/scoped-slot-text/karma.spec.ts
```diff
@@ -30,7 +30,7 @@ describe('scoped-slot-text', () => {
 
     cmpLabel.textContent = 'New text to go in the slot';
 
-    expect(cmpLabel.textContent).toBe('New text to go in the slot');
+    expect(cmpLabel.textContent.trim()).toBe('New text to go in the slot');
```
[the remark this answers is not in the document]
tanner-reits: Whitespace is the annoying part while doing these text patches. I did my best to produce the same output that the exact same component using `shadow: true` would return. The resulted in the the string always (in my finding) having spaces padding the start and end.

rwaskiewicz (approved):

test/karma/test-app/scoped-slot-text/karma.spec.ts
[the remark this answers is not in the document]
christian-bromann: Would it make sense to call `trim()` in the `textContent` getter?

christian-bromann (approved):

christian-bromann:

PR #5148, opened 2023-12-08
fix(runtime): patch `removeChild` for `scoped` components

tanner-reits:

Stencil does not patch the HtmlElement `removeChild()` method. This is causing issues in some of our Framework wrappers as frameworks are using these methods during rendering dynamic content (like looping through arrays and generating/removing elements).

Fixes: #3278, #2259

## What is the new behavior?

We provide a patch for `removeChild()` that works with our pseudo-slot behavior for `scoped` components.

- [ ] Yes
- [X] No

## Testing

**Manual testing**
Gonna give abbreviated testing steps since most of it can be achieved following the [React framework wrapper guide](https://stenciljs.com/docs/react) and the associated issues' repro cases.
1. Follow the [React framework wrapper guide](https://stenciljs.com/docs/react) to get a working React app with Stencil proxied components.
2. Create the Stencil components and React app from [this](https://github.com/jeco123/stencilbug.git) repro from #3278 or the PR summary of #2259 (**Note**: make all components `scoped`)
3. Run the React app and interact with the button to remove nodes. The component will disappear and a console error will be thrown
4. Install a build of this branch into the Stencil component library.
5. Enable `extras.experimentalSlotFixes` in the Stencil config.
6. Re-build the Stencil library and React component library
7. Repeat step 3 and notice that no console errors are thrown and the nodes are removed correctly 🎉 

**Automated testing**
Existing unit & e2e tests continue to pass
Added e2e tests for the patched method.

## Other information

Only available with the `extras.experimentalSlotFixes` config flag enabled.

Only available for Stencil component with `scoped: true` encapsulation.

### DIFF
--- a/src/runtime/dom-extras.ts
+++ b/src/runtime/dom-extras.ts
@@ -96,7 +96,7 @@ export const patchSlotAppendChild = (HostElementPrototype: any) => {
 /**
  * Patches the `removeChild` method on a `scoped` Stencil component.
  * This patch attempts to remove the specified node from a slot reference
- * if the the slot exists. Otherwise, it falls-back to the original `removeChild` method.
+ * if the slot exists. Otherwise, it falls-back to the original `removeChild` method.
  *
  * @param ElementPrototype The Stencil component to be patched
  */
--- a/src/runtime/dom-extras.ts
+++ b/src/runtime/dom-extras.ts
@@ -5,6 +5,7 @@ import { CMP_FLAGS, HOST_FLAGS } from '@utils';
 
 import type * as d from '../declarations';
 import { PLATFORM_FLAGS } from './runtime-constants';
+import { updateFallbackSlotVisibility } from './vdom/vdom-render';
 
 export const patchPseudoShadowDom = (
   hostElementPrototype: HTMLElement,
@@ -19,6 +20,7 @@ export const patchPseudoShadowDom = (
   patchSlotInsertAdjacentText(hostElementPrototype);
   patchTextContent(hostElementPrototype, descriptorPrototype);
   patchChildSlotNodes(hostElementPrototype, descriptorPrototype);
+  patchSlotRemoveChild(hostElementPrototype);
 };
 
 export const patchCloneNode = (HostElementPrototype: HTMLElement) => {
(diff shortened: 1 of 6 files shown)


src/runtime/dom-extras.ts
```diff
@@ -19,6 +20,7 @@ export const patchPseudoShadowDom = (
   patchSlotInsertAdjacentText(hostElementPrototype);
   patchTextContent(hostElementPrototype, descriptorPrototype);
   patchChildSlotNodes(hostElementPrototype, descriptorPrototype);
+  patchSlotRemoveChild(hostElementPrototype);
```
tanner-reits, line 23: Just wanna call out that this will _only_ be patched for **`scoped`** components. The associated issues had some examples using no encapsulation, but we're moving away from patching this behavior for that case.

src/runtime/vdom/vdom-render.ts
```diff
@@ -695,7 +695,7 @@ export const patch = (oldVNode: d.VNode, newVNode: d.VNode) => {
  *
  * @param elm the element of interest
  */
-const updateFallbackSlotVisibility = (elm: d.RenderNode) => {
+export const updateFallbackSlotVisibility = (elm: d.RenderNode) => {
```
tanner-reits, line 698: Had to export this to use in the patches in `dom-extras`. Can move this somewhere else if we want

tanner-reits (commented):
Two small notes

src/runtime/dom-extras.ts
```diff
@@ -74,12 +84,46 @@ export const patchSlotAppendChild = (HostElementPrototype: any) => {
...
+      return;
     }
     return (this as any).__appendChild(newChild);
   };
 };
 
+/**
+ * Patches the `removeChild` method on a `scoped` Stencil component.
+ * This patch attempts to remove the specified node from a slot reference
+ * if the the slot exists. Otherwise, it falls-back to the original `removeChild` method.
```
rwaskiewicz, line 99: Non-blocking nit:
```suggestion
 * if the slot exists. Otherwise, it falls-back to the original `removeChild` method.
```

rwaskiewicz (approved):

christian-bromann (approved):

PR #5159, opened 2023-12-12
chore(snc): resolve errors for optimize-css tests

rwaskiewicz:

We have quite a few SNC violations in our code!

GitHub Issue Number: N/A

## What is the new behavior?

resolve `strictNullCheck` violations for `optimize-css#optimizeCss` unit tests. the `fileName` argument doesn't appear to be used by `optimizeCss` in any meaningful way that (other than an additional path normalization call).

`optimizeCss` is a part of stencil's public api. as a result, we cannot remove this otherwise unused function parameter without it constituting a breaking change. add a TODO for now, to be completed in the next major version of stencil.

add a type assertion to invalid styling input. this allows us to test invalid css being passed to the optimizatiton function, without widening the type signature of `optimizeCss`

- [ ] Yes
- [x] No

## Testing

`filePath` only seems to be used here (highlighted in green):
Screenshot 2023-12-12 at 8 19 11 AM
where the only semi-meaningful use I can see is in the worker's `optimizeCss` fn, which never seems to use it: 
Screenshot 2023-12-12 at 8 19 45 AM

### DIFF
--- a/src/compiler/style/optimize-css.ts
+++ b/src/compiler/style/optimize-css.ts
@@ -8,6 +8,8 @@ export const optimizeCss = async (
   compilerCtx: d.CompilerCtx,
   diagnostics: d.Diagnostic[],
   styleText: string,
+  // TODO(STENCIL-1076): Investigate removing this parameter, which appears to be unused. This function is exported by
+  // the compiler, making this a breaking change should we remove it.
   filePath: string,
 ) => {
   if (typeof styleText !== 'string' || !styleText.length) {
--- a/src/compiler/style/test/optimize-css.spec.ts
+++ b/src/compiler/style/test/optimize-css.spec.ts
@@ -6,6 +6,8 @@ import path from 'path';
 import { optimizeCss } from '../optimize-css';
 
 describe('optimizeCss', () => {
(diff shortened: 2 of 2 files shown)


christian-bromann (approved):
Nice one 👌

src/compiler/style/optimize-css.ts
```diff
@@ -8,6 +8,8 @@ export const optimizeCss = async (
   compilerCtx: d.CompilerCtx,
   diagnostics: d.Diagnostic[],
   styleText: string,
+  // TODO(STENCIL-1076): Investigate removing this parameter, which appears to be unused. This function is exported by
+  // the compiler, making this a breaking change should be remove it.
```
tanner-reits, line 12:
```suggestion
  // the compiler, making this a breaking change should we remove it.
```

tanner-reits (approved):

src/compiler/style/optimize-css.ts
rwaskiewicz: Ah, good catch, thanks! (Done, and squashed the change since this was non-blocking)

PR #5160, opened 2023-12-12
chore(snc): resolve compiler event.ts violation

rwaskiewicz:

we have several SNC violations in the codebase
GitHub Issue Number: N/A

## What is the new behavior?

remove a single strictNullCheck violation for `events.ts`. this change only affects the type system, no functional changes have been made to the codebase.

specifically:
- the type of the internal (non-exported) `EventCallback#eventName` property has been widened to accept `null` in addition to `string`s.
- the type of the `eventName` local variable when parsing `arg0` in `on()` has been changed from `string`->`null`. this aligns us closer with the value we're assigning, and prevents 'implicit `any`' errors from the type checker

- [ ] Yes
- [x] No

## Testing

The change here only affects the type system. Since the interface that I'm changing is local only to the file it's contained in, and we have the value `null` assigned to the property I'm widening, I think that running the type checker here is sufficient. However, please let me know if there's anything else you'd like to see me try to verify this change.

### DIFF
--- a/src/compiler/events.ts
+++ b/src/compiler/events.ts
@@ -14,7 +14,7 @@ export const buildEvents = (): d.BuildEvents => {
 
   const on = (arg0: any, arg1?: any): d.BuildOnEventRemove => {
     if (typeof arg0 === 'function') {
-      const eventName: string = null;
+      const eventName: null = null;
       const callback = arg0;
       evCallbacks.push({
         eventName,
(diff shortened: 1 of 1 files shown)


christian-bromann (approved):
LGTM 👍

alicewriteswrongs (approved):

Issue #3137, opened 2021-11-09
feat: Allow ”as” ES6 keyword for Stencil’s decorators


splitinfinities:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/.github/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already include this feature request, without success.

### Describe the Feature Request

The `@Event` and `@Element` decorators can be imported and can interfere with global types or be mistakenly used within the components.d.ts file. In those circumstances, it would be helpful for component authors to be able to alias the import like with the `as` keyword of es6 modules. 

```ts
import { h, Event as StencilEvent, Element as StencilElement } from “@stencil/core”;

//etc
```

### Describe the Use Case

This feature allows component authors to use an aliased decorator that would be referred to within the components.d.ts file so that collisions can be reduced when interoping with global browser types. 

### Describe Preferred Solution

Open to perspective here, I’m not an expert on this. 

### Describe Alternatives

Maybe we could rename all the decorators? Or have alternate names that could be imported? 

Open to perspective here, I’m not an expert on this.

rwaskiewicz:
Potentially a part of the same problem/implementation space, but not the same problem: https://github.com/ionic-team/stencil/issues/3037

rwaskiewicz:
Related: https://github.com/ionic-team/stencil-eslint/issues/28, although not the same issue

rwaskiewicz:
This issue has been labeled to be added to our internal backlog to be discussed as a team.

rwaskiewicz:
The fix for this issue has been released as a part of today's [Stencil 4.9.0 release](https://github.com/ionic-team/stencil/releases/tag/v4.9.0).

PR #5161, opened 2023-12-12
feat(compiler): Stencil decorator import aliasing

tanner-reits:

It is not possible to use aliases (i.e. "as") when importing Stencil decorators because of the compiler's meta-to-static conversions of all Stencil decorators.

GitHub Issue Number: #3137 

## What is the new behavior?

Import aliasing is now possible as we map aliased imports back to their original decorator name. The aliased name is then passed to all the transformation functions.

- [ ] Yes
- [X] No

## Testing

**Manual testing**
- This can be easily tested in a Stencil component starter by aliasing any Stencil decorator used in a component. 
- Also manually tested in a project using a CE build

**Automated testing**
Added e2e test that tests a subset of Stencil decorators with aliased import names.

## Other information

PR to update relevant docs: https://github.com/ionic-team/stencil-site/pull/1292

### DIFF
--- a/src/compiler/transformers/decorators-to-static/import-alias-map.ts
+++ b/src/compiler/transformers/decorators-to-static/import-alias-map.ts
@@ -1,0 +1,42 @@
+import ts from 'typescript';
+
+import { STENCIL_DECORATORS, StencilDecorator } from './decorators-constants';
+
+export class ImportAliasMap extends Map<StencilDecorator, string> {
+  constructor(sourceFile: ts.SourceFile) {
+    super();
+    this.generateImportAliasMap(sourceFile);
+  }
+
+  /**
+   * Parses a {@link ts.SourceFile} and generates a map of all imported Stencil decorators
+   * to their aliases import name (if one exists).
+   *
+   * @param sourceFile The source file to parse
+   */
+  private generateImportAliasMap(sourceFile: ts.SourceFile) {
+    const importDeclarations = sourceFile.statements.filter(ts.isImportDeclaration);
+
+    for (const importDeclaration of importDeclarations) {
+      if (importDeclaration.moduleSpecifier.getText().includes('@stencil/core')) {
+        const namedBindings = importDeclaration.importClause?.namedBindings;
+
+        if (ts.isNamedImports(namedBindings)) {
+          for (const element of namedBindings.elements) {
+            const importName = element.name.getText();
+            const originalImportName = element.propertyName?.getText() ?? importName;
+
+            // We only care to generate a map for the Stencil decorators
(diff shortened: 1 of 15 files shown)


src/compiler/transformers/decorators-to-static/import-alias-map.ts
```diff
@@ -0,0 +1,42 @@
...
+              this.set(originalImportName as StencilDecorator, importName);
+            }
+          }
+        }
+      }
+    }
+  }
+
+  override get(key: StencilDecorator): string {
+    return super.get(key) ?? key;
```
christian-bromann, line 40: Is there a valid use case to ask for a decorator key we don't know?

christian-bromann (approved):
One non blocking question, rest LGTM 👍

src/compiler/transformers/decorators-to-static/import-alias-map.ts
tanner-reits: This isn't for the case of a decorator we don't know, but for a decorator that wasn't used in the component. If we don't override this, when we pass decorator names into the functions responsible for transforming them, we'd need to do `importAliasMap.get('Watch') ?? 'Watch'`. Since if the component didn't have a watch function, the map would not have a key for `Watch` and would return `undefined`

tanner-reits: We could, alternatively, change the class so that it always has all the Stencil decorators and we just change the value for any aliased keys, works the same.

christian-bromann: Thanks for clarifying. This makes sense to me and can stay as is.

rwaskiewicz (changes requested):

test/karma/test-app/import-aliasing/karma.spec.ts
```diff
@@ -0,0 +1,29 @@
+import { setupDomTests, waitForChanges } from '../util';
+
+/**
+ * Test cases for using import aliases for Stencil decorators. Only
+ * tests a subset of all Stencil decorators.
```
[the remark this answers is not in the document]
tanner-reits: I didn't just because we have so many decorators it felt overkill to slam them all in a test, but I can

rwaskiewicz (commented):
@tanner-reits LGTM, thanks! Can you do me a favor and open a complimentary PR to this one (and link it in the PR Description) to update the verbiage in our documentation regarding type conflicts? Specifically, I'm looking at https://github.com/ionic-team/stencil-site/blob/e16218ad879ad1d7aedd5365abf2b764e5357329/docs/components/events.md?plain=1#L85-L90

rwaskiewicz (approved):

PR #5163, opened 2023-12-12
test(compiler): test bundle sorting

rwaskiewicz:

see 'new behavior'
GitHub Issue Number: N/A

## What is the new behavior?

add comprehensive testing to how component metadata is sorted. the motivation for this pull request is there are several strict null check violations in the code that is put under test. however, i didn't feel comfortable making any changes without tests in place, hence this commit
 

- [ ] Yes
- [x] No

## Testing

I ran each test through my local debugger, and set breakpoints to validate that the returned values were coming from the correct branch.

We do now have 100% code coverage in this function now (but obviously, we need the former testing check to verify the tests actually evaluate the branches that they say they do)


alicewriteswrongs (approved):

christian-bromann (approved):
🙌 👍

PR #5164, opened 2023-12-12
fix(runtime): allow setting `key` attr on nested Stencil components

alicewriteswrongs:
This makes it possible to set a `key` attribute on a nested Stencil component without messing up how hydration works. I noticed that this was not working correctly when working on #5143.

There's a regression test which will reproduce the issue if run in the context of `main`. In short, if you have two components like so:

```tsx
    @Component({ tag: 'cmp-a' })
    class CmpA {
      render() {
        return (
          <Host key="test">
            <cmp-b key="my-test-key">light-dom</cmp-b>
          </Host>
        );
      }
    }
    @Component({ tag: 'cmp-b' })
    class CmpB {
      render() {
        return (
          <Host>
            <slot key="key1"></slot>
            <footer key="key2"></footer>
          </Host>
        );
      }
    }
```

You'll get rendering issues where comments that need to be present for slot relocation and whatnot to work aren't appearing. Not good!

## What is the new behavior?

This ensures that we only start to check whether `vnode.$key$` matches between vnodes _after_ the first render. This fixes a situation where on the first render the `.$key$` property of the _old_ root vnode would _always_ be `null`, leading the comparison to result in always re-rendering the component.

- [ ] Yes
- [x] No

## Testing

I've run this in Framework (the unit tests) and haven't seen any issues, and I've also checked in out in an [example hydrate app](https://github.com/alicewriteswrongs/stencil-hydrate-example) I created the other day.

### DIFF
--- a/src/runtime/test/hydrate-no-encapsulation.spec.tsx
+++ b/src/runtime/test/hydrate-no-encapsulation.spec.tsx
@@ -13,7 +13,6 @@ describe('hydrate no encapsulation', () => {
         );
       }
     }
-    // @ts-ignore
     const serverHydrated = await newSpecPage({
       components: [CmpA],
       html: `<cmp-a></cmp-a>`,
@@ -38,7 +37,6 @@ describe('hydrate no encapsulation', () => {
         );
       }
     }
-    // @ts-ignore
     const serverHydrated = await newSpecPage({
       components: [CmpA],
       html: `<cmp-a></cmp-a>`,
@@ -54,7 +52,6 @@ describe('hydrate no encapsulation', () => {
       </cmp-a>
     `);
 
-    // @ts-ignore
     const clientHydrated = await newSpecPage({
       components: [CmpA],
       html: serverHydrated.root.outerHTML,
@@ -81,7 +78,6 @@ describe('hydrate no encapsulation', () => {
         return <Host>Hello</Host>;
       }
     }
-    // @ts-ignore
     const serverHydrated = await newSpecPage({
       components: [CmpA],
       html: `<cmp-a></cmp-a>`,
@@ -95,7 +91,6 @@ describe('hydrate no encapsulation', () => {
       </cmp-a>
     `);
 
-    // @ts-ignore
     const clientHydrated = await newSpecPage({
       components: [CmpA],
       html: serverHydrated.root.outerHTML,
@@ -126,7 +121,6 @@ describe('hydrate no encapsulation', () => {
         );
       }
     }
-    // @ts-ignore
     const serverHydrated = await newSpecPage({
       components: [CmpA],
       html: `<cmp-a></cmp-a>`,
@@ -146,7 +140,6 @@ describe('hydrate no encapsulation', () => {
       </cmp-a>
     `);
 
-    // @ts-ignore
     const clientHydrated = await newSpecPage({
       components: [CmpA],
       html: serverHydrated.root.outerHTML,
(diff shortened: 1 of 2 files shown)


src/runtime/test/hydrate-no-encapsulation.spec.tsx
```diff
@@ -166,6 +166,69 @@ describe('hydrate no encapsulation', () => {
     `);
   });
 
+  it('nested text slot with key', async () => {
```
alicewriteswrongs, line 169: this test is a reproduction of the issue I was seeing, if you wanted to see what happens you could check out this branch locally and either run the test without rebuilding or

```sh
git checkout origin/main -- src/runtime/vdom/vdom-render.ts
npm run build
npx jest --clearCache
npx --node-options=--experimental-vm-modules jest --coverage=false -- hydrate-no-encapsulation
```

which will run this test without the fix so you can see what happens

alicewriteswrongs (commented):
one note

src/runtime/test/hydrate-no-encapsulation.spec.tsx
```diff
@@ -166,6 +166,69 @@ describe('hydrate no encapsulation', () => {
...
+      render() {
+        return (
+          <Host>
+            <slot key="key1"></slot>
+            <footer key="key2"></footer>
+          </Host>
+        );
+      }
+    }
+    // @ts-ignore
```
rwaskiewicz, line 191: For my own knowledge, why do we need this `@ts-ignore`? If we need it, can we add a justification/reason why we need it in the comment for future us?

src/runtime/test/hydrate-no-encapsulation.spec.tsx
```diff
@@ -166,6 +166,69 @@ describe('hydrate no encapsulation', () => {
...
+          <!--o.1.1-->
+          <!--s.2.0.0.0.-->
+          <!--t.1.1.1.0-->
+          light-dom
+          <footer c-id="2.1.0.1"></footer>
+        </cmp-b>
+      </cmp-a>
+    `);
+
+    // @ts-ignore
```
rwaskiewicz, line 211: For my own knowledge, why do we need this `@ts-ignore`? If we need it, can we add a justification/reason why we need it in the comment for future us?

src/runtime/test/hydrate-no-encapsulation.spec.tsx
rwaskiewicz: I ran this locally, adding what I found for posterity:
```
  ● hydrate no encapsulation › nested text slot with key

    expect(received).toBe(expected) // Object.is equality

    - Expected  - 3
    + Received  + 1

      <cmp-a class="hydrated">
        <!--r.1-->
        <cmp-b class="hydrated">
    -     <!--r.2-->
    -     <!---->
    -     <!--s.2.0.0.0.-->
    +     <!---->
          light-dom
          <footer></footer>
        </cmp-b>
      </cmp-a>
```

rwaskiewicz (approved):
LGTM - added a few non-blocking ocmments

src/runtime/test/hydrate-no-encapsulation.spec.tsx
alicewriteswrongs: Yup that's what I expect to see 👍

src/runtime/test/hydrate-no-encapsulation.spec.tsx
alicewriteswrongs: We don't! I copy-pasted an existing test which has this (every test in this file does) and just assumed it was necessary. However, it's not! That's for two reasons:

- we don't type-check the `spec.tsx` files anyway (they're not included in `tsconfig.include`) so removing these doesn't cause any new errors
- it appears that additionally whatever errors these were added to suppress aren't present, since if I add `src/**/*.tsx` to `include` in `tsconfig.json` and then start removing these `@ts-ignore` comments I don't see any new errors

So I'm going to go ahead and just remove them all from the file

rwaskiewicz (approved):

PR #5165, opened 2023-12-12
fix(declarations): bundle child_process type for portability

### LINKED ISSUE #5162: bug: Compiler error because of node type reference without node types
### Prerequisites

### Stencil Version

4.8.2

### Current Behavior

The compiler errors during a build because of a type import from `node:child_process`, which cannot be found. This seems to be not found because the stencil project is set up to not include node types.

### Expected Behavior

Do not error the build because of a type import from node in the internal compiler implementation when the stencil project itself does not have node types.

### System Info

```shell
      System: node 19.6.0
    Platform: windows (10.0.19045)
   CPU Model: Intel(R) Core(TM) i5-7300HQ CPU @ 2.50GHz (4 cpus)
    Compiler: C:\stencil-bug\node_modules\@stencil\core\compiler\stencil.js
       Build: 1702322155
     Stencil: 4.8.2
  TypeScript: 5.2.2
      Rollup: 2.42.3
      Parse5: 7.1.2
      Sizzle: 2.42.3
      Terser: 5.26.0
```

### Steps to Reproduce

```sh
npm ci
# ...
npm run build
# ... ok
npm run build
# ...
[ ERROR ]  TypeScript: node_modules/@stencil/core/internal/stencil-private.d.ts:2:53
           Cannot find module 'child_process' or its corresponding type declarations.

      L1:  import { result } from './utils';
      L2:  import type { Serializable as CPSerializable } from 'child_process';
```

Happens when `lib` or `typeRoots` are set in `tsconfig.json` without node types

### Code Reproduction URL

https://github.com/silvertech-daniel/stencil-bug

### Additional Information

_No response_

alicewriteswrongs:
We need to make sure our type declarations don't depend on types we don't bundle!

fixes #5162

STENCIL-1077

GitHub Issue Number: N/A

## What is the new behavior?

- 

- [ ] Yes
- [x] No

## Testing

There's a reproduction case on #5162. You could test that this fixes the issue by building and packing this locally, installing it in that reproduction case, and confirming that you don't get a type error when building.

### DIFF
--- a/scripts/bundles/internal.ts
+++ b/scripts/bundles/internal.ts
@@ -1,3 +1,4 @@
+import { generateDtsBundle } from 'dts-bundle-generator';
 import fs from 'fs-extra';
 import { join } from 'path';
 
@@ -55,6 +56,29 @@ async function copyStencilInternalDts(opts: BuildOptions, outputInternalDir: str
   const privateDtsSrcPath = join(declarationsInputDir, 'stencil-private.d.ts');
   const privateDtsDestPath = join(outputInternalDir, 'stencil-private.d.ts');
   let privateDts = cleanDts(await fs.readFile(privateDtsSrcPath, 'utf8'));
+
+  // @stencil/core/internal/child_process.d.ts
+  const childProcessSrcPath = join(declarationsInputDir, 'child_process.d.ts');
+  const childProcessDestPath = join(outputInternalDir, 'child_process.d.ts');
+
+  // we generate a tiny tiny bundle here of just
+  // `src/declarations/child_process.ts` so that `internal/stencil-private.d.ts`
+  // can import from `'./child_process'` without worrying about resolving the
+  // types from `node_modules`.
+  const childProcessDts = generateDtsBundle([
+    {
+      filePath: childProcessSrcPath,
+      libraries: {
+        // we need to mark this library so that types imported from it are inlined
+        inlinedLibraries: ['child_process'],
+      },
+      output: {
+        noBanner: true,
+        exportReferencedTypes: false,
+      },
+    },
+  ]).join('\n');
+  await fs.writeFile(childProcessDestPath, childProcessDts);
 
   // the private `.d.ts` imports the `Result` type from the `@utils` module, so
   // we need to rewrite the path so it imports from the right relative path
--- a/src/declarations/stencil-private.ts
+++ b/src/declarations/stencil-private.ts
@@ -1,7 +1,7 @@
 import { result } from '@utils';
-import type { Serializable as CPSerializable } from 'child_process';
 
 import type { InMemoryFileSystem } from '../compiler/sys/in-memory-fs';
+import type { CPSerializable } from './child_process';
 import type {
   BuildEvents,
   BuildLog,
--- a/scripts/utils/bundle-dts.ts
+++ b/scripts/utils/bundle-dts.ts
@@ -1,4 +1,4 @@
-import { EntryPointConfig, generateDtsBundle, OutputOptions } from 'dts-bundle-generator/dist/bundle-generator.js';
+import { EntryPointConfig, generateDtsBundle, OutputOptions } from 'dts-bundle-generator';
 import fs from 'fs-extra';
 
 import { BuildOptions } from './options';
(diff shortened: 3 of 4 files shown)


src/declarations/stencil-private.ts
```diff
@@ -1,7 +1,7 @@
 import { result } from '@utils';
-import type { Serializable as CPSerializable } from 'child_process';
 
 import type { InMemoryFileSystem } from '../compiler/sys/in-memory-fs';
+import type { CPSerializable } from './child_process';
```
alicewriteswrongs, line 4: this is a bit convoluted, I know, but the alternative is to keep importing directly from `child_process` in this file and then run `dts-bundle-generator` on the whole thing (`build/declarations/stencil-private.d.ts`). In some rough testing this took the size of `internal/stencil-private.d.ts` from ~69k to ~160k, which I thought was too much. So instead we stand up a module that just re-exports one type from `child_process`, and we can then bundle _that_ with `dts-bundle-generator` and ship it right next to `stencil-private.d.ts` so it can import the inlined type from it, making `stencil-private.d.ts` portable.

alicewriteswrongs (commented):
one note

scripts/bundles/internal.ts
```diff
@@ -56,6 +57,29 @@ async function copyStencilInternalDts(opts: BuildOptions, outputInternalDir: str
...
+        // we need to mark this library so that types imported from it are inlined
+        inlinedLibraries: ['child_process'],
+      },
+      output: {
+        noBanner: true,
+        exportReferencedTypes: false,
+      },
+    },
+  ]).join('\n');
+  await fs.writeFile(childProcessDestPath, childProcessDts);
```
christian-bromann, line 81: If someone wonders what this results to: it creates a file in `internal/child_process.d.ts` with this content:

```ts


type Serializable = string | object | number | boolean | bigint;

export {
	Serializable as CPSerializable,
};

export {};

```

scripts/bundles/internal.ts
```diff
@@ -1,3 +1,4 @@
+import { generateDtsBundle } from 'dts-bundle-generator/dist/bundle-generator.js';
```
christian-bromann, line 1: There is probably no other way around this but I am afraid this can break at any time. Should we at least raise an issue in `dts-bundle-generator` and ask if they could expose the function from the package?

christian-bromann (commented):
Is there a way we can test this so in case we use another Node.js interface we get alerted to re-export it properly?

scripts/bundles/internal.ts
alicewriteswrongs: I actually just copied that from our `bundle-dts.ts` script, here:

https://github.com/ionic-team/stencil/blob/97dcb45d44751d239b0afb6380bea217818b211a/scripts/utils/bundle-dts.ts#L1

But I don't think it's necessary to do the relative import like that, since importing those things directly from `dts-bundle-generator` seems to work fine. I'll change it in both locations.

alicewriteswrongs:
For testing this overall, i.e. putting a regression test in place, I thought that the component starter tests _should_ catch problems of this type but evidently not. I'll look into how we might modify them to do that a bit

alicewriteswrongs:
I realized why most users wouldn't have run into this issue. If you start a Stencil project doing the following:

```sh
npm init stencil@latest component test-node-type-failure
```

and then do

```sh
npm i
npm run build
```

you won't see the error about missing types, which makes sense since if you check `node_modules/@types` you'll see that it's full of types!

_However_ if your project is only depending on `@stencil/core` alone then you won't pull in any type declarations and you will see the issue.

So I'm going to add a smoke test which just checks that the project runs without errors when it is the only dependency installed in a project

alicewriteswrongs:
(I'll add that in a separate PR so we can first confirm that the test fails, merge this, rebase, and confirm it then passes)

christian-bromann:
@alicewriteswrongs afaik another way to potentially test this is to define `types` in `tsconfig.json` which causes TypeScript to only rely on the package types defined in there instead of the ones it can find within `node_modules`.

alicewriteswrongs:
Oh true, well I already put together #5170 which I kind of think we'll want anyway (it covers a broader range of potential issues than just types missing)

tanner-reits (commented):
Should be good pending the import update from [this](https://github.com/ionic-team/stencil/pull/5165/files#r1425721040) comment

alicewriteswrongs:
@tanner-reits @christian-bromann just updated that import (no idea why it was like that in the first place!)

tanner-reits (approved):

christian-bromann (approved):

Issue #3382, opened 2022-05-25
bug: HTMLUListElement is not defined


erwinheitzman:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/.github/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

2.15.2

### Current Behavior

When running unit tests and the HTMLUListElement is used in either your testcode or tested code, it will throw an error that the HTMLUListElement is not defined.

### Expected Behavior

HTMLUListElement is supported by both NodeJS and browsers so I expect this to work.
I also noted that all elements have the constructor MockHTMLElement and I expect elements to have their correct types.

### Steps to Reproduce

Run the reproduction application test or run the following test example to reproduce:
```js
it('HTMLUListElement is not defined', async () => {
    const unorderedList = document.createElement('ul');
    console.log(unorderedList instanceof HTMLUListElement);
});
```

### Code Reproduction URL

https://github.com/erwinheitzman/stencil-htmlulistelement

### Additional Information

I feel that it is likely related to the environment that Stencil runs it's tests in or related to a mock.

_No response_

johnjenkins:
> HTMLUListElement is supported by both NodeJS ...

What do you mean by supported by NodeJS here? 
Node is an independent js environment detached from a browser, therefore without a DOM and therefore without a native `HTMLUListElement` instance. 

Perhaps in another 'pretend' dom this test will work (see jsdom), but at the moment this is how MockDoc (stencil's 'pretend' dom) is designed to work.

I'd suggest your bug report should probably be a feature request.

erwinheitzman:
You can ignore the NodeJS mention really, the most important part is that the MockDoc assigns the the MockHTMLElement while instead there should probably be a MockHTMLUListElement and a similar version for all the supported elements.

For reference I am talking about this: https://developer.mozilla.org/en-US/docs/Web/API/HTMLUListElement?retiredLocale=nl

This is widely supported and since our Stencil code has a snippet like:
```js
if (element instanceof HTMLUListElement) {
  // do stuff
}
```
We are not able to unit test this functionality even though the code is valid.

theo-staizen:
[...]

PR #5169, opened 2023-12-13
fix(mock-doc): add HTMLUListElement

rwaskiewicz:

When Stencil's mock-doc creates an `HTMLUListElement`, we create an element that does not descend from `HTMLUListElement`

Fixes: #3382

## What is the new behavior?

add a definition for `HTMLUListElement`. this allows `createElement` calls to return a mocked version of `HTMLUListElement` that passes `instanceof` calls.

prior to this commit, folks attempting to use an `HTMLUListElement` in a stencil test like so would receive an error:
```ts
describe('my-component', () => {
  it('using HTMLUListElement works', async () => {
    const unorderedList = document.createElement('ul');
    console.log(unorderedList instanceof HTMLUListElement);
  });
});
```
```
[46:08.8]  jest args: --spec --e2e --max-workers=8
 FAIL  src/my-component.spec.ts
  my-component
    ✕ using HTMLUListElement works (4 ms)

  ● my-component › using HTMLUListElement works

    ReferenceError: HTMLUListElement is not defined

      2 |   it('using HTMLUListElement works', async () => {
      3 |     const unorderedList = document.createElement('ul');
    > 4 |     console.log(unorderedList instanceof HTMLUListElement);
        |                                          ^
      5 |   });
      6 | });
      7 |

      at Object.<anonymous> (src/my-component.spec.ts:4:42)
```

- [ ] Yes
- [x] No

## Testing

I used the original reproduction case in the linked issue.
For posterity, it had a test that matched the one in the "New Behavior"/commit message:
```ts
describe('my-component', () => {
  it('using HTMLUListElement works', async () => {
    const unorderedList = document.createElement('ul');
    console.log(unorderedList instanceof HTMLUListElement);
  });
});
```
If this test were placed in a new Stencil Component Starter project (`npm init stencil@latest component ul-test && cd $_ && npm i`, then copy the test to any `*.spec.ts` file), then the tests run (`npm t`), we'd see the failure above.

After building this branch and packing it (`npm run clean && npm ci && npm run build && npm pack`) and installing the tarball into the component starter project (`npm i <TARBALL>`), the tests pass (`npm t`)
## Other information
STENCIL-466 bug: HTMLUListElement is not defined

### DIFF
--- a/src/mock-doc/test/element.spec.ts
+++ b/src/mock-doc/test/element.spec.ts
@@ -1,5 +1,5 @@
 import { MockDocument } from '../document';
-import { MockAnchorElement, MockMetaElement, MockSVGElement } from '../element';
+import { MockAnchorElement, MockMetaElement, MockSVGElement, MockUListElement } from '../element';
 import { MockElement, MockHTMLElement } from '../node';
 import { cloneWindow, MockWindow } from '../window';
 
@@ -477,6 +477,14 @@ describe('element', () => {
       const elm: MockAnchorElement = doc.createElement('a');
       elm.href = 'http://stenciljs.com/path/to/page';
       expect(elm.pathname).toBe('/path/to/page');
+    });
+  });
+
+  describe('ul', () => {
+    it('textContent', () => {
+      const elm: MockUListElement = doc.createElement('ul');
+      elm.textContent = 'this is an item in an unordered list';
+      expect(elm.textContent).toBe('this is an item in an unordered list');
     });
   });
 });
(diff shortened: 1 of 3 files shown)


src/mock-doc/test/element.spec.ts
```diff
@@ -479,4 +479,12 @@ describe('element', () => {
       expect(elm.pathname).toBe('/path/to/page');
     });
   });
+
+  describe('ul', () => {
+    it('textContent', () => {
+      const elm: MockUListElement = doc.createElement('ul');
```
rwaskiewicz, line 485: There aren't any unique properties here to test - so I just went with `textContent` to test the creation of _something_

christian-bromann (approved):
👍

tanner-reits (approved):
LGTM

PR #5170, opened 2023-12-13
test(ci): smoke test that Stencil doesn't error when installed alone

alicewriteswrongs:
This adds a smoke test that checks that a project with only `@stencil/core` installed will build a simple component without errors.

When working on #5165 it became clear we should have a regression test in place to ensure we know right away if a change makes a 'bare' stencil project (i.e. with only `@stencil/core` installed) is going to error out.

GitHub Issue Number: N/A

## What is the new behavior?

- 

- [ ] Yes
- [ ] No

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -44,6 +44,13 @@ jobs:
     with:
       build_name: stencil-core
 
+  no_starter_smoke_tests:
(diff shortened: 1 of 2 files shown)


alicewriteswrongs:
This might be useful at some point, but I'm going to close it now

PR #5177, opened 2023-12-14
chore(snc): fix some snc errors in puppeteer-emulate.ts

alicewriteswrongs:
This fixes 6 SNC errors by switching from explicit annotation to using the `satisfies` keyword.

This is a type-level only change, no runtime behavior should change.

- [ ] Yes
- [x] No

## Testing

This is only changing type annotations, so I think as long as the build is passing and unit tests are green we should be good!


christian-bromann (approved):
💪

rwaskiewicz (approved):

PR #5178, opened 2023-12-14
chore(snc): fix four snc errors relating to `sys.details` in `cli/`

alicewriteswrongs:
This fixes a few snc errors resulting from the typing of `sys.details` as `SystemDetails | undefined` that show up in `src/cli`.

Ideally we'd fix the typing but it's a public type, so for now the best thing is just to add some optional chaining.

- [ ] Yes
- [x] No

## Testing

This is just adding some optional chaining property accesses, so I believe there should not be any runtime changes here to worry about. If the build is passing and the tests are green I think we're good.

### DIFF
--- a/src/cli/task-info.ts
+++ b/src/cli/task-info.ts
@@ -13,12 +13,14 @@ export const taskInfo = (coreCompiler: CoreCompiler, sys: CompilerSystem, logger
 
   console.log(``);
   console.log(`${logger.cyan('      System:')} ${sys.name} ${sys.version}`);
-  console.log(`${logger.cyan('    Platform:')} ${details.platform} (${details.release})`);
-  console.log(
-    `${logger.cyan('   CPU Model:')} ${details.cpuModel} (${sys.hardwareConcurrency} cpu${
-      sys.hardwareConcurrency !== 1 ? 's' : ''
-    })`,
-  );
+  if (details) {
+    console.log(`${logger.cyan('    Platform:')} ${details.platform} (${details.release})`);
+    console.log(
+      `${logger.cyan('   CPU Model:')} ${details.cpuModel} (${sys.hardwareConcurrency} cpu${
+        sys.hardwareConcurrency !== 1 ? 's' : ''
+      })`,
+    );
+  }
   console.log(`${logger.cyan('    Compiler:')} ${sys.getCompilerExecutingPath()}`);
   console.log(`${logger.cyan('       Build:')} ${coreCompiler.buildId}`);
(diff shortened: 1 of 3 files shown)


rwaskiewicz (approved):

src/cli/task-info.ts
```diff
@@ -13,9 +13,9 @@ export const taskInfo = (coreCompiler: CoreCompiler, sys: CompilerSystem, logger
 
   console.log(``);
   console.log(`${logger.cyan('      System:')} ${sys.name} ${sys.version}`);
-  console.log(`${logger.cyan('    Platform:')} ${details.platform} (${details.release})`);
+  console.log(`${logger.cyan('    Platform:')} ${details?.platform} (${details?.release})`);
```
christian-bromann, line 16: to avoid having `undefined` in the logs
```suggestion
  if (details) {
    console.log(`${logger.cyan('    Platform:')} ${details?.platform} (${details?.release})`);
  }
```

src/cli/task-info.ts
```diff
@@ -13,9 +13,9 @@ export const taskInfo = (coreCompiler: CoreCompiler, sys: CompilerSystem, logger
...
   console.log(
-    `${logger.cyan('   CPU Model:')} ${details.cpuModel} (${sys.hardwareConcurrency} cpu${
+    `${logger.cyan('   CPU Model:')} ${details?.cpuModel} (${sys.hardwareConcurrency} cpu${
```
christian-bromann, line 18:
```suggestion
    `${logger.cyan('   CPU Model:')} ${details?.cpuModel || 'unknown'} (${sys.hardwareConcurrency} cpu${
```

christian-bromann (commented):
Do we expect these values to be actually undefined? If so we might just want to skip logging them.

src/cli/task-info.ts
alicewriteswrongs: I thought about doing something like this, however I don't think it will ever be undefined in practice. In the `cli/` module we're always operating with a node-specific `CompilerSystem`, on which the `details` property is always defined:

https://github.com/ionic-team/stencil/blob/654d7213c1c21647b5df83c14b90fdeed3ed47f1/src/sys/node/node-sys.ts#L652-L661

so I don't think in practice it will ever happen. I'd like to change the type to more accurately reflect the fact that it's always defined, but this is a public type so that would be a breaking change unfortunately

alicewriteswrongs: anyway - added a check in just for the heck of it

christian-bromann (approved):
👍

PR #5179, opened 2023-12-14
chore(snc): make optional property required on `PrintLine`

alicewriteswrongs:
the `.text` property on the `PrintLine` interface was previously marked optional, but there are no locations in the Stencil codebase where we do not supply a value for it, so we can safely remove a few snc violations by typing it as just `string` instead of `string | undefined`.

I believe this is safe to do because this is not a public type, but instead is an internal type used for logging and so on.

- [ ] Yes
- [x] No

## Testing

This is just a type-level change so we should be good as long as the build is green!

### DIFF
--- a/src/declarations/stencil-private.ts
+++ b/src/declarations/stencil-private.ts
@@ -44,7 +44,7 @@ export interface SourceMap {
 export interface PrintLine {
   lineIndex: number;
   lineNumber: number;
-  text?: string;
+  text: string;
   errorCharStart: number;
   errorLength?: number;
 }


christian-bromann (approved):
👍

rwaskiewicz:
@alicewriteswrongs 

Double checking here:

> I believe this is safe to do because this is not a public type, but instead is an internal type used for logging and so on.

Doesn't this get exported as a [property on `Diagnostic` though](https://github.com/ionic-team/stencil/blob/22ad258ae39701237d949e98cf9e850627798622/src/declarations/stencil-public-compiler.ts#L2481)?

alicewriteswrongs:
ahh you're right, I forgot about that. I guess this is a 'grey' type then haha.

I would still maybe make the case for this change because I doubt anyone is going to get any utility out of the `PrintLine` interface without supplying `text` _anyway_, but if we want to hold off I'm fine with that too

rwaskiewicz:
I can buy that - let's give it a go

rwaskiewicz (approved):

PR #5180, opened 2023-12-14
chore(snc): make properties on `CompilerWorkerTask` required

alicewriteswrongs:
We had some optional properties on this interface which are always supplied in practice, so we can drop a few snc errors by making them required.

- [ ] Yes
- [x] No

## Testing

This PR is _mainly_ a type-level change, but does also require a small change to a mock. As long as the build is green I believe we're safe!


christian-bromann (approved):
👍

rwaskiewicz (approved):

PR #5182, opened 2023-12-15
chore(snc): resolve validate-dev-server.spec errors

rwaskiewicz:

We have quite a few SNC violations in the code base!

GitHub Issue Number: N/A

## What is the new behavior?

update testing for `devServer#port` values:
1. add a test for `undefined` to verify the desired behavior (setting a
default port if one is not specified)
2. add a type assertion to suppress SNC violation when the port is set
to null

update testing for `devServer#historyApiFallback` values:
1. add a test for `undefined` to verify the desired behavior (setting a
default port if one is not specified)
2. add a type assertion to suppress SNC violation when the port is set
to null

add the bang operator to assertions where we have previously verified
the existence of a property (using `toBeDefined`)

- [ ] Yes
- [x] No

## Testing

I've run the new tests locally to verify they pass and are hitting the code paths they intend to (with a local debugger)

Otherwise, all other changes are affecting the type system. Locally running `npx tsc --strictNullChecks --noEmit` and `npm run build` verifies the drop in SNCs and the build still passes.

## Other information

I'll squash this before merging to pretty up the commit msg

### DIFF
--- a/src/compiler/config/test/validate-dev-server.spec.ts
+++ b/src/compiler/config/test/validate-dev-server.spec.ts
@@ -123,9 +123,16 @@ describe('validateDevServer', () => {
   );
 
   it('should not set default port if null', () => {
-    inputConfig.devServer = { ...inputDevServerConfig, port: null };
+    // we intentionally set the value to `null` for the purposes of this test, hence the type assertion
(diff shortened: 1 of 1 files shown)


christian-bromann (approved):
👍

alicewriteswrongs (approved):

PR #5193, opened 2023-12-18
chore(renovate): group download & upload artifact

rwaskiewicz:

see 'new behavior' section

GitHub Issue Number: N/A

## What is the new behavior?

group github's `actions/download-artifact` and `actions/upload-artifact` into a single renovate job. in #5191 and #5192, renovate created a pr for each of these individually for the v4.0 release of each library. however, they have a peer dependency on one another, making it such that we could merge neither.

this was solved by cherry-picking the commit from the latter onto the former. going forward, let's group these for all dependencies. we should (hopefully) only run into this across majors, but we'll err on the 'safe' side of things for now

- [ ] Yes
- [x] No

## Testing

Not much we can do, other than tail the logs once we merge it.

I did run `renovate-config-validator` locally, which validated our config conforms to the JSON schema as a quick check (but that doesn't verify everything/is just a spot check)

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -88,6 +88,11 @@
       schedule: ["before 11am on monday"]
     },
     {
+      // group these two, as they may rely on one another during major version bumps (see #5191)
+      matchPackageNames: ['actions/download-artifact', 'actions/upload-artifact'],
+      groupName: 'Download + Upload Artifacts',
+    },
+    {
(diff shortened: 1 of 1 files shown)


tanner-reits (approved):

christian-bromann (approved):
👍

PR #5199, opened 2023-12-20
fix(build): compile pixel-match file correctly

christian-bromann:

The ESBuild task to build the screenshot component of Stencil doesn't re-compile pixel-match.

GitHub Issue Number: N/A

## What is the new behavior?
Make sure that the core screenshot component and pixel-match are compiled correctly. I changed the implementation of it to make this type of error less likely to happen again with the compromise of making the code less flexible. Happy to revert that if seen needed.

- [ ] Yes
- [x] No

## Testing

n/a

## Other information

n/a

### DIFF
--- a/scripts/esbuild/screenshot.ts
+++ b/scripts/esbuild/screenshot.ts
@@ -7,6 +7,11 @@ import { BuildOptions } from '../utils/options';
 import { writePkgJson } from '../utils/write-pkg-json';
 import { getBaseEsbuildOptions, getEsbuildAliases, getEsbuildExternalModules, runBuilds } from './util';
 
+const screenshotBuilds = {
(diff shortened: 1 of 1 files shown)


rwaskiewicz (approved):

rwaskiewicz:
Non-blocking ask - can you please make sure the commit/pr title doesn't use "fix()" in the title? This probably shouldn't be in the user-facing changelog

rwaskiewicz:
For posterity - I was able to confirm this works by running `npm run build.esbuild` locally

christian-bromann:
> Non-blocking ask - can you please make sure the commit/pr title doesn't use "fix()" in the title?

Squashed commits and renamed commit message.

alicewriteswrongs (approved):

PR #5200, opened 2023-12-21
chore(esbuild): add esbuild-based testing bundle

christian-bromann:
Testing module is compiled with Rollup.

GitHub Issue Number: N/A
STENCIL-993

## What is the new behavior?
Testing module is compiled with ESLint.

- [ ] Yes
- [x] No

## Testing

n/a

## Other information

n/a

### DIFF
--- a/scripts/esbuild/mock-doc.ts
+++ b/scripts/esbuild/mock-doc.ts
@@ -53,7 +53,6 @@ export async function buildMockDoc(opts: BuildOptions) {
     bundle: true,
     alias: mockDocAliases,
     logLevel: 'info',
-    target: 'node16',
   };
 
   const esmOptions: ESBuildOptions = {
--- a/scripts/esbuild/testing.ts
+++ b/scripts/esbuild/testing.ts
@@ -1,0 +1,144 @@
+import type { BuildOptions as ESBuildOptions, Plugin } from 'esbuild';
+import fs from 'fs-extra';
+import { join } from 'path';
+
+import { copyTestingInternalDts } from '../bundles/testing';
+import { getBanner } from '../utils/banner';
+import type { BuildOptions } from '../utils/options';
+import { writePkgJson } from '../utils/write-pkg-json';
+import { getBaseEsbuildOptions, getEsbuildAliases, getEsbuildExternalModules, runBuilds } from './util';
+
+const EXTERNAL_TESTING_MODULES = [
+  'assert',
+  'buffer',
+  'child_process',
+  'console',
+  'constants',
+  'crypto',
+  'fs',
+  '@jest/core',
+  'jest-cli',
+  'jest',
+  'expect',
+  '@jest/reporters',
+  'jest-environment-node',
+  'jest-message-id',
+  'jest-runner',
+  'net',
+  'os',
+  'path',
+  'process',
+  'puppeteer',
+  'puppeteer-core',
+  'readline',
+  'rollup',
+  '@rollup/plugin-commonjs',
+  '@rollup/plugin-node-resolve',
+  'stream',
+  'tty',
+  'url',
+  'util',
+  'vm',
+  'yargs',
+  'zlib',
+];
+
+export async function buildTesting(opts: BuildOptions) {
+  const inputDir = join(opts.buildDir, 'testing');
+  const sourceDir = join(opts.srcDir, 'testing');
+
+  await Promise.all([
+    // copy jest testing entry files
+    fs.copy(join(opts.scriptsBundlesDir, 'helpers', 'jest'), opts.output.testingDir),
+    copyTestingInternalDts(opts, inputDir),
+  ]);
+
+  // write package.json
+  writePkgJson(opts, opts.output.testingDir, {
+    name: '@stencil/core/testing',
+    description: 'Stencil testing suite.',
+    main: 'index.js',
+    types: 'index.d.ts',
+  });
+
+  const external = [
+    ...EXTERNAL_TESTING_MODULES,
+    ...getEsbuildExternalModules(opts, opts.output.testingDir),
+    '@rollup/plugin-commonjs',
+    '@rollup/plugin-node-resolve',
+    'rollup',
+    '../compiler/stencil.js',
+  ];
+
+  const testingEsbuildOptions: ESBuildOptions = {
+    ...getBaseEsbuildOptions(),
+    entryPoints: [join(sourceDir, 'index.ts')],
+    bundle: true,
+    format: 'cjs',
+    outfile: join(opts.output.testingDir, 'index.js'),
+    platform: 'node',
+    logLevel: 'info',
+    external,
+    write: false,
+    alias: getEsbuildAliases(),
+    banner: { js: getBanner(opts, `Stencil Testing`, true) },
+    plugins: [
+      externalAliases('@app-data', '@stencil/core/internal/app-data'),
+      externalAliases('@platform', '@stencil/core/internal/testing'),
+      externalAliases('../internal/testing/index.js', '@stencil/core/internal/testing'),
+      externalAliases('@stencil/core/dev-server', '../dev-server/index.js'),
+      lazyRequirePlugin(
+        opts,
+        [
+          '@stencil/core/internal/app-data',
+          '@stencil/core/internal/testing',
+          '../dev-server/index.js',
+          '../internal/testing/index.js',
+          '../mock-doc/index.cjs'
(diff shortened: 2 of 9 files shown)


scripts/esbuild/testing.ts
```diff
@@ -0,0 +1,141 @@
...
+import fs from 'fs-extra';
+import { join } from 'path';
+
+import { copyTestingInternalDts } from '../bundles/testing';
+import { getBanner } from '../utils/banner';
+import type { BuildOptions } from '../utils/options';
+import { writePkgJson } from '../utils/write-pkg-json';
+import { getBaseEsbuildOptions, getEsbuildAliases, getEsbuildExternalModules, runBuilds } from './util';
+
+const EXTERNAL_TESTING_MODULES = [
```
alicewriteswrongs, line 11: I think the built-in node modules should be marked as external when we set `platform: 'node'` (based [on the esbuild docs](https://esbuild.github.io/api/#platform)) so I think we should only need to put the npm packages in here

alicewriteswrongs: or put another way, did you run into an issue where you needed to specifically mark things like `os` and `path` as external?

scripts/esbuild/testing.ts
```diff
@@ -0,0 +1,141 @@
...
+    main: 'index.js',
+    types: 'index.d.ts',
+  });
+
+  const external = [
+    ...EXTERNAL_TESTING_MODULES,
+    ...getEsbuildExternalModules(opts, opts.output.testingDir),
+    '@rollup/plugin-commonjs',
+    '@rollup/plugin-node-resolve',
+    'rollup',
```
alicewriteswrongs, line 69: I think these are in `EXTERNAL_TESTING_MODULES` already

scripts/esbuild/testing.ts
```diff
@@ -0,0 +1,141 @@
...
+  'url',
+  'util',
+  'vm',
+  'yargs',
+  'zlib',
+];
+
+export async function buildTesting(opts: BuildOptions) {
+  const inputDir = join(opts.buildDir, 'testing');
+  const sourceDir = join(opts.srcDir, 'testing');
```
alicewriteswrongs, line 48: What do you think about adding an `fs.emptyDir` call before we starting writing files, with `opts.output.testingDir` as the argument, so we can be sure the esbuild-based build is 'clean' and we don't have files from previous builds hanging around?

scripts/esbuild/testing.ts
```diff
@@ -0,0 +1,141 @@
...
+  const testingEsbuildOptions: ESBuildOptions = {
+    ...getBaseEsbuildOptions(),
+    entryPoints: [join(sourceDir, 'index.ts')],
+    bundle: true,
+    format: 'cjs',
+    outfile: join(opts.output.testingDir, 'index.js'),
+    platform: 'node',
+    logLevel: 'info',
+    external,
+    write: false,
```
alicewriteswrongs, line 82: just curious, what's the reason for `write: false` here? based on https://esbuild.github.io/api/#write it looks like that should prevent the bundle from being written to disk, but strangely it doesn't seem to do that 🤔

scripts/esbuild/testing.ts
alicewriteswrongs: we do that here, for instance: 

https://github.com/ionic-team/stencil/blob/276cc7370b6bf22326b74f7a314a67a27f23cf7b/scripts/esbuild/sys-node.ts#L17-L18

alicewriteswrongs (changes requested):
a few questions and comments!

scripts/esbuild/testing.ts
```diff
@@ -0,0 +1,141 @@
...
+      externalAliases('../internal/testing/index.js', '@stencil/core/internal/testing'),
+      externalAliases('@stencil/core/dev-server', '../dev-server/index.js'),
+      lazyRequirePlugin(opts, [
+        '@stencil/core/internal/app-data',
+        '@stencil/core/internal/testing',
+        '../dev-server/index.js',
+        '../internal/testing/index.js',
+        '../mock-doc/index.cjs',
+      ]),
+    ],
```
rwaskiewicz, line 97: I'm not sure I fully understand the differences between this section and what we have in `bundles/testing.ts` today https://github.com/ionic-team/stencil/blob/865ae1d36e4d436d9011ddfb09690d41a7709684/scripts/bundles/testing.ts#L80-L83

Specifically: Why do we not externalize `@stencil/core/mock-doc` to `../mock-doc/index.cjs` (like we do for the ESBuild scripts)? Do we no longer need this?

rwaskiewicz (changes requested):
Mostly questions to better my own understanding.

Can you do me a favor and add steps for testing this change to the PR summary?

scripts/esbuild/mock-doc.ts
```diff
@@ -50,7 +50,6 @@ export async function buildMockDoc(opts: BuildOptions) {
     bundle: true,
     alias: mockDocAliases,
     logLevel: 'info',
-    target: 'node16',
```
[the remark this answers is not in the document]
alicewriteswrongs: I'm not sure, it's possible there was some reason I had this where it was emitting JS that didn't work with node 16 _without_ it but I fear whatever justification there was may have been lost to the sands of time (we're talking about 3 months ago here)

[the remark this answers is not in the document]
alicewriteswrongs: if it works well on our supported version without this I have no concern about removing it

scripts/esbuild/testing.ts
christian-bromann: We have to do some last modifications on the output code in `lazyRequirePlugin` to enable lazy require to work. I am not sure if there is a better way do to that but it seems like I can either have a plugin that reads the file from disk to modify the source and push it along the pipeline or do it before it is being written to disk.

scripts/esbuild/mock-doc.ts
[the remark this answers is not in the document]
christian-bromann: I created the following function:

```ts
export function getEsbuildTargets(): string[] {
  return ['node16', 'chrome79', 'edge79', 'firefox70', 'safari14'];
}
```
to ensure we set the targets based on what we need to support. I thought unifying the target property simplifies things but I can see that it might create confusion why we would set `node16` and other targets for something we only run in Node.js.

Happy to revert this change but it I think it doesn't change anything on the resulting mock-doc bundle.

scripts/esbuild/testing.ts
christian-bromann: I saw that a bunch have been already defined in [`externalNodeModules`](https://github.com/ionic-team/stencil/blob/2131f07bc30786669a34ac6dbef526d3cdfe0c86/scripts/esbuild/util.ts#L48) so will remove them from the list 👌

scripts/esbuild/testing.ts
christian-bromann: Update: I was assuming that `mock-doc` was externalised but further investigation has shown that it is part of the bundle. I debugged through the build to find why that is and hope to find a resolution today.

scripts/esbuild/testing.ts
christian-bromann: added

```ts
await fs.emptyDir(opts.output.testingDir);
```

scripts/esbuild/util.ts
```diff
@@ -112,5 +112,15 @@ export function getBaseEsbuildOptions(): ESBuildOptions {
...
   };
 }
+
+/**
+ * Get build targets with minimal supported browser versions
+ * @see https://stenciljs.com/docs/support-policy#browser-support
+ * @returns an array of build targets
+ */
+export function getEsbuildTargets(): string[] {
+  return ['node16', 'chrome79', 'edge79', 'firefox70', 'safari14'];
```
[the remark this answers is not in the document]
christian-bromann: My intention is to have it streamlined across all builds to add more clarity. While the runtime for the compiler is obvious it might not be for other parts, e.g. testing. I don't believe adding browser build targets to outputs that run in Node.js only has a big effect if any at all as we aren't using any WebAPIs in the source anyway.

I don't have a strong opinion on this one and happy to revert it if we prefer to keep specific targets for specific outputs.

tanner-reits (approved):
Nothing in addition to Ryan and Alice's reviews. Only thing I'm not clear on is the multiple targets (mainly the browser specific ones) for `getEsBuildTargets`

Issue #4866, opened 2023-09-29
bug: getMismatchedPixels timeout when using clip options in compareScreenshot


Spirielle:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

3.4.0

### Current Behavior

We have e2e tests with screenshots failing with the error "Error: Failed: "getMismatchedPixels timeout: 15000ms"
I have noticed that it happens only when
1. There's a visual difference between the 2 screenshots 
AND
2. The test uses the "clip" option. For example like this:
`await page.compareScreenshot('', { clip: { x: 0, y: 0, width: 800, height: 800 } });`

### Expected Behavior

We expect e2e with screenshots using clip options to be comparable.

### Steps to Reproduce

- Use clip option in a e2e with compare screenshot
 `await page.compareScreenshot('', { clip: { x: 0, y: 0, width: 800, height: 800 }` });
- Run test a first time to have an initial screenshot
- Change e2e slightly to trigger a visual change
- Run test a second time -> error occurs

### Code Reproduction URL

https://github.com/Spirielle/bug-stencil-getMismatchedPixels

### Investigation done so far
**About \# 1,** I believe it only happens when there's a difference because when the images are identical, the code which compares pixels is not reached. [See here](https://github.com/ionic-team/stencil/blob/4cd86f350d1840fd61db95dc267684faa540caa7/src/screenshot/screenshot-compare.ts#L89C2-L89C2) 

**About \# 2,** I think the width and height given to pixel-match does not take the clip options into consideration.

We use the clip options when calling the function CompareScreenshots [here](https://github.com/ionic-team/stencil/blob/cb3eac3a2b7795a5c8d3f9100812337d1a8d8a9d/src/testing/puppeteer/puppeteer-screenshot.ts#L117)

However, it seems we don't use the clip size when calling pixelmatch [here](https://github.com/ionic-team/stencil/blob/4cd86f350d1840fd61db95dc267684faa540caa7/src/screenshot/screenshot-compare.ts#L108)

[...]

PR #5205, opened 2023-12-22
fix(screenshot): recognise clip options

### LINKED ISSUE #5208: bug: Image data size does not match width/height in compareScreenshot
### Prerequisites

### Stencil Version

4.9.0

### Current Behavior

e2e tests with screenshots failing with the error "Image data size does not match width/height." when calling method `compareScreenshot` .

I have noticed that it happens only when

1. There's a visual difference between the 2 screenshots
AND
2. The screenshot is taken in a different size than 600 x 600.
    
   
So it happens if we don't change anything when calling `compareScreenshot` like this:
` await page.compareScreenshot();`
Or if the image is clipped to something else than 600 x 600, for example:
` await page.compareScreenshot('', { clip: { x: 0, y: 0, width: 800, height: 800 } });` 
 
 However, using the following does not cause any error:
`  await page.compareScreenshot('', { clip: { x: 0, y: 0, width: 600, height: 600 } });` 
    
I noticed that when not setting any clip size, the screenshot saved in my screnshot folder is 800 x 600.

### Expected Behavior

When the 2 screenshots compared are of the same size, I would expect no unhandled exception when calling `compareScreenshot`. 

### System Info

_No response_

### Steps to Reproduce

- Use clip option in a e2e with compare screenshot
 `await page.compareScreenshot('', { clip: { x: 0, y: 0, width: 800, height: 800 }` });
- Run test a first time to have an initial screenshot
- Change e2e slightly to trigger a visual change
- Run test a second time -> error occurs

### Code Reproduction URL

https://github.com/Spirielle/bug-stencil-getMismatchedPixels

### Additional Information

This issue follows this other issue: https://github.com/ionic-team/stencil/issues/4866

In the other one, the same underlying error caused a timeout when calling compareScreenshot.
It has been fixed, so we fortunately do not get any timeout anymore.

christian-bromann:
Currently the `ScreenshotOptions` passed into the `compareScreenshot` command aren't recognised as fully as `width` and `height` are overwritten by the value stored in the `emulateConfig` which is hard coded in [`validateConfig`](https://github.com/ionic-team/stencil/blob/bb0b1d46f63175dc09d0a23445be4d4a0d891a01/src/compiler/config/validate-testing.ts#L173-L174).

refs #4866
fixes #5208

## What is the new behavior?
Just take `width` and `height` provided by the command.

- [ ] Yes
- [x] No

## Testing

To verify this change:

- `git clone git@github.com:Spirielle/bug-stencil-getMismatchedPixels.git`
- `cd bug-stencil-getMismatchedPixels`
- `npm i`
- `npm test` -> test pass
- go into `src/components/my-component/my-component.e2e.ts` and comment out line 13 and comment in line 16
- `npm run test` -> test fail due to `Image data size does not match width/height.`

This is because the clip parameter were not recognised so the image size differs from the baseline image.

Verify fix:

- install dev build: `npm install @stencil/core@4.9.1-dev.1704841621.8f08cf1`
- `npm test` -> test now pass

## Other information

n/a

### DIFF
--- a/src/screenshot/screenshot-compare.ts
+++ b/src/screenshot/screenshot-compare.ts
@@ -104,15 +104,11 @@ export async function compareScreenshot(
       // compare the two images pixel by pixel to
       // figure out a mismatch value
 
-      // figure out the actual width and height of the screenshot
-      const naturalWidth = Math.round(emulateConfig.viewport.width * emulateConfig.viewport.deviceScaleFactor);
-      const naturalHeight = Math.round(emulateConfig.viewport.height * emulateConfig.viewport.deviceScaleFactor);
-
       const pixelMatchInput: d.PixelMatchInput = {
         imageAPath: join(screenshotBuildData.imagesDir, screenshot.diff.imageA),
         imageBPath: join(screenshotBuildData.imagesDir, screenshot.diff.imageB),
-        width: naturalWidth,
-        height: naturalHeight,
+        width,
+        height,
         pixelmatchThreshold: pixelmatchThreshold,
       };
 
--- a/src/screenshot/screenshot-compare.ts
+++ b/src/screenshot/screenshot-compare.ts
@@ -107,8 +107,8 @@ export async function compareScreenshot(
       const pixelMatchInput: d.PixelMatchInput = {
         imageAPath: join(screenshotBuildData.imagesDir, screenshot.diff.imageA),
         imageBPath: join(screenshotBuildData.imagesDir, screenshot.diff.imageB),
-        width,
-        height,
+        width: Math.round(width),
+        height: Math.round(height),
         pixelmatchThreshold: pixelmatchThreshold,
       };
 


src/screenshot/screenshot-compare.ts
```diff
@@ -104,15 +104,11 @@ export async function compareScreenshot(
...
-      const naturalWidth = Math.round(emulateConfig.viewport.width * emulateConfig.viewport.deviceScaleFactor);
-      const naturalHeight = Math.round(emulateConfig.viewport.height * emulateConfig.viewport.deviceScaleFactor);
-
       const pixelMatchInput: d.PixelMatchInput = {
         imageAPath: join(screenshotBuildData.imagesDir, screenshot.diff.imageA),
         imageBPath: join(screenshotBuildData.imagesDir, screenshot.diff.imageB),
-        width: naturalWidth,
-        height: naturalHeight,
+        width,
+        height,
```
rwaskiewicz, line 111: GH's having some issues this morning, the first time I submitted something I got a 5xx error. I don't think the original went through, but if it did, apologies if this is a duplicate review!

Previously, we were rounding out the values to the nearest integer for this value - for my own understanding:
1. Does it matter if floats can be provided here?
2. If so, do we have any guarantee the values of `width` or `height` are integers?

rwaskiewicz (changes requested):
Can you do me a favor and update the testing instructions in the PR summary? Right now, they allude to a few steps to take care of in the new year

src/screenshot/screenshot-compare.ts
christian-bromann: > 1. Does it matter if floats can be provided here?

I don't believe so, let me verify that this doesn't cause an issue.

> 2\. If so, do we have any guarantee the values of `width` or `height` are integers?

Since this is user input we can't guarantee the values to be integer. Let me verify if using float values has an impact on pixelmatch and if so add a warning or error out in case a float value was provided.

christian-bromann: I was able to verify that a rounding operation is done when storing the images causing a failure when comparing the image sizes with float values. I pushed a patch to round values and everything works as expected. Thanks!

christian-bromann:
@rwaskiewicz I created a dev build `4.9.1-dev.1704841621.8f08cf1` , mind triggering a Ionic build with it?

rwaskiewicz:
@christian-bromann Sent you a DM with the action run

christian-bromann:
@rwaskiewicz I updated the description to help verify this change. Please let me know if you can verify the resolution.

I've discovered some more problems in this area, e.g. not providing a clip property at all causes issues too since Puppeteer starts headless with a 800x600px size while the default width/height properties are 600x600. So while this patch [fixes](https://github.com/ionic-team/stencil/issues/4866#issuecomment-1868140405) the users issue, there are certainly more problems in this area.

rwaskiewicz (approved):
Works as expected!

christian-bromann:
The fix for this issue has been released as a part of today's [Stencil v4.11.0 release](https://github.com/ionic-team/stencil/releases/tag/v4.11.0).

PR #5218, opened 2024-01-02
chore(repo): prevent renovate from upgrading open

rwaskiewicz:

Renovate tries to upgrade `open`
GitHub Issue Number: N/A

## What is the new behavior?

Open v10 has a single breaking change, in that it requires Node v18+ to be used. Stencil v4 still supports Node 16, and must continue to use Open v9 as a result. We cannot drop support for Node 16 without a breaking change, so we’ll hold off on this work until Stencil v5.

- [ ] Yes
- [x] No

## Testing

We can't fully test this until it's merged unfortunately.
I did validate the config is correct locally with: `npm i -g renovate && renovate-config-validator`.

## Other information

Replaces: #5213

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -139,6 +139,11 @@
       matchPackageNames: ['rollup'],
       matchPackagePrefixes: ['@rollup'],
       groupName: 'rollup,'
+    },
+    // TODO(STENCIL-1088): remove once support for Node v16 is dropped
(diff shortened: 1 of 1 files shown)


alicewriteswrongs (approved):
looks good!

tanner-reits (approved):

PR #5219, opened 2024-01-02
chore(esbuild): ensure that compiler dir is empty before we run esbuild

alicewriteswrongs:
This adds a call to the fs-extra `emptyDir` before we run the whole esbuild-based bundle and build step for the `compiler/` and `mock-doc/` dirs to ensure that there's nothing hanging around from a previous build.

Just something I noticed was missing when reviewing #5200!

Currently we don't clear out the previous build before we run esbuild to bundle the compiler. Since we normally run the rollup-based build first (because we don't yet build everything with esbuild) this could potentially mean something from the rollup build is sticking around for the `compiler/` bundle, which could in turn give us a false confidence in the ability of the esbuild-based script to stand alone. This fixes that issue!

## What is the new behavior?

we call `fs.emptyDir` when we should.

- [ ] Yes
- [x] No

## Testing

Exercise the relevant functionality after building with `esbuild`.

To do so, you should be able to just

```sh
npm run build.esbuild && npm pack
```

and then install into a sample project. If the tests run and things are compiling and whatnot then I think we're good!

### DIFF
--- a/scripts/esbuild/compiler.ts
+++ b/scripts/esbuild/compiler.ts
@@ -17,6 +17,9 @@ export async function buildCompiler(opts: BuildOptions) {
   const compilerFileName = 'stencil.js';
   const compilerDtsName = compilerFileName.replace('.js', '.d.ts');
 
+  // clear out rollup stuff and ensure directory exists
+  await fs.emptyDir(opts.output.compilerDir);
+
   // create public d.ts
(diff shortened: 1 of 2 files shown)


rwaskiewicz (approved):
Non-blocking ask - can you make sure the commit msg + pr name doesn't use 'fix()'? Otherwise, that'll show up in the changelogs as a user-facing fix

christian-bromann (approved):
👍

PR #5220, opened 2024-01-03
feat(runtime): add extras flag for scoped slot changes

tanner-reits:

We made changes in #5146 and #5135 to alter the behavior of `slot`s inside scoped components. The changes aligned Stencil's slot polyfill with behavior in the native shadow DOM and were placed behind the `experimentalSlotFixes` extra flag in a project's Stencil config. However, these changes negatively impacted the Ionic Framework causing automated tests to fail.

GitHub Issue Number: N/A

## What is the new behavior?

This commit adds a new extras flag, `experimentalScopedSlotChanges`, that is used to change logic in a few places to align Stencil's `scoped` encapsulation to work more like the native Shadow DOM when using slots.

Changes in #5146 and #5135 are changed from being gated by the `experimentalSlotFixes` flag to this new flag so they can be toggled independently.

- [ ] Yes
- [X] No

This will be considered a breaking change once we remove the ability to opt-in/opt-out

## Testing

**Manual tests**
Changes can be tested by enabling `experimentalScopedSlotChanges` in a starter project's Stencil config and following the test cases from either #5146 and/or #5135

**Automated tests**
Stencil unit and e2e tests continue to pass (with the new flag enabled). Ionic Framework screenshot tests pass with the flag disabled (expected). 

## Other information

We'll need to coordinate with the Framework team to update their components before we can enable this flag by default (and eventually remove it altogether).

Related doc changes: https://github.com/ionic-team/stencil-site/pull/1307

### DIFF
--- a/src/compiler/app-core/app-data.ts
+++ b/src/compiler/app-core/app-data.ts
@@ -169,6 +169,8 @@ export const updateBuildConditionals = (config: ValidatedConfig, b: BuildConditi
   b.slotChildNodesFix = config.extras.slotChildNodesFix;
   // TODO(STENCIL-914): remove this option when `experimentalSlotFixes` is the default behavior
   b.experimentalSlotFixes = config.extras.experimentalSlotFixes;
+  // TODO(STENCIL-1086): remove this option when it's the default behavior
+  b.experimentalScopedSlotChanges = config.extras.experimentalScopedSlotChanges;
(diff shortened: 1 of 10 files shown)


rwaskiewicz (approved):

christian-bromann (approved):
👍

PR #5221, opened 2024-01-03
fix(runtime): slot regressions from experimental slot fixes

tanner-reits:

A few changes made behind the `experimentalSlotFixes` flag were resulting in screenshot test failures in the Ionic Framework

GitHub Issue Number: N/A

## What is the new behavior?

- Fixes the conditional responsible for patching element methods for the pseudo shadow DOM when `experimentalSlotFixes` is enabled.
- Updated the `appendChild` patch to force a re-render of the patched element

See commit messages for more information

- [ ] Yes
- [X] No

## Testing

**Automated tests**
- Stencil unit and e2e tests continue to pass.
- A build of Stencil containing these changes (and changes from the other "slot regressions" branches) was used in a manually triggered run of the Stencil nightly build and succeeded. Run results: https://github.com/ionic-team/ionic-framework/actions/runs/7280624127

### DIFF
--- a/src/runtime/dom-extras.ts
+++ b/src/runtime/dom-extras.ts
@@ -88,7 +88,7 @@ export const patchSlotAppendChild = (HostElementPrototype: any) => {
 
       // Check if there is fallback content that should be hidden
       updateFallbackSlotVisibility(this);
-      // Fore a re-render of the host element
+      // Force a re-render of the host element
       forceUpdate(this);
 
       return insertedNode;
--- a/src/runtime/dom-extras.ts
+++ b/src/runtime/dom-extras.ts
@@ -1,5 +1,5 @@
 import { BUILD } from '@app-data';
-import { getHostRef, plt, supportsShadow } from '@platform';
+import { forceUpdate, getHostRef, plt, supportsShadow } from '@platform';
 import { NODE_TYPES } from '@stencil/core/mock-doc';
 import { CMP_FLAGS, HOST_FLAGS } from '@utils';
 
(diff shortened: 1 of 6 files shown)


src/runtime/dom-extras.ts
```diff
@@ -84,10 +84,14 @@ export const patchSlotAppendChild = (HostElementPrototype: any) => {
     if (slotNode) {
       const slotChildNodes = getHostSlotChildNodes(slotNode, slotName);
       const appendAfter = slotChildNodes[slotChildNodes.length - 1];
-      appendAfter.parentNode.insertBefore(newChild, appendAfter.nextSibling);
+      const insertedNode = appendAfter.parentNode.insertBefore(newChild, appendAfter.nextSibling);
+
       // Check if there is fallback content that should be hidden
       updateFallbackSlotVisibility(this);
-      return;
+      // Fore a re-render of the host element
```
rwaskiewicz, line 91: Non-blocking nit:
```suggestion
      // Force a re-render of the host element
```

src/runtime/dom-extras.ts
```diff
@@ -84,10 +84,14 @@ export const patchSlotAppendChild = (HostElementPrototype: any) => {
...
+      forceUpdate(this);
```
rwaskiewicz, line 92: Can you help me understand why we need to do force an update here?

test/karma/test-app/dom-reattach-clone/cmp-deep-slot.tsx
```diff
@@ -2,6 +2,7 @@ import { Component, h } from '@stencil/core';
 
 @Component({
   tag: 'dom-reattach-clone-deep-slot',
+  scoped: true,
```
rwaskiewicz, line 5: For my own understanding, why do we need to add `scoped` to the components in this test suite?

rwaskiewicz (changes requested):

src/runtime/dom-extras.ts
rwaskiewicz: Note: I want a little additional clarification here, after reading the git commit msg

christian-bromann (approved):
Don't anything in addition to what @rwaskiewicz already pointed out 👍

test/karma/test-app/dom-reattach-clone/cmp-deep-slot.tsx
tanner-reits: The test uses the `cloneNode` method on a Stencil component, but with the changes to the bootstrap methods in this PR, we only patch native methods for `scoped` components when `experimentalSlotFixes` is enabled.

src/runtime/dom-extras.ts
tanner-reits: I ran into an issue when testing this out with Framework screenshot tests. For some components that had `appendChild` patched, the component would include the append child node after the call until something triggered a re-render of the component (this was as simple as hovering the component). This made it so the component would immediately re-render to reflect the changes and get the tests to pass

rwaskiewicz (approved):

PR #5222, opened 2024-01-03
fix(runtime): revert slot relocation forwarding

### LINKED ISSUE #5215: bug: Slot Attribute is lost
### Prerequisites

### Stencil Version

4.8.2

### Current Behavior

if i set in the stencil config `experimentalSlotFixes: true` the <div slot="content" /> is relocated correctly but its loses the attribute [slot].

```html
<my-comp>
<div slot="content">Hello</div>
</my-comp>
```

-> will render into:

```html
<my-comp>
<div>Hello</div>
</my-comp>
```

### Expected Behavior

```html
<my-comp>
<div slot="content">Hello</div>
</my-comp>
```

-> will render into:

```html
<my-comp>
<div slot="content">Hello</div>
</my-comp>
```

### System Info

```shell
Node 18, Windows 11, Chrome 120
```

### Steps to Reproduce

Create a snippet with slot attribute. After first render inspect in browser the rendered component and the slot attribute is lost.
Without the experiemental flag (the default) the slot attribute is every time there.

### Code Reproduction URL

https://stackblitz.com/edit/node-uahzga?file=src%2Findex.html,src%2Fcomponents%2Fmy-comp.tsx

### Additional Information

in the 4.8.2 commit i see only a default slot unit test. maybe a named slot unit test will show this issue aswell :)
https://github.com/ionic-team/stencil/commit/4303d6af1bbcd995e3e02891b5e50768e8eeaffd

cheers

tanner-reits:

Changes made in #4940 and #4993 are causing issues in the Ionic Framework screenshot testing with `experimentalSlotFixes` enabled. With this behavior, the `slot` attribute can get removed from elements which is an issue if styles target the attribute (i.e. something like `::slotted([slot="start"])`).

Fixes #5215

## What is the new behavior?

This commit reverts some logic in the runtime attempting to handle the use case of slots being slotted into other slots where the slot name changes along the way. The Stencil runtime cannot handle this use case elegantly because it relocates the node in the DOM rather than having references to nodes like the native Shadow DOM.

For this use case, it is recommended to either use shadow encapsulation for all components involved, or to keep the slot name consistent through the component tree.

Reverts 21b3bf0fcbfdaee1aef6df8f8b01cd112c96da38, 9adcfab1c184db84c644a09ad85289b5db50376a, and part of 0106e0b01f038349796cba436d8625559be41e26

- [ ] Yes
- [X] No

Not a breaking change since this behavior was being the `experimentalSlotFixes` flag anyway

## Testing

Automated tests are passing and Framework screenshot tests are passing as can be seen in [this run](https://github.com/ionic-team/ionic-framework/actions/runs/7280624127) of the nightly workflow

## Other information

Stencil is not setup to handle this use case with the current slot polyfill logic. This limitation is called out in the Stencil docs: https://github.com/ionic-team/stencil-site/pull/1306


christian-bromann (approved):

dutscher:
maybe a unit test like in this commit? 
https://github.com/ionic-team/stencil/commit/4303d6af1bbcd995e3e02891b5e50768e8eeaffd#diff-7ac2ac45ec60f7bfdb3329273c9d1ce24b0b508e62ef72623876b84521c611c1

cheers and thanks for your effort

src/runtime/vdom/vdom-render.ts
```diff
@@ -192,13 +189,6 @@ const relocateToHostRoot = (parentElm: Element) => {
...
-        // When putting an element node back in its original location,
-        // we need to reset the `slot` attribute back to the value it originally had
-        // so we can correctly relocate it again in the future
-        if (childNode.nodeType === NODE_TYPE.ElementNode && !!childNode['s-sn']) {
-          childNode.setAttribute('slot', childNode['s-sn']);
-        }
-
         // Need to tell the render pipeline to check to relocate slot content again
         checkSlotRelocate = true;
       }
```
rwaskiewicz, line 194: I can't comment on the line itself, but do we still need [this if statement](https://github.com/ionic-team/stencil/pull/5222/files#diff-4828a657f521302051295a582e81299dde7354857095b41f6ae9b062d835670fL233), also introduced in #4993?

tanner-reits: Yep that can go too. Good call. Merged into commit d2ab3314e7117f370d049dc1b600d69074774328

rwaskiewicz (approved):

PR #5231, opened 2024-01-04
chore(snc): fix snc errors in vdom-annotations.ts

alicewriteswrongs:
Just fixing a few SNC errors in `src/runtime/vdom/vdom-annotations.ts`. These mainly related to possibly-undefined or possibly-null values which could be addressed with optional chaining.

We have a fair number of snc errors in this file

## What is the new behavior?

I don't believe this will change any behavior, other than possibly preventing some runtime errors. I think though that, as is often the case in the Stencil codebase, things are typed `Foo | undefined` when they're always `Foo` in practice, so I don't think we'll have too much to worry about.

- [ ] Yes
- [x] No

## Testing

I think if the automated tests are all passing we should be good.

### DIFF
--- a/src/runtime/vdom/vdom-annotations.ts
+++ b/src/runtime/vdom/vdom-annotations.ts
@@ -32,7 +32,7 @@ export const insertVdomAnnotations = (doc: Document, staticComponents: string[])
     parseVNodeAnnotations(doc, doc.body, docData, orgLocationNodes);
 
     orgLocationNodes.forEach((orgLocationNode) => {
-      if (orgLocationNode != null) {
+      if (orgLocationNode != null && orgLocationNode['s-nr']) {
         const nodeRef = orgLocationNode['s-nr'];
 
         let hostId = nodeRef['s-host-id'];
@@ -49,7 +49,7 @@ export const insertVdomAnnotations = (doc: Document, staticComponents: string[])
             nodeRef.setAttribute(HYDRATE_CHILD_ID, childId);
           } else if (nodeRef.nodeType === NODE_TYPE.TextNode) {
             if (hostId === 0) {
-              const textContent = nodeRef.nodeValue.trim();
+              const textContent = nodeRef.nodeValue?.trim();
               if (textContent === '') {
                 // useless whitespace node at the document root
                 orgLocationNode.remove();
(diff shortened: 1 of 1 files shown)


src/runtime/vdom/vdom-annotations.ts
```diff
@@ -35,8 +35,8 @@ export const insertVdomAnnotations = (doc: Document, staticComponents: string[])
       if (orgLocationNode != null) {
         const nodeRef = orgLocationNode['s-nr'];
 
-        let hostId = nodeRef['s-host-id'];
-        let nodeId = nodeRef['s-node-id'];
+        let hostId = nodeRef?.['s-host-id'];
```
christian-bromann, line 38: I wonder if we should just not enter the if block entirely if `nodeRef` is undefined, e.g.

```patch
- if (orgLocationNode != null) {
+ if (orgLocationNode != null && orgLocationNode['s-nr']) {
```

christian-bromann (approved):
One small nit, not blocking.

src/runtime/vdom/vdom-annotations.ts
```diff
@@ -35,8 +35,8 @@ export const insertVdomAnnotations = (doc: Document, staticComponents: string[])
...
+        let nodeId = nodeRef?.['s-node-id'];
         let childId = `${hostId}.${nodeId}`;
```
rwaskiewicz, line 40: With these changes, it's possible for [`childId`](https://github.com/ionic-team/stencil/pull/5231/files#diff-330540c04820aa83be13838701b0dab6068df5ba4a560f7d24382e247f7816edR40) to be assigned to the string `"undefined.undefined"` (as opposed to the runtime failing on attempting to access a property on an `undefined` object (`nodeRef`)). 

We could introduce some funky cases where [`orgLocationNodeId` gets assigned](https://github.com/ionic-team/stencil/pull/5231/files#diff-330540c04820aa83be13838701b0dab6068df5ba4a560f7d24382e247f7816edR67) to either:
- `undefined.undefined.`
- `undefined.undefined.c`

This _probably_ isn't happening today, otherwise I'd expect we'd be seeing a lot more issues in the GH repo.  

Do we know if it's possible for `orgLocationNode.parentElement` to be truthy, but `orgLocationNode['s-nr']` to be falsy? If not, we can probably proceed with @christian-bromann's [change proposal](https://github.com/ionic-team/stencil/pull/5231/files#r1443040184)

rwaskiewicz (changes requested):

src/runtime/vdom/vdom-annotations.ts
alicewriteswrongs: yeah I think that makes sense to me, I'll make that change

src/runtime/vdom/vdom-annotations.ts
alicewriteswrongs: followed @christian-bromann's suggestion!

rwaskiewicz (approved):

PR #5232, opened 2024-01-05
chore(ci): add a version number output from the release-dev task

alicewriteswrongs:
This adds an output from the `release-dev` workflow which will let other calling workflows get the version which has just been released.

Right now if the `release-dev` workflow is called by another workflow there isn't a way for the calling workflow to get the version number which was just released.

## What is the new behavior?

This adds an `output` to the whole workflow so that a calling workflow can get the version number, for instance so that something like #5229 can run a dev release, get the version number, and then write a comment out to the user.

See the documentation for workflow outputs (as opposed to _job_ outputs) here: https://docs.github.com/en/actions/using-workflows/reusing-workflows#using-outputs-from-a-reusable-workflow

- [ ] Yes
- [x] No

## Testing

There's not a great way to test this without it being on `main` unfortunately.

However! I think that:

- this is a pretty straightforward change and I believe is something we can hand-check against the [relevant documentation](https://docs.github.com/en/actions/using-workflows/reusing-workflows#using-outputs-from-a-reusable-workflow)
- we could introduce this on `main` and then test it (I think we should be able to see the outputs using [debug logging](https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/enabling-debug-logging))

### DIFF
--- a/.github/workflows/release-dev.yml
+++ b/.github/workflows/release-dev.yml
@@ -4,6 +4,11 @@ on:
   workflow_dispatch:
     # Make this a reusable workflow, no value needed
     # https://docs.github.com/en/actions/using-workflows/reusing-workflows
+  workflow_call:
+    outputs:
+      dev-version:
+        description: The version that was just published to npm.
+        value: ${{ jobs.get-dev-version.outputs.dev-version }}
 
 jobs:
   build_core:


.github/workflows/release-dev.yml
```diff
@@ -4,6 +4,10 @@ on:
   workflow_dispatch:
     # Make this a reusable workflow, no value needed
     # https://docs.github.com/en/actions/using-workflows/reusing-workflows
+    outputs:
+      dev-version:
+        description: The version that was just published to npm.
+        value: ${{ jobs.get-dev-version.dev-version }}
```
rwaskiewicz, line 10: Do we need to add `outputs` to this value? E.g.
```suggestion
        value: ${{ jobs.get-dev-version.outputs.dev-version }}
```

The docs you linked in the PR summary seem to reference/use that key/accessor like so (there wasn't a good way to reference this code block in the docs, inlined with comments by me):

```yaml
name: Reusable workflow

on:
  workflow_call:
    # Map the workflow outputs to job outputs
    outputs:
      firstword:
        description: "The first output string"
        value: ${{ jobs.example_job.outputs.output1 }} # <- Uses outputs
      secondword:
        description: "The second output string"
        value: ${{ jobs.example_job.outputs.output2 }} <- Also uses outputs

jobs:
  example_job:
    name: Generate output
    runs-on: ubuntu-latest
    # Map the job outputs to step outputs
    outputs: # <- outputs defined on the job level 
      output1: ${{ steps.step1.outputs.firstword }} 
      output2: ${{ steps.step2.outputs.secondword }}
    steps:
      - id: step1
        run: echo "firstword=hello" >> $GITHUB_OUTPUT
      - id: step2
        run: echo "secondword=world" >> $GITHUB_OUTPUT
```

alicewriteswrongs: oh yup we probably do!

rwaskiewicz:
@alicewriteswrongs have you tried cutting a release from this branch? This shouldn't try to pull code from `main` (including the GH Actions infrastructure). I [just gave it a go](https://github.com/ionic-team/stencil/actions/runs/7423239333) to check my assumption there and hit an error:

> Invalid workflow file: .github/workflows/release-dev.yml#L8
The workflow is not valid. .github/workflows/release-dev.yml (Line: 8, Col: 16): Unrecognized named-value: 'jobs'. Located at position 1 within expression: jobs.get-dev-version.outputs.dev-version

alicewriteswrongs:
yeah I'm trying to run the same workflow in my test repo (https://github.com/alicewriteswrongs/test-pr-comment-action/) and getting the same error there, which I don't understand based on the docs: https://github.com/alicewriteswrongs/test-pr-comment-action/actions/runs/7423347859/workflow

I'll ping here when I figure that out

alicewriteswrongs:
oh it was just a typo, `workflow_call` vs `workflow_dispatch`

alicewriteswrongs:
Ah so unfortunately the outputs are only set when the workflow is called from another workflow (in the `workflow_call`) so if it's called via `workflow_dispatch` (i.e. by calling it from the GH actions web interface I don't think it will set them

alicewriteswrongs:
So I don't think we have a great way to test this unless we want to add a test workflow that will call it

I'm going to try to get a mockup of all of it working in my test repo for now

rwaskiewicz (approved):

christian-bromann (approved):

PR #5241, opened 2024-01-08
chore(deps): upgrade rollup/plugin-replace to v5

rwaskiewicz:

We're on quite an old version of `@rollup/plugin-replace`! The version here is from 2020 😱 

GitHub Issue Number: N/A

## What is the new behavior?

upgrade stencil's dependency on `@rollup/plugin-replace` from v2.x to
v5.x. this package is being done as a part of a larger effort to upgrade
stencil's version of rollup to v2 (which itself is a part of a larger
effort to move to v4). this package has been updated on its own, as:

1. doing so allows us to cut scope/size of other upgrade PRs
2. this package relies on rollup v1.20.0+, and no other sibling
   packages, making this upgrade less involved than others.

update replace plugin calls to use preventAssignment.

`preventAssignment` was added in rollup v2.40.0. when we attempt to
build stencil with the latest v5.x release of the replace-plugin, we
received the following error:

```
(!) Plugin replace: @rollup/plugin-replace: 'preventAssignment'
currently defaults to false. It is recommended to set this option to
`true`, as the next major version will default this option to `true`.
```

explicitly set it to suppress this error. doing so does not change the
stencil binary itself, nor does it appear to change the output of the
generated code.

remove @rollup/plugin-replace & rollup-plugin-sourcemaps from renovate config

remove the former so that renovate can begin to update this package. do
the same for `rollup-plugin-sourcemaps`, which is already up to date.

- [ ] Yes
- [x] No

### Stencil Output

I cut three different local builds of Stencil using `npm run clean && npm ci && npm run build`:
1. One from `main` (https://github.com/ionic-team/stencil/commit/af6e45bc87e18c44da841a1abe11c4d388957326)
2. One prior to setting `preventAssignment` (commit 1)
3. One after setting `preventAssignment` (commit 2)

Comparing the Builds:

1 vs 2 - In upgrading the plugin, a good deal of boilerplate was able to be removed that the plugin was inlining. This led to deduplicated identifiers (e.g. `Chunk$2`) to have their numeric aspect decremented (and in some cases, removed - e.g. `Chunk$2` -> `Chunk$1` || `Chunk$1` -> `Chunk`). There's some ~460 instances of this occurring, but all seem to map back to the deleted code

2 vs 3 - Only code related to inserting the `preventAssignment` configurations have changed. To test this further, I needed to test with the Ionic Framework.

### Ionic Framework

I built framework with Stencil v4.9.1 and [a build of this branch](
https://github.com/ionic-team/stencil/actions/runs/7452395185). The only differences between the generated code was the Stencil version between the two builds.

I [ran the Stencil Nightly Build with this build](https://github.com/ionic-team/ionic-framework/actions/runs/7452455311) and everything passed.
 
## Other information

**NOTE TO SELF**: SQUASH THIS PLEASE

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,7 +16,7 @@
         "@rollup/plugin-commonjs": "15.1.0",
         "@rollup/plugin-json": "6.1.0",
         "@rollup/plugin-node-resolve": "9.0.0",
-        "@rollup/plugin-replace": "2.3.4",
+        "@rollup/plugin-replace": "5.0.5",
         "@rollup/pluginutils": "5.1.0",
         "@types/eslint": "^8.4.6",
         "@types/exit": "^0.1.31",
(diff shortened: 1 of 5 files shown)


alicewriteswrongs (approved):
lgtm!

christian-bromann (approved):
👍

PR #5264, opened 2024-01-16
chore(compiler): resolve strictNullCheck for element-decorator

rwaskiewicz:

We have many strictNullChecks violations

GitHub Issue Number: N/A

## What is the new behavior?

fix a single strictNullCheck for the file by adding a type guard to the final `filter` call against a class's decorated members. this allows the typescript compiler to understand that the collection `elementRefs` is a collection of `string`, rather than a `(string | null)[]`, based on the return type of the local function `parseElementDecorator`

- [ ] Yes
- [x] No

## Testing

This only affects the type system, and therefore shouldn't change any runtime values. As such, I only tested this by running our strict null checks counter, unit and karma tests locally.

### DIFF
--- a/src/compiler/transformers/decorators-to-static/element-decorator.ts
+++ b/src/compiler/transformers/decorators-to-static/element-decorator.ts
@@ -14,7 +14,7 @@ export const elementDecoratorsToStatic = (
   const elementRefs = decoratedMembers
     .filter(ts.isPropertyDeclaration)
     .map((prop) => parseElementDecorator(prop, decoratorName))
-    .filter((element) => !!element);
+    .filter((element): element is string => !!element);
 
   if (elementRefs.length > 0) {
     newMembers.push(createStaticGetter('elementRef', ts.factory.createStringLiteral(elementRefs[0])));


christian-bromann (approved):

src/compiler/transformers/decorators-to-static/element-decorator.ts
```diff
@@ -14,7 +14,7 @@ export const elementDecoratorsToStatic = (
   const elementRefs = decoratedMembers
     .filter(ts.isPropertyDeclaration)
     .map((prop) => parseElementDecorator(prop, decoratorName))
```
alicewriteswrongs, line 16: is there any circumstance in which `parseElementDecorator` can return `""`? it looks like it's returning the decorated property's name:'

https://github.com/ionic-team/stencil/blob/53a92bc61431ce05122702bbf1ebd33086f8ee47/src/compiler/transformers/decorators-to-static/element-decorator.ts#L28-L35

which I _think_ couldn't be `""`, but just wondering because

```js
!!("")
```

evaluates to `false`, so it's not exactly a typeguard there - should we possibly substitute `typeof element === "string"` here instead?

alicewriteswrongs (changes requested):
one (possibly pedantic) question 😄

src/compiler/transformers/decorators-to-static/element-decorator.ts
rwaskiewicz: > is there any circumstance in which parseElementDecorator can return ""? it looks like it's returning the decorated property's name

> which I think couldn't be ""

_Maybe_, I agree that I think it couldn't be "", BUT... I'm not sure of the internals of `getText` here, other than I'd think a property could/must have a name, save for:
1. An error in the TS compiler
2. Stencil misidentifying a `prop` parameter as a `PropertyDeclaration` when it is fact, not (and therefore may have a `getText()` return value of an empty string.

So it _may_ be possible, but if it is, I'd err on the side of "it's a bug" should this happen.

> should we possibly substitute typeof element === "string" here instead?

I'm not opposed, but wouldn't `!!element` be preferable here, since that eliminate the possibility of us working on empty strings (if the edge-ish case exists here)?

alicewriteswrongs: yeah, that's true -- I think in any case where it is `""` we want to filter it out anyway! so I think this is good as-is 👍

PR #5269, opened 2024-01-17
feat(compiler): deprecate customResolveOptions config option

rwaskiewicz:

we have a stencil configuration option that is part of our public configuration api, `nodeResolve#customResolveOptions`. This option will no longer be honored by newer options of `@rollup/plugin-node-resolve` - we need to remove it in the future (a breaking change). for now, we can start the process by logging a warning.

GitHub Issue Number: N/A

## What is the new behavior?

deprecate the `nodeResolve#customResolveOptions` configuration option in a Stencil configuration. this field is used to configure `@rollup/plugin-node-resolve` when generating any non-documentation output target artifacts. the contents of this option get propagated by the rollup plugin to `resolve`, which does the actual resolution of files in stencil's in-memory filesystem.

in newer versions of `@rollup/plugin-node-resolve`, this option has been removed, which no option to override defaults (creating a separation of concerns between the plugin and `resolve`). since stencil exposes this configuration in its public api, we cannot remove it without causing a breaking change (nor could we find a suitable way to reuse an existing configuration).

this change is a prerequisite to upgrading `@rollup/plugin-node-resolve`. we will not remove the configuration option/upgrade the plugin until stencil v5. in the interim, we add a warning message at config validation time to try to elicit feedback on this change

- [ ] Yes
- [x] No

## Testing

1. Create a new Stencil project in your directory of choice with the create-stencil cli - `npm init stencil@latest component`
2. In the project, install the following dev build `npm i @stencil/core@4.10.0-dev.1705521704.ea55603`
3. In the project's `stencil.config.ts` file, add the top level configuration `nodeResolve: customRollupOptions: {}`. Observe a deprecation notice.
```diff
  testing: {
    browserHeadless: "new",
  },
+  nodeResolve: {customResolveOptions: {}}
```
Screenshot 2024-01-17 at 3 21 44 PM
4. Build the project - `npm run build`. Observe the following warning:
```
> warn-custom-resolve@0.0.1 build
> stencil build

[22:12.8]  @stencil/core
[22:13.0]  [LOCAL DEV] v4.10.0-dev.1705521704.ea55603 🚘

[ WARN  ]  Build Warn
           nodeResolve.customResolveOptions is a deprecated option in a Stencil Configuration file. If you need this
           option, please open a new issue in the Stencil repository
           (https://github.com/ionic-team/stencil/issues/new/choose)
```
Note: The build will fail since we're providing empty options to `resolve`
5. Revert the changes to `stencil.config.ts`
```diff
  testing: {
    browserHeadless: "new",
  },
-  nodeResolve: {customResolveOptions: {}}
```
6. Build the project - `npm run build`. Observe the warning is gone.
## Other information

i've marked this as a 'feat' to raise awareness to it (i want it surfaced in the changelog). if you think 'fix' is more appropriate, happy to change it.

### DIFF
--- a/src/compiler/config/test/validate-custom.spec.ts
+++ b/src/compiler/config/test/validate-custom.spec.ts
@@ -26,6 +26,24 @@ describe('validateCustom', () => {
       },
     ];
     const { diagnostics } = validateConfig(userConfig, mockLoadConfigInit());
-    expect(diagnostics.length).toBe(1);
+    // TODO(STENCIL-1107): Decrement the right-hand side value from 2 to 1
+    expect(diagnostics.length).toBe(2);
+    // TODO(STENCIL-1107): Keep this assertion
+    expect(diagnostics[0]).toEqual({
+      header: 'Build Warn',
+      level: 'warn',
+      lines: [],
+      messageText: 'test warning',
+      type: 'build',
+    });
+    // TODO(STENCIL-1107): Remove this assertion
+    expect(diagnostics[1]).toEqual({
+      header: 'Build Warn',
+      level: 'warn',
+      lines: [],
+      messageText:
+        'nodeResolve.customResolveOptions is a deprecated option in a Stencil Configuration file. If you need this option, please open a new issue in the Stencil repository (https://github.com/ionic-team/stencil/issues/new/choose)',
+      type: 'build',
+    });
   });
 });
--- a/src/compiler/config/validate-config.ts
+++ b/src/compiler/config/validate-config.ts
@@ -1,5 +1,5 @@
 import { createNodeLogger, createNodeSys } from '@sys-api-node';
-import { buildError, isBoolean, isNumber, isString, sortBy } from '@utils';
+import { buildError, buildWarn, isBoolean, isNumber, isString, sortBy } from '@utils';
 
 import {
   ConfigBundle,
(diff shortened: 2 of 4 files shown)


src/compiler/config/test/validate-custom.spec.ts
```diff
@@ -26,6 +26,24 @@ describe('validateCustom', () => {
...
+    // TODO(STENCIL-1107): Keep this assertion
+    expect(diagnostics[0]).toEqual({
+      header: 'Build Warn',
+      level: 'warn',
+      lines: [],
+      messageText: 'test warning',
+      type: 'build',
+    });
+    // TODO(STENCIL-1107): Remove this assertion
+    expect(diagnostics[1]).toEqual({
```
rwaskiewicz, line 40: We need to add this as [this field](https://github.com/ionic-team/stencil/pull/5269/files#diff-cfdd0f19925ea4e2870c4390ad870f820215b69acf2b4bba2eef548036fe95ebL102) is being used by our tests still. In adding the warning [here](https://github.com/ionic-team/stencil/pull/5269/files#diff-53c120bef3205f89229fbe98a5aadfcb51012ead144729ade2fc597c1e9ad5e9R270-R275), that caused an additional warning to trigger.

I figured it'd be best to just:
1. Increment the count of warns we assert against for now, since there's no other good way to 'split up' the number of assertions we get 
2. Add explicit assertions for the existing warning and here. That way, in STENCIL-1107, we can leave the first one and we're a little better off in this test suite than where we started

christian-bromann (approved):
LGTM 👍

alicewriteswrongs (approved):
lgtm!

christian-bromann:
The fix for this issue has been released as a part of today's [Stencil v4.11.0 release](https://github.com/ionic-team/stencil/releases/tag/v4.11.0).

PR #5273, opened 2024-01-18
docs(contributing): fix path setting in tsconfig.json

christian-bromann:

Copy-pasting the `tsconfig.json` patch from the contributing guidelines causes an error:

```sh
[ ERROR ]  TypeScript: tsconfig.json:19:34
           Non-relative paths are not allowed when 'baseUrl' is not set. Did you forget a
           leading './'?

     L18:  "paths": {
     L19:    "@stencil/core/internal": ["node_modules/@stencil/core/internal"],
     L20:    "@stencil/core/internal/*": ["node_modules/@stencil/core/internal/*"]
```

Seems that this has changed at some point in TypeScript.

## What is the new behavior?

Making it a true relative path.

- [ ] Yes
- [x] No

## Testing

- Create a new stencil project `npm init stencil`
- link stencil dependency to a local stencil project
- copy & paste the current snippet into the `tsconfig.json` and observe the error
- now apply the suggested change
- build passes now

## Other information

n/a

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -78,8 +78,8 @@ Using `npm link` is beneficial to the development cycle in that consecutive buil
   "compilerOptions": {
     "baseUrl": ".",
     "paths": {
-      "@stencil/core/internal": ["node_modules/@stencil/core/internal"],
-      "@stencil/core/internal/*": ["node_modules/@stencil/core/internal/*"]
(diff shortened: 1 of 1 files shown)


tanner-reits (commented):
I tried this out and was able to build with the current paths. The existing snippet does include a field for `baseUrl` which I saw your error said was missing, did you include that in your config?

Side note: I did have to include a path for `mock-doc` for it to build, which does not currently exist in the guide

PR #5288, opened 2024-01-23
chore(jsdoc): add JSDoc to `src/compiler/build/compiler-ctx.ts`

alicewriteswrongs:
This adds some documentation for the two functions in here.

Just noticed this randomly and in particular wanted to call out how `getModuleLegacy` works.

- [ ] Yes
- [x] No

## Testing

Documentation only change, so as long as lint and spellcheck are passing I think we're good.

### DIFF
--- a/src/compiler/build/compiler-ctx.ts
+++ b/src/compiler/build/compiler-ctx.ts
@@ -62,7 +62,17 @@ export class CompilerContext implements d.CompilerCtx {
   }
 }
 
-export const getModuleLegacy = (compilerCtx: d.CompilerCtx, sourceFilePath: string) => {
+/**
+ * Get a {@link d.Module} from the current compiler context which corresponds
+ * to a supplied source file path. If a module record corresponding to the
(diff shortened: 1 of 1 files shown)


rwaskiewicz (approved):

christian-bromann (approved):
I wonder if the "Legacy" has a meaning here, do you know? Otherwise LGTM 👍

alicewriteswrongs:
the 'legacy' naming I'm not sure of, it looks like the function was renamed in this commit: https://github.com/ionic-team/stencil/commit/0fbc2f9854f074b24a7c8e0f45b6a58c32486f02

but I can't find a non-legacy `getModule` around, so maybe we should rename it 🤔

christian-bromann:
Could it be this one: https://github.com/ionic-team/stencil/blob/9d2d0a0612caafb88c559907c6dec1a39e760615/src/compiler/transpile/transpiled-module.ts#L6-L13?

PR #5289, opened 2024-01-23
chore(jsdoc): clarify how `parseStaticComponentMeta` behaves

alicewriteswrongs:
This fleshes out a bit further how this functions behaves, in particular that it will perform a side-effect of creating a `ComponentCompilerMeta` object and then sticking that into a few other objects.

- [ ] Yes
- [x] No

## Testing

Documentation-only change, so if lint, spellcheck, and format are passing we're good I think.

### DIFF
--- a/src/compiler/transformers/static-to-meta/component.ts
+++ b/src/compiler/transformers/static-to-meta/component.ts
@@ -22,10 +22,17 @@ import { parseStaticStyles } from './styles';
 import { parseStaticWatchers } from './watchers';
 
 /**
- * Given an instance of TypeScript's Intermediate Representation (IR) for a
- * class declaration ({@see ts.ClassDeclaration}) which represents a Stencil
- * component class declaration, parse and format various pieces of data about
- * static class members which we use in the compilation process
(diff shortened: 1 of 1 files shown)


tanner-reits (approved):

src/compiler/transformers/static-to-meta/component.ts
```diff
@@ -22,10 +22,17 @@ import { parseStaticStyles } from './styles';
 import { parseStaticWatchers } from './watchers';
 
 /**
- * Given an instance of TypeScript's Intermediate Representation (IR) for a
- * class declaration ({@see ts.ClassDeclaration}) which represents a Stencil
- * component class declaration, parse and format various pieces of data about
- * static class members which we use in the compilation process
+ * Given a {@see ts.ClassDeclaration} which represents a Stencil component
+ * class declaration, parse and format various pieces of data about static class
+ * members which we use in the compilation process
```
christian-bromann, line 27:
```suggestion
 * members which we use in the compilation process.
```

christian-bromann (approved):
LGTM 👍

PR #3771, opened 2022-10-26
Only add HYDRATED_CSS when cmpTags are set

raldenhoven:
## Pull request checklist

Please check if your PR fulfills the following requirements:
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [ ] Build (`npm run build`) was run locally and any changes were pushed
- [ ] Unit tests (`npm test`) were run locally and passed
- [ ] E2E Tests (`npm run test.karma.prod`) were run locally and passed
- [ ] Prettier (`npm run prettier`) was run locally and passed

## Pull request type

 

 

Please check the type of change your PR introduces:
- [x] Bugfix
- [ ] Feature
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

## What is the current behavior?
An invalid piece of CSS is injected when cmpTags are empty. If this happens in JSDOM it will throw an invalid css error.

GitHub Issue Number: N/A

## What is the new behavior?
The CSS will only be added to the head when cmpTags are defined (making the css always valid).

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Testing

Manually tested the change in the compiled code, no error where thrown again.

PR #5305, opened 2024-01-29
fix(runtime): only generate lazy build CSS when there are component tags

tanner-reits:

If a developer accidentally calls `defineCustomElements()` multiple times with the same component metadata, invalid CSS will be generated and injected into the DOM. This happens because we only create a `cmpTags` array based on new component tags added to the Custom Element registry, but we do not check if the array has a length before appending additional CSS rules to the tag selectors.

Closes: #3771

## What is the new behavior?

This commit fixes an issue where invalid CSS could be generated and inserted into the DOM if `bootstrapLazy` was called with no component metadata or if no new components are registered in the custom element registry (i.e. it is called multiple times with the same metadata).

## Documentation

N/A

- [ ] Yes
- [X] No

## Testing

**Automated testing**
Added some new unit tests to `bootstrap-lazy.spec.tsx`

**Manual testing**
This can be tested using the reproduction [in this comment](https://github.com/ionic-team/stencil/pull/3771#issuecomment-1300036253) from #3771 and a build of this branch.

**Note:** When testing I was experiencing some caching issues with the wrong version of the runtime being used so you may need to nuke `node_modules` and lock files when installing this build into the project


christian-bromann (approved):
👍

tanner-reits:
@rwaskiewicz Removed those tests as discussed

src/runtime/bootstrap-lazy.ts
```diff
@@ -196,29 +196,32 @@ export const bootstrapLazy = (lazyBundles: d.LazyBundlesRuntimeData, options: d.
...
+    }
 
-  // Add hydration styles
-  if (BUILD.invisiblePrehydration && (BUILD.hydratedClass || BUILD.hydratedAttribute)) {
-    dataStyles.innerHTML += cmpTags + HYDRATED_CSS;
-  }
+    // Add hydration styles
+    if (BUILD.invisiblePrehydration && (BUILD.hydratedClass || BUILD.hydratedAttribute)) {
+      dataStyles.innerHTML += cmpTags + HYDRATED_CSS;
+    }
```
rwaskiewicz, line 210: Non-blocking - Can we add a TODO to get this part under unit test?

rwaskiewicz:
@tanner-reits I found a few additional files for the invisible prehydration testing that I removed in cdafaa0 - can you & @christian-bromann PTAL? Otherwise, this looks good to me, I had just one non-blocking comment

rwaskiewicz (approved):

tanner-reits:
@rwaskiewicz Thanks for that, LGTM. Wasn't sure if we'd wanna bring these tests back if the ability to remove an element from the registry became a thing, but I suppose we can revert all those changes if that time comes.

christian-bromann (approved):

Issue #5306, opened 2024-01-31
bug: Incorrect TypeScript typings for JSX


maxpatiiuk:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

4.12.0

### Current Behavior

It appears that the TypeScript definitions for JSX support in Stencil are not completely correct, and are causing issue when used with third-party tools (like @typescript-eslint)

### Expected Behavior

JSX.Element type is resolved correctly

### System Info

```shell
System: node 20.11.0
    Platform: darwin (23.2.0)
   CPU Model: Apple M1 Pro (10 cpus)
    Compiler: /Users/mak13180/site/esri/arcgis-web-components/node_modules/@stencil/core/compiler/stencil.js
       Build: 1705333241
     Stencil: 4.10.0 🍪
  TypeScript: 5.3.3
      Rollup: 2.42.3
      Parse5: 7.1.2
      jQuery: 4.0.0-pre
      Terser: 5.26.0
```
```


### Steps to Reproduce

[...]

PR #5307, opened 2024-01-31
fix(declarations): use correct TypeScript JSX typings

maxpatiiuk:

Fix incorrect JSX typings, as described in great detail in https://github.com/ionic-team/stencil/issues/5306

GitHub Issue Number: #5306

## What is the new behavior?

TypeScript compiler is able to resolve Stencil's JSX typings correctly, thus working well with third-party tools that use TypeScript compiler types

## Documentation

Described in great detail in https://github.com/ionic-team/stencil/issues/5306

- [ ] Yes
- [x] No

## Testing

The error affects tools that use the TypeScript compiler. It is unfortunate that typescript does not report this error directly - instead, it just settles for an implicit `any`, which is how I discovered this bug in the first place.

Thus suggested testing:
1. In an app that is using Stencil and TypeScript, install typescript-eslint and enable the [@typescript-eslint/no-unsafe-return](https://typescript-eslint.io/rules/no-unsafe-return) rule
2. Create a small test file like this:
  ```tsx
  import { type VNode, h } from "@stencil/core/internal";
  
  export function a(): VNode {
    return <br />;
  }
  ```

  Before this PR there would be an error on line `return <br />;`. After this PR, there should not be an error there

### DIFF
--- a/src/declarations/stencil-public-runtime.ts
+++ b/src/declarations/stencil-public-runtime.ts
@@ -596,12 +596,12 @@ export interface ChildNode {
  *
  * For further information: https://stenciljs.com/docs/host-element
  */
-export declare const Host: FunctionalComponent<HostAttributes>;
+export declare const Host: (props: HostAttributes) => VNode;
 
 /**
  * Fragment
  */
-export declare const Fragment: FunctionalComponent<{}>;
+export declare const Fragment: (props: {}) => VNode;
 
 /* eslint-disable jsdoc/require-param, jsdoc/require-returns -- we don't want to JSDoc these overloads at this time */
 /**
@@ -619,6 +619,7 @@ export declare namespace h {
   export function h(sel: any, data: VNodeData | null, children: VNode): VNode;
 
   export namespace JSX {
+    type Element = VNode;
     interface IntrinsicElements extends LocalJSX.IntrinsicElements, JSXBase.IntrinsicElements {
       [tagName: string]: any;
     }


src/declarations/stencil-public-runtime.ts
```diff
@@ -596,12 +596,12 @@ export interface ChildNode {
  *
  * For further information: https://stenciljs.com/docs/host-element
  */
-export declare const Host: FunctionalComponent<HostAttributes>;
```
rwaskiewicz, line 599: While I understand the motivation for this change, I don't think we can make this change in Stencil v4. This changes the output of `typeof Host`, which would constitute a breaking change.

In Stencil v4, the following code type checks without error:
```ts
import { Host } from '@stencil/core';

const Hello: typeof Host = (_props: any, children: any, utils: any) =>
  utils.map(children, child => ({
    ...child,
    vattrs: {
      ...child.vattrs,
      class: `${child.vattrs.class} add-class`,
    },
  }));
```

Whereas with this commit/change, it fails with:
```
[ ERROR ]  TypeScript: src/components/my-component/my-component.tsx:4:7
           Type '(_props: any, children: any, utils: any) => any' is not assignable to type '(props: HostAttributes) =>
           VNode'.Target signature provides too few arguments. Expected 3 or more, but got 1.

      L4:  const Hello: typeof Host = (_props: any, children: any, utils: any) =>
      L5:    utils.map(children, child => ({
```

src/declarations/stencil-public-runtime.ts
```diff
@@ -596,12 +596,12 @@ export interface ChildNode {
...
+export declare const Host: (props: HostAttributes) => VNode;
 
 /**
  * Fragment
  */
-export declare const Fragment: FunctionalComponent<{}>;
```
rwaskiewicz, line 604: Similar to my comment above - while I understand the motivation for this change, it would also be a breaking change.

In Stencil v4, the following code type checks without error:
```ts
import { Fragment } from '@stencil/core';

const Hello: typeof Fragment = (_props: any, children: any, utils: any) =>
  utils.map(children, child => ({
    ...child,
    vattrs: {
      ...child.vattrs,
      class: `${child.vattrs.class} add-class`,
    },
  }));
```

Whereas with this commit/change, it fails with:
```
[ ERROR ]  TypeScript: src/components/my-component/my-component.tsx:4:7
           Type '(_props: any, children: any, utils: any) => any' is not assignable to type '(props: {}) =>
           VNode'.Target signature provides too few arguments. Expected 3 or more, but got 1.

      L4:  const Hello: typeof Fragment = (_props: any, children: any, utils: any) =>
      L5:    utils.map(children, child => ({
```

rwaskiewicz (changes requested):
Hey @maxpatiiuk 👋 

I created a dev build based on this branch to test it out (you can grab it yourself using `npm i @stencil/core@4.12.0-dev.1706795674.f7ef54c`). 

Using this dev build, I tested the PR by using it to build the Ionic Framework. There are a few cases where I think this would improve type checking, I noticed a few cases where setting `type Element = VNode` caused a few breaking changes with the typings used in the Framework today. I'm concerned this might be confusing to developers here if they were to upgrade Stencil and the type checker began to complain.

With that in mind and the current state of this PR, I'd like to revisit this for Stencil v5. I think this is the best case scenario for improving the typings of Stencil, while trying to minimize breaking changes in minor releases.

Now that I understand the issue here/how to reproduce it, I'm going to combine the two issues linked here. In the interim, I suggest adding an `any` return type to your render functions, as that appears to appease the ESLint rule `no-unsafe-return` and [aligns with the currently typed return value of `render`](https://github.com/ionic-team/stencil/blob/f73aa149f06dd3014bfbc2ab7223f8363b859b41/src/declarations/stencil-public-runtime.ts#L498)

Thanks for your patience and let me know if you have any questions!

Issue #3319, opened 2022-04-07
bug: client-side hydration of pre-rendered web component is broken for bundles that use `dist-custom-element`


weaintplastic:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/.github/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

2.15.0

### Current Behavior

It seems that the client-side hydration of pre-rendered web components leads to different results when comparing the use of Stencil's `dist` bundle and a custom bundle based on `dist-custom-element` output.

#### Pre-rendered HTML

```html
<my-component class="sc-my-component-h hydrated" s-id="1">
  <!--r.1-->
  <!--o.0.1.-->
  <label class="sc-my-component sc-my-component-s" c-id="1.0.0.0">
    <!--s.1.1.1.0.-->
    <!--t.0.1-->
    This is the label
  </label>
</my-component>
```

After client-side hydration with Stencil's `dist` bundle, the html looks correct and is working as expected.

```html
<my-component class="hydrated">
  <!-- #shadow-root (open) -->
  <label class="sc-my-component sc-my-component-s">
    <slot>
      <!-- ↳ #text reveal -->
    </slot>
  </label>
  <!---->
  This is the label
</my-component>
```

After client-side hydration with the custom `app` bundle that is using Stencil's `dist-custom-element` distribution, the html looks very different due to the slotted content not being hydrated correctly. 

```html
<my-component class="sc-my-component-h hydrated" s-id="1">
  <!-- #shadow-root (open) -->
  <label>
    <slot>
      <!-- ↳ <label> reveal -->
    </slot>
  </label>
  <!--r.1-->
  <!--o.0.1.-->
  <label class="sc-my-component sc-my-component-s" c-id="1.0.0.0">
    <!--s.1.1.1.0.-->
    <!--t.0.1-->
    This is the label
  </label>
</my-component>
```

### Expected Behavior

I'd expect the client-side hydrated results to be exactly the same and slots are resolved correctly.

### Steps to Reproduce

I've set up a project to reproduce the bug in two different scenarios and included a thorough description on how to reproduce there.

[...]

PR #3330, opened 2022-04-14
fix(runtime): importing custom elements build in lazy build

sean-perkins:
## Pull request checklist

Please check if your PR fulfills the following requirements:
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [x] Build (`npm run build`) was run locally and any changes were pushed
- [x] Unit tests (`npm test`) were run locally and passed
- [x] E2E Tests (`npm run test.karma.prod`) were run locally and passed
- [x] Prettier (`npm run prettier`) was run locally and passed

## Pull request type

 

 

Please check the type of change your PR introduces:
- [x] Bugfix
- [ ] Feature
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

## What is the current behavior?

You are unable to import and use a custom elements build within a lazy-hydrated bundle Stencil project. When trying to use a Stencil compiled library within another Stencil project, you will receive numerous exceptions & the component will fail to render.

GitHub Issue Number: #2531, #3233

## What is the new behavior?

Stencil libraries can be consumed within other Stencil libraries. For example, developers can use `@ionic/core/components` within their own Stencil libraries. 

- Hydrated web component behavior is skipped when the element being rendered is a custom element
- Introduces `$customElement$` flag on the compiler metadata for the component, so that the runtime can make decisions based on if the component being rendered is a custom element or part of the lazy/hydrated build.

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Testing

You can either clone this repository: https://github.com/sean-perkins/stencil-ce-project or follow these steps:

1. Create a new Stencil project `npm stencil init`
2. Install another Stencil library as a dependency, i.e. `@ionic/core`
3. For Ionic, you will need a global script to initialize required functionality:
```ts
// src/global/app.ts
import { initialize } from '@ionic/core/components';

initialize();
```
 4. Update `stencil.config.ts` to use the global script:
```ts
// stencil.config.ts
{
  ...
  globalScript: './src/global/app.ts'
}
```
 5. In your new Stencil project, use an Ionic component from the custom elements bundle
```tsx
// src/components/my-component/my-component.tsx
import { defineCustomElement  } from '@ionic/core/components/ion-button.js';

@Component(...) 
export class MyComponent {
  connectedCallback() {
    defineCustomElement();
  }

  render() {
    return (
[...]

PR #5317, opened 2024-02-01
fix(custom-elements): hydrate on client side

andrew9994:

Components hydrated on the server are hydrated wrongly on the client using the `dist-custom-elements` output target.

GitHub Issue Number: #3319 

## What is the new behavior?

By setting `build.hydrateClientSide = true` in `custom-elements-build-conditionals.ts` and by disabling lazy loading, the components are rendered correctly.

## Documentation

Reproduction repo:

https://github.com/andrew9994/stencil-hydrate-dist-custom-elements-bug

- [ ] Yes
- [x] No

## Testing

To completely unit test these modifications a change in the `spec-page.ts` would be needed and I was not sure that I should change that. Instead I added unit tests to check for the default values, and added missing unit tests for the`hasHydrateOutputTargets` logic in the `lazy-build-conditionals.ts`

## Screenshots from reproduction repo

Before:
Bildschirmfoto 2024-02-01 um 11 05 14

After:
Bildschirmfoto 2024-02-01 um 11 03 26

### DIFF
--- a/src/compiler/output-targets/dist-custom-elements/custom-elements-build-conditionals.ts
+++ b/src/compiler/output-targets/dist-custom-elements/custom-elements-build-conditionals.ts
@@ -1,6 +1,7 @@
+import { isOutputTargetHydrate } from '@utils';
+
 import type * as d from '../../../declarations';
 import { getBuildFeatures, updateBuildConditionals } from '../../app-core/app-data';
-
 /**
  * Get build conditions appropriate for the `dist-custom-elements` Output
  * Target, including disabling lazy loading and hydration.
@@ -17,9 +18,10 @@ export const getCustomElementsBuildConditionals = (
   // then the default in "import { BUILD, NAMESPACE } from '@stencil/core/internal/app-data'"
   // needs to have the static build conditionals set for the custom elements build
   const build = getBuildFeatures(cmps) as d.BuildConditionals;
+  const hasHydrateOutputTargets = config.outputTargets.some(isOutputTargetHydrate);
 
   build.lazyLoad = false;
-  build.hydrateClientSide = false;
+  build.hydrateClientSide = hasHydrateOutputTargets;
   build.hydrateServerSide = false;
   build.asyncQueue = config.taskQueue === 'congestionAsync';
   build.taskQueue = config.taskQueue !== 'immediate';
--- a/src/runtime/initialize-component.ts
+++ b/src/runtime/initialize-component.ts
@@ -32,7 +32,8 @@ export const initializeComponent = async (
     // Let the runtime know that the component has been initialized
     hostRef.$flags$ |= HOST_FLAGS.hasInitializedComponent;
 
-    if (BUILD.lazyLoad || BUILD.hydrateClientSide) {
+    const bundleId = cmpMeta.$lazyBundleId$;
+    if ((BUILD.lazyLoad || BUILD.hydrateClientSide) && bundleId) {
       // lazy loaded components
       // request the component's implementation to be
       // wired up with the host element
--- a/src/compiler/output-targets/test/build-conditionals.spec.ts
+++ b/src/compiler/output-targets/test/build-conditionals.spec.ts
@@ -15,6 +15,16 @@ describe('build-conditionals', () => {
   });
 
   describe('getCustomElementsBuildConditionals', () => {
+    it('default', () => {
+      const { config } = validateConfig(userConfig, mockLoadConfigInit());
+      const bc = getCustomElementsBuildConditionals(config, cmps);
+      expect(bc).toMatchObject({
+        lazyLoad: false,
+        hydrateClientSide: false,
+        hydrateServerSide: false,
+      });
+    });
+
     it('taskQueue async', () => {
       userConfig.taskQueue = 'async';
       const { config } = validateConfig(userConfig, mockLoadConfigInit());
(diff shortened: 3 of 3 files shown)


christian-bromann:
@andrew9994 thanks for raising the PR, mind resolving the prettier issues by running `npm run prettier`?

@ionic-team/stencil I was able to reproduce the bug and verified that the patch resolves the problem. Removing `BUILD.hydrateClientSide` from the `if` statement seems like a very deliberate change. I found it being introduced in a [massive PR](https://github.com/ionic-team/stencil/pull/1953) many years ago. I wonder if this might have an impact or side effect on other hydration cases. That said, the unit tests seem to be passing. I don't know enough to make confident decision here but from POV the change looks fine. Curious about your thoughts.

christian-bromann:
I similar PR was introduced by @sean-perkins some time ago: https://github.com/ionic-team/stencil/pull/3330

andrew9994:
Thanks for the quick review @christian-bromann! I resolved the code-style issues 👍

andrew9994:
@tanner-reits did you run a build on the fix branch and linked it properly via npm? I described the steps [here](https://github.com/andrew9994/stencil-hydrate-dist-custom-elements-bug?tab=readme-ov-file#try-the-fix-locally). For me and my colleagues the fix is working, and also @christian-bromann confirmed this. I also added some before/after screenshots to the description of this PR.

andrew9994:
@tanner-reits, good catch, it turns out the fix wasn't working in the original reproduction repository. Apologies for not testing it there. I identified the main issue: the `defineCustomElement` function was being called while the document was still loading. To address this, I moved the built bundle into the `<body>` - with [this change](https://github.com/weaintplastic/stencil-hydrate-bug-demo/commit/dfb68d352dcfc4ffeda6d27b7bac8b8497c80ed5) the fix now works as expected. (I also made additional commits in that repository to simplify the local linking of @stencil/core)

It's important to note that this fix will only work when the document has finished loading and is in the `interactive/complete` state. Do you see an issue regarding this?

andrew9994:
@tanner-reits, @rwaskiewicz  could you give an update on this? Can we expect a merge in the foreseeable future? If not, should I look for a fix in the runtime, instead of changing the build-flags?

christian-bromann:
> could you give an update on this? Can we expect a merge in the foreseeable future?

I am currently reviewing this with the team and will post updates as soon as we have them.

src/runtime/initialize-component.ts
```diff
@@ -32,7 +32,7 @@ export const initializeComponent = async (
     // Let the runtime know that the component has been initialized
     hostRef.$flags$ |= HOST_FLAGS.hasInitializedComponent;
 
-    if (BUILD.lazyLoad || BUILD.hydrateClientSide) {
+    if (BUILD.lazyLoad) {
```
christian-bromann, line 35: Can you explain the reasoning for this change? For me it seems that keeping `|| BUILD.hydrateClientSide` results in the same result 🤔

christian-bromann: Ok, I see it prevents Stencil from lazily loading a component file.

src/compiler/output-targets/dist-custom-elements/custom-elements-build-conditionals.ts
```diff
@@ -19,7 +19,7 @@ export const getCustomElementsBuildConditionals = (
   const build = getBuildFeatures(cmps) as d.BuildConditionals;
 
   build.lazyLoad = false;
-  build.hydrateClientSide = false;
+  build.hydrateClientSide = true;
```
christian-bromann, line 22: Instead of always setting this value to `true`, which can cause Stencil to ship unnecessary code related to hydration even though the user is not intending to use this feature, I suggest to make this value depending on whether a `dist-hydrate-script` output target is set.

Similar to how we approach this in [`dist`](https://github.com/ionic-team/stencil/blob/9d2d0a0612caafb88c559907c6dec1a39e760615/src/compiler/output-targets/dist-lazy/lazy-build-conditionals.ts):

```ts
import { isOutputTargetHydrate } from '@utils';

const hasHydrateOutputTargets = config.outputTargets.some(isOutputTargetHydrate);
build.hydrateClientSide = hasHydrateOutputTargets;
```

src/runtime/initialize-component.ts
christian-bromann, line 35: We feel a bit unsure about removing `BUILD.hydrateClientSide` from the if statement as we can't tell what side effects that has. We suggest to achieve the same result by looking up the `lazyBundleId` of `cmpData` which guarantees that we only enter the if statement in the case of transpiling a lazy loaded component:
```suggestion
    const bundleId = cmpMeta.$lazyBundleId$;
    if ((BUILD.lazyLoad || BUILD.hydrateClientSide) && bundleId) {
```

christian-bromann (changes requested):
@andrew9994 thanks for taking a stab at this. The team has looked into ramifications of this change and has identified two suggestion to reduce risk of introducing any side effects. Let me know what you think of these suggestions.

src/compiler/output-targets/dist-custom-elements/custom-elements-build-conditionals.ts
andrew9994: Good points, this way we do not introduce unattended changes for the people who are not using SSR. Implemented your suggestion and also added tests for it 👍

src/runtime/initialize-component.ts
andrew9994: My feeling is that this was originally an unintendedly introduced condition - but just to be sure I implemented your suggestion, to make sure we don't introduce breaking changes.

andrew9994:
Thanks @christian-bromann and to the team for the review 🙏  I implemented your suggestions

christian-bromann (approved):
LGTM 👍 

I will have someone else from the @ionic-team/stencil team review this.

rwaskiewicz (approved):
Thanks!

PR #5318, opened 2024-02-01
fix(mock-doc): improve error message when `:scope` selector is used

alicewriteswrongs:
This improves the error message shown when there is an error on the jQuery side when running `querySelector` and friends by selectively making an addendum to the error message to just explain a bit better what's going on.

STENCIL-1108

If you try to use the [`:scope` pseudo-selector](https://developer.mozilla.org/en-US/docs/Web/CSS/:scope) you'll get an error message which is a little less than helpful.

## What is the new behavior?

This adds to that error message by adding a `try / catch` around `jQuery.find` and in cases where 1) an error was thrown and 2) the selector used has `:scope` in it we prepend some explanatory text to the error message before re-throwing it.

- [ ] Yes
- [x] No

## Testing

I added some unit tests which exercise this functionality and check for the right error message.

It's also a good idea to check out how this works in Framework.

You can add the following test to `core/src/components/accordion/test/accordion.spec.ts`:

```tsx
it('group should properly set readonly on child accordions', async () => {
  const page = await newSpecPage({
    components: [Item, Accordion, AccordionGroup],
    html: `
      <ion-accordion-group animated="false">
        <ion-accordion>
          <ion-item slot="header">Label</ion-item>
          <div slot="content">Content</div>
        </ion-accordion>
      </ion-accordion-group>
    `,
  });

  const accordionGroup = page.body.querySelector('ion-accordion-group')!;
  // This causes the exception, since it invokes underlying implementation code that uses the :scope selector
  accordionGroup.readonly = true;
});
```

on `@stencil/core@latest` this will let you then reproduce the less-than-helpful message. If you then build and install this branch you can check out what the improved version looks like.

### DIFF
--- a/src/mock-doc/selector.ts
+++ b/src/mock-doc/selector.ts
@@ -1,16 +1,96 @@
 import { MockElement } from './node';
 import jQuery from './third-party/jquery';
 
-export function matches(selector: string, elm: MockElement) {
-  const r = jQuery.find(selector, undefined, undefined, [elm]);
-  return r.length > 0;
+/**
+ * Check whether an element of interest matches a given selector.
+ *
+ * @param selector the selector of interest
+ * @param elm an element within which to find matching elements
+ * @returns whether the element matches the selector
+ */
+export function matches(selector: string, elm: MockElement): boolean {
+  try {
+    const r = jQuery.find(selector, undefined, undefined, [elm]);
+    return r.length > 0;
+  } catch (e) {
+    updateSelectorError(selector, e);
+    throw e;
+  }
 }
 
+/**
+ * Select the first element that matches a given selector
+ *
+ * @param selector the selector of interest
+ * @param elm the element within which to find a matching element
+ * @returns the first matching element, or null if none is found
+ */
 export function selectOne(selector: string, elm: MockElement) {
-  const r = jQuery.find(selector, elm, undefined, undefined);
-  return r[0] || null;
+  try {
+    const r = jQuery.find(selector, elm, undefined, undefined);
+    return r[0] || null;
+  } catch (e) {
+    updateSelectorError(selector, e);
+    throw e;
+  }
 }
 
-export function selectAll(selector: string, elm: MockElement) {
-  return jQuery.find(selector, elm, undefined, undefined);
+/**
+ * Select all elements that match a given selector
+ *
+ * @param selector the selector of interest
+ * @param elm an element within which to find matching elements
+ * @returns all matching elements
+ */
+export function selectAll(selector: string, elm: MockElement): any {
+  try {
+    return jQuery.find(selector, elm, undefined, undefined);
+  } catch (e) {
+    updateSelectorError(selector, e);
+    throw e;
+  }
+}
+
+/**
+ * A manifest of selectors which are known to be problematic in jQuery. See
+ * here to track implementation and support:
+ * https://github.com/jquery/jquery/issues/5111
+ */
+export const PROBLEMATIC_SELECTORS = [':scope', ':where', ':is'] as const;
+
+/**
+ * Given a selector and an error object thrown by jQuery, annotate the
+ * error's message to add some context as to the probable reason for the error.
+ * In particular, if the selector includes a selector which is known to be
+ * unsupported in jQuery, then we know that was likely the cause of the
+ * error.
+ *
+ * @param selector our selector of interest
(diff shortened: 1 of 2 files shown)


src/mock-doc/selector.ts
```diff
@@ -1,16 +1,72 @@
...
+ * @param e an error object that was thrown in the course of using jQuery
+ */
+function updateScopeSelectorError(selector: string, e: unknown) {
+  if (selector.includes(':scope') && (e as Error).message) {
+    (e as Error).message = JQUERY_SCOPE_ADVISORY + '\n' + (e as Error).message;
+  }
+}
+
+/**
+ * description
```
rwaskiewicz, line 25: Non-blocking nit - can we fill this in?

src/mock-doc/test/selector.spec.ts
```diff
@@ -225,4 +225,27 @@ describe('selector', () => {
...
+
+    expect(() => {
+      doc.querySelector(':scope');
+    }).toThrow(
+      [
+        "At present jQuery does not support the ':scope' selector.",
+        'If you need to use it in your test, consider writing an end-to-end test instead.',
+        'Syntax error, unrecognized expression: unsupported pseudo: scope',
+      ].join('\n'),
+    );
```
rwaskiewicz, line 249: Is this assertion the same as the one above it? Do we need it?

src/mock-doc/selector.ts
```diff
@@ -1,16 +1,72 @@
...
+ * Given a selector and an error object thrown by jQuery, annotate the
+ * error's message to add some context as to the probable reason for the error.
+ * In particular, if the selector includes the `:scope` selector, then we know
+ * that was likely the cause of the error.
+ *
+ * @param selector our selector of interest
+ * @param e an error object that was thrown in the course of using jQuery
+ */
+function updateScopeSelectorError(selector: string, e: unknown) {
+  if (selector.includes(':scope') && (e as Error).message) {
```
rwaskiewicz, line 19: I think there are a couple other L4 selectors that Sizzle doesn't handle on its own right now - `:where()` and `:is()`. What do you think about expanding this to query a list of known selectors?

I threw together a rough little patch since it'd take a little change on a couple of different lines/make it hard to look at in GH. TAL and let me know what you think!
[multi-l4-selector.patch](https://github.com/ionic-team/stencil/files/14131219/multi-l4-selector.patch)

rwaskiewicz (commented):
Looks good! I had 2 non-blocking comments and one larger-picture question. LMK what you think

christian-bromann (approved):
Besides what has been already mentioned, LGTM 👍 

Thanks for adding additional documentation for `selectOne` and `selectAll` 👌

src/mock-doc/selector.ts
alicewriteswrongs: makes sense - just pushed that change

src/mock-doc/test/selector.spec.ts
```diff
@@ -225,4 +226,50 @@ describe('selector', () => {
...
+
+  it.each(PROBLEMATIC_SELECTORS)("should error for '%p' selector", (selector) => {
+    const doc = new MockDocument();
+
+    const expectedMessage = [
+      `At present jQuery does not support the ${selector} selector.`,
+      'If you need this in your test, consider writing an end-to-end test instead.',
+      `Syntax error, unrecognized expression: unsupported pseudo: ${selector.replace(':', '')}`,
+    ].join('\n');
+
```
rwaskiewicz, line 238: Can we add an assertion for `matches` as well?
```suggestion

    expect(() => doc.matches(selector)).toThrow(expectedMessage);
```

rwaskiewicz (approved):
LGTM and is working as expected. I had one non-blocking ask for one more assertion in our tests. Otherwise, should be good to go

rwaskiewicz:
Ah, we can also probably consider this a 'fix', rather than a chore (and `dock` -> `doc` in the PR summary/final commit)

src/mock-doc/test/selector.spec.ts
alicewriteswrongs: oh good idea 👍

alicewriteswrongs:
I _always_ write / think `mock-dock` haha, as if we're building some sort of boat simulator 🚢 ⚓️ 🌊

PR #5333, opened 2024-02-05
chore(ci): update comment actions to same versions as tech debt burndown

alicewriteswrongs:
Just a quick version bump to sync the versions in the `pack-and-comment` workflow with the ones we already use in the tech debt burndown.

I also realized we want to ensure that this workflow doesn't run on CI when not in the context of a pull request.

GitHub Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/.github/workflows/pack-and-comment.yml
+++ b/.github/workflows/pack-and-comment.yml
@@ -15,6 +15,13 @@ jobs:
     name: Pack and Comment
     runs-on: 'ubuntu-22.04'
     steps:
+      - name: Check event type
+        run: |
+          if [ "${{ github.event_name }}" != "pull_request" ]; then
+            echo "This workflow was not triggered by a pull request. Exiting..."
+            exit 0
+          fi
+
       - name: Checkout Code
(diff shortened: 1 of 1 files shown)


.github/workflows/pack-and-comment.yml
```diff
@@ -15,6 +15,13 @@ jobs:
     name: Pack and Comment
     runs-on: 'ubuntu-22.04'
     steps:
+      - name: Check event type
+        run: |
+          if [ "${{ github.event_name }}" != "pull_request" ]; then
+            echo "This workflow was not triggered by a pull request. Exiting..."
+            exit 0
+          fi
+
```
alicewriteswrongs, line 24: this is to ensure that when CI is running because of e.g. a push to `main` or in the context of the merge queue we don't have this action running (and failing because it can't find or create a comment)

alicewriteswrongs (commented):
one note

christian-bromann (approved):

rwaskiewicz (approved):

PR #5336, opened 2024-02-07
fix(mock-doc): overwrite parentElement in MockHTMLElement to return null

### LINKED ISSUE #5252: bug: Error when using testing-library/dom getByRole
### Prerequisites

### Stencil Version

4.9.1

### Current Behavior

When using `testing-library/dom` with stencil an error occurs when using `getByRole`:

```
Cannot read properties of null (reading 'defaultView')
```

### Expected Behavior

`getByRole` should find elements with the given role, or say it could not be found, but it should not lead to the above error.

### System Info

```shell
@testing-library/dom: 9.3.4
@testing-library/jest-dom: 6.2.0
```

```
System: node 20.10.0
    Compiler: [...]/node_modules/@stencil/core/compiler/stencil.js
       Build: 1704730339
     Stencil: 4.9.1 🍬
  TypeScript: 5.2.2
      Rollup: 2.42.3
      Parse5: 7.1.2
      jQuery: 4.0.0-pre
      Terser: 5.26.0
```

### Steps to Reproduce

Component under test:

```typescript

@Component({
  tag: 'test-component',
  shadow: false,
})
export class TestComponent {
  render() {
    return <button>test</button>;
  }
}
```

test:

```typescript
import { newSpecPage } from '@stencil/core/testing';
import { screen } from '@testing-library/dom';

import '@testing-library/jest-dom';
import { TestComponent } from './test-component';

describe('test component', () => {
  it('can find button using getByRole', async () => {
    await newSpecPage({
      components: [TestComponent],
      html: `<test-component/>`,
    });
    const button = screen.getByRole('button');
    expect(button).toEqualHtml('<button>test</button>');
  });
});

```

### Code Reproduction URL

https://github.com/angelo-v/stencil-testing-library-dom-error/blob/main/src/components/my-component/my-component.spec.ts

### Additional Information

It works when setting the `hidden` option:

```
screen.getByRole('button', {hidden: true});
```

christian-bromann:

`MockHTMLElement` returns `MockDocument` when accessing `parentElement`.

fixes #5252
STENCIL-1104

## What is the new behavior?

Accessing `parentElement` in from the `<html />` node always returns `null` as defined by the spec.

According to the [spec](https://dom.spec.whatwg.org/#parent-element):

> A [node](https://dom.spec.whatwg.org/#concept-node)’s [parent](https://dom.spec.whatwg.org/#concept-tree-parent) of type [Element](https://dom.spec.whatwg.org/#element) is known as its parent element. If the [node](https://dom.spec.whatwg.org/#concept-node) has a [parent](https://dom.spec.whatwg.org/#concept-tree-parent) of a different type, its [parent element](https://dom.spec.whatwg.org/#parent-element) is null.

The HTML document is _not_ considered to be an element.

- [ ] Yes
- [x] No

## Testing

Added a simple unit test for this. I also created a reproducible example based on the original issue:

- `git clone git@github.com:christian-bromann/getbyrole-repro.git`
- `cd getbyrole-repro`
- `npm install`
- `npm test`

The test will pass. The project uses a Stencil dev build `4.12.1-dev.1707312078.2f2e3e3` based on this PR.

## Other information

n/a

### DIFF
--- a/src/mock-doc/test/event.spec.ts
+++ b/src/mock-doc/test/event.spec.ts
@@ -291,10 +291,12 @@ describe('event', () => {
 
       const composedPath = event.composedPath();
 
-      expect(composedPath).toHaveLength(3);
+      expect(composedPath).toHaveLength(5);
       expect(composedPath[0]).toBe(divElm);
       expect(composedPath[1]).toBe(doc.body);
       expect(composedPath[2]).toBe(doc.documentElement);
+      expect(composedPath[3]).toBe(doc);
+      expect(composedPath[4]).toBe(win);
     });
 
     it('returns the correct path for a nested element', () => {
@@ -311,11 +313,13 @@ describe('event', () => {
 
       const composedPath = event.composedPath();
 
-      expect(composedPath).toHaveLength(4);
+      expect(composedPath).toHaveLength(6);
       expect(composedPath[0]).toBe(pElm);
       expect(composedPath[1]).toBe(divElm);
       expect(composedPath[2]).toBe(doc.body);
       expect(composedPath[3]).toBe(doc.documentElement);
+      expect(composedPath[4]).toBe(doc);
+      expect(composedPath[5]).toBe(win);
     });
   });
 });
--- a/src/mock-doc/test/event.spec.ts
+++ b/src/mock-doc/test/event.spec.ts
@@ -291,12 +291,10 @@ describe('event', () => {
 
       const composedPath = event.composedPath();
 
-      expect(composedPath).toHaveLength(5);
+      expect(composedPath).toHaveLength(3);
       expect(composedPath[0]).toBe(divElm);
       expect(composedPath[1]).toBe(doc.body);
       expect(composedPath[2]).toBe(doc.documentElement);
-      expect(composedPath[3]).toBe(doc);
-      expect(composedPath[4]).toBe(win);
     });
 
     it('returns the correct path for a nested element', () => {
@@ -313,13 +311,11 @@ describe('event', () => {
 
       const composedPath = event.composedPath();
 
-      expect(composedPath).toHaveLength(6);
+      expect(composedPath).toHaveLength(4);
       expect(composedPath[0]).toBe(pElm);
       expect(composedPath[1]).toBe(divElm);
       expect(composedPath[2]).toBe(doc.body);
       expect(composedPath[3]).toBe(doc.documentElement);
-      expect(composedPath[4]).toBe(doc);
-      expect(composedPath[5]).toBe(win);
     });
   });
 });
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -57,3 +57,10 @@ jobs:
     uses: ./.github/workflows/test-unit.yml
     with:
       build_name: stencil-core
+
+  pack_and_comment:
+    name: Pack and Comment
+    needs: [build_core]
+    uses: ./.github/workflows/pack-and-comment.yml
+    with:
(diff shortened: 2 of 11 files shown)


src/mock-doc/test/event.spec.ts
```diff
@@ -313,13 +311,11 @@ describe('event', () => {
...
       const composedPath = event.composedPath();
 
-      expect(composedPath).toHaveLength(6);
+      expect(composedPath).toHaveLength(4);
       expect(composedPath[0]).toBe(pElm);
       expect(composedPath[1]).toBe(divElm);
       expect(composedPath[2]).toBe(doc.body);
       expect(composedPath[3]).toBe(doc.documentElement);
-      expect(composedPath[4]).toBe(doc);
-      expect(composedPath[5]).toBe(win);
```
rwaskiewicz, line 322: I think we still need `composedPath`'s return value to propagate all the way up to `Window`, per https://developer.mozilla.org/en-US/docs/Web/API/Event/composedPath. Is that not the case?

rwaskiewicz (changes requested):
Are there any instructions for verifying this against the reproduction case?

I installed a locally built tarball in the repro and got:
```
npm t

> testing-library-test@0.0.1 test
> stencil test --spec

[19:31.2]  @stencil/core
[19:31.4]  [LOCAL DEV] v4.12.1-dev.1707311711.2f2e3e3 😸
[19:31.4]  testing spec files
[19:31.4]  jest args: --spec --max-workers=8
  my-component
    ✕ renders (8 ms)

  ● my-component › renders

    TypeError: Right-hand side of 'instanceof' is not an object

      14 |     const button = getByRole(root, 'button');
      15 |
    > 16 |     expect(button).toBeInTheDocument();
         |                    ^
      17 |   });
      18 |
      19 | });

      at checkHtmlElement (node_modules/@testing-library/jest-dom/dist/matchers-57b266e9.js:85:19)
      at Object.toBeInTheDocument (node_modules/@testing-library/jest-dom/dist/matchers-57b266e9.js:274:5)
      at __EXTERNAL_MATCHER_TRAP__ (node_modules/expect/build/index.js:325:30)
      at Object.throwingMatcher [as toBeInTheDocument] (node_modules/expect/build/index.js:326:15)
      at Object.<anonymous> (src/components/my-component/my-component.spec.ts:16:20)

Test Suites: 1 failed, 1 total
Tests:       1 failed, 1 total
Snapshots:   0 total
Time:        0.425 s, estimated 1 s
Ran all test suites.
```

This insinuates there's some issue pulling in that `toBeInTheDocument`, rather than an issue with the fix here - curious if you're seeing that as well

christian-bromann:
> This insinuates there's some issue pulling in that `toBeInTheDocument`, rather than an issue with the fix here - curious if you're seeing that as well

I haven't. I pushed a reproducible example and added some instructions to run it in the PR description.

src/mock-doc/test/event.spec.ts
tanner-reits: I think you're right. I recreated the test "returns the correct path for a simple div element" in the browser and that generated a path of `div`, `body`, `html`, `document`, `window`

tanner-reits (changes requested):
Just blocking for those couple tests

christian-bromann:
@rwaskiewicz @tanner-reits thanks for the review, I was able to revert all test changes with some further updates on the event handler.

rwaskiewicz (approved):
LGTM/works as expected on Christian's repro

tanner-reits (approved):

PR #5337, opened 2024-02-07
fix(renderer): fix missing slot ref callback handling

yigityuce:
fixes one of the issues raised in #5335

adds a new internal variable which is stored during vnode rendering to reuse it within the relocatedNodes logic

## What is the current behavior?

Slot element's `ref` attribute is not handled correctly.

GitHub Issue Number:  #5335 

## What is the new behavior?

Fixed

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

PR #5343, opened 2024-02-09
chore(ci): refactor pack-and-comment workflow to support community PRs

alicewriteswrongs:
This changes the workflow from being a `workflow_call` triggered workflow which is intended to be called from _another_ workflow (`main`) and instead makes it a `pull_request_target` workflow which is "standalone". This means that no special privileges will be necessary to run it and that will in turn allow the `create-or-update-comment` action to work properly.

See these resources for more details:

- https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#restrictions-on-repository-forks
- https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/

Our little tarball packing / commenting workflow doesn't work on community PRs! E.g. https://github.com/ionic-team/stencil/pull/5337

## What is the new behavior?

This should enable it to work on any PR by moving to `pull_request_target`.

- [ ] Yes
- [x] No

## Testing

As I understand it this has to be merged to `main` before it will start to be triggered on newer PRs, so we'll have to do something like merge this and then open another PR to test whether it does what we think it should.

### DIFF
--- a/.github/workflows/pack-and-comment.yml
+++ b/.github/workflows/pack-and-comment.yml
@@ -1,33 +1,32 @@
 name: Pack and Comment
+# for each PR this workflow builds and then packs a Stencil tarball and then
+# posts a comment on the PR with a link to download the tarball, as well as an
+# example command for installing the tarball in a project.
 
 on:
-  workflow_call:
-    # Make this a reusable workflow, no value needed
-    # https://docs.github.com/en/actions/using-workflows/reusing-workflows
-    inputs:
-      build_name:
-        description: Name for the build, used to resolve the correct build artifact
-        required: true
-        type: string
+  pull_request_target:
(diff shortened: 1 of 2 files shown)


rwaskiewicz (approved):

.github/workflows/pack-and-comment.yml
```diff
@@ -1,33 +1,30 @@
 name: Pack and Comment
+# this workflow builds and then posts on a PR a
```
rwaskiewicz, line 2: Non blocking - comment ends short here

alicewriteswrongs: oops!

christian-bromann (approved):

Issue #5335, opened 2024-02-06
bug: v4 Scoped Component Issues


yigityuce:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

4.12.1

### Current Behavior

Regarding the discussion in [here](https://discord.com/channels/520266681499779082/1202888778549694524) we started to test out the "scoped" component approach. Please take a look at our checklist below and the status.

## Terms:
- "with slot fixes": `experimentalSlotFixes` and `scopedSlotTextContentFix` configs are applied
- "without slot fixes": `experimentalSlotFixes` and `scopedSlotTextContentFix` configs are **NOT** applied

## Status:
| Test Case | Description | with slot fixes | without slot fixes | PR |
| --- | --- | --- | --- | --- |
| :white_check_mark: Dynamic Tag | We need to have a dynamically updatable slot parent tag name to pass the accessibility audits | OK | OK | - |
| :white_check_mark: Dynamic Sibling Children | We need to have dynamically re-orderable or conditionally rendered default slot sibling elements | OK | OK | - |
| Nested Relocated & Dynamic Sibling Children | We need to have dynamically re-orderable children and also these children elements can be passed through several components via using default slot | FAIL | FAIL | #5403  |
| Conditional Rendering (named slot) | We need to have conditionally rendered named slot element(s) | FAIL | OK | #5365 |
| Conditional Rendering (default slot) | We need to have conditionally rendered default slot element(s) | PARTIALLY<sup>1</sup> | PARTIALLY<sup>1</sup> | #5365 |
| Slot fallback | We need to have stable slot fallback | FAIL<sup>1</sup> | FAIL<sup>1</sup> | - |
| Slot ref handling | We need to be able to handle the `ref` attribute for slots | FAIL | FAIL | #5337 |
| Update "textContent" | We need to be able to update the default slot text by using component's `textContent` attribute | FAIL | FAIL | #5354 |
| Update "innerText" | We need to be able to update the default slot text by using component's `innerText` attribute | FAIL | FAIL | - |

<br />
<sup>1</sup>: slot contents are hidden but slot fallback content is not visible

[...]

PR #5354, opened 2024-02-12
feat(renderer): add flag to update default slot text content

yigityuce:
add "experimentalDefaultSlotTextContentFix" flag to update text content of the default slot instead of update all the content

fixes one of the issues raised in #5335

As of now, when the "experimentalScopedSlotChanges" flag is set, it removes all the content of the custom component (including named slots) and then inserts the text.

GitHub Issue Number: #5335 

## What is the new behavior?

When the newly introduced "experimentalDefaultSlotTextContentFix" is set, updating `textContent` updates only the "default slot" content while keeping the named slot contents.

- [ ] Yes
- [ ] No


christian-bromann:
Thanks for the contribution @yigityuce 🙏 

We have discussed this idea/proposal with the team and have decided that this is __not__ something we would like to support. It creates a behavior that contradicts with how the Shadow DOM behaves which is not a direction we would like to take. Furthermore we believe that you can achieve your goals by modifying the query in your example, e.g.:

```html
<body>
  <my-component>
    <span slot="prefix">Prefix</span>
    <span slot="suffix">Suffix</span>
    <span>Content text</span>
  </my-component>
  <button onclick="onButtonClick()" style="margin-top: 48px">Update content</button>
</body>
<script>
  function onButtonClick() {
    const defaultSlot = document.querySelector('my-component span:not([slot])');
    defaultSlot && (defaultSlot.textContent = `${Date.now()}`);
  }
</script>
```

Is there any reason this wouldn't work for you?

yigityuce:
> Is there any reason this wouldn't work for you?

What if I don't wrap the "Content text" with `span` element and want to update it? what would be the selector to get that textNode

yigityuce:
and one more example that comes up my mind is like that:


```tsx
// component lib file
@Component({  tag: 'my-component' })
export class MyComponent {
  @Event() randomize: EventEmitter<void>;
  @Event() reset: EventEmitter<void>;
 
  render() {
    return (
       <Host>
         <button onClick={() => this.randomize.emit()}>Randomize</button>
         <slot />
         <button onClick={() => this.reset.emit()}>Reset</button>
       </Host>
    );
  }
}
```

```tsx
// in react context

export const Example: FC = () => {
    const [counter, setCounter] = useState(0);
    return (
      <>
        <my-component onReset={() => setCounter(0)} onRandomize={() => setCounter(Date.now() % 10)}>
          {counter}
        </my-component>
        <button onClick={() => setCounter(counter + 1)}>Increment</button>
      </>
    );
}
```



```html
<!-- Expected rendered DOM after clicking increment button -->
<my-component>
  <button>Randomize</button>
  1
  <button>Reset</button>
</my-component>
<button>Increment</button>


<!-- Actual rendered DOM after clicking increment button -->
<my-component>
  1
</my-component>
<button>Increment</button>

```

because current behavior is to remove all nodes first and then insert an new text to to the host

christian-bromann:
> and one more example that comes up my mind is like that:

Can you share this as a Stackblitz project?

yigityuce:
Hi @christian-bromann 

> Can you share this as a Stackblitz project?

I tried to reproduce but I realized it is my bad, it is **NOT a valid case**, so pls ignore that example. 


However, during this try-out, I found out another edge case with the given query selector to query the not slotted element. `document.querySelector('my-component span:not([slot])')`


I directly copied the content of the files here to explain and also created a [new branch](https://github.com/yigityuce/stencil-v4-sc-default-slot-textcontent-update/commit/c7f4fe7df1478ee6972634f880487fcd573f2e04) with an updated version of the repro repo.

```tsx
@Component({
  tag: 'my-component',
  styleUrl: 'my-component.css',
  shadow: false,
  scoped: true,
})
export class MyComponent {
  render() {
    return (
      <Host>
        <div class="some-fancy-styles" id="prefix-wrapper">
          <slot name="prefix" />
        </div>
        <slot />
        <div class="some-fancy-styles" id="suffix-wrapper">
          <slot name="suffix" />
        </div>
      </Host>
    );
  }
}

```




```html
<!DOCTYPE html>
<html dir="ltr" lang="en">
  <head>
    <meta charset="utf-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1.0, minimum-scale=1.0, maximum-scale=5.0" />

    <script type="module" src="/build/stencil-v4-sc-default-slot-textcontent-update.esm.js"></script>
    <script nomodule src="/build/stencil-v4-sc-default-slot-textcontent-update.js"></script>
  </head>
  <body>
    <my-component>
      <span slot="prefix">Prefix</span>
      <span slot="suffix">Suffix</span>
      <div>Content text</div>
    </my-component>
    <button onclick="onButtonClick()" style="margin-top: 48px">Update content</button>
  </body>
  <script>
    function onButtonClick() {
      const contentTextWrapper = document.querySelector('my-component div:not([slot])');
      console.log('not slotted:', contentTextWrapper)
      contentTextWrapper && (contentTextWrapper.textContent = `${Date.now()}`);
    }
  </script>
</html>
```

so the thing is when you hit the button, the `prefix-wrapper` div's textContent gets updated because it matches with `'my-component div:not([slot])'` selector. 

The only way that I see to accomplish this is to provide a unique selector to select `<div>Content text</div>` element. In that option, you should be aware of the custom component's internal DOM structure to avoid collision in the selectors which should not be the end users' concern for those who consume the UI kit.


Or, a big or, being able to update only the default slot element while mutating the textContent (which is the solution I propose). What would be the drawbacks of this proposal and the mitigation steps:
1. Case: If someone wants to change named slot content -> Mitigation: they can always use this selector: `'component-name [slot="slot-name"]'`.
2. Case: If someone wants to put just the text by clearing out all the slots -> Mitigation: they can remove all the slot elements in the custom component and update the custom component's textContent attr. By applying my proposal, this will work as well since only the default slotted element will be updated and user won't need to query this not slotted element which has some potential collision as of now.


Summary: I mean updating the named slot textContent or removing the content & updating the custom component's textContent has a workaround, but it is not the case for the default slot with textNode.


One more thing might work in future but it's not usable as of today because of the browser compatibility: [::target-text](https://caniuse.com/mdn-css_selectors_target-text)

christian-bromann:
> so the thing is when you hit the button, the `prefix-wrapper` div's textContent gets updated because it matches with `'my-component div:not([slot])'` selector.

Try to make the selector more strict, e.g. `my-component > div:not([slot])`

> What would be the drawbacks

The drawbacks that caused the team to push back on this proposal are:

- this introduces a diverge in how the component behaves compared to the Shadow DOM spec
- while this seems harmless at first glance it requires us to teach the community that this specific behavior differs from the specification
- it requires us to maintain this specific edge case when there is no technical reason to introduce this

> One more thing might work in future but it's not usable as of today because of the browser compatibility: [::target-text](https://caniuse.com/mdn-css_selectors_target-text)

I am not sure if this selector actually addresses this.

I feel inclined to close the issue. Please see above reasons why we don't feel comfortable merging the proposal. The DOM API provides various options that allows to query the element you like to update. We don't see any need to provide workarounds that introduce contradicting behavior to the Shadow DOM spec.

yigityuce:
hey @christian-bromann 

I understand your concerns and you have a valid point so I'm closing this PR. Thanks for the discussion and for showing the point of your view clearly.

Issue #5353, opened 2024-02-12
bug: `emulate`is not used for E2E Screenshot tests


mlechler:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

4.12.1

### Current Behavior

We have four viewports defined in the `emulate` array in the `stencil.config.ts`. When running our e2e screenshot tests these defined viewports aren't used. As mentioned [here](https://github.com/ionic-team/stencil/issues/4601#issuecomment-1648447341), I understood, that the `emulate` array should still be used for screenshot tests, which isn't the case.

### Expected Behavior

Use the `emulate` array of the `stencil.config.ts` or provide a easy way to use multiple viewports for every e2e screenshot test, without setting every single one of it.

### System Info

```shell
System: node 20.11.0
    Platform: darwin (21.6.0)
   CPU Model: Apple M1 Pro (10 cpus)
    Compiler: /node_modules/@stencil/core/compiler/stencil.js
       Build: 1707148558
     Stencil: 4.12.1 🏸
  TypeScript: 5.3.3
      Rollup: 2.56.3
      Parse5: 7.1.2
      jQuery: 4.0.0-pre
      Terser: 5.27.0
```

### Steps to Reproduce

Run `npm run test.screenshot` in the test repo.
Should generate screenshot with 600x600, but generates 800x600

### Code Reproduction URL

https://github.com/mlechler/stencil-e2e-test-bug

tanner-reits:
@mlechler I confirmed that the generated screenshots do not have the same dimensions as what is being set in the Stencil config. I'll get this ingested in our backlog for someone to take a look at.

mlechler:
Hi @tanner-reits, what is the time horizon on this issue?

tanner-reits:
@mlechler I don't have a definitive timeline right now. I spent some more time playing around with a fix, but we uncovered some additional discrepancies in the generated output. The team needs to dig in deeper to the screenshot testing process so we're confident in the change(s) we make. This is an experimental feature in Stencil after all, so there are many kinks and nuances we need to work through.

christian-bromann:
[...]

PR #5359, opened 2024-02-13
fix(testing): use viewport for Puppeteer screenshot clip dimensions

tanner-reits:

Screenshots generated by Stencil's screenshot testing (via Puppeteer) do not have the dimensions of the viewport specified in the Stencil config's `testing.emulate` option.

Fixes: #5353

## What is the new behavior?

If [`clip` options](https://pptr.dev/api/puppeteer.screenshotclip) are not explicitly passed to `page.compareScreenshot()`, Stencil will default the clip option `width` and `height` to the `width` and `height` dimensions of the viewport in the emulate config.

## Documentation

N/A

- [ ] Yes
- [X] No


christian-bromann (approved):
👍

src/testing/puppeteer/puppeteer-screenshot.ts
```diff
@@ -137,7 +137,7 @@ export async function pageCompareScreenshot(
   return results;
 }
 
-function createPuppeteerScreenshotOptions(opts: ScreenshotOptions) {
+export function createPuppeteerScreenshotOptions(opts: ScreenshotOptions, width: number, height: number) {
```
rwaskiewicz, line 140: Can we adjust this function a bit so that it's clear that `width` and `height` are meant to be provided together? I know its a bit nitpicky, but I think that it'll help to demonstrate that these two value are related to one another:
```suggestion
export function createPuppeteerScreenshotOptions(opts: ScreenshotOptions, {width, height}: {width: number, height: number}) {
```

src/testing/puppeteer/puppeteer-screenshot.ts
```diff
@@ -123,6 +118,11 @@ export async function pageCompareScreenshot(
     }
   }
 
+  const screenshotOpts = createPuppeteerScreenshotOptions(opts, width, height);
```
rwaskiewicz, line 121: Can we add a small comment here that clarifies that we just updated `width` and `height`, and that the order of operations is important here (since `opts.clip` can influence `width` and `height` in the conditional directly above this)

christian-bromann:
@tanner-reits I verified this change while looking into #5421 and it seems the behavior changed with this patch. While before the baseline for [this screenshot test](https://github.com/ionic-team/stencil/blob/b5a2536d0ac40a6ad940c55fcc52c8cdfcfa8b15/test/end-to-end/src/app-root/app-root.e2e.ts#L34-L44) had a dimension of 800x600px, the new captured image has a dimension of 600x600px. I can't spot an `emulate` configuration within a `stencil.config.ts` anywhere in our repository so I am not sure if it is desired to have the screenshot taken at this dimension.

I am not sure if this is related to a patch I made some month ago: https://github.com/ionic-team/stencil/pull/5205.

tanner-reits:
@christian-bromann Got to the bottom of that. If no emulate config is specified in a Stencil config's `testing` object, a default config is provided by [`validateTesting`](https://github.com/ionic-team/stencil/blob/main/src/compiler/config/validate-testing.ts#L168-L183) during config validation.

This is a bit confusing, however, because we also have a stock emulate config in [`setScreenshotEmulateData`](https://github.com/ionic-team/stencil/blob/main/src/testing/puppeteer/puppeteer-emulate.ts#L5-L16), but based on the validation logic this will always get overridden.

That being said, I am still confused why the screenshot being generated on `main` is 600x800 and not 600x600. The `master.json` file that gets generated says the dimensions are 600x600. So, truthfully not sure what the correct behavior here is.

christian-bromann:
> That being said, I am still confused why the screenshot being generated on `main` is 600x800 and not 600x600.

I was able to verify that with this branch the screenshots when running `❯ npm run test.end-to-end` are correctly taken with 600x600 pixel as noted in the `master.json`. 🙌 

This leaves with the one last remaining confusion which is when does Stencil default back to the values set in [`setScreenshotEmulateData`](https://github.com/ionic-team/stencil/blob/main/src/testing/puppeteer/puppeteer-emulate.ts#L5-L16). Regardless of this, we should move forward with merging this.

rwaskiewicz (approved):

PR #5366, opened 2024-02-15
chore(repo): add puppeteer to the renovate ignore list

rwaskiewicz:

See 'new behavior' section

GitHub Issue Number: N/A

## What is the new behavior?

ignore dependency bumps for puppeteer for versions 22 and higher. starting with v22, puppeteer dropped support for node 16, which stencil v4 still supports.

## Documentation

N/A - internal change

- [ ] Yes
- [x] No

## Testing

I tested the using Renovate's validator tool:
```
`npm i -g renovate`
`renovate-config-validator`
```

This section also matches others that we have for ignoring Node 16-related packages

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -149,6 +149,11 @@
     {
       matchPackageNames: ['open'],
       allowedVersions: '<10',
+    },
+    // TODO(STENCIL-1141): remove once support for Node v16 is dropped
+    {
+      matchPackageNames: ['puppeteer'],
+      allowedVersions: '<21',
     }
   ],
   // Never rebase the branch or update it unless manually requested to avoid noisy PR emails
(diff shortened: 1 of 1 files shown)


renovate.json5
```diff
@@ -149,6 +149,11 @@
     {
       matchPackageNames: ['open'],
       allowedVersions: '<10',
+    },
+    // TODO(STENCIL-1141): remove once support for Node v16 is dropped
+    {
+      matchPackageNames: ['puppeteer'],
+      allowedVersions: '<21',
```
tanner-reits, line 156: Should this be 
```suggestion
      allowedVersions: '<=21',
```

tanner-reits (changes requested):

renovate.json5
rwaskiewicz: Ahh good catch!

christian-bromann (approved):

PR #5368, opened 2024-02-15
chore(repo): remove old release scripts pt2

rwaskiewicz:
**Do not merge, waiting for parent to land first**

We have a good deal of code related to releases that no longer runs.

GitHub Issue Number: N/A

## What is the new behavior?

now that stencil releases are largely automated, release tasks that
were skipped during manual releases can be removed. this commit removes
all tasks that were skipped with `isCI` is was `true` (as it's always
`true` these days).

additional conditionals that use logical-AND with `isCI` have been
updated, as that part of the conditional is now always `true`.

remove the `--otp` flag, as it is only used for manual releases

as a result of these removals, the `--any-branch` flag can be safely
removed from our release pipeline, as its no longer used.

## Documentation

N/A - this old process being removed had no documentation 🙈 

- [ ] Yes
- [x] No

## Testing

I kicked off a dev publish of Stencil using this branch, which ran successfully (and published to NPM) https://github.com/ionic-team/stencil/actions/runs/7917891689

I was able to run this build (`4.12.2-dev.1708009206.09089a5`) locally in a basic Stencil project.

There's _some_ risk here as I haven't cut a prod release, but could cut a prerelease of our next version if someone wants to test that with me 🙂

### DIFF
--- a/scripts/release-tasks.ts
+++ b/scripts/release-tasks.ts
@@ -2,15 +2,8 @@ import color from 'ansi-colors';
 import Listr, { ListrTask } from 'listr';
 
 import { bundleBuild } from './build';
-import { validateBuild } from './test/validate-build';
 import { BuildOptions } from './utils/options';
-import {
-  isPrereleaseVersion,
-  isValidVersionInput,
-  postGithubRelease,
-  SEMVER_INCREMENTS,
-  updateChangeLog,
-} from './utils/release-utils';
+import { isPrereleaseVersion, isValidVersionInput, SEMVER_INCREMENTS, updateChangeLog } from './utils/release-utils';
 
 /**
  * Runs a litany of tasks used to ensure a safe release of a new version of Stencil
@@ -23,7 +16,6 @@ export async function runReleaseTasks(opts: BuildOptions, args: ReadonlyArray<st
   const tasks: ListrTask[] = [];
   const newVersion = opts.version;
   const isDryRun = args.includes('--dry-run') || opts.version.includes('dryrun');
-  const isAnyBranch = args.includes('--any-branch');
   let tagPrefix: string;
 
   const { execa } = await import('execa');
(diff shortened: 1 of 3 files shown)


scripts/release-tasks.ts
```diff
@@ -2,15 +2,8 @@ import color from 'ansi-colors';
 import Listr, { ListrTask } from 'listr';
 
 import { bundleBuild } from './build';
-import { validateBuild } from './test/validate-build';
 import { BuildOptions } from './utils/options';
-import {
-  isPrereleaseVersion,
-  isValidVersionInput,
-  postGithubRelease,
```
rwaskiewicz, line 10: `postGithubRelease` is still in the codebase, but technically no longer used. Leaving it here for the moment to see if I can get GitHub to do this for us

scripts/release-tasks.ts
```diff
@@ -61,113 +53,66 @@ export async function runReleaseTasks(opts: BuildOptions, args: ReadonlyArray<st
...
-            (err) => {
-              // Command fails with code 1 and no output if the tag does not exist, even though `--quiet` is provided
-              // https://github.com/sindresorhus/np/pull/73#discussion_r72385685
-              if (err.stdout !== '' || err.stderr !== '') {
-                throw err;
-              }
-            },
-          ),
-      skip: () => isDryRun,
-    },
```
rwaskiewicz, line 97: I recommend checking this section out with whitespace disabled.

scripts/release-tasks.ts
```diff
@@ -61,113 +53,66 @@ export async function runReleaseTasks(opts: BuildOptions, args: ReadonlyArray<st
...
+          },
+          (err) => {
+            // Command fails with code 1 and no output if the tag does not exist, even though `--quiet` is provided
+            // https://github.com/sindresorhus/np/pull/73#discussion_r72385685
+            if (err.stdout !== '' || err.stderr !== '') {
+              throw err;
+            }
+          },
+        ),
+    skip: () => isDryRun,
```
christian-bromann, line 87: Non blocking suggestion to make it also a bit more readable using `async/await` instead of promise chaining:

```suggestion
    task: async () => {
      await execa('git', ['fetch'])
      // Retrieve the prefix for a version string - https://docs.npmjs.com/cli/v7/using-npm/config#tag-version-prefix
      const { stdout: tagPrefix } = await execa('npm', ['config', 'get', 'tag-version-prefix'])
        .catch(() => ({}))
      // verify that a tag for the new version string does not already exist by checking the output of
      // `git rev-parse --verify`
      const { stdout: hasGitTag } = await execa('git', ['rev-parse', '--quiet', '--verify', `refs/tags/${tagPrefix}${newVersion}`]).catch((err) => {
        // Command fails with code 1 and no output if the tag does not exist, even though `--quiet` is provided
        // https://github.com/sindresorhus/np/pull/73#discussion_r72385685
        if (err.stdout !== '' || err.stderr !== '') {
          throw err;
        }
      });
      if (hasGitTag) {
        throw new Error(`Git tag \`${tagPrefix}${newVersion}\` already exists.`);
      }
    },
    skip: () => isDryRun,
```

christian-bromann (approved):
🔪  LGTM 👍

tanner-reits (approved):

scripts/release-tasks.ts
rwaskiewicz: I'm gonna hold off on this for the moment - I want to see if we can eliminate `Listr` here entirely. I started to do this (`async` work) and a few `if` statement reorders and felt like I was optimizing something that might be a better off being rewritten with some testing under it (and I didn't get that far on JEDI days 😉)

rwaskiewicz:
I'll merge this after today's Stencil release

PR #5370, opened 2024-02-15
feat(ci): remove need to compile scripts

christian-bromann:
Currently a developer working on Stencil needs to compile the entire scripts folder first before running the files in them. This seems cumbersome and there is an opportunity to improve the DX here.

## What is the new behavior?
By adding [`tsx`](https://www.npmjs.com/package/tsx) we can get rid of the intermediate build step by running TS files on demand. This simplifies a lot of internal scripts and confusion when it comes to resolving paths.

- [ ] Yes
- [x] No

## Testing

Checkout the branch and run `npm run build`.

## Other information

n/a


christian-bromann:
> Tech Debt Burndown / Download error files and report (pull_request_target) Failing after 31s

This is due to the fact that we are running the workflow from `main` which still uses the compiled script.

christian-bromann:
@tanner-reits @alicewriteswrongs this is ready for review now.

rollup.config.js
tanner-reits, line 1: Can we remove the `rollup.prod.ci` and `rollup.watch` scripts in the `package.json` too if we're killing this?

package.json
```diff
@@ -122,6 +119,7 @@
     "rollup-plugin-sourcemaps": "^0.6.3",
     "semver": "^7.3.7",
     "terser": "5.31.0",
+    "tsx": "^4.10.3",
```
tanner-reits, line 122: Any reasoning on choosing `tsx` over `ts-node`? Just curious.

tanner-reits (changes requested):

package.json
christian-bromann: `tsx` is much better maintained as `ts-node`. I am switching in all my projects to tsx now.

rollup.config.js
christian-bromann: done

alicewriteswrongs:
Definitely a nicer DX here! It's a bit less mysterious why we're doing what we doing for the build here I think.

One thing I'm wondering about is type checking for the `scripts/` directory. In our current setup on `main` we run a full typescript build of the directory which means that we'll never end up running the scripts if there's a type error. From what I can tell `tsx` [takes the stance as a project that type checking in e.g. your editor is good enough](https://tsx.is/usage#no-type-checking) and does not do any type checking when you go to execute a file (I confirmed that by putting `const stringy:  string = 1;` into `scripts/index.ts` and I was able to run it with no error).

Anywhooo - just wondering if this is the direction we want to go, since it does open up the possibility of some type errors slipping through in for instance the release scripts and then only showing up as runtime error (or wrong behavior!) when we run on CI.

The `tsx` docs suggest possibly doing a `tsc --noEmit` before you run the scripts - does that make sense here?

package.json
```diff
@@ -54,10 +52,8 @@
     "test.docs-build": "cd test && npm run build.docs-json && npm run build.docs-readme",
     "test.watch": "node --experimental-vm-modules ./node_modules/jest/bin/jest.js --watch",
     "test.watch-all": "node --experimental-vm-modules ./node_modules/jest/bin/jest.js --watchAll --coverage",
-    "tsc": "tsc --incremental",
-    "tsc.prod": "tsc",
-    "tsc.scripts": "tsc --build --force scripts/tsconfig.json",
-    "tsc.watch": "tsc --incremental --watch"
+    "ts": "tsc --noEmit --project scripts/tsconfig.json && tsx",
```
alicewriteswrongs, line 55: this is a nice way to do this 👍

alicewriteswrongs (approved):

package.json
```diff
@@ -23,28 +23,26 @@
...
+    "clean:scripts": "rm -rf scripts/build",
     "lint": "eslint 'bin/*' 'scripts/*.ts' 'scripts/**/*.ts' 'src/*.ts' 'src/**/*.ts' 'src/**/*.tsx' 'test/wdio/**/*.tsx'",
     "install.jest": "bash ./src/testing/jest/install-dependencies.sh",
     "prettier": "npm run prettier.base -- --write",
     "prettier.base": "prettier --cache \"./({bin,scripts,src,test}/**/*.{ts,tsx,js,jsx})|bin/stencil|.github/(**/)?*.(yml|yaml)|*.js\"",
     "prettier.dry-run": "npm run prettier.base -- --list-different",
-    "release.ci.prepare": "npm run tsc.scripts && node scripts/build --release --ci-prepare",
-    "release.ci": "npm run tsc.scripts && NODE_OPTIONS=--max-old-space-size=4096 node scripts/build --release --ci-publish",
-    "rollup.prod.ci": "rollup --config --config-prod --config-ci",
-    "rollup.watch": "rollup --watch --config",
```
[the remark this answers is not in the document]
christian-bromann: 😉

christian-bromann:
> CI error, but otherwise looks good

This is due to the fact that we are running the workflow from main which still uses the compiled script. Therefore I suggest to suggest merging by bypassing the protection.

PR #5375, opened 2024-02-16
fix(sys): fix expected types for `createNodeLogger` and `createNodeSys`

christian-bromann:
The methods `createNodeLogger` and `createNodeSys` currently have a wrong parameter definition.

## What is the new behavior?
Update types to match with implementation:

- `createNodeLogger`: see https://github.com/ionic-team/stencil/blob/9062e1a967461dd962881b925b0c55aebddd8975/src/sys/node/logger/index.ts#L12-L16
- `createNodeSys`, see https://github.com/ionic-team/stencil/blob/4959295c24ec3effcc8d63a8305dffd6e07a617d/src/sys/node/node-sys.ts#L37

- [ ] Yes
- [x] No

Yes and no: yes we change public API interfaces but also no since these have been wrong in the first place.

## Testing

n/a

## Other information

n/a


src/sys/node/public.ts
```diff
@@ -3,19 +3,25 @@ import type { CompilerSystem, Logger } from '@stencil/core/internal';
 /**
  * Creates a "logger", based off of NodeJS APIs, that will be used by the compiler and dev-server.
  * The NodeJS "process" object must be provided as a property in the first argument's object.
- * @param c an object containing a `Process` for Stencil to use
  * @returns a {@link Logger} object
  */
-export declare function createNodeLogger(c: { process: any }): Logger;
+export declare function createNodeLogger(): Logger;
```
rwaskiewicz, line 8: It looks like we removed this in https://github.com/ionic-team/stencil/commit/c114f0a9ce4352672d86a8605df53f8b7d81aa4f - I don't see any comments to this effect, but I _think_ we held off here due to breaking a public API (`createNodeLogger` is a available from `@stencil/core/sys/node`). @alicewriteswrongs do you remember by chance?

rwaskiewicz, line 8: If we do move forward with this change, we should update https://github.com/ionic-team/stencil/blob/b91f94a122dc9add876f0277f99afc92c96e66dd/scripts/test/validate-build.ts#L267 as well

src/sys/node/public.ts
```diff
@@ -3,19 +3,25 @@ import type { CompilerSystem, Logger } from '@stencil/core/internal';
...
  * Creates the "system", based off of NodeJS APIs, used by the compiler. This includes any and
  * all file system reads and writes using NodeJS. The compiler itself is unaware of Node's
  * `fs` module. Other system APIs include any use of `crypto` to hash content. The NodeJS
  * "process" object must be provided as a property in the first argument's object.
- * @param c an object containing a `Process` for Stencil to use
+ * @param c an object containing a `Process` for Stencil to use and a logger instanced created by {@link createNodeLogger}
  * @returns a {@link CompilerSystem} object
  */
-export declare function createNodeSys(c: { process: any }): CompilerSystem;
+export declare function createNodeSys(c: { process: any; logger: any }): CompilerSystem;
```
rwaskiewicz, line 18: Should we make both `process` and `logger` optional here to align with https://github.com/ionic-team/stencil/blob/4959295c24ec3effcc8d63a8305dffd6e07a617d/src/sys/node/node-sys.ts#L37?

```suggestion
export declare function createNodeSys(c: { process?: any; logger?: any }): CompilerSystem;
```

rwaskiewicz (changes requested):

src/sys/node/public.ts
alicewriteswrongs: yeah I think that might be why we left it like that, I don't remember exactly tbh

scripts/test/validate-build.ts
```diff
@@ -3,9 +3,10 @@ import { dirname, join, relative } from 'path';
 import { rollup } from 'rollup';
 import ts, { ModuleResolutionKind, ScriptTarget } from 'typescript';
 
-import { NODE_BUILTINS } from '../utils/constants';
-import { BuildOptions, getOptions } from '../utils/options';
-import { PackageData } from '../utils/write-pkg-json';
+import { createNodeLogger, createNodeSys } from '../../src/sys/node/index.js';
```
rwaskiewicz, line 6: I think this new import is causing CI to fail - would importing just the types fix it?
```suggestion
import type { createNodeLogger, createNodeSys } from '../../src/sys/node/index.js';
```

src/sys/node/public.ts
christian-bromann: I created two versions and deprecated one which will result in the following:
Screenshot 2024-02-21 at 7 49 19 AM

alicewriteswrongs (approved):

src/sys/node/public.ts
```diff
@@ -3,19 +3,35 @@ import type { CompilerSystem, Logger } from '@stencil/core/internal';
 /**
  * Creates a "logger", based off of NodeJS APIs, that will be used by the compiler and dev-server.
  * The NodeJS "process" object must be provided as a property in the first argument's object.
- * @param c an object containing a `Process` for Stencil to use
  * @returns a {@link Logger} object
  */
+export declare function createNodeLogger(): Logger;
+
+// TODO: Remove in Stencil v5, use `createNodeLogger` without parameters instead
```
rwaskiewicz, line 10: Can we tie this to a ticket in Jira? Ideally, this should be formatted in an easily grep-able format:
```suggestion
// TODO(STENCIL-123): Remove in Stencil v5, use `createNodeLogger` without parameters instead
```
Otherwise, we end up spending time trying to marry this TODO to something we may/may not have in Jira already

rwaskiewicz (approved):
One non-blocking ask

src/sys/node/public.ts
christian-bromann: created `STENCIL-1196`

christian-bromann:

PR #5376, opened 2024-02-16
chore(scripts): turn on 'strict' tsconfig option

alicewriteswrongs:
This turns on the `strict` option in the tsconfig for our build scripts. I then fixed the resulting errors by:

- making optional types required on `BuildOptions`
- reconfiguring the function that builds a `BuildOptions` object so that when the object is created all those fields are present
- fixed various other random errors here and there, like the `.code` property on minification results and a few things like that.

GitHub Issue Number: N/A

- [ ] Yes
- [x] No

## Testing

There are a few non-type-level changes but they're pretty contained, so I'm not terribly worried about the impact. I built and packed Stencil and installed it in Framework with no issues after these changes as well.

### DIFF
--- a/scripts/build.ts
+++ b/scripts/build.ts
@@ -87,7 +87,7 @@ export async function bundleBuild(opts: BuildOptions): Promise<void> {
             await bundle.write(output);
           }),
         );
-      } else {
+      } else if (rollupOption.output) {
         await bundle.write(rollupOption.output);
       }
     }),
--- a/scripts/bundles/plugins/content-types-plugin.ts
+++ b/scripts/bundles/plugins/content-types-plugin.ts
@@ -33,7 +33,7 @@ export async function createContentTypeData(opts: BuildOptions) {
   Object.keys(mimeDbJson).forEach((mimeType) => {
     const mimeTypeData = mimeDbJson[mimeType];
     if (Array.isArray(mimeTypeData.extensions)) {
-      mimeTypeData.extensions.forEach((ext) => {
+      mimeTypeData.extensions.forEach((ext: string) => {
         extData.push({
           ext,
           mimeType,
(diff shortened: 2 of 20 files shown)


scripts/build.ts
```diff
@@ -88,7 +88,7 @@ export async function bundleBuild(opts: BuildOptions): Promise<void> {
           }),
         );
       } else {
-        await bundle.write(rollupOption.output);
+        await bundle.write(rollupOption.output!);
       }
```
christian-bromann, line 92: How about:
```ts
      } else if (rollupOption.output) {
        await bundle.write(rollupOption.output);
      }
```

scripts/bundles/plugins/content-types-plugin.ts
```diff
@@ -33,7 +33,7 @@ export async function createContentTypeData(opts: BuildOptions) {
   Object.keys(mimeDbJson).forEach((mimeType) => {
     const mimeTypeData = mimeDbJson[mimeType];
     if (Array.isArray(mimeTypeData.extensions)) {
-      mimeTypeData.extensions.forEach((ext) => {
+      mimeTypeData.extensions.forEach((ext: any) => {
```
christian-bromann, line 36:
```suggestion
      mimeTypeData.extensions.forEach((ext: string) => {
```

scripts/release-prompts.ts
```diff
@@ -126,10 +126,15 @@ export async function promptRelease(opts: BuildOptions): Promise<NormalizedRelea
 
   try {
     const answers = await inquirer.prompt<ReleasePromptAnswers>(prompts);
+    const npmTag = determineAnsweredTagToUse(answers);
+
+    if (npmTag === null) {
+      throw 'no proper tag found!';
```
christian-bromann, line 132: Any reason we wouldn't throw an error?
```suggestion
      throw new Error('no proper tag found!');
```

alicewriteswrongs: oh yeah meant to do that! forgot to circle back

scripts/bundles/plugins/content-types-plugin.ts
alicewriteswrongs: yup, that works!

tanner-reits (approved):

PR #5385, opened 2024-02-20
chore(build): make esbuild standalone

alicewriteswrongs:
This removes the dependence of the esbuild-based build on the existing Rollup-based build. Previously we needed to run the Rollup-based build first so that certain files which could not yet be built with Esbuild would be present on disk, but since we can now build everything with Esbuild we don't need to take this step anymore.

We do, however, continue to need to run the TypeScript compiler in order to generate up-to-date typedef files.

Part of STENCIL-1016

This is blocked by #5276 and CI will not pass until that merges

GitHub Issue Number: N/A

- [ ] Yes
- [x] No

## Testing

If CI is passing this is a strong indication that we're in a good position to make this change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,8 +24,8 @@
   ],
   "scripts": {
     "build": "npm run tsc.scripts && npm run tsc.prod && npm run rollup.prod.ci",
-    "build.esbuild": "npm run clean && npm run build && node scripts/build/esbuild/build.js",
-    "build.esbuild.watch": "npm run clean && npm run build && node scripts/build/esbuild/build.js --watch",
+    "build.esbuild": "npm run clean && npm run tsc.scripts && npm run tsc.prod && node scripts/build/esbuild/build.js",
+    "build.esbuild.watch": "npm run build.esbuild -- --watch",
     "build.updateSelectorEngine": "node scripts/build/updateSelectorEngine.js",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -24,8 +24,8 @@
   ],
   "scripts": {
     "build": "npm run tsc.scripts && npm run tsc.prod && npm run rollup.prod.ci",
-    "build.esbuild": "npm run clean && npm run build && node scripts/build/esbuild/build.js",
-    "build.esbuild.watch": "npm run clean && npm run build && node scripts/build/esbuild/build.js --watch",
+    "build.esbuild": "npm run clean && npm run tsc.scripts && npm run tsc.prod && node scripts/build/esbuild/build.js",
+    "build.esbuild.watch": "npm run clean && npm run tsc.scripts && npm run tsc.prod && node scripts/build/esbuild/build.js --watch",
```
christian-bromann, line 28: I wonder if we can pass in the watch flag in some way like this:
```suggestion
    "build.esbuild.watch": "npm run build.esbuild -- --watch",
```

and have it attached to the last command.

christian-bromann (approved):
Minor suggestion but otherwise LGTM 👍

rwaskiewicz (approved):

tanner-reits (approved):

package.json
alicewriteswrongs: ohh seems like that does work, nice!

PR #5390, opened 2024-02-21
chore(scripts): remove an unused option from BuildOptions

alicewriteswrongs:
After we removed the old release scripts recently this option is no longer used anywhere, so we can delete it.

## What is the new behavior?

No behavior change, just removing an unused field from an interface.

- [ ] Yes
- [x] No

## Testing

This is just a type-level change, so if CI is passing I think we're good!


christian-bromann (approved):
👍

rwaskiewicz (approved):

PR #5398, opened 2024-02-22
chore(ci): fix puppeteer downloads

rwaskiewicz:

See the 'new bheavior' section

GitHub Issue Number: N/A

## What is the new behavior?

the url that pulls down chrome images seems to be intermittently failing
in github actions. when we go to install dependencies (`npm ci`), we're
greeted with an error:

```
npm ERR! code 1
npm ERR! path /home/runner/work/stencil/stencil/node_modules/puppeteer
npm ERR! command failed
npm ERR! command sh -c node install.mjs
npm ERR! Error: ERROR: Failed to set up chrome-headless-shell v121.0.6167.85! Set "PUPPETEER_SKIP_DOWNLOAD" env variable to skip download.
```

this appears to be a result of an intermittent failure (if it's
intermittent it to-be-determined). reports online suggest bumping to
puppeteer v22, which we cannot do at this moment due to it dropping
support for node 16. instead, set the base url environment variable to
tell puppeteer where to pull chrome from

## Documentation

N/A - CI fix

- [ ] Yes
- [x] No

## Testing

CI begins to pass again (everything but the tech debt burndown)

## Other information

We will need to force this PR through, as the tech debt burndown (which pulls from `main` for some of its steps) will not have the fix applied to it (because we haven't landed the code). That's the only section of CI that should fail though

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -4,3 +4,5 @@
 # If this value is changed, please ensure that it works both locally and in a continuous integration environment in
 # a repeatable manner (i.e. it can run many times, one after the other, without failing due to out-of-memory errors).
 node_options=--max-old-space-size=4096
+# TODO(STENCIL-1141): remove `PUPPETEER_DOWNLOAD_BASE_URL` once support for Node v16 is dropped
+PUPPETEER_DOWNLOAD_BASE_URL=https://storage.googleapis.com/chrome-for-testing-public


christian-bromann (approved):
👍

.github/workflows/actions/get-core-dependencies/action.yml
```diff
@@ -14,7 +14,7 @@ runs:
 
     - name: Install Dependencies
       run: |
-        npm ci \
+        PUPPETEER_DOWNLOAD_BASE_URL=https://storage.googleapis.com/chrome-for-testing-public npm ci \
```
rwaskiewicz, line 17: whoops, we don't actually need this

alicewriteswrongs (approved):

PR #5399, opened 2024-02-22
fix(build): address issue with dynamic import and vite

### LINKED ISSUE #5389: bug: latest v4.12.3 causes: "dynamic import cannot be analyzed by Vite." error with Storybook
### Prerequisites

### Stencil Version

4.12.3

### Current Behavior

When running in a Storybook environment Stencil v4.12.3 produces a build error.

> The above dynamic import cannot be analyzed by Vite.
See https://github.com/rollup/plugins/tree/master/packages/dynamic-import-vars#limitations for supported dynamic import formats. If this is intended to be left as-is, you can use the /* @vite-ignore */ comment inside the import() call to suppress this warning.

### Expected Behavior

The error should not appear.

### System Info

```shell
Package.json modules and versions: 


"@storybook/addon-essentials": "7.6.17",
"@storybook/addon-interactions": "7.6.17",
"@storybook/addon-links": "7.6.17",
"@storybook/blocks": "7.6.17",
"@storybook/html": "7.6.17",
"@storybook/html-vite": "7.6.17",
"@storybook/test": "7.6.17",
"@types/jest": "^29.5.6",
"@types/node": "^16.18.11",
"jest": "^29.7.0",
"jest-cli": "^29.7.0",
"puppeteer": "^21.9.0",
"react": "18.2.0",
"react-dom": "18.2.0",
"storybook": "7.6.17"
```

### Steps to Reproduce

Create a new StencilJS project using `npm init stencil`.

Add storybook using `npx storybook@latest init`.
Choose **html** and **Vite** as the project options.

Create a `preview-head.html` file and add the following lines:
```html
<link rel="stylesheet" href="dist/test/test.css" />
<script type="module" src="dist/test/test.esm.js"></script>
```
_test_ being your project name.

Add a `my-component.stories.ts` file to your component directory with the follow straightforward storybook config:
```
export default {
  title: 'Example/Button',
};

export const Default = {
  render: `<my-component first="Stencil" last="'Don't call me a framework' JS"></my-component>`,
};
```
 
`npm run build` to compile the Stencil code.

`npm run storybook` to run Storybook and the error will occur.

### Code Reproduction URL

https://github.com/gregorypratt/stencil-issue

### Additional Information

_No response_

alicewriteswrongs:
When we added a script for building the modules in `internal/` with Esbuild in #5276 we needed to make a change to the function that Stencil uses at runtime to lazy-load components (in
`src/client/client-load-module.ts`). Prior to #5276 we had a dynamic import statement which looked like so:

```ts
import(
  `./${bundleId}.entry.js${BUILD.hotModuleReplacement && hmrVersionId ? '?s-hmr=' + hmrVersionId : ''}`
)
```

This constructs a filepath to the module for a given Stencil component, accounting for HMR versioning, and then imports the module. All well and good, but unfortunately this dynamic import does not play well with Esbuild. As described
[here](https://esbuild.github.io/api/#non-analyzable-imports) when Esbuild is in 'bundle' mode and it encounters an `import()` _and_ the imported path or identifier looks "analyzable" it will attempt to resolve the corresponding file and incorporate it into the bundle.

This is not always what you want! In particular, in our situation the leading `"./"` in the template literal we had in `client-load-module.ts` caused Esbuild to consider the `import()` an "analyzable" import and it then tried to resolve and bundle the import instead of just leaving the dynamic import in the code (as Rollup does in this case).

This created an issue because at _compile time_ (i.e. when Stencil itself is built) this import does not resolve to anything, so Esbuild would essentially transform that line into an empty import. This caused runtime issues because the side-effect of the dynamic import was no longer happening, so the modules containing Stencil component classes and so on were not longer being loaded in.

To get this working for #5276 we pulled out the `"./"` string as a separate variable, changing the template literal so it looks something like this:

```ts
const MODULE_IMPORT_PREFIX = './';

import(
  `${MODULE_IMPORT_PREFIX}${bundleId}.entry.js${BUILD.hotModuleReplacement && hmrVersionId ? '?s-hmr=' + hmrVersionId : ''}`
)
```

This causes Esbuild to conclude that the import is "non-analyzable", which addresses the issue and causes both Rollup and Esbuild to emit equivalent code for this snippet, where both retain the dynamic import, allowing for the runtime module resolution that we want here.

_However_, this broke the ability to use Stencil with Vite, which will complain about non-analyzable imports if it sees a dynamic import which does _not_ begin with `"./"`. See #5389 for details.

So essentially we have a situation where the behavior of Rollup, Esbuild, and Vite is incompatible. The solution is to figure out a way for both the Esbuild and Rollup builds to emit code in this case which retains the dynamic import _and_ retains the leading `"./"` in the template literal.

This is accomplished by retaining the `${MODULE_IMPORT_PREFIX}` in the template literal, so that Esbuild does not attempt to analyze and bundle the import, and adding plugins to both the Rollup and Esbuild bundles to transform the emitted code before it is written to disk.

fixes #5389
STENCIL-1181

GitHub Issue Number: N/A

- [ ] Yes
- [ ] No

## Testing

To test this the reproduction case linked in #5389 should be tried with both the ESbuild and Rollup based builds. It should work fine with either one.

The reproduction is here: https://github.com/gregorypratt/stencil-issue

After cloning locally, build and pack with Rollup and test to see that the issue doesn't repro anymore, and then do the same with Esbuild.

### DIFF
--- a/scripts/esbuild/internal-platform-client.ts
+++ b/scripts/esbuild/internal-platform-client.ts
@@ -1,4 +1,4 @@
-import type { BuildOptions as ESBuildOptions } from 'esbuild';
+import type { BuildOptions as ESBuildOptions, Plugin } from 'esbuild';
 import { replace } from 'esbuild-plugin-replace';
 import fs from 'fs-extra';
 import glob from 'glob';
@@ -45,6 +45,9 @@ export async function getInternalClientBundle(opts: BuildOptions): Promise<ESBui
     ...getBaseEsbuildOptions(),
     entryPoints: [join(inputClientDir, 'index.ts')],
     format: 'esm',
+    // we do 'write: false' here because we write the build to disk in our
+    // `findAndReplaceLoadModule` plugin below
+    write: false,
     outfile: join(outputInternalClientDir, 'index.js'),
     platform: 'node',
     external: clientExternal,
@@ -59,6 +62,7 @@ export async function getInternalClientBundle(opts: BuildOptions): Promise<ESBui
       // we want to get the esm, not the cjs, since we're creating an esm
       // bundle here
       externalAlias('@stencil/core/mock-doc', '../../mock-doc/index.js'),
+      findAndReplaceLoadModule(),
     ],
   };
 
@@ -94,6 +98,31 @@ export async function getInternalClientBundle(opts: BuildOptions): Promise<ESBui
   };
 
   return [internalClientBundle, internalClientPatchBrowserBundle];
+}
+
+/**
+ * We need to manually find-and-replace a bit of code in
+ * `client-load-module.ts` in order to prevent Esbuild from analyzing /
+ * transforming the input by ensuring it does not start with `"./"`. However
+ * some _other_ bundlers will _not_ work with such an import if it _lacks_ a
+ * leading `"./"`, so we thus we have to do a little dance where we manually
+ * replace it here after it's been run through Esbuild.
+ *
+ * @returns an Esbuild plugin
+ */
+export function findAndReplaceLoadModule(): Plugin {
+  return {
+    name: 'findAndReplaceLoadModule',
+    setup(build) {
+      build.onEnd(async (result) => {
+        for (const file of result.outputFiles!) {
+          const { path, text } = file;
+
+          await fs.writeFile(path, text.replace(/\${MODULE_IMPORT_PREFIX}/, './'));
+        }
+      });
+    },
+  };
 }
 
 async function copyPolyfills(opts: BuildOptions, outputInternalClientPolyfillsDir: string) {
--- a/scripts/bundles/internal-platform-client.ts
+++ b/scripts/bundles/internal-platform-client.ts
@@ -8,7 +8,7 @@ import type { BuildOptions } from '../utils/options';
 import { writePkgJson } from '../utils/write-pkg-json';
 import { aliasPlugin } from './plugins/alias-plugin';
 import { reorderCoreStatementsPlugin } from './plugins/reorder-statements';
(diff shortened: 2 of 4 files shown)


scripts/esbuild/internal-platform-client.ts
```diff
@@ -96,6 +100,33 @@ export async function getInternalClientBundle(opts: BuildOptions): Promise<ESBui
...
+ * replace it here after it's been run through Esbuild.
+ *
+ * @returns an Esbuild plugin
+ */
+export function findAndReplaceLoadModule(): Plugin {
+  return {
+    name: 'externalAliases',
+    setup(build) {
+      console.log('ARGASDGASDF');
+
```
christian-bromann, line 118:
```suggestion
```

christian-bromann (approved):
Is there a way to test this?

LGTM otherwise 👍

scripts/utils/options.ts
```diff
@@ -147,6 +147,9 @@ export function createReplaceData(opts: BuildOptions): Record<string, any> {
     '__VERSION:TERSER__': opts.terserVersion,
     '__VERSION:TYPESCRIPT__': opts.typescriptVersion,
 
+    // this is a hack
```
christian-bromann, line 150: Maybe worth extending on the docs here and mentioned why it is needed or pointing to a comment made in other files.

rwaskiewicz (commented):
I agree with @christian-bromann's comments here, nothing blocking on my end.

scripts/utils/options.ts
alicewriteswrongs: that was just an oversight actually, we don't need to do anything in the `createReplaceData` function

PR #5145, opened 2023-12-08
fix(testing): make Puppeteer an optional dependency

christian-bromann:
## What is the current behavior?
Currently it seems required to have Puppeteer installed to build a Stencil component. It seems that the compiler imports parts of the testing module to compile a project which causes "/testing/puppeteer/puppeteer-declarations.d.ts" to be loaded.

STENCIL-881
fixes #4526

## What is the new behavior?

To solve the type dependency the easiest is to add a Rollup plugin that simply adds a "@ts-ignore" before the import. If Puppeteer is not installed, it will not throw any errors, if it is installed, all types are propagated correctly.

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Testing

To verify this change:

- create a new stencil project `npm init stencil`
- install all dependencies via `npm install`
- build project `npm run build` and see build succeeding
- now remove Puppeteer as dependency `rm -r ./node_modules/puppeteer`
- build project `npm run build` and see build failing
- now install this branch to the project
- remove Puppeteer as dependency `rm -r ./node_modules/puppeteer`
- build project `npm run build` and see build succeeding

demo-types

## Other information

While this is a rather undesired thing to do, in hopes we eventually get rid of Jest and Puppeteer as required test harness, this will be code that will be removed eventually anyway.

Issue #4526, opened 2023-06-29
bug: Stencil should not require puppeteer dependency in new projects


sean-perkins:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

v4

### Current Behavior

When creating a new project and removing the `puppeteer` dev dependency, Stencil will fail to start with the following exception:

```ts
[ ERROR ]  TypeScript: node_modules/@stencil/core/testing/jest/jest-environment.d.ts:9:44
           Cannot find module 'puppeteer' or its corresponding type declarations.

      L8:  setup(): Promise<void>;
      L9:  newPuppeteerPage(): Promise<import("puppeteer").Page>;
     L10:  closeOpenPages(): Promise<void>;

[ ERROR ]  TypeScript: node_modules/@stencil/core/testing/puppeteer/puppeteer-declarations.d.ts:3:120
           Cannot find module 'puppeteer' or its corresponding type declarations.

      L2:  import type { EventInitDict, EventSpy, ScreenshotDiff, ScreenshotOptions } from '@stencil/core/internal';
      L3:  ions, HTTPResponse, Page, ScreenshotOptions as PuppeteerScreenshotOptions, WaitForOptions } from 'puppeteer';
      L4:  /**

[27:32.9]  build failed, watching for changes... in 4.68 s
```

### Expected Behavior

Stencil should not require puppeteer as a dependency. Stencil projects may choose to use Playwright to test their component libraries and currently will require the dev dependency to satisfy Stencil. 

This worked previously in v2. Unsure if it regressed in v3 or v4. 

### System Info

```shell
System: node 16.15.1
    Platform: darwin (22.2.0)
   CPU Model: Apple M1 (8 cpus)
    Compiler: /Users/sean/Documents/ionic/issues/stencil-without-puppeteer/node_modules/@stencil/core/compiler/stencil.js
       Build: 1687953951
     Stencil: 4.0.1 ⛹
  TypeScript: 5.0.4
      Rollup: 2.42.3
      Parse5: 7.1.2
      Sizzle: 2.42.3
      Terser: 5.18.1
```

### Steps to Reproduce

[...]

PR #5415, opened 2024-02-26
chore(): port puppeteer.d.ts patch fix to esbuild

alicewriteswrongs:
This ports the fix which we landed in #5145 from the Rollup build to Esbuild. This will ensure that if Stencil is built with Esbuild and then installed in a puppeteer-free context it will still work!

We have a previous issue (#4526) which reported this issue with the Rollup build (well, with a published version of Stencil) but the fix for that issue (#5145) wasn't ported over to Esbuild SO if you build Stencil with Esbuild and then, for instance, install it in Framework you can reproduce the same problem. No good!

## What is the new behavior?

This pulls out the fix from the Rollup build as a separate function and then adds it to the Esbuild script for building `testing/`.

- [ ] Yes
- [x] No

## Testing

First make sure you can repro the issue on `main` with Framework.

On `main`:

- do
  ```sh
  npm run build.esbuild && npm pack
  ```
- install the resulting tarball in Framework (well in `core/`) and confirm that you get a type error when you
  ```sh
  npm run build
  ```
  
Then check out this branch and re-run the above steps. The issue should now be resolved! Additionally, you do `npm run build` _instead_ of `npm run build.esbuild` that should still work as well.

### DIFF
--- a/scripts/bundles/testing.ts
+++ b/scripts/bundles/testing.ts
@@ -106,7 +106,7 @@ export async function testing(opts: BuildOptions) {
         preferConst: true,
       }),
       prettyMinifyPlugin(opts, getBanner(opts, `Stencil Testing`, true)),
-      ignorePuppteerDependency(opts),
+      ignorePuppeteerDependency(opts),
     ],
     treeshake: {
       moduleSideEffects: false,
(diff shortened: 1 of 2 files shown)


christian-bromann (approved):

rwaskiewicz:
Tested locally with the instructions in the PR summary, works as expected 👍 

I also built the following combinations of Stencil:
- `main`, w rollup
- this branch, w esbuild
- `main`, w rollup
- this branch, w esbuild

Comparing the rollup-based builds yielded no noteworthy changes.

Likewise for the esbuild-based builds, except for the one we obviously want:
Screenshot 2024-02-26 at 5 22 24 PM

rwaskiewicz (approved):

tanner-reits (approved):

PR #5419, opened 2024-02-28
fix(compiler): fix generated import statement

christian-bromann:
If you have a plain component like this:

```ts
@Component({
  tag: 'my-component',
  // styleUrl: 'my-component.css',
  shadow: true,
})
export class MyComponent {
  /**
   * The first name
   */
  first: string;
}
```

and try to export it via:

```ts
{
  type: 'dist-custom-elements',
  customElementsExportBehavior: 'single-export-module'
},
```

the compiler will fail with:

```
[ ERROR ]  Rollup: Parse Error
           Unexpected token (Note that you need plugins to import files that are not
           JavaScript) 2: export { getAssetPath, setAssetPath, setNonce,
           setPlatformOptions } from '@stencil/core/internal/client'; 3: export * from
           '@user-index-entrypoint'; 4: export { {MyComponent } from 'MyComponent'; ^ 5:
           6: globalScripts();
```

The change was originally introduced here: https://github.com/ionic-team/stencil/pull/3368/files#diff-df2741ac09fef4bbb750d864b2b96bdc4f8e649f347c7cc7def8edc977e1004eR198 and seems to be not intentional.

## What is the new behavior?
Removed the additional bracket.

## Documentation

n/a

- [ ] Yes
- [x] No

## Testing

I added a unit test for this particular use case.

## Other information

n/a

### DIFF
--- a/src/compiler/output-targets/dist-custom-elements/test/dist-custom-elements.spec.ts
+++ b/src/compiler/output-targets/dist-custom-elements/test/dist-custom-elements.spec.ts
@@ -3,22 +3,23 @@ import { addCustomElementInputs } from '../index';
 describe('dist-custom-elements', () => {
   it('should add custom element inputs', () => {
     const buildCtx: any = {
-      components: [{
-        isPlain: true,
-        sourceFilePath: './foo/bar.tsx',
-        tagName: 'my-tag'
-      }]
+      components: [
+        {
+          isPlain: true,
+          sourceFilePath: './foo/bar.tsx',
+          tagName: 'my-tag',
+        },
+      ],
     };
     const bundleOpts: any = {
       inputs: {},
-      loader: {}
-    }
+      loader: {},
+    };
     const outputTarget: any = {
       type: 'dist-custom-elements',
-      customElementsExportBehavior: 'single-export-module'
+      customElementsExportBehavior: 'single-export-module',
     };
     addCustomElementInputs(buildCtx, bundleOpts, outputTarget);
-    expect(bundleOpts.loader['\x00core']).toContain(
-      `export { MyTag } from '\x00MyTag';\n`)
+    expect(bundleOpts.loader['\x00core']).toContain(`export { MyTag } from '\x00MyTag';\n`);
   });
 });
--- a/src/compiler/output-targets/dist-custom-elements/test/dist-custom-elements.spec.ts
+++ b/src/compiler/output-targets/dist-custom-elements/test/dist-custom-elements.spec.ts
@@ -1,21 +1,22 @@
+import type * as d from '@stencil/core/declarations';
+import { mockBuildCtx } from '@stencil/core/testing';
+
+import { BundleOptions } from '../../../bundle/bundle-interface';
+import { stubComponentCompilerMeta } from '../../../types/tests/ComponentCompilerMeta.stub';
 import { addCustomElementInputs } from '../index';
 
 describe('dist-custom-elements', () => {
   it('should add custom element inputs', () => {
-    const buildCtx: any = {
-      components: [
-        {
-          isPlain: true,
-          sourceFilePath: './foo/bar.tsx',
-          tagName: 'my-tag',
-        },
-      ],
-    };
-    const bundleOpts: any = {
+    const cmpMeta = stubComponentCompilerMeta({ isPlain: true, sourceFilePath: './foo/bar.tsx', tagName: 'my-tag' });
+    const buildCtx = mockBuildCtx();
+    buildCtx.components = [cmpMeta];
+    const bundleOpts: BundleOptions = {
+      id: 'customElements',
+      platform: 'client',
       inputs: {},
       loader: {},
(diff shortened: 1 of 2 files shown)


src/compiler/output-targets/dist-custom-elements/test/dist-custom-elements.spec.ts
```diff
@@ -0,0 +1,25 @@
+import { addCustomElementInputs } from '../index';
```
rwaskiewicz, line 1: I'd like to add a bit more type safety to these unit tests if possible. Stencil has a lot of interfaces/types that have _several_ properties on them, making them a bit cumbersome to build out for each and every test suite (or even on a test-by-test basis).

We have a handful of stubs/mock helper functions to provide us with a "known good" baseline for some of these types, which we can override at construction time (most of the time). The API of these mocks/stubs is a little inconsistent due to the project changing hands over the years, but I _think_ we have a few that could be of use here. Take a look at the included patch, and let me know what you think - I'll add comments here in the PR as well, but they'll be a bit fragmented (hence including that patch).

[addtl-type-safety.patch](https://github.com/ionic-team/stencil/files/14434817/addtl-type-safety.patch)

src/compiler/output-targets/dist-custom-elements/test/dist-custom-elements.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+  it('should add custom element inputs', () => {
+    const buildCtx: any = {
+      components: [
+        {
+          isPlain: true,
+          sourceFilePath: './foo/bar.tsx',
+          tagName: 'my-tag',
+        },
+      ],
+    };
```
rwaskiewicz, line 13: Note: This is included in the patch file in my first comment on this PR


To build out a mock `BuildContext`, we have `mockBuildCtx` ([GH Link](https://github.com/ionic-team/stencil/blob/6faf746990330da4369e0d73725b0fc2becebb33/src/testing/mocks.ts#L199)). This builds a build context with a `ValidatedConfig` and `CompilerCtx` for us, and defaults to known good versions if neither are provided.

At the moment, we can't override values like `components` on it, but we can directly assign that property on the `buildCtx` object, which is the only property used by `addCustomElementIntputs`.

To build out an array of one of `ComponentCompilerMeta`s, we can use `stubComponentCompilerMeta` ([GH Link](https://github.com/ionic-team/stencil/blob/5976c9b6a6e7b49d470390021b9c31e4d3cbbf4b/src/compiler/types/tests/ComponentCompilerMeta.stub.ts#L10)). Unlike `mockBuildCtx`, we can override just about anything on this object.

All in all, I think we could accomplish the same thing with a bit more type safety with:
```ts
    const cmpMeta = stubComponentCompilerMeta({ isPlain: true, sourceFilePath: './foo/bar.tsx', tagName: 'my-tag' });
    const buildCtx = mockBuildCtx();
    buildCtx.components = [cmpMeta];
```

src/compiler/output-targets/dist-custom-elements/test/dist-custom-elements.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+    const bundleOpts: any = {
```
rwaskiewicz, line 14: Note: This is included in the patch file in my first comment on this PR

I think this could be made a `BundleOptions` object by adding an `id` and `platform` property here:
```ts
    const bundleOpts: BundleOptions = {
      id: 'customElements',
      platform: 'client',
      inputs: {},
      loader: {},
    };
```

src/compiler/output-targets/dist-custom-elements/test/dist-custom-elements.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+      inputs: {},
+      loader: {},
+    };
+    const outputTarget: any = {
```
rwaskiewicz, line 18: Note: This is included in the patch file in my first comment on this PR

I think we have enough here to make this a `OutputTargetDistCustomElements`, we just need to import the type
```ts
import type * as d from '@stencil/core/declarations';
```
Then:
```ts
    const outputTarget: d.OutputTargetDistCustomElements = {
```

rwaskiewicz (changes requested):
👍 for more tests! I had some asks around type safety, take a look at the included patch and LMK what you think

src/compiler/output-targets/dist-custom-elements/test/dist-custom-elements.spec.ts
christian-bromann: This is amazing, thank you!

christian-bromann:
@rwaskiewicz thanks for the patch, good to know these primitives exist, that helps a lot. I added another test for the "non-plain" use case.

rwaskiewicz (approved):

tanner-reits (approved):
LGTM

PR #5420, opened 2024-02-28
chore(scripts): swap from rollup to esbuild for main build

alicewriteswrongs:
This swaps over from Rollup to Esbuild for the main `npm run build` script in `package.json`. The existing Rollup based script is preserved as `build.rollup`.

This change causes our main CI run (`.github/workflows/main.yml`) to now use the Esbuild-based build, so the existing, supplemental CI run which previously was based on the Esbuild-based build pipeline now uses the 'legacy' (or nearly so) Rollup-based version. This requires a bit of renaming and so on.

STENCIL-1016

GitHub Issue Number: N/A

- [ ] Yes
- [x] No

## Testing

To test this out in the context of our GH-based releases I ran the dev and nightly releases against this branch:

- dev release: https://github.com/ionic-team/stencil/actions/runs/8087583103
- nightly release: https://github.com/ionic-team/stencil/actions/runs/8087594210

Both builds finished without error. The versions look _crazily_ similar:

- dev: `4.12.4-dev.1709157324.4664b01`
- nightly: `4.12.4-dev.1709157393.4664b01`

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,9 +23,9 @@
     "testing/"
   ],
   "scripts": {
-    "build": "npm run tsc.scripts && npm run tsc.prod && npm run rollup.prod.ci",
-    "build.esbuild": "npm run clean && npm run tsc.scripts && npm run tsc.prod && node scripts/build/esbuild/build.js --prod --ci",
-    "build.esbuild.watch": "npm run build.esbuild -- --watch",
+    "build.rollup": "npm run tsc.scripts && npm run tsc.prod && npm run rollup.prod.ci",
+    "build": "npm run clean && npm run tsc.scripts && npm run tsc.prod && node scripts/build/esbuild/build.js --prod --ci",
+    "build.watch": "npm run build -- --watch",
     "build.updateSelectorEngine": "node scripts/build/updateSelectorEngine.js",
     "clean": "rm -rf build/ cli/ compiler/ dev-server/ internal/ mock-doc/ sys/ testing/ && npm run clean-scripts",
     "clean-scripts": "rm -rf scripts/build",
--- a/.github/workflows/build-rollup.yml
+++ b/.github/workflows/build-rollup.yml
@@ -1,4 +1,4 @@
-name: Build Stencil with Esbuild
+name: Build Stencil with Rollup
 
 on:
   workflow_call:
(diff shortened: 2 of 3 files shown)


alicewriteswrongs:
One question I have with this PR for the team: does it make sense for us to keep the nightly and dev releases on Rollup, or to also swap them over to Esbuild?

Right now they are using Esbuild since they use the `build.yml` workflow, which previously called Rollup but now calls Esbuild.

rwaskiewicz (changes requested):
Changes LGTM - I think everything here _should_ work; Prod releases still appear to use rollup (as we intended) and this shouldn't require any changes to our status checks in CI, nice!

Only ask here is that we manually kick off a dev and nightly build based on this branch just to kick the tires/make sure everything works as expected. Can you run both workflows and post the results/runs in the PR summary?

alicewriteswrongs:
yeah good idea, PR summary updated with links to the gh actions runs for those

rwaskiewicz (approved):

package.json
```diff
@@ -23,9 +23,9 @@
     "testing/"
   ],
   "scripts": {
-    "build": "npm run tsc.scripts && npm run tsc.prod && npm run rollup.prod.ci",
-    "build.esbuild": "npm run clean && npm run tsc.scripts && npm run tsc.prod && node scripts/build/esbuild/build.js --prod --ci",
-    "build.esbuild.watch": "npm run build.esbuild -- --watch",
+    "build.rollup": "npm run tsc.scripts && npm run tsc.prod && npm run rollup.prod.ci",
```
tanner-reits, line 26: Do we need `npm run clean` at the start of this?

tanner-reits (approved):
One question, non-blocking if not needed

package.json
alicewriteswrongs: I don't think so, since the Rollup build clears out the directories here:

https://github.com/ionic-team/stencil/blob/1eec9a82a4d623d46228e84798adbf2f88d7f3e8/scripts/build.ts#L44-L52

alicewriteswrongs: it does not - went ahead and just added it because why not

although I'll point out this is our current build script, so we're currently not clearing out `scripts/` stuff between builds

although 🤔 maybe that's intentional so that the caching business that happens in there sticks around?

rwaskiewicz: Yeah, exactly - I typically run it (`clean`) whenever I switch branches locally (`npm run clean && npm ci && npm run build`), but it does slow down some overall build time

alicewriteswrongs: yeah, ok I'm going to go ahead and leave the command as it is on `main` for now

PR #5436, opened 2024-03-05
feat(compiler): update glob and minimatch dependencies

christian-bromann:
We are currently using old versions of `glob` and `minimatch`.

## What is the new behavior?
Update them to latest.

## Documentation

n/a

- [ ] Yes
- [x] No

The latest versions of these package require Node.js v16 and up which is what we require for Stencil v4 as well.

## Testing

I would assume as long as the build passes we are good to go.

## Other information

n/a


alicewriteswrongs (approved):
tried out the build locally in a sample project and in framework and everything looks good!

rwaskiewicz:
Can we update renovate's config to stop ignoring these deps please?

tanner-reits (changes requested):
Just blocking for Ryan's comment about Renovate

christian-bromann:
Updated renovate file.

PR #5437, opened 2024-03-05
fix(runtime): remove `forceUpdate` in `appendChild` patch

tanner-reits:

Stencil's patch for `appendChild` for slot emulation includes a call to `forceUpdate()` that is no longer needed and should not have been included in the first place (see new behavior section for more info).

GitHub Issue Number: N/A

## What is the new behavior?

This commit removes a call to `forceUpdate` that was added to Stencil's patch for `appendChild` for slot emulation. The call was originally added to fix a failing test in Framework that we've now identified was the result of the patch's DOM manipulation in tandem with a mutation observer in framework.

Instead, the mutation observer in framework was modified to observe changes in the element subtree, removing the need for this call.

STENCIL-1192

## Documentation

N/A

- [ ] Yes
- [x] No

## Testing

Automated unit and e2e tests continue to pass.

Tested this in a use case in Framework (see [STENCIL-1192](https://outsystemsrd.atlassian.net/browse/STENCIL-1192))

### DIFF
--- a/src/runtime/dom-extras.ts
+++ b/src/runtime/dom-extras.ts
@@ -1,5 +1,5 @@
 import { BUILD } from '@app-data';
-import { forceUpdate, getHostRef, plt, supportsShadow } from '@platform';
+import { getHostRef, plt, supportsShadow } from '@platform';
 import { NODE_TYPES } from '@stencil/core/mock-doc';
 import { CMP_FLAGS, HOST_FLAGS } from '@utils';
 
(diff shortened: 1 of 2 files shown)


christian-bromann (approved):
I recon that we added the `forceUpdate` without putting it behind the `experimentialSlotFixes` flag and removing it could have a an impact on users that for some reasons rely on it. I don't think anyone does but there is a chance.

LGTM 👍

alicewriteswrongs (approved):

christian-bromann:

PR #5448, opened 2024-03-08
chore(jsdoc): fix some JSDoc `@link` tags

alicewriteswrongs:
This fixes some `@link` tags in JSDoc which weren't resolving correctly. Since the Typedoc tool reports on `@link` tags which it can't resolve we can use that to find them.

Some of these links don't actually resolve! It's a bit hard to see this, but one way that I can test if the `{@link Foo}` is correctly resolving the symbol `Foo` is to ask the LSP about it in my editor. If the name can be resolved then I'll see some info about it, otherwise I won't. This shows the before and after of this in `src/cli/run.ts`:

https://github.com/ionic-team/stencil/assets/6207644/5b416c3b-af86-4c17-85d2-08407fbc8271

We have `import type * as d from '../declarations';` at the top of the file, so `d.CompilerSystem` can be resolved but `CompilerSystem` cannot. 

## What is the new behavior?

Several of these are fixed! A few where it would be required to add a new import to a module in order to get something to resolve I just turned into comments instead.

- [ ] Yes
- [x] No

## Testing

Try to make sure that these actually resolve, possibly using the method outlined above with the LSP in an editor. Or just try to reason about it.

### DIFF
--- a/src/cli/run.ts
+++ b/src/cli/run.ts
@@ -121,7 +121,7 @@ export const run = async (init: d.CliInitOptions) => {
  * @param coreCompiler an instance of a minimal, bootstrap compiler for running the specified task
  * @param config a configuration for the Stencil project to apply to the task run
  * @param task the task to run
- * @param sys the {@link CompilerSystem} for interacting with the operating system
+ * @param sys the {@link d.CompilerSystem} for interacting with the operating system
  * @public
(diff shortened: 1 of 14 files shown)


christian-bromann (approved):
LGTM 👍

PR #5456, opened 2024-03-11
chore(test): add initial WebdriverIO infrastructure

christian-bromann:
The team has decided to press forward with WebdriverIO as the replacement for Karma/Browserstack

In order to make it such that the team can swarm on replacing these ~150 tests, we need some initial infrastructure in place.

## What is the new behavior?
This patch adds initial infrastructure to run WebdriverIO tests locally and in CI.

## Documentation

Adding some docs on how to migrate tests.

- [ ] Yes
- [x] No

## Testing

I assume as long as the pipeline passes, we should be good.

## Other information

STENCIL-1190


.github/workflows/test-wdio.yml
```diff
@@ -0,0 +1,51 @@
+name: BrowserStack Tests
+
+on:
```
rwaskiewicz, line 3: Does this run ATM? I'm not seeing any logs from GH Actions

test/wdio/package.json
```diff
@@ -0,0 +1,28 @@
...
+  "version": "0.0.0",
+  "scripts": {
+    "build": "node ../../bin/stencil build --debug --es5",
+    "test": "run-s build wdio",
+    "wdio": "wdio run ./wdio.conf.ts"
+  },
+  "devDependencies": {
+    "@stencil/core": "file:../..",
+    "@stencil/sass": "^3.0.10",
+    "@types/node": "^20.11.25",
```
rwaskiewicz, line 13: Can this be tied to v16 to match the value we have below for `volta.node`?

rwaskiewicz (changes requested):
I took a first pass at this PR - I looked at most of the infra here and had some questions about how it's running today/how we might shrink this diff a bit. I haven't looked at the tests themselves in a ton of detail yet - I'll take a second pass (including the tests) once we resolve the first round of questions - LMK if I can add any more detail!

.github/workflows/test-wdio.yml
```diff
@@ -0,0 +1,51 @@
...
+  merge_group:
+  pull_request_target:
+    branches:
+      - 'main'
+  push:
+    branches:
+      - 'main'
+
+permissions: read-all
```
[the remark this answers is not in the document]
christian-bromann: We can remove it for now to see if it has any impact.

test/karma/stencil.config.ts
```diff
@@ -24,7 +24,6 @@ export const config: Config = {
     },
   ],
   globalScript: 'test-app/global.ts',
-  globalStyle: 'test-app/style-plugin/global-sass-entry.scss',
```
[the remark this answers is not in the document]
christian-bromann: Correct, this lives in https://github.com/ionic-team/stencil/blob/6d17f2689df4028da4f177086feab78525ed7d51/test/wdio/stencil.config.ts#L15 now

test/karma/package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "karma",
-  "lockfileVersion": 2,
```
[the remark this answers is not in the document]
christian-bromann: good catch, will revert

test/karma/test-app/attribute-basic/karma.spec.ts
```diff
@@ -1,27 +0,0 @@
-import { setupDomTests, waitForChanges } from '../util';
```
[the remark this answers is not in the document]
christian-bromann: I tried that, I actually re-created this PR fresh from `main` and only used `g mv` to move files and I ended up still getting these files flagged as `removed`:

```
On branch cb/karma-webdriverio-replace-with-history
Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
        new file:   .github/workflows/test-wdio.yml
        modified:   package.json
        modified:   test/karma/stencil.config.ts
        deleted:    test/karma/test-app/attribute-basic/index.html
        deleted:    test/karma/test-app/attribute-basic/karma.spec.ts
        deleted:    test/karma/test-app/attribute-boolean/index.html
        deleted:    test/karma/test-app/attribute-boolean/karma.spec.ts
        deleted:    test/karma/test-app/attribute-complex/index.html
        deleted:    test/karma/test-app/attribute-host/index.html
        deleted:    test/karma/test-app/attribute-host/karma.spec.ts
        deleted:    test/karma/test-app/attribute-html/index.html
        deleted:    test/karma/test-app/attribute-html/karma.spec.ts
        deleted:    test/karma/test-app/build-data/index.html
        deleted:    test/karma/test-app/build-data/karma.spec.ts
...
```

.github/workflows/test-wdio.yml
christian-bromann: The workflow is running and passing now: https://github.com/ionic-team/stencil/actions/runs/8252971259/job/22573834057?pr=5456

.github/workflows/test-wdio.yml
```diff
@@ -0,0 +1,45 @@
...
+
+      - name: Download Build Archive
+        uses: ./.github/workflows/actions/download-archive
+        with:
+          name: stencil-core
+          path: .
+          filename: stencil-core-build.zip
+
+      - name: Run WebdriverIO Component Tests
+        run: npm run test.wdio
```
alicewriteswrongs, line 42: I think it's sort of six of one half a dozen of another but would it make sense for us to kick off the wdio tests from within the `main` workflow, like we do for the unit tests and whatnot? instead of having a separate workflow here that runs independently

alicewriteswrongs: also is the plan to circle back later to do some sort of matrix on browsers and browser versions?

alicewriteswrongs (approved):
tried this out locally and it works great! looks like CI is working too

I just had one question, which I don't think is blocking, about how it makes sense for us to set up the CI workflow stuff

excited to get to converting the rest of these tests!

test/karma/test-app/attribute-basic/karma.spec.ts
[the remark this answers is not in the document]
rwaskiewicz: Ugh sorry about that. We probably don't hit git's ratio of "sameness" when we mv/rename the file & update its contents, which would explain that 😞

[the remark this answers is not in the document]
christian-bromann: I updated the branch and pushed commits that split the change set into multiple ones with file location and content changes separated. It seems that GitHub still marks them as "deleted" and "new" but git should contain the history now.

.github/workflows/test-wdio.yml
christian-bromann: > to kick off the wdio tests from within the main workflow, like we do for the unit tests and whatnot?

The test run is cheap and can be run as part of the unit test.

> also is the plan to circle back later to do some sort of matrix on browsers and browser versions?

Just pushed a fix to include Firefox and Safari when running in CI

christian-bromann: I recognize that there are some browser differences with some assertions. I had to change some assertion values based on whether we run in Safari or not. This could be either an actual browser issue or something with the driver underneath. I would like to investigate this separately. This could become ultimately a burden to maintain at some point. I can bring this up at the next Jam session.

christian-bromann:
I had to bring back `test/karma/test-app/components.d.ts` as it is required for the Karma test application to build. I don't see that this could pose a problem to us given the fact that we mostly just remove lines from that file.

test/wdio/computed-properties-watch-decorator/cmp.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+    render({
+      template: () => (
+        <>
+          <computed-properties-watch-decorator></computed-properties-watch-decorator>
+          <button type="button">Trigger watch callbacks</button>
+        </>
+      ),
+    });
+    document.querySelector('button').addEventListener('click', () => {
+      const cmp = document.querySelector('computed-properties-watch-decorator');
```
rwaskiewicz, line 15: For my own understanding, is there any difference between this and say,
```ts
    const el = $('computed-properties-watch-decorator');
```
Would using `$` be unnecessarily `async` and add overhead to the test?

test/wdio/computed-properties-watch-decorator/cmp.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+      cmp.setAttribute('first', 'Bob');
+      cmp.setAttribute('last', 'Builder');
+    });
+  });
+
+  it('triggers the watch callback when the associated prop changes', async () => {
+    const el = document.querySelector('computed-properties-watch-decorator');
+    await expect(el).toHaveText(['First name called with: not yet', 'Last name called with: not yet'].join('\n'));
```
rwaskiewicz, line 23: In the [Karma version of the test](
https://github.com/ionic-team/stencil/pull/5456/files#diff-0bc7d13655d1b50de6b97c648eccd3328381c75b048b436f146a54906b065e65L14) (where we didn't use `toHaveText`):
```
    expect(el.innerText).toBe('First name called with: not yet\n\nLast name called with: not yet');
```

I noticed the double `\n` in the old test, whereas we have a single one in this test. I'm guessing this has to do with how `toHaveText` operates (I'm guessing not on `innerText`, perhaps `textContent`?), is that the reason for this (tiny) difference? 

FWIW, the rendered component looks correct to me, which is why I assume this is just an API difference, but I could be wrong!

test/wdio/conditional-rerender/cmp.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+        <>
+          <style>{css}</style>
+          <conditional-rerender-root></conditional-rerender-root>
+        </>
+      ),
+    });
+  });
+
+  it('contains a button as a child', async () => {
+    await $('main').waitForExist();
```
rwaskiewicz, line 38: Previously, we had been using a [`setTimeout` call](https://github.com/ionic-team/stencil/pull/5456/files#diff-e96ae002fbd5cf28368a0849203fec35d2d7d7a4d19f86b3724090fe39cbb427L15) with a value of 500 to ensure that the `setTimeout` call in the component (which uses a value of 20). IIUC, `waitForExist` has a default wait time of 500ms. I know that 20ms is a _short_ amount of time, but is it possible that this call ends before that `setTimeout` call in the component completes?

test/wdio/image-import/cmp.test.tsx
```diff
@@ -0,0 +1,20 @@
+import { h } from '@stencil/core';
+import { render, type StencilEnvironment } from '@wdio/browser-runner/stencil';
+
+describe('svg attr', () => {
+  let env: StencilEnvironment | undefined;
```
rwaskiewicz, line 5: For my own understanding, what would the benefit of storing the return value `render` in `env` here, as opposed to doing something like:
```diff
diff --git a/test/wdio/image-import/cmp.test.tsx b/test/wdio/image-import/cmp.test.tsx
index 8be6be201..043418999 100644
--- a/test/wdio/image-import/cmp.test.tsx
+++ b/test/wdio/image-import/cmp.test.tsx
@@ -1,17 +1,15 @@
 import { h } from '@stencil/core';
-import { render, type StencilEnvironment } from '@wdio/browser-runner/stencil';
+import { render } from '@wdio/browser-runner/stencil';
 
 describe('svg attr', () => {
-  let env: StencilEnvironment | undefined;
-
   beforeEach(async () => {
-    env = render({
+    render({
       template: () => <image-import></image-import>,
     });
   });
 
   it('adds and removes attribute', async () => {
-    const img = $(env.root).$('img');
+    const img = $('img');
     await img.waitForExist();
 
     const src = await img.getAttribute('src');
```

test/wdio/stencil.config.ts
```diff
@@ -0,0 +1,22 @@
...
+export const config: Config = {
+  namespace: 'TestApp',
+  tsconfig: 'tsconfig-stencil.json',
+  outputTargets: [
+    {
+      type: 'dist',
+    },
+  ],
+  plugins: [sass()],
+  buildDist: true,
```
rwaskiewicz, line 14: Comparing this to `karma/stencil.config.ts`, it looks like we're not bringing `buildEs5` over to this config. I'm OK with that, but wanted to make sure I understand what the motivation for that was (I assume it's speed/warning related, but wanted to make sure I'm not missing anything).

test/karma/test-app/components.d.ts
```diff
@@ -5,49 +5,11 @@
  * It contains typing information for all components that exist in this project.
  */
 import { HTMLStencilElement, JSXBase } from "@stencil/core/internal";
-import { SomeTypes } from "./util";
```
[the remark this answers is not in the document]
rwaskiewicz: Just tying [this comment](https://github.com/ionic-team/stencil/pull/5456#issuecomment-1993300117) back to this thread.

For posterity, we're gonna ignore the `components.d.ts` file in the `wdio` dir, but can't do that ATM for the Karma tests.

rwaskiewicz (changes requested):

test/wdio/attribute-complex/cmp.test.tsx
```diff
@@ -1,16 +1,26 @@
...
 
+describe('attribute-complex', function () {
   beforeEach(async () => {
-    app = await setupDom('/attribute-complex/index.html');
+    render({
+      template: () => <attribute-complex></attribute-complex>,
+    });
+  });
+
+  this.afterEach(() => {
```
tanner-reits, line 17:
```suggestion
  afterEach(() => {
```

tanner-reits (changes requested):
Couple questions, nothing major

test/wdio/attribute-boolean/cmp.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+    await expect(root).toHaveAttribute('aria-hidden', 'false');
+    await expect(root).toHaveAttribute('fixedtrue', 'true');
+    await expect(root).toHaveAttribute('fixedfalse', 'false');
+    await expect(root).not.toHaveAttribute('readonly');
+    await expect(root).not.toHaveAttribute('tappable');
+    await expect(root).not.toHaveAttribute('str');
+    await expect(root).not.toHaveAttribute('no-appear');
+    await expect(root).not.toHaveAttribute('no-appear-two');
+
+    await browser.pause(100);
```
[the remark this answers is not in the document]
christian-bromann: nope, will remove.

test/wdio/attribute-boolean/cmp.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+    await expect(child).not.toHaveAttribute('noreflect');
+    await expect(child).not.toHaveAttribute('tappable');
+
+    const button = await $('button');
+    await button.click();
+
+    await expect($(root)).toHaveAttribute('aria-hidden', 'true');
+    await expect(root).toHaveAttribute('fixedtrue', 'true');
+    await expect(root).toHaveAttribute('fixedfalse', 'false');
+    await expect(root).toHaveAttribute('readonly', 'true');
```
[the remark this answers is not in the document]
christian-bromann: The `readonly` property is a [boolean value](http://www.w3.org/TR/html5/forms.html#attr-input-readonly) which can be set in the following ways:

```html
<input type="text" readonly />
<input type="text" readonly="" />
<input type="text" readonly="readonly" />
<input type="text" readonly="ReAdOnLy" />
```

The WebDriver protocol transforms this property into its actual value (e.g. `true`) while calling `getAttribute` just returns the value set to the property (e.g. `''`). This is different from e.g. `tappable` which is not a reserved HTML property and can contain arbitrary strings.

That said, it is worth investigating constraints using reserved attributes on Stencil components. One could definitely expect to have it work like any custom property, especially if applied to elements where `readonly` has no effect on (e.g. [text](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/text), [search](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search), [url](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/url), [tel](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/tel), [email](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/email), [password](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/password), [date](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date), [month](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/month), [week](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/week), [time](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/time), [datetime-local](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local), and [number](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number) [<input>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input) types and the [<textarea>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea) form control elements).

test/wdio/attribute-complex/cmp.test.tsx
```diff
@@ -51,15 +61,18 @@ describe('attribute-complex', function () {
...
-    await waitForChanges();
-
     expect(instance.nu0).toBe(1234);
     expect(instance.nu1).toBe(-111.1);
 
     expect(instance.bool0).toBe(true);
     expect(instance.bool1).toBe(false);
     expect(instance.bool2).toBe(false);
 
+    await browser.pause(100);
```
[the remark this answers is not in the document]
christian-bromann: It seems like we can remove this.

test/wdio/attribute-host/cmp.test.tsx
```diff
@@ -0,0 +1,103 @@
...
+      // get default border color from body element as it might differ between different OS
+      'border-color': defaultBorderColor,
+      display: 'inline-block',
+      'font-size': '16px',
+    });
+
+    // this tests CSS custom properties in inline style, but CSS var are
+    // not supported natively in IE11, so let's skip the test
+    const win = window as any;
+    if (win.CSS && win.CSS.supports && win.CSS.supports('--prop', 'value')) {
```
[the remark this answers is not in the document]
christian-bromann: good point, will remove.

test/wdio/attribute-host/cmp.test.tsx
```diff
@@ -0,0 +1,103 @@
...
+    await expect(elm).not.toHaveAttribute('color');
+    await expect(elm).not.toHaveAttribute('no-attr');
+
+    let defaultBorderColor = getComputedStyle(document.body).borderColor.replaceAll(' ', '');
+    if (browser.isChromium) {
+      defaultBorderColor = defaultBorderColor.replace('rgb(', 'rgba(').replace(')', ',1)');
+    }
+    await expect(elm).toHaveStyle({
+      // get default border color from body element as it might differ between different OS
+      'border-color': defaultBorderColor,
```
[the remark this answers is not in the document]
christian-bromann: This is unfortunately a limitation within [the spec](https://github.com/w3c/webdriver/issues/1447) which hasn't been addressed yet. I raised a PR to the wpt.fyi to push on this topic. Anyway for now we have to do some browser checks to validate the right value.

The reason why I am doing `getComputedStyle(document.body).borderColor.replaceAll(' ', '')` is because we are validating the "before" state here which refers to the system default color for this HTML element. Running this in CI has shown that this default color differs between Mac (my machine) and Linux (the machine we use in CI).

test/wdio/computed-properties-prop-decorator/cmp.test.tsx
```diff
@@ -0,0 +1,39 @@
+import { h } from '@stencil/core';
+import { render } from '@wdio/browser-runner/stencil';
+
+describe('computed-properties-prop-decorator', function () {
+  beforeEach(async () => {
+    render({
+      template: () => (
+        <div>
```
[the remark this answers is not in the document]
christian-bromann: Sure thing.

test/wdio/computed-properties-watch-decorator/cmp.test.tsx
christian-bromann: Querying an element via `$` gives you  a `WebdriverIO.Element` instance that allows you to call WebdriverIO specific commands for it, e.g. `$('button').click()`. Now, it is not necessary to use `await` if you only use `$` since WebdriverIO only resolves the Promise behind the `$` call once you interact with the element. Basically:

```ts
// 👎 await for $ call not needed
const elem = await $('buttom')
await elem.click()
// 👍 better
await $('button').click()
```

Now, the speed improvement is marginal so both is essentially the same operation.

test/wdio/computed-properties-watch-decorator/cmp.test.tsx
christian-bromann: The WebDriver "[get element test](https://www.w3.org/TR/webdriver2/#get-element-text)" command returns a representation of the "visible" state of a text. While it is not possible to define this concretely, they are relying on an implementation for it that has been used for many years. This implementation seems to remove double `\n` as they are not impacting how the text is rendered. `innerText` on the other hand actually gives you the characters of the DOM node which doesn't contains any filtering.

test/wdio/conditional-rerender/cmp.test.tsx
christian-bromann: Actually I will remove this as the async `expect` call will do some retries until it hits the right content. I don't think we will run into a race condition here. The timeout is just too small.

test/wdio/image-import/cmp.test.tsx
christian-bromann: none, I initially tried to use `env.root` to ensure we capture the right elements on the stage but then discovered that this doesn't behave reliably sometimes (which I will address in the WebdriverIO project separately) so I moved away from it. The only other element on the page is the reporter which is encapsulated in a shadow dom and we shouldn't be able to reach into it for now. I will update as suggested above.

test/wdio/stencil.config.ts
christian-bromann: It seemed to me that this option generates more output that is not relevant to these tests so I left it out. Frankly, I'm not entirely certain about the specific differences it entails.

test/wdio/attribute-complex/cmp.test.tsx
```diff
@@ -1,16 +1,26 @@
...
+import { render } from '@wdio/browser-runner/stencil';
+import { $, browser } from '@wdio/globals';
 
-describe('attribute-complex', function () {
-  const { setupDom, tearDownDom } = setupDomTests(document);
-  let app: HTMLElement;
+customElements.whenDefined('attribute-complex').then(() => {
+  const el: any = document.querySelector('attribute-complex');
+  el?.getInstance();
+});
```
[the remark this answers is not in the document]
christian-bromann: I copied it over from the Karma test but it seems it is not relevant for the test at all, will remove.

test/wdio/tsconfig-stencil.json
```diff
@@ -0,0 +1,39 @@
+{
+  "compilerOptions": {
+    // "strict": true,
+    // "alwaysStrict": true,
```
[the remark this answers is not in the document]
christian-bromann: Nothing can stop us!

christian-bromann:
@rwaskiewicz @tanner-reits thanks for the feedback. I addressed the comments.

test/wdio/computed-properties-prop-decorator/cmp.test.tsx
```diff
@@ -1,15 +1,15 @@
...
+import { Fragment, h } from '@stencil/core';
 import { render } from '@wdio/browser-runner/stencil';
 
-describe('computed-properties-prop-decorator', function () {
+describe('computed-properties-prop-decorator', () => {
   beforeEach(async () => {
     render({
       template: () => (
-        <div>
+        <>
```
tanner-reits, line 8: Should `Fragment` be in this tag?

tanner-reits (approved):
One more question, non-blocking

test/wdio/computed-properties-prop-decorator/cmp.test.tsx
christian-bromann: Isn't this done automatically?

rwaskiewicz (approved):

PR #5465, opened 2024-03-14
chore(test): migrate Karma tests to WebdriverIO

christian-bromann:
This patch migrates:

- [custom-elements-delegates-focus](https://github.com/ionic-team/stencil/blob/b4230b6939b016299d78cc4d16c175e73c703733/test/karma/test-app/custom-elements-delegates-focus)
- [custom-elements-output-tag-class-different](https://github.com/ionic-team/stencil/blob/b4230b6939b016299d78cc4d16c175e73c703733/test/karma/test-app/custom-elements-output-tag-class-different)
- [custom-elements-output-webpack](https://github.com/ionic-team/stencil/blob/b4230b6939b016299d78cc4d16c175e73c703733/test/karma/test-app/custom-elements-output-webpack)


test/karma/test-app/custom-elements-output-webpack/webpack.config.js
```diff
@@ -1,20 +0,0 @@
-const path = require('path');
-const { WWW_OUT_DIR } = require('../../constants');
-
-module.exports = {
-  entry: path.resolve(__dirname, 'index.esm.js'),
```
christian-bromann, line 5: Does someone know the point of having Webpack involved here? It doesn't seem to do anything other than re-bundling the package and importing it in the browser. I believe we cover this feature with the [setup script](https://github.com/ionic-team/stencil/blob/908f7780f5979beeb99a4b5f0aedb626590a8ec8/test/wdio/setup.ts) automatically.

alicewriteswrongs: I think @rwaskiewicz is familiar with the basis for the setup -- if it's possible for us to do without it I'm all for it because it's super confusing haha

alicewriteswrongs (approved):

tanner-reits:
@christian-bromann looks like there are some CI failures

christian-bromann:
> looks like there are some CI failures

Addressed them, they sneaked it after the last rebase

test/karma/test-app/custom-elements-output-webpack/webpack.config.js
rwaskiewicz: IIRC, reason for this was two-fold:
1. to test Stencil could go through the webpack process (this is a regression Stencil had at some point)
2. To use the `dist-custom-elements` output target, as opposed to `www`

test/wdio/package-lock.json
```diff
@@ -23,8 +23,7 @@
       }
     },
     "../..": {
-      "name": "@stencil/core",
```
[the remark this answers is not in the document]
christian-bromann: not sure how this slipped in

christian-bromann:
> Some of these tests were built specifically for DCE or Webpack regressions/features - with that in mind, I think there's some value in bringing them back.

@rwaskiewicz I tracked down the source of the test which was introduced in https://github.com/ionic-team/stencil/pull/3039. From what I can see is that this verifies that the DCE output target provides a `defineCustomElement` method to register the custom element. I would suggest to tweak the setup script to __not__ run for this test and have the component imported and registered manually and verify that the component can be initialized properly.

I don't think it makes sense anymore to keep Webpack as a dependency for this test stack.

christian-bromann:
> I think I'd be amenable to that 👍

I made changes accordingly. Let me know what you think.

alicewriteswrongs (approved):

test/wdio/slot-reorder/cmp.test.tsx
```diff
@@ -131,21 +131,21 @@ describe('slot-reorder', () => {
...
-    await browser.waitUntil(() => {
+    await browser.waitUntil(async () => {
       return !document.querySelector('div.reordered');
     });
 
     ordered();
 
     await $('button').click();
-    await browser.waitUntil(() => {
+    await browser.waitUntil(async () => {
```
rwaskiewicz, line 148: Does passing an `async` function to `waitUntil` do anything special? Or do we not need these `async` fns?

rwaskiewicz (approved):
One non-blocking question

test/wdio/slot-reorder/cmp.test.tsx
christian-bromann: I've seen race conditions saying `.then is not a function` and my assumptions are that this has to do with WebdriverIO. It is on the list of things I'ld like to look into eventually. In the meeting, I am adding this to avoid the race conditions.

PR #5472, opened 2024-03-14
chore(wdio): port key-reorder test suite to wdio

alicewriteswrongs:
Porting over the key-reorder test suite from Karma to Wdio

- [ ] Yes
- [x] No

## Testing

This one is pretty straightforward, the only thing was that I realized that we had a component that wasn't being used that I deleted.


test/karma/test-app/key-reorder/cmp.tsx
alicewriteswrongs, line 1: We had two components, `<key-reorder>` and `<key-reorder-root>`, but we were only rendering the `<key-reorder-root>` component. So I renamed the existing `<key-reorder-root>` component to `<key-reorder>` and then deleted this component.

alicewriteswrongs (commented):
just noting the one funny thing

tanner-reits (approved):

Issue #1352, opened 2019-01-31
Compile time error when EventEmitter type contains a period (static analysis error?)


snaptopixel:
**Stencil version:**

```
 @stencil/core@0.17.0<version>
```

**I'm submitting a:**

[x] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**
I'm using ambient type declarations with namespaces like `myComponent.SomeType` in my stencil lib. Works great, but the second I do this: `EventEmitter<myComponent.SomeType>` I get a cryptic error when compiling:
```
[ ERROR ]  /path/to/my-component.tsx
           Cannot read property 'toString' of undefined
```
The problem goes away when there is no period, if I do something like:
```ts
type SomeType = myComponent.SomeType;
// Omitted code
@Event() foo: EventEmitter<SomeType>;
```

**Expected behavior:**
If it is valid TypeScript it works with Stencil

christian-bromann:
This has been released with Stencil 🚞 [v4.13.0](https://github.com/ionic-team/stencil/releases/tag/v4.13.0)

PR #5475, opened 2024-03-14
fix(compiler): don't validate references for @Prop, @Method and @Event decorator

christian-bromann:
It currently seems that we have implemented a validation that would prevent returning a type that collides with the name of any imported Stencil type. As far as I understood the situation is as following: given I have a type name that has the same name as one of Stencils decorators (e.g. `Element`, `Event`, `Listen`, `Method`, `Prop`, `PropDidChange`, `PropWillChange`, `State` or `Watch`) it will throw a meaningless error. This can be easily reproduced in many ways:

```ts
// in src/utils/utils.ts
interface Method {}

export function returnSomethingInvalid(): Method {
  return {}
}
```

Then having the following component decorator will fail:

```ts
// in src/components/my-component/my-component.tsx
import { Component, Prop, h, Method, EventEmitter, Event } from '@stencil/core';
import { returnSomethingInvalid } from '../../utils/utils';

@Component({
  tag: 'my-component',
  styleUrl: 'my-component.css',
  shadow: true,
})
export class MyComponent {
  // ...
  @Method()
  async didDismiss() {
    return returnSomethingInvalid()
  }
  // ...
}
```

This behavior was introduced in https://github.com/ionic-team/stencil/commit/ca60073da7733d4e40ccc7bd1ad35b891462aeb2 and was suppose to fix an issue reported in #1352.

## What is the new behavior?
Remove this validation as I don't think it makes sense here. I don't totally understand the reasoning behind it but it seems unnecessary to validate if the returning type matches with one of Stencils decorators. It seems safe to me to just remove this completely. 

STENCIL-980

- [ ] Yes
- [x] No

## Testing

I can add some compiler tests for this if we think we want to move forward with this.

## Other information

n/a

### DIFF
--- a/src/compiler/transformers/decorators-to-static/event-decorator.ts
+++ b/src/compiler/transformers/decorators-to-static/event-decorator.ts
@@ -9,7 +9,6 @@ import {
   resolveType,
   retrieveTsDecorators,
   serializeSymbol,
-  validateReferences,
 } from '../transform-utils';
 import { getDecoratorParameters, isDecoratorNamed } from './decorator-utils';
 
@@ -76,7 +75,6 @@ const parseEventDecorator = (
     docs: serializeSymbol(typeChecker, symbol),
     complexType: getComplexType(typeChecker, program, prop),
   };
-  validateReferences(diagnostics, eventMeta.complexType.references, prop.type);
   return eventMeta;
 };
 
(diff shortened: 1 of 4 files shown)


rwaskiewicz (approved):

alicewriteswrongs (approved):

christian-bromann:

PR #5477, opened 2024-03-14
chore(wdio): port json-basic suite from karma to wdio

alicewriteswrongs:
This one is real straightforward! Just one test.

- [ ] Yes
- [x] No

## Testing

If CI passes I think we're good!


test/wdio/json-basic/cmp.test.tsx
```diff
@@ -0,0 +1,14 @@
+import { h, Fragment } from '@stencil/core';
```
rwaskiewicz, line 1: Non-blocking - `Fragment` isn't used in this test and can be removed
```suggestion
import { h } from '@stencil/core';
```

rwaskiewicz (approved):

christian-bromann (approved):

PR #5479, opened 2024-03-14
chore(wdio): port lifecycle-async tests from karma to wdio

alicewriteswrongs:
Porting the `lifecycle-async` test suite from Karma to Wdio!

- [ ] Yes
- [x] No

## Testing

Nothing too crazy going on! Some comments pointing out a few things though.


test/wdio/lifecycle-async/cmp-a.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+  beforeEach(() => {
+    render({
+      template: () => <lifecycle-async-a></lifecycle-async-a>,
+    });
+  });
+
+  it('fire load methods in order', async () => {
+    await $('lifecycle-async-a').waitForExist();
+
+    await browser.pause(500);
```
alicewriteswrongs, line 14: unfortunately without these `borwser.pause` calls here things aren't in place by the time we do all the `toHaveText` checks

alicewriteswrongs (commented):
just pointing out one thing

test/wdio/lifecycle-async/util.ts
```diff
@@ -0,0 +1,5 @@
+export function timeout(ms: number): Promise<void> {
```
christian-bromann, line 1: Do we need this?

test/wdio/lifecycle-async/cmp-a.test.tsx
christian-bromann, line 14: Would this work?
```suggestion
    await $('lifecycle-async-a li').waitForExist();
```

christian-bromann (changes requested):

test/wdio/lifecycle-async/util.ts
alicewriteswrongs: yeah two of the components use it, it's not used in the test file but in `cmp-b` and `cmp-c`

christian-bromann: ah, ok didn't recognize that. I thought it was a way to "pause" the test. All good then.

test/wdio/lifecycle-async/cmp-a.test.tsx
alicewriteswrongs: I do still need the pause here, if I just do the `await $('lifecycle-async-a li')` then the tests ends up being flaky and seems to faily around 50% of the time

alicewriteswrongs: It looks like what can happen is that it stops waiting when only one `li` is on the DOM, so the check for `.length` doesn't work :/

rwaskiewicz (approved):

test/wdio/lifecycle-async/cmp-a.test.tsx
christian-bromann: I raised an issue for this and will take a look eventually: https://github.com/webdriverio/expect-webdriverio/issues/1493

christian-bromann (approved):
👍

PR #5481, opened 2024-03-14
feat(testing): support deep piercing with Puppeteer

christian-bromann:
Currently Stencil only supports piercing Shadow DOMs one level deep. However users might want to look up more nested components. This is already supported by Puppeteer, e.g. using Puppeteer primitives, you can do:

```ts
const span = await page.$('cmp-a >>> cmp-b >>> cmp-c >>> span');
expect(await span.evaluate((el) => el.textContent)).toBe('I am in component C');
```

However Stencil wraps this API for convenience purposes and had a custom implementation for piercing through the Shadow DOM. Likely this feature was implemented in Puppeteer after this interface was implemented in Stencil.

I mentioned on our Jam session that I had concerns about limitations about this approach. As we are now fully rely on Puppeteers implementation for piercing, these limitations are gone.

## What is the new behavior?
This patch makes piercing through Shadow DOM available using Stencil primitives.

- [ ] Yes
- [x] No

## Testing

Add some e2e tests for this.

## Other information

n/a


christian-bromann:
@rwaskiewicz this is no ready for review.

christian-bromann:
Raised a subsequent PR to update the docs: https://github.com/ionic-team/stencil-site/pull/1383

rwaskiewicz (approved):

tanner-reits (approved):

PR #5484, opened 2024-03-15
chore(test): migrate `es5-addclass-svg`

christian-bromann:
Migrate `es5-addclass-svg` from Karma to WDIO!


test/karma/test-app/es5-addclass-svg/karma.spec.ts
```diff
@@ -1,31 +0,0 @@
...
-    }
-  });
-
-  function testNativeShadow(result: HTMLElement) {
-    const svg = result.shadowRoot.querySelector('svg');
-    expect(svg.getAttribute('class')).toContain('existing-css-class');
-    expect(svg.getAttribute('class')).not.toContain('sc-es5-addclass-svg');
-  }
-
-  function testPolyfilledShadow(result: HTMLElement) {
```
rwaskiewicz, line 27: Can we add a note to the git commit as to why we're no longer testing this please?

rwaskiewicz (changes requested):

test/karma/test-app/es5-addclass-svg/karma.spec.ts
christian-bromann: Rebased the branch and squashed all commits into a single one that contains a note that we removed testing this behavior as we expect Shadow DOM to be widely supported by now.

rwaskiewicz (approved):

alicewriteswrongs (approved):

PR #5485, opened 2024-03-15
chore(test): migrate `esm-import` and `esm-webpack`

christian-bromann:
migrate `esm-import` and `esm-webpack`


rwaskiewicz (dismissed):
Can you add some context for us to review this? This doesn't look like a straight port of tests, but rather a coalescing of the two that strips out some webpack-configuration. I'd like to understand better what's changed & why here for these non-trivial ports before reviewing

christian-bromann:
Yeah so these two were a bit weird as `test/karma/test-app/esm-import/karma.spec.ts` did define a function that contained all assertions:

```ts
export async function testEsmImport(app: HTMLElement) {
  // ...
```
```

Which then got imported in `test/karma/test-app/esm-webpack/karma.spec.ts` like so:

```ts
    await testEsmImport(app);	
    const hydratedElm = app.querySelector('esm-import.hydrated');	
    expect(hydratedElm).not.toBe(null
```

What I've done is to move the content of the function into each test, therefor duplicated the content of the function but I think for tests that this is fine since it removes confusion and dependencies. I tried to migrate how the component is being initialized as close as possible to the original but left out the CustomEvent polyfill since we don't need this anymore afaik.

christian-bromann:
@alicewriteswrongs @rwaskiewicz any objections to merge?

alicewriteswrongs (approved):

alicewriteswrongs:
sorry somehow I thought I had put the review!

PR #5487, opened 2024-03-15
chore(wdio): port listen-jsx suite from karma to wdio

alicewriteswrongs:
Another day, another wdio port

Nothing super notable about this one, we've got one test which translated to wdio-ese quite easily.

- [ ] Yes
- [x] No

## Testing

If CI passes we're good!


christian-bromann (approved):

test/wdio/listen-jsx/cmp-root.test.tsx
```diff
@@ -0,0 +1,16 @@
+import { h, Fragment } from '@stencil/core';
```
rwaskiewicz, line 1: Non-blocking - Fragment isn't used and can be deleted

```suggestion
import { h } from '@stencil/core';
```

rwaskiewicz (approved):

PR #5493, opened 2024-03-15
chore(wdio): convert event-custom-type to wdio

rwaskiewicz:

GitHub Issue Number: N/A

## What is the new behavior?

port the `test/karma/test-app/event-custom-type` test suite to webdriverio.

## Documentation

N/A

- [ ] Yes
- [x] No

## Testing

## Other information

STENCIL-1191


test/wdio/event-custom-type/cmp.test.tsx
```diff
@@ -0,0 +1,20 @@
...
+    });
+  });
+
+  it('should dispatch an event on load', async () => {
+    await expect($('#counter')).toHaveText('1');
+  });
+
+  it('should emit a complex type', async () => {
+    // wait to ensure the event it emitted in the component
+    await browser.pause(100);
```
christian-bromann, line 17: This should not be needed as WebdriverIO will retry the assertion in case it fails
```suggestion
```

christian-bromann (changes requested):

test/wdio/event-custom-type/cmp.test.tsx
rwaskiewicz: makes sense, thanks!

christian-bromann (approved):
👍

test/wdio/event-custom-type/index.html
alicewriteswrongs, line 1: I think maybe you missed deleting the html file here

alicewriteswrongs (changes requested):
oops, meant for that comment to be a review!

alicewriteswrongs (approved):

PR #5495, opened 2024-03-15
chore(test): migrate `external-imports`

christian-bromann:
migrate external-imports


test/wdio/key-reorder/cmp.test.tsx
```diff
@@ -31,6 +31,7 @@ describe('key-reorder', function () {
 
     const button = root.$('button');
     await button.click();
+    await $('#item-4').waitForExist();
```
christian-bromann, line 34: This caused my test to fail locally, so thought of adding this to make it less flaky.

rwaskiewicz: Can we note this in a git commit message please?

test/wdio/external-imports/cmp.test.tsx
```diff
@@ -0,0 +1,26 @@
...
+        </>
+      )
+    })
+  });
+  it('render all components without errors', async () => {
+    let elm = $('external-import-a');
+    await expect(elm).toHaveText('Marty McFly');
+
+    elm = $('external-import-b');
+    await expect(elm).toHaveText('Marty McFly');
```
rwaskiewicz, line 21: I know this was how it was in the original test, but can we use a separate variable assignment here in the off-chance that this assignment doesn't resolve/has a reference to `external-import-a` (unsure how possible that actually is in practice)? I want to make sure we're asserting against `external-import-b` instead of `external-import-a`:
```suggestion
    const elm2 = $('external-import-b');
    await expect(elm2).toHaveText('Marty McFly');
```

test/wdio/external-imports/cmp.test.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+    elm = $('external-import-c');
+    await expect(elm).toHaveText('Marty McFly');
```
rwaskiewicz, line 24: Similar to my comment above - 
```suggestion
    const elm3 = $('external-import-c');
    await expect(elm3).toHaveText('Marty McFly');
```

rwaskiewicz (changes requested):
One small ask in terms of

test/wdio/key-reorder/cmp.test.tsx
christian-bromann: Add comment in the git message

alicewriteswrongs (approved):

test/wdio/external-imports/cmp.test.tsx
```diff
@@ -0,0 +1,26 @@
...
+        <>
+          <external-import-a></external-import-a>
+          <external-import-b></external-import-b>
+          <external-import-c></external-import-c>
+        </>
+      ),
+    });
+  });
+  it('render all components without errors', async () => {
+    let elm = $('external-import-a');
```
rwaskiewicz, line 17: Non-blocking - this can be `const` now
```suggestion
    const elm = $('external-import-a');
```

rwaskiewicz (approved):

PR #5516, opened 2024-03-18
chore(wdio): migrate import-aliasing to wdio

rwaskiewicz:

GitHub Issue Number: N/A

## What is the new behavior?

port the `test/karma/test-app/import-aliasing` test suite to webdriverio

- [ ] Yes
- [x] No

## Testing

## Other information

STENCIL-1191


test/wdio/import-aliasing/cmp.test.tsx
```diff
@@ -1,37 +1,41 @@
...
+    });
   });
 
-  afterEach(tearDownDom);
+  it('should render correctly with aliased imports', async () => {
+    await $('import-aliasing').waitForExist();
 
-  it('should render correctly with alised imports', async () => {
-    const host = app.querySelector('import-aliasing');
+    const host = document.querySelector('import-aliasing');
```
rwaskiewicz, line 22: I struggled to convert this to something like:
```
const host = await $('import-aliasing');
```
for a couple of reasons:
1. I wasn't sure how to get the element itself. I wanted this rather than using the webdriver protocol for two reasons
    1. I wanted to set the attribute 'user' to 'Peter' here - https://github.com/ionic-team/stencil/pull/5516/files#diff-e821c874ede28b67d6d626b7e6b9f648f2a7905d0b059e46415cc914c0d7aa6dL25. As I understand it, I should do something like `browser.execute((elm) => elm.setAttribute('user', 'Peter'), host);`. But, the type checker didn't like that, as the 2nd argument there, `host`, isn't an HTMLElement. A type assertion like `host as unknown as HTMLElement` "worked", but I didn't love it for just setting an attribute
    2. Similarly, I didn't know how to call `myMethod` here - https://github.com/ionic-team/stencil/pull/5516/files#diff-e821c874ede28b67d6d626b7e6b9f648f2a7905d0b059e46415cc914c0d7aa6dR37. I suppose `browser.execute()` could be used here, but I'm not sure if that's the right way to go about things.

If there's a better way to do this, I'm happy to hear suggestions!

test/wdio/import-aliasing/cmp.test.tsx
```diff
@@ -41,7 +45,8 @@ describe('import aliasing', function () {
   });
 
   it('should link up to the surrounding form', async () => {
-    const formEl = app.querySelector('form');
+    await $('import-aliasing').waitForExist();
+    const formEl = document.querySelector('form');
     expect(new FormData(formEl).get('test-input')).toBe('my default value');
```
rwaskiewicz, line 50: Similar to my problem above, I didn't know how to get the form element to create the FormData here. Happy to hear suggestions

test/wdio/import-aliasing/cmp.test.tsx
christian-bromann: There is no reason to "feel" forced to use the WebDriver primitives. These are some known limitation/DX issues that are still unresolved.

christian-bromann (approved):
LGTM 👍

tanner-reits (approved):

PR #5517, opened 2024-03-18
chore(wdio): fix build issues

rwaskiewicz:

GitHub Issue Number: N/A

See "New Behavior"

## What is the new behavior?

fix a build issue where running `npm run build` twice in the `test/wdio` directory would result in the following failure:
```
 [47:48.5]  aborted build, 343ms  MEM: 359.2MB

[ ERROR ]  TypeScript: event-custom-type/cmp.tsx:3:44
           Cannot find module '../components' or its corresponding type declarations.

      L3:  import { EventCustomTypeCustomEvent } from '../components';

[47:48.5]  build failed in 343 ms
```

This was introduced in https://github.com/ionic-team/stencil/pull/5493, where the import statement was not correctly ported over. As a result, type resolution would fail under `tsconfig.json#moduleResoution` having a value of 'Node16'

- [ ] Yes
- [x] No

## Testing

Check out this branch and run `npm run build` 2+ times. You should not see an error

### DIFF
--- a/test/wdio/event-custom-type/cmp.tsx
+++ b/test/wdio/event-custom-type/cmp.tsx
@@ -1,6 +1,6 @@
 import { Component, h, Event, EventEmitter, State, Listen } from '@stencil/core';
 
-import { EventCustomTypeCustomEvent } from '../components';
+import { EventCustomTypeCustomEvent } from '../src/components.js';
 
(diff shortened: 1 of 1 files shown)


christian-bromann (approved):
👍

tanner-reits (approved):

PR #5519, opened 2024-03-18
chore(test): migrate scoped-slot-append-and-prepend from Karma to WebdriverIO

christian-bromann:
Migrate scoped-slot-append-and-prepend from Karma to WebdriverIO


alicewriteswrongs (approved):
lgtm!

PR #5522, opened 2024-03-18
chore(test): migrate scoped-slot-text

christian-bromann:
Migrate scoped-slot-text

### DIFF
--- a/test/karma/test-app/components.d.ts
+++ b/test/karma/test-app/components.d.ts
@@ -14,8 +14,6 @@ export namespace Components {
     interface ChildWithReflection {
         "val": number | any;
     }
-    interface CmpLabel {
-    }
     interface CustomElementChild {
     }
     interface CustomElementChildDifferentNameThanClass {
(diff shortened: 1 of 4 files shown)


test/karma/test-app/scoped-slot-text-with-sibling/karma.spec.ts
```diff
@@ -15,9 +15,9 @@ describe('scoped-slot-text-with-sibling', () => {
...
    * @returns the custom element
    */
-  function getCmpLabel(): HTMLCmpLabelElement {
+  function getCmpLabel(): HTMLCmpLabelWithSlotSiblingElement {
     const customElementSelector = 'cmp-label-with-slot-sibling';
-    const cmpLabel: HTMLCmpLabelElement = app.querySelector(customElementSelector);
+    const cmpLabel: HTMLCmpLabelWithSlotSiblingElement = app.querySelector(customElementSelector);
     if (!cmpLabel) {
       fail(`Unable to find element using query selector '${customElementSelector}'`);
     }
```
alicewriteswrongs, line 23: did you mean to include these edits to a different karma test?

christian-bromann: Nope, for some reason this Karma test was using types generated by the directory I migrated. I updated the test to use the right class.

rwaskiewicz: I `blame` me 😆 
Screenshot 2024-03-19 at 8 36 23 AM

Can we add a note in the final commit message about this? A lot of these `scoped-` prefixed tests are look pretty similar in name. I think it'd be worth a clarification note for another 2.5 years from now 😉

rwaskiewicz (changes requested):
One small ask regarding that type change RE: adding it to the git commit msg

christian-bromann:
Thanks for the review. @rwaskiewicz added a note to the commit message.

rwaskiewicz (approved):
Thanks!

alicewriteswrongs (approved):

PR #5524, opened 2024-03-19
chore(lint): turn on linting for the `test/wdio` directory tree

alicewriteswrongs:
This just adds this to the `lint` command in `package.json`. Two details:

- I added an entry to `overrides` in `.eslintrc.js` to prevent the `jest/expect-expect` rule from being applied to `test/wdio`. If violations of other jest rules pop up in there too we should turn them off as well.
- It automatically fixed a `let` to `const` and reordered some imports

No linting!

## What is the new behavior?

Linting!

- [ ] Yes
- [x] No

## Testing

If CI passes we should be good.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -89,6 +89,13 @@ module.exports = {
         'no-var': 'off',
       },
     },
+    {
+      // we don't want to use jest-related lint rules in the wdio tests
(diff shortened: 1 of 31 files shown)


christian-bromann (approved):
Nice on 👌

rwaskiewicz (approved):

PR #5525, opened 2024-03-19
chore(test): make scoped-basic less flakey

alicewriteswrongs:
This de-flakes things a bit. This test was sometimes not passing on CI in the style assertions where it checks things like `color` and `backgroundColor`, in particular I think because of some differences in how computer styles for these things are set and formatted.

- [ ] Yes
- [x] No

## Testing

I wasn't seeing this test failure locally, but only on CI when the tests are running in Firefox.

To run the tests locally you can do

```sh
CI=true npm run test.wdio
```


tanner-reits (approved):
kinda goofy, but it works 😂

rwaskiewicz (approved):

PR #5529, opened 2024-03-19
chore(test): migrate `shadow-dom-basic` tests to wdio

tanner-reits:

GitHub Issue Number: N/A

## What is the new behavior?

Migrates `test/karma/test-app/shadow-dom-basic` to WDIO

- [ ] Yes
- [x] No


test/wdio/shadow-dom-basic/cmp.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+    await expect(shadowDiv.textContent.trim()).toBe('shadow');
+
+    const shadowBG = window.getComputedStyle(shadowDiv).backgroundColor;
+    await expect(shadowBG).toBe('rgb(0, 0, 0)');
+
+    const lightBG = window.getComputedStyle(shadowDomBasic.lastElementChild).backgroundColor;
+    await expect(lightBG).toBe('rgb(255, 255, 0)');
+  }
+
+  async function testPolyfilledShadow(elm: Element) {
```
rwaskiewicz, line 43: Do these tests run? I wonder if these were just for IE 11 support 🤔

test/wdio/shadow-dom-basic/cmp.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+    await expect(shadowDomBasic.nodeName).toBe('SHADOW-DOM-BASIC');
+    const shadowDiv = shadowDomBasic.shadowRoot.lastElementChild.previousElementSibling;
+    await expect(shadowDiv.nodeName).toBe('DIV');
+    await expect(shadowDiv.textContent.trim()).toBe('shadow');
+
+    const shadowBG = window.getComputedStyle(shadowDiv).backgroundColor;
+    await expect(shadowBG).toBe('rgb(0, 0, 0)');
+
+    const lightBG = window.getComputedStyle(shadowDomBasic.lastElementChild).backgroundColor;
+    await expect(lightBG).toBe('rgb(255, 255, 0)');
```
rwaskiewicz, line 40: I don't think we need to `await` these `expect`s, since we're not performing any `$` or other WebdriverIO queries:
```suggestion
    expect(lightDiv.nodeName).toBe('DIV');
    expect(lightDiv.textContent.trim()).toBe('light');

    expect(shadowDomBasic.nodeName).toBe('SHADOW-DOM-BASIC');
    const shadowDiv = shadowDomBasic.shadowRoot.lastElementChild.previousElementSibling;
    expect(shadowDiv.nodeName).toBe('DIV');
    expect(shadowDiv.textContent.trim()).toBe('shadow');

    const shadowBG = window.getComputedStyle(shadowDiv).backgroundColor;
    expect(shadowBG).toBe('rgb(0, 0, 0)');

    const lightBG = window.getComputedStyle(shadowDomBasic.lastElementChild).backgroundColor;
    expect(lightBG).toBe('rgb(255, 255, 0)');
```

rwaskiewicz (changes requested):

test/wdio/shadow-dom-basic/cmp.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+      template: () => <shadow-dom-basic-root></shadow-dom-basic-root>,
+    });
+  });
+
+  it('render', async () => {
+    await $('shadow-dom-basic-root').waitForExist();
+    const elm = document.querySelector('shadow-dom-basic-root');
+
+    if ('attachShadow' in HTMLElement.prototype) {
+      await testNativeShadow(elm);
```
christian-bromann, line 16: I migrated a similar test and just got rid of the polyfilled version since today we expect Shadow DOM support when using Stencil, right?

test/wdio/shadow-dom-basic/cmp.test.tsx
tanner-reits: I believe that's true, but with https://github.com/ionic-team/stencil/pull/5530 I think we'll start getting lint errors (even in the case of not using wdio queries)

test/wdio/shadow-dom-basic/cmp.test.tsx
tanner-reits: Good call, I'll remove these

test/wdio/shadow-dom-basic/cmp.test.tsx
tanner-reits: Yep, removing the polyfilled tests now

test/wdio/shadow-dom-basic/cmp.test.tsx
rwaskiewicz: Hmmmm @alicewriteswrongs is that the case?

alicewriteswrongs: yep! it will make us just `await` them all

but if you `await` a value which is not a promise it doesn't do anything, so I think the trade-off might be worth it if the lint rule lets us never miss an `expect` which _should_ be awaited

alicewriteswrongs: ah actually I take it back! it seems the rule is smart enough to know when it's necessary

it doesn't complain about a missing `await` in

```ts
expect(2).toBe(2);
```

but it does complain about it in

```ts
    expect($('#json-foo')).toHaveText('bar');
```

alicewriteswrongs: I guess it looks for `$` or at the matcher function called on the return value of `expect` or something

alicewriteswrongs: looks like it checks a list of matchers which are async:

https://github.com/webdriverio/webdriverio/blob/51ac8482284ad34dea3cc899872397fb734de617/packages/eslint-plugin-wdio/src/rules/await-expect.ts#L21-L50

list here: https://github.com/webdriverio/webdriverio/blob/main/packages/eslint-plugin-wdio/src/constants.ts

tanner-reits: oh fancy! I can update these assertions then

rwaskiewicz (approved):

test/wdio/shadow-dom-basic/cmp.test.tsx
rwaskiewicz: Sweet! Thanks!

christian-bromann (approved):

alicewriteswrongs (approved):

PR #5530, opened 2024-03-19
chore(lint): add eslint-plugin-wdio, turn on rule for awaiting `expect`

alicewriteswrongs:
This adds `eslint-plugin-wdio` so that we can get some Webdriver-specific linting rules added to our configuration. In particular, we can specify that the `expect` function is prefixed with `await`, ensuring that we don't have dangling promises.

This rule is activated only for files in `test/wdio`.

We don't have a linting rule for this and I, in particular, keep forgetting to `await` everything.

## What is the new behavior?

Everything must be awaited!

- [ ] Yes
- [x] No

## Testing

If CI passes we're good.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,7 +1,7 @@
 module.exports = {
   root: true,
   parser: '@typescript-eslint/parser',
-  plugins: ['@typescript-eslint', 'jsdoc', 'jest', 'simple-import-sort'],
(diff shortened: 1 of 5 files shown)


christian-bromann (approved):
🙌

tanner-reits (approved):

rwaskiewicz (approved):

PR #5536, opened 2024-03-20
chore(test): migrate reflect-single-render

christian-bromann:
migrate reflect-single-render


test/wdio/reflect-single-render/cmp.test.tsx
```diff
@@ -0,0 +1,21 @@
...
+    render({
+      template: () => <parent-with-reflect-child></parent-with-reflect-child>,
+    });
+  });
+
+  it('renders the parent and child the correct number of times', async () => {
+    const parentShadowRoot = await $('parent-with-reflect-child');
+    await parentShadowRoot.waitForExist();
+
+    const childShadowRoot = parentShadowRoot.shadow$('child-with-reflection');
```
rwaskiewicz, line 15: For my own understanding, is there a case where it's "better" to use [`shadow$`](https://webdriver.io/docs/api/element/shadow$/) instead of [deep selectors](https://webdriver.io/docs/selectors/#deep-selectors)?  My only knowledge of their differences comes from https://webdriver.io/docs/selectors/#deep-selectors:

> [shadow$](https://webdriver.io/docs/api/element/shadow$) and [shadow$$](https://webdriver.io/docs/api/element/shadow$$) have been such workarounds that had their [limitations](https://github.com/Georgegriff/query-selector-shadow-dom#how-is-this-different-to-shadow). With the deep selector you can now query all elements within any shadow DOM using the common query command.

alicewriteswrongs (approved):

test/wdio/reflect-single-render/cmp.test.tsx
christian-bromann: The major difference is that:

- `deep selector`: only supports CSS queries but is able to pierce through all nested shadow DOMs
- `shadow$` and `shadow$$` supports other types of WebDriver selector types, e.g. xpath but can only pierce into the shadow DOM of one element at a time

In this use case I only had to query within the shadow DOM of one element and no further which made me choose this command.

rwaskiewicz (approved):

PR #5540, opened 2024-03-20
chore(wdio): migrate reflect-nan-attribute-with-child to wdio

rwaskiewicz:

GitHub Issue Number: N/A

## What is the new behavior?

port the `test/karma/test-app/reflect-nan-attribute-with-child` test suite to webdriverio

- [ ] Yes
- [x] No

## Testing

## Other information
STENCIL-1191


test/wdio/reflect-nan-attribute-with-child/cmp.test.tsx
```diff
@@ -0,0 +1,21 @@
...
+  beforeEach(() => {
+    render({
+      template: () => <parent-reflect-nan-attribute></parent-reflect-nan-attribute>,
+    });
+  });
+
+  it('renders the parent and child the correct number of times', async () => {
+    await $('parent-reflect-nan-attribute').waitForExist();
+
+    await expect($('parent-reflect-nan-attribute').$('>>>div:first-of-type').$('div:first-of-type')).toHaveText(
```
rwaskiewicz, line 14: Doing something like:
```tsx
    await expect($('parent-reflect-nan-attribute')).toHaveText("...");
```
Gave me the text of both the parent and child component here, which was a bit unexpected, as I thought I'd have to pierce the shadow dom to get the text of the child. Maybe that's not the case?

Anyway, that's why these selectors are a bit longer - happy to change them if there's a better way!

alicewriteswrongs (approved):
lgtm!

christian-bromann (approved):
LGTM 👍

Issue #5514, opened 2024-03-18
feat: Allow to specify in CustomOuputTarget if the output have to be re-executed on each build


Sukaato:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already include this feature request, without success.

### Describe the Feature Request

I'd like to have a parameter that lets you choose whether the custom output target can be called at every change or only during the "final" build.

### Describe the Use Case

I've created a little plugin to retrieve all `@config` docsTags in a `component-config.d.ts` so it's easyer to customise the default state of a component.

stencil.config.ts
```ts
export const config: Config = {
  ... // Config
  outputTargets: [
    {
      type: 'custom',
      name: 'component-option',
     execType: 'run', // new prop in CustomOutputTarget
     async generator(config, compilerCtx: any, buildCtx, docs): Promise<void> {
       // Logique to create a .d.ts file
     }
    },
  ]
}


```
component-config.d.ts
```ts
export interface ComponentOptions {
  "pop-drawer"?: {
    triggerAction?: "click" | "context-menu" | "hover";
    side?: "end" | "start";
    open?: boolean;
  }
}

```
These options are extracted from the component, but if a modification is made to this component, the output target is not updated.

### Describe Preferred Solution

```ts
export const config: Config = {
  ... // Config
  outputTargets: [
    {
      type: 'custom',
      name: 'component-option',
     execType: 'run' | 'build', // new prop in CustomOutputTarget
     async generator(config, compilerCtx: any, buildCtx, docs): Promise<void> { }
    },
  ]
}

// 'run' : run on every change
// 'build' : run only on "final" build
```

### Related Code

https://github.com/Sukaato/stencil-custom-target-example

christian-bromann:
> I'd like to have a parameter that lets you choose whether the custom output target can be called at every change or only during the "final" build.

Can you elaborate on this? When you say "at every change" are you talking about having the output target run during watch mode? Because this works for me given your example project. I guess, could you explain in detail what the current behavior is and your expected/desired behavior?

Sukaato:
Yes when a mean at every change, i mean in dev mode.

[...]

PR #5541, opened 2024-03-20
fix(compiler): Allow OutputTargetCustom to be called on devMode

Sukaato:

GitHub Issue Number: #5514

## What is the new behavior?

In this PR, I've added logic so that OutputTargetCustom can be called in dev mode and not just during build.
OutputTargetCustom can choose when it runs by adding a `task` property to the declaration object.

- [ ] Yes
- [x] No

## Testing

Run stencil tests + manual test with npm link

### DIFF
--- a/src/declarations/stencil-public-compiler.ts
+++ b/src/declarations/stencil-public-compiler.ts
@@ -2241,12 +2241,14 @@ export interface OutputTargetCustom extends OutputTargetBase {
   type: 'custom';
   name: string;
   /**
-   * Indicate in wich mode the output target has to be executed.
-   * By default if nothing is specified it run for both.
-   * - `"run"`: Executed in dev mode on every change in your code
-   * - `"build"`: Executed only on build
+   * Indicate when the output target should be executed.
+   * 
+   * - `"onBuildOnly"`: Executed only when `stencil build` is called without `--watch`.
+   * - `"always"`: Executed on every build, including in `watch` mode.
+   *
+   * Defaults to "always".
    */
-  task?: 'run' | 'build';
+  taskShouldRun?: 'onBuildOnly' | 'always';
   validate?: (config: Config, diagnostics: Diagnostic[]) => void;
   generator: (config: Config, compilerCtx: CompilerCtx, buildCtx: BuildCtx, docs: JsonDocs) => Promise<void>;
   copy?: CopyTask[];
--- a/src/declarations/stencil-public-compiler.ts
+++ b/src/declarations/stencil-public-compiler.ts
@@ -1,5 +1,6 @@
 import type { ConfigFlags } from '../cli/config-flags';
 import type { PrerenderUrlResults, PrintLine } from '../internal';
+import type { BuildCtx, CompilerCtx } from './stencil-private';
 import type { JsonDocs } from './stencil-public-docs';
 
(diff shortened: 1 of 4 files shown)


christian-bromann (approved):
The change looks good to me 👍 

I will have someone else from the @ionic-team/stencil team review this as well.

src/declarations/stencil-public-compiler.ts
```diff
@@ -2233,8 +2234,15 @@ export interface OutputTargetHydrate extends OutputTargetBase {
 export interface OutputTargetCustom extends OutputTargetBase {
   type: 'custom';
   name: string;
+  /**
+   * Indicate in wich mode the output target has to be executed.
+   * By default if nothing is specified it run for both.
+   * - `"run"`: Executed in dev mode on every change in your code
+   * - `"build"`: Executed only on build
+   */
+  task?: 'run' | 'build';
```
christian-bromann, line 2243: I don't see any docs for custom output targets so having the additional TS doc comment seems sufficient at this point.

alicewriteswrongs, line 2243: For the comment, just a few wording things, and I'm wondering if the variants here and the configuration name should change? Just brainstorming but I'm thinking something like this:

```suggestion
  /**
   * Indicate when the output target should be executed.
   * 
   * - `"onBuildOnly"`: Executed only when `stencil build` is called without `--watch`.
   * - `"always"`: Executed on every build, including in `watch` mode.
   *
   * Defaults to "always".
   */
  taskShouldRun?: 'onBuildOnly' | 'always';
```

alicewriteswrongs (changes requested):
code itself looks good and I think this change makes sense! Just had a question about the configuration naming

Sukaato:
How can I find out if the compiler is in build mode to retrieve only tasks that have `taskShouldRun: 'onBuildOnly'

christian-bromann:
@Sukaato the [`BuildOptions`](https://github.com/ionic-team/stencil/blob/641c07a7bddfa63ba521fd157b58a1d284a6badc/scripts/utils/options.ts#L200) have a `isWatch` flag that can help you to determine whether we are in watch mode or not. So if the flag is set you are only call it if the user has set `taskShouldRun` to `always`, otherwise only if `onBuildOnly` is set.

Sukaato:
Correction done

Sukaato:
any update @alicewriteswrongs ?

alicewriteswrongs (approved):

christian-bromann:
@Sukaato there seems to be a linter issue in `src/declarations/stencil-public-compiler.ts`, mind running `npm run prettier`?

christian-bromann:
@Sukaato seems like still some lint issues:

```
> npm run prettier.base -- --list-different


> @stencil/core@4.1[7](https://github.com/ionic-team/stencil/actions/runs/8913029708/job/24479044933?pr=5541#step:5:8).2 prettier.base
> prettier --cache "./({bin,scripts,src,test}/**/*.{ts,tsx,js,jsx})|bin/stencil|.github/(**/)?*.(yml|yaml)|*.js" --list-different

src/declarations/stencil-public-compiler.ts
```

alicewriteswrongs (approved):

PR #5542, opened 2024-03-20
chore(wdio): port global-script test suite to wdio

alicewriteswrongs:
This one is more complicated than some of our other test suites, mainly because the Karma test in `test/karma/test-app/global-script` was actually testing a separate application in `test/karma/test-global-script` which was built as a separate project, in order that a separate build incorporating a [global
script](https://stenciljs.com/docs/config#globalscript) could be tested.

This retains this as a new separate project but keeps the same 'feel' as our existing WebdriverIO tests by:

- locating the component and the spec file in a subdirectory of `test/wdio`, like our other tests
- adding a new build command in `test/wdio`, `build.global-script`, which build just the project in `test/wdio/global-script` with a separate stencil configuration (`test/wdio/global-script.stencil.config.ts`)
  - this special separate build both excludes the existing Stencil components and is, in turn, excluded from the existing Stencil project
- the `www` output target bundle is then selectively imported into WebdriverIO tests (via `test/wdio/setup.ts`) only when we're running the `global-script.test.tsx` test file

Overall this gives us the ability to test different Stencil configuration options using the same setup and with a bit less complexity than we had in the original Karma setup. This can serve as a template for other Karma tests where we'll need a similar setup.

GitHub Issue Number: N/A

- [ ] Yes
- [ ] No


test/wdio/global-script.stencil.config.ts
alicewriteswrongs, line 1: This is a Stencil configuration which is specific just to `test/wdio/global-script`, allowing us to test setting `globalScript: 'global-script/global.ts'` in isolation from the rest of the WebdriverIO test application

test/wdio/global-script/global.ts
alicewriteswrongs, line 1: The test is simple, but it basically sets a value on `window.__testStart` and then we check that we can read that value in the test

test/wdio/global-script/test-cmp.tsx
```diff
@@ -1,7 +1,7 @@
 import { Component, h } from '@stencil/core';
 
 @Component({
-  tag: 'test-cmp',
+  tag: 'global-script-test-cmp',
```
alicewriteswrongs, line 4: I changed the tag to avoid any collisions

test/wdio/package.json
```diff
@@ -3,7 +3,9 @@
   "type": "module",
   "version": "0.0.0",
   "scripts": {
-    "build": "node ../../bin/stencil build --debug --es5",
+    "build": "run-s build.main build.global-script",
+    "build.main": "node ../../bin/stencil build --debug --es5",
+    "build.global-script": "node ../../bin/stencil build --debug --es5 --config global-script.stencil.config.ts",
```
alicewriteswrongs, line 8: expanded the `build` script here by moving the existing build to `build.main` and then adding a new one `build.global-script` - I think this can be a decent pattern to follow for the other similar tests

test/wdio/setup.ts
```diff
@@ -2,3 +2,18 @@
...
+    __wdioSpec__: string;
+  }
+}
+
+/**
+ * load the separate test app for the global script tests, if appropriate to do so
+ */
+if (window.__wdioSpec__.includes('global-script.test.tsx')) {
+  await import('./www-global-script/build/testglobalscript.esm.js');
+}
```
alicewriteswrongs, line 19: this ensures that the `global-script/`-specific bundle isn't loaded when we're running other tests

test/wdio/tsconfig-global-script.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "./tsconfig-stencil.json",
+  "include": [
+    "global-script"
+  ],
+  "exclude": [
+    "global-script/*.test.tsx"
+  ]
+}
```
alicewriteswrongs, line 9: A minimal tsconfig just for the global scripts - note that this is the one which is specified in the `global-script`-specific Stencil configuration file

test/wdio/tsconfig-stencil.json
```diff
@@ -35,6 +35,8 @@
     "./global.ts"
   ],
   "exclude": [
-    "./**/*.test.tsx"
+    "./**/*.test.tsx",
+    "./global-script/**/*.tsx",
+    "./global-script/**/*.ts"
```
alicewriteswrongs, line 40: This is just to ensure that `global-script` stuff isn't included in the existing project

alicewriteswrongs (commented):
Noting a few things!

test/wdio/setup.ts
```diff
@@ -2,3 +2,18 @@
  * load the testapp so that we don't have to import the components within the tests
  */
 await import('./dist/testapp/testapp.esm.js');
+
+// this is well justified!
```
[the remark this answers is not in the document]
alicewriteswrongs: sounds good!

test/wdio/package.json
```diff
@@ -3,7 +3,9 @@
   "type": "module",
   "version": "0.0.0",
   "scripts": {
-    "build": "node ../../bin/stencil build --debug --es5",
+    "build": "run-s build.main build.global-script",
```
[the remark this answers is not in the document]
alicewriteswrongs: I thought about it, I don't think that the two builds will conflict or anything, but I thought that maybe the memory usage might be a little high on the gh actions runners so I went with sequential

just pushed a change to use `run-p` to see what it looks like

test/wdio/tsconfig-stencil.json
rwaskiewicz: Can we turn this into a comment that lives above this configuration? That way folks who modify this file next can get a little additional context when making changes

test/wdio/package.json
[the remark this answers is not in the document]
alicewriteswrongs: you can see an example of the output here: https://github.com/ionic-team/stencil/actions/runs/8363897946/job/22898026360?pr=5542

I think I might want to stick with `run-s` here, not because of a performance issue but just I think the interleaved build output is somewhat confusing to look at

test/wdio/setup.ts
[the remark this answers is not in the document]
rwaskiewicz: @christian-bromann When you raise that issue, can you or @alicewriteswrongs  add a link to it in the comment here?

test/wdio/global-script.stencil.config.ts
```diff
@@ -0,0 +1,19 @@
...
+export const config: Config = {
+  namespace: 'TestGlobalScript',
+  tsconfig: 'tsconfig-global-script.json',
+  outputTargets: [
+    {
+      type: 'www',
+      dir: `www-global-script`,
+    },
+  ],
+  plugins: [sass()],
```
rwaskiewicz, line 16: Do we need the `sass()` plugin here? I don't see any stylesheets here, but I could be missing some other reason for this 🤔

test/wdio/tsconfig-global-script.json
rwaskiewicz: For my own understanding, why not copy the existing one that's pre-configured/existing? There are some niceties like `alwaysStrict` in there already, why not just take what we have and iterate on it at a later point?

rwaskiewicz (changes requested):

test/wdio/setup.ts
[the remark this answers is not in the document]
christian-bromann: see https://github.com/webdriverio/webdriverio/issues/12535

test/wdio/tsconfig-global-script.json
alicewriteswrongs: I'm not sure I follow - this one inherits from `tsconfig-stenci.json` so it will get all the same compiler options

test/wdio/tsconfig-stencil.json
alicewriteswrongs: sure, makes sense

test/wdio/global-script.stencil.config.ts
alicewriteswrongs: ah! yeah we don't need it, I think in an earlier version of the diff I got an error without it but I must have since removed whatever was causing an error

rwaskiewicz (approved):

PR #5543, opened 2024-03-20
fix(mock-doc): re-arrange inheritance tree to fix type issue

christian-bromann:

We see a breaking nightly build in https://github.com/ionic-team/ionic-framework/actions/runs/8338316083/job/22818424122 which happened due to the fact that I initially had to set `localName` different for Document and Element types. This fails because Document inherits from Element (wrongly IMHO) and so TypeScript complains because it overwrites `localName` in Document to a property that returns `undefined` when the parent class `Element` expects a `string` return type.

## What is the new behavior?

I shuffled around the inheritance structure a bit to make this work. Document should inherit from `MockNode` as a super class and both `Document` and `Element` should inherit from there.

## Documentation

n/a

- [ ] Yes
- [x] No

I expect no breaking changes technically but can't guarantee.

## Testing

I will make a dev build an re-run this in framework to double check.

### DIFF
--- a/src/mock-doc/document.ts
+++ b/src/mock-doc/document.ts
@@ -5,11 +5,10 @@ import { MockDocumentFragment } from './document-fragment';
 import { MockDocumentTypeNode } from './document-type-node';
 import { createElement, createElementNS, MockBaseElement } from './element';
 import { resetEventListeners } from './event';
-import { MockElement, MockHTMLElement, MockTextNode, resetElement } from './node';
+import { MockElement, MockHTMLElement, MockNode, MockTextNode, resetElement } from './node';
 import { parseHtmlToFragment } from './parse-html';
 import { parseDocumentUtil } from './parse-util';
 import { MockWindow } from './window';
-import { MockNode } from './node';
 
 export class MockDocument extends MockNode {
   defaultView: any;
--- a/src/mock-doc/document.ts
+++ b/src/mock-doc/document.ts
@@ -9,16 +9,15 @@ import { MockElement, MockHTMLElement, MockTextNode, resetElement } from './node
 import { parseHtmlToFragment } from './parse-html';
 import { parseDocumentUtil } from './parse-util';
 import { MockWindow } from './window';
+import { MockNode } from './node';
 
-export class MockDocument extends MockHTMLElement {
+export class MockDocument extends MockNode {
   defaultView: any;
   cookie: string;
   referrer: string;
 
   constructor(html: string | boolean = null, win: any = null) {
-    super(null, null);
-    this.nodeName = NODE_NAMES.DOCUMENT_NODE;
-    this.nodeType = NODE_TYPES.DOCUMENT_NODE;
+    super(null, NODE_TYPES.DOCUMENT_NODE, NODE_NAMES.DOCUMENT_NODE, null);
     this.defaultView = win;
     this.cookie = '';
     this.referrer = '';
@@ -42,15 +41,11 @@ export class MockDocument extends MockHTMLElement {
     }
   }
 
-  override get dir() {
+  get dir() {
     return this.documentElement.dir;
   }
-  override set dir(value: string) {
+  set dir(value: string) {
     this.documentElement.dir = value;
-  }
-
-  override get localName(): undefined {
-    return undefined;
   }
 
   get location() {
@@ -225,14 +220,14 @@ export class MockDocument extends MockHTMLElement {
     return getElementsByName(this, elmName.toLowerCase());
   }
 
-  override get title() {
+  get title() {
     const title = this.head.childNodes.find((elm) => elm.nodeName === 'TITLE') as MockElement;
     if (title != null && typeof title.textContent === 'string') {
       return title.textContent.trim();
     }
     return '';
   }
-  override set title(value: string) {
+  set title(value: string) {
     const head = this.head;
     let title = head.childNodes.find((elm) => elm.nodeName === 'TITLE') as MockElement;
     if (title == null) {
@@ -297,7 +292,7 @@ const DOC_KEY_KEEPERS = new Set([
   '_shadowRoot',
 ]);
 
-export function getElementById(elm: MockElement, id: string): MockElement {
+export function getElementById(elm: MockElement | MockDocument, id: string): MockElement {
   const children = elm.children;
   for (let i = 0, ii = children.length; i < ii; i++) {
     const childElm = children[i];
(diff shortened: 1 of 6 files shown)


rwaskiewicz (changes requested):
While I agree this inheritance model as currently written doesn't seem to align well with the standard, I don't think we can change a class hierarchy like this without it being a breaking change - we end up removing all methods that are provided by `MockHTMLElement` & `MockNode` out of instances of `MockDocument` here.

I propose we take one of two different approaches here - 

## Option 1 - Throw for `localName`

We can change `MockDocument`'s implementation of `localName` to `throw` instead of return `undefined`. Since `localName` doesn't/shouldn't actually exist on real `Document` instances, we can appease the type checker by "returning a `never` top-type" here:
```diff
diff --git a/src/mock-doc/document.ts b/src/mock-doc/document.ts
index e97c1bd3b..a5ada8570 100644
--- a/src/mock-doc/document.ts
+++ b/src/mock-doc/document.ts
@@ -49,8 +49,8 @@ export class MockDocument extends MockHTMLElement {
     this.documentElement.dir = value;
   }
 
-  override get localName(): undefined {
-    return undefined;
+  override get localName(): never {
+    throw new Error("Unimplemented");
   }
 
   get location() {

```


## Option 2 - Widen `localName` return type

This one's a little more disingenuous, if not a little deceiving. In it, we can widen the `MockElement` return type to `string | undefined`. This would equally appease the type checker, and aligns with calling a getter that doesn't exist on an object:


```diff
diff --git a/src/mock-doc/node.ts b/src/mock-doc/node.ts
index 6c775831f..2f000d0ad 100644
--- a/src/mock-doc/node.ts
+++ b/src/mock-doc/node.ts
@@ -266,12 +266,22 @@ Testing components with ElementInternals is fully supported in e2e tests.`,
     );
   }
 
-  get localName() {
+
+  // TODO(STENCIL-ABC): Refactor the inheritance tree
+  /**
+   * Implementation of the {@link https://developer.mozilla.org/en-US/docs/Web/API/Element/localName localName property}.
+   *
+   * Due to the way MockDoc is constructed, we widen the return type of this function to include `undefined`.
+   * This is not standards compliant, and will be fixed in a major version release of Stencil.
+   *
+   * @returns the node name, represented as a string
+   * @throws a new Error is thrown if an instance's `nodeName` property is not set.
+   */
+  get localName(): string | undefined {
     /**
      * The `localName` of an element should be always given, however the way
      * MockDoc is constructed, it won't allow us to guarantee that. Let's throw
      * and error we get into the situation where we don't have a `nodeName` set.
-     *
      */
     if (!this.nodeName) {
       throw new Error(`Can't compute elements localName without nodeName`);
```

The patches for each can be found below:

- Option 1 - [widen-localname.patch](https://github.com/ionic-team/stencil/files/14743632/widen-localname.patch)
- Option 2 - [throw-localname-doc.patch](https://github.com/ionic-team/stencil/files/14743633/throw-localname-doc.patch)


I tested both options by building Stencil locally with each of the patches below, building the tarball, and installing it into an instance of Ionic Framework Core locally. I haven't run it through the entire Nightly pipeline, which might reveal one (or both) aren't actually tenable. LMK if you have any reservations or simply know these won't work 😅

christian-bromann:
> I don't think we can change a class hierarchy like this without it being a breaking change

@rwaskiewicz, could you please provide further explanation for your statement? In my opinion, I believe we are not removing any features that we advertise to provide. Our documentation does not explicitly mention that Stencil utilizes a JSDOM-like implementation of the DOM API. This absence may lead users to assume that the environment in which the unit tests run adheres strictly to HTML specifications. In fact by keeping the methods attached to the wrong type of objects we might cause false positives, so having this change (while possibly breaking) results in a better state after all. In addition to that, I don't see that much changing anyway, e.g.:

- changing the inheritance for `MockDocument` to `MockNode` may remove functions like `click` or `getAttribute` which unlikely someone used on a document object, and even if, pointing to my argument above: chances are that the user is confusing a MockDocument with an actual element maintaining a possibly false/positive test
- all elements removed from `MockElement` have been attached to `MockNode` which is the class `MockElement` inherits from, so this change will have no impact on the user
- it is unlikely that users use these primitives directly, most of them use the component methods to interact with elements

That said, I am all for carefully evaluating if changes we introduce can cause issues for users updating to latest version of Stencil. In this case however we are changing the interface towards a state that we advertise in our docs anyway. I would argue that fixing certain slot behavior are breaking changes too but we still move forward with them, without introducing a flag for every fix and even as patches, as we orient ourselves to the behavior defined by the Shadow DOM specification. I think this case is no different.

I would prefer to move forward with this. The suggested alternatives punt on the issue at hand and even introduce more divergence between what we actually suggest to provide to user. That said, I am also not dying on a hill for this 😉 so you call!

rwaskiewicz:
@christian-bromann 

Completely understandable! If I'm honest, our documentation hasn't always been great, and we've put more burden on our users to figure things out than we should have in the past. While we've been working to fix that for some time now, we can't guarantee that folks haven't been relying on the existing incorrect behavior for some time. I agree continuing to do this may lead to/potentially compound some of confusion, but I think we should do this in a way that's potentially less invasive for folks.

I can't **prove** that people are relying on this behavior, which is the tricky part here. It potentially makes all our arguments surrounding what is/is not a breaking change a judgement call.

Here, I'd like to err on the side of caution until we've designed reordering these classes a bit more. As is, the current state of this branch removes support for a few things like [`hidden` on `Document`](https://developer.mozilla.org/en-US/docs/Web/API/Document/hidden) (we're relying on [getting that from `MockElement`](https://github.com/ionic-team/stencil/blob/0ca92e8d0ef08075c6ff9423e31fec14d4fa4630/src/mock-doc/node.ts#L541-L550)). 

We can see this with the following patch on a Stencil Component Starter output:

```diff
diff --git a/src/components/my-component/my-component.tsx b/src/components/my-component/my-component.tsx
index 56d51d9..24d66d5 100644
--- a/src/components/my-component/my-component.tsx
+++ b/src/components/my-component/my-component.tsx
@@ -1,5 +1,6 @@
 import { Component, Prop, h } from '@stencil/core';
 import { format } from '../../utils/utils';
+import { MockDocument } from '@stencil/core/mock-doc';
 
 @Component({
   tag: 'my-component',
@@ -23,6 +24,12 @@ export class MyComponent {
   @Prop() last: string;
 
   private getText(): string {
+    const doc = new MockDocument();
+    // hidden is no longer available
+    doc.hidden
+    // neither is attachInternals, although this would be a very
+    // wrong thing for someone to do
+    doc.attachInternals()
     return format(this.first, this.middle, this.last);
   }
```

I'm very open to pulling this into v5 - for the purposes of unblocking the nightly build, I think a simpler solution here is the better way to go

christian-bromann:
@rwaskiewicz sounds good to me, let's label this PR so we don't forget to merge it in v5. In the meantime we can move forward with https://github.com/ionic-team/stencil/pull/5595. A dev release was cut that we can use to test against framework: `4.13.0-dev.1711491223.b404b98`.

PR #5544, opened 2024-03-20
chore(test): migrate `watch-native-attributes` to wdio

tanner-reits:

GitHub Issue Number: N/A

## What is the new behavior?

Migratest `test/karma/test-app/watch-native-attributes` to wdio

- [ ] Yes
- [x] No


test/wdio/watch-native-attributes/cmp.test.tsx
```diff
@@ -0,0 +1,22 @@
...
+  it('triggers the callback for the watched attribute', async () => {
+    await $('watch-native-attributes').waitForExist();
+
+    const cmp = document.querySelector('watch-native-attributes');
+    expect(cmp.innerText).toBe('Label: myStartingLabel\n\nCallback triggered: false');
+
+    cmp.setAttribute('aria-label', 'myNewLabel');
+    await browser.pause(100);
+
+    expect(cmp.innerText).toBe('Label: myNewLabel\n\nCallback triggered: true');
```
christian-bromann, line 20: I would suggest something like this to leverage the async matchers.
```suggestion
    const $cmp = $('watch-native-attributes')
    await $cmp.waitForExist();

    await expect($cmp).toHaveText('Label: myStartingLabel\n\nCallback triggered: false');

    const cmp = document.querySelector('watch-native-attributes');
    cmp.setAttribute('aria-label', 'myNewLabel');

    await expect(cmp).toHaveText('Label: myNewLabel\n\nCallback triggered: true');
```

I looking into ways to easier retrieve the raw DOM element from the `WebdriverIO.Element` object to allow us to just define the element once.

christian-bromann (commented):
One suggestion

test/wdio/watch-native-attributes/cmp.test.tsx
tanner-reits: Is the `$` variable prefix the recommended convention when using async queries? Just curious!

alicewriteswrongs (approved):
lgtm!

test/wdio/watch-native-attributes/cmp.test.tsx
christian-bromann: not really .. but in the context of using `HTMLElement` and `WebdriverIO.Element` objects it seems like a good idea to do but not a convention.

christian-bromann (approved):
👍

PR #5545, opened 2024-03-20
chore(wdio): migrate slot-array-complex to wdio

rwaskiewicz:

GitHub Issue Number: N/A

## What is the new behavior?

port the `test/karma/test-app/slot-array-complex` test suite to webdriverio

- [ ] Yes
- [x] No

## Testing

## Other information

STENCIL-1191


christian-bromann (approved):
LGTM 👍

test/wdio/slot-array-complex/cmp.test.tsx
```diff
@@ -0,0 +1,21 @@
...
+    });
+
+    await $('main slot-array-complex').waitForExist();
+  });
+
+  it('renders slotted content', async () => {
+    await expect($('main slot-array-complex')).toHaveChildren(3);
+    await expect($('main slot-array-complex').$$('*')[0]).toHaveText('slot - start');
+    await expect($('main slot-array-complex').$$('*')[1]).toHaveText('slot - default');
+    await expect($('main slot-array-complex').$$('*')[2]).toHaveText('slot - end');
```
christian-bromann, line 17: This will be a bit more effective:
```suggestion
    const elems = $('main slot-array-complex').$$('*')
    await expect(elems[0]).toHaveText('slot - start');
    await expect(elems[1]).toHaveText('slot - default');
    await expect(elems[2]).toHaveText('slot - end');
```

christian-bromann (approved):
👍

PR #5546, opened 2024-03-20
chore(test): migrate `slot-array-basic` to wdio

tanner-reits:

GitHub Issue Number: N/A

## What is the new behavior?

Migrates `test/karma/test-app/slot-array-basic` to wdio

- [ ] Yes
- [x] No


christian-bromann (approved):
LGTM 👍

rwaskiewicz (changes requested):
Can we re-add the styles in these tests? I did a quick test comparing what we have now, and what we had, and there are some pretty big differences that help differentiate elements (even if we don't assert on them)

Before:
Screenshot 2024-03-21 at 10 56 24 AM

Now:
Screenshot 2024-03-21 at 10 56 57 AM

This is the patch that I think we can apply directly to this branch to get the "before" results:
[css.patch](https://github.com/ionic-team/stencil/files/14697017/css.patch)

rwaskiewicz (approved):

PR #5553, opened 2024-03-21
chore(test): migrate slot-array-top

christian-bromann:
migrate slot-array-top


test/wdio/package.json
```diff
@@ -11,10 +11,10 @@
     "@stencil/core": "file:../..",
     "@stencil/sass": "^3.0.10",
     "@types/node": "^16.11.7",
-    "@wdio/browser-runner": "^8.33.1",
-    "@wdio/cli": "^8.33.1",
-    "@wdio/globals": "^8.33.1",
-    "@wdio/mocha-framework": "^8.33.1",
+    "@wdio/browser-runner": "^8.35.1",
```
rwaskiewicz, line 14: Are the `@wdio` upgrades strictly needed for this PR? If possible, I'd prefer we let Renovatebot handle these - that way, we can separate test implementations from dependency upgrades, which makes bisecting/debugging commits a bit easier (especially as we're all swarming on this area of the codebase right now)

rwaskiewicz (changes requested):
One small ask/question

test/karma/test-app/slot-array-top/index.html
```diff
@@ -1,15 +0,0 @@
-<!DOCTYPE html>
-<meta charset="utf8">
-<script src="./build/testapp.esm.js" type="module"></script>
-<script src="./build/testapp.js" nomodule></script>
-
-<style>
```
rwaskiewicz, line 6: Can we restore these styles as well? I think they could help with some visual debugging, even if we don't assert on them

test/wdio/package.json
christian-bromann: yes, there are some bugs I fixed in WebdriverIO that would break the tests without the update. Alternatively we can merge https://github.com/ionic-team/stencil/pull/5498 and I can just rebase and move forward without dep updates (as the other PR includes them)

alicewriteswrongs (approved):

test/karma/test-app/slot-array-top/index.html
christian-bromann: Migrated the styles.

rwaskiewicz (approved):
Thanks! LGTM!

PR #5554, opened 2024-03-21
chore(test): migrate shadow-dom-slot-nested

christian-bromann:
migrate shadow-dom-slot-nested


test/wdio/shadow-dom-slot-nested/cmp.test.tsx
```diff
@@ -1,19 +1,23 @@
...
+        </main>
+      ),
+    });
   });
-  afterEach(tearDownDom);
 
   it('renders children', async () => {
-    let elm = app.querySelector('main');
+    let elm = document.querySelector('main');
     expect(window.getComputedStyle(elm).color).toBe('rgb(0, 0, 255)');
```
rwaskiewicz, line 18: Wouldn't we need [these styles](https://github.com/ionic-team/stencil/pull/5554/files#diff-5167e5130b3d8467e0db353baf3f708a63407ec2e309d1aec56d58b17f675807L7-L10) applied at the root in order for this assertion to pass? Or is there somewhere else these are set that I'm not seeing?

test/wdio/shadow-dom-slot-nested/cmp.test.tsx
```diff
@@ -1,19 +1,23 @@
...
 
-    elm = app.querySelector('shadow-dom-slot-nested-root');
+    elm = document.querySelector('shadow-dom-slot-nested-root');
     expect(elm.shadowRoot).toBeDefined();
 
     if ('attachShadow' in HTMLElement.prototype) {
```
rwaskiewicz, line 23: Would this always be true for us now? Could we remove the conditional here and the whole `else` block?

rwaskiewicz (changes requested):

test/wdio/shadow-dom-slot-nested/cmp.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+    `);
+
+    const cmp = $('shadow-dom-slot-nested-root');
+    const section = cmp.shadow$('section');
+    await expect(section.getCSSProperty('color')).toMatchInlineSnapshot(`
+      {
+        "parsed": {
+          "alpha": 1,
+          "hex": "#008000",
+          "rgba": "rgba(0,128,0,1)",
```
rwaskiewicz, line 46: It looks like Firefox would prefer this to be `rgb` values instead of `rgba` 😢

christian-bromann: bummer .. updated test to validate against hex values

rwaskiewicz (approved):

PR #5559, opened 2024-03-21
chore(wdio): migrate slot-fallback tests

rwaskiewicz:

GitHub Issue Number: N/A

## What is the new behavior?

port the `test/karma/test-app/slot-fallback` test suite to webdriverio

- [ ] Yes
- [x] No

## Testing

## Other information
STENCIL-1191


test/wdio/slot-fallback/cmp.test.tsx
```diff
@@ -0,0 +1,131 @@
...
+    await expect($('.results1 slot-fb[name="start"]:not([hidden])')).toHaveText('slot start fallback 1');
+    await expect($('.results1 section slot-fb:not([hidden])')).toHaveText('slot default fallback 1');
+    await expect($('.results1 article span slot-fb[name="end"]:not([hidden])')).toHaveText('slot end fallback 1');
+
+    // set light dom instead and hide fallback content
+    await $('button.change-light-dom').click();
+    await browser.pause();
+
+    // fallback content hidden but still the same
+    // webdriverio seems to have some trouble with the `[hidden]` selector, which is why we don't use it here
```
christian-bromann, line 30: Making the comment more clear. WebDriver's `getText` command is defined to only fetch the `visible` text from an element
```suggestion
    // WebdriverIO's `getText` command can only fetch a visible text, which is why we don't use it here
```

test/wdio/slot-fallback/cmp.test.tsx
```diff
@@ -0,0 +1,131 @@
...
+    await expect($('.results1 content-start')).toHaveText('slot light dom 0 : start');
+    await expect($('.results1 section content-default')).toHaveText('slot light dom 0 : default');
+    await expect($('.results1 article span content-end')).toHaveText('slot light dom 0 : end');
+
+    await $('button.change-fallback-content').click();
+    await $('button.change-slot-content').click();
+    await browser.pause();
+
+    // fallback content hidden and updated content
+    // webdriverio seems to have some trouble with the `[hidden]` selector, which is why we don't use it here
```
christian-bromann, line 51:
```suggestion
    // WebdriverIO's `getText` command can only fetch a visible text, which is why we don't use it here
```

test/wdio/slot-fallback/cmp.test.tsx
```diff
@@ -0,0 +1,131 @@
...
+    await expect($('.results1 content-start')).not.toBeExisting();
+    await expect($('.results1 section content-default')).not.toBeExisting();
+    await expect($('.results1 article span content-end')).not.toBeExisting();
+
+    // change back to showing slot content
+    await $('button.change-light-dom').click();
+    await browser.pause();
+
+    // fallback content hidden and updated content
+    // webdriverio seems to have some trouble with the `[hidden]` selector, which is why we don't use it here
```
christian-bromann, line 101:
```suggestion
    // WebdriverIO's `getText` command can only fetch a visible text, which is why we don't use it here
```

christian-bromann (approved):
A suggestion to make the error message more clear, otherwise LGTM 👍

PR #5560, opened 2024-03-21
 chore(wdio): migrate slot-hide-content tests 

rwaskiewicz:

GitHub Issue Number: N/A

## What is the new behavior?

port the `test/karma/test-app/slot-hide-content` test suite to
webdriverio

git history has been preserved as much as possible here, using `git mv`
to move files from the `karma` dir to `wdio/` in all possible cases.
however, git may fail to mark some files as 'moved' if we change a file
enough that it exceeds git's "sameness threshold".

- [ ] Yes
- [x] No

## Testing

## Other information
STENCIL-1191


test/wdio/slot-hide-content/cmp.test.tsx
```diff
@@ -1,31 +1,52 @@
...
+
+    await $('button').waitForExist();
+
+    document.querySelector('button').addEventListener('click', clickHandler);
   });
 
-  afterEach(tearDownDom);
+  afterEach(() => {
+    document.removeEventListener('click', clickHandler);
+  });
```
tanner-reits, line 37: You can remove the need for this if you bind to a function in the `beforeEach` rather than querying and adding the event listener

tanner-reits (approved):

Issue #5458, opened 2024-03-12
[Tracking] WebdriverIO Migration


christian-bromann:
In #5456 we introduced WebdriverIO as testing framework of choice for testing Stencil components. This issue tracks our progress migrating our current existing Karma test suite to WebdriverIO. To migrate a test, follow these instructions:

1. assign a test to you by clicking on the checkbox next to the test you like to migrate
1. move the whole directory to the `wdio` folder, e.g. `git mv test/karma/test-app/conditional-rerender test/wdio/conditional-rerender`
1. rename `karma.spec.ts` into e.g. `cmp.test.tsx` (note the __.tsx__)
1. replace the component test up by using WebdriverIOs `render` function, e.g. 
    ```tsx
    import { h, Fragment } from '@stencil/core';
    import { render } from '@wdio/browser-runner/stencil';

    describe('conditional-rerender', function () {
      beforeEach(() => {
        render({
          template: () => <>...</>
        })
      })
    })
    ```
    Copy everything from the included `.html` file , except doctype, meta and script tags that import the test app:
    ```patch
    - <!DOCTYPE html>
    - <meta charset="utf8">
    - <script src="./build/testapp.esm.js" type="module"></script>
    - <script src="./build/testapp.js" nomodule></script>
    ```
1. run the test in watch mode to modify it, e.g. `cd test/wdio && npm ci && npm run build && npm run wdio -- --spec conditional-rerender --watch`
1. Make updates to the test
    - if you see `querySelector` calls, replace them with WebdriverIO queries, e.g.
       ```patch
       - const main = app.querySelector('main');
       + const main = $('main');
       ```
    - either replace all synchronous assertions with asynchronous ones provided by WebdriverIO or ensure that the component is ready for assertion by adding a `await main.waitForExist()` after every time you call `render` or interact with the component via `click`
    - use the chance to optimize assertions e.g. by consolidating them into a `toMatchInlineSnapshot`
1. Lastly, remove the `.html` file from the directory and other unnecessary files
1. Run the whole test suite from the root directory via: `npm run test.wdio`
1. Build Karma to compile updated types for its testapp `cd test/karma && npm run build.all`
1. Run Prettier to ensure formatting is alright: `npm run prettier`
1. If all passes, raise a PR, move to the next test and start from the beginning.

Tests:

[...]

PR #5561, opened 2024-03-21
chore(wdio): port svg-class test suite to wdio

alicewriteswrongs:
More wdio!

For #5458 

- [ ] Yes
- [x] No


test/wdio/svg-class/cmp.test.tsx
```diff
@@ -1,15 +1,28 @@
...
+          <svg-class></svg-class>
+        </>
+      ),
+    });
   });
-  afterEach(tearDownDom);
 
   it('toggles svg class', async () => {
+    const app = document.body;
+    await $('svg-class').waitForStable();
```
rwaskiewicz, line 25: FWIW, I had some flakiness trying to run this test locally - changing this to
```suggestion
    await $('svg-class').waitForExist();
```
seemed to help.

@christian-bromann is there much a difference between these two in practice? Do they 'wait' for the same amount of time?

test/wdio/svg-class/cmp.test.tsx
```diff
@@ -18,10 +31,7 @@ describe('svg class', () => {
     expect(circle.getAttribute('class')).toBe(null);
     expect(rect.getAttribute('class')).toBe(null);
 
-    const button = app.querySelector('button');
-    button.click();
-
-    await waitForChanges();
+    await $('button').click();
```
rwaskiewicz, line 34: FWIW, I had some flakiness trying to run this test locally - adding the following helped:
```suggestion
    await $('button').click();
    await $('svg-class').waitForStable();
```

test/wdio/svg-class/cmp.test.tsx
christian-bromann: "stable" in this context means whether or not the element moves or not. I think we can just check for `waitForExist` as the location of the element mostly doesn't matter since they aren't animated.

christian-bromann (approved):

test/wdio/svg-class/cmp.test.tsx
alicewriteswrongs: ah ok, I sort of imputed my own definition of `waitForStable` which had to do more with no more children being added or something. Good to know!

rwaskiewicz (approved):

PR #5565, opened 2024-03-21
chore(wdio): port the static-styles test suite to wdio

alicewriteswrongs:
It's a karma -> wdio port, what can I say!

GitHub Issue Number: #5458 

- [ ] Yes
- [x] No

## Testing

If CI is passing we should be good here I think!


tanner-reits (approved):

test/wdio/static-styles/cmp.test.tsx
```diff
@@ -0,0 +1,17 @@
...
+  beforeEach(() => {
+    render({
+      template: () => <static-styles></static-styles>,
+    });
+  });
+
+  it('applies styles from static getter', async () => {
+    await $('static-styles').waitForExist();
+    await expect($('h1')).toHaveStyle({
+      color: browser.isChromium ? 'rgba(255,0,0,1)' : browser.isFirefox ? 'rgb(255,0,0)' : 'rgb(255,0,0)',
```
christian-bromann, line 14: Wouldn't this be sufficient?
```suggestion
      color: browser.isChromium ? 'rgba(255,0,0,1)' : 'rgb(255,0,0)',
```

christian-bromann (approved):
One comment, non blocking

LGTM 👍

test/wdio/static-styles/cmp.test.tsx
alicewriteswrongs: haha yeah that looks like it would be 😅

PR #5569, opened 2024-03-22
chore(test): migrate slot-nested-default-order

christian-bromann:
migrate slot-nested-default-order


test/wdio/slot-nested-default-order/cmp.test.tsx
```diff
@@ -0,0 +1,25 @@
...
+      template: () => (
+        <slot-nested-default-order-parent>
+          <p>Hello</p>
+        </slot-nested-default-order-parent>
+      ),
+    });
+  });
+
+  it('should render the slot content after the div', async () => {
+    const childCmps = $('slot-nested-default-order-parent').shadow$('slot-nested-default-order-child').shadow$$('*');
```
rwaskiewicz, line 16: It doesn't look like either component that's a part of this test uses the shadow dom. For my own understanding, why do we use `shadow$` and `shadow$$` here? 

It looks like this could be shortened to something like:
```ts
    const childCmps = $('slot-nested-default-order-parent slot-nested-default-order-child').$$('*');
```
But, maybe I'm missing something - either way LMK!

christian-bromann: You are right, I was assuming for some reason we would deal with a shadow DOM here, fixed the selector

rwaskiewicz (approved):

alicewriteswrongs (approved):

PR #5572, opened 2024-03-22
chore(wdio): migrate slot-html to wdio

rwaskiewicz:

GitHub Issue Number: N/A

## What is the new behavior?

port the `test/karma/test-app/slot-html` test suite to webdriverio

some of the indices for accessing default slots on non-encapsulated components changed in this commit, specifically for the `.results5`, `.results10`, `.results11`, and `.results12` CSS queries/their related tests. when running under webdriverio, we get differnt values here - however, we're not changing the underlying implementation, so hopefully this is safe!

- [ ] Yes
- [ ] No

## Testing

I made every assertion "red" here, which as a bit painstaking 😆 But it's all green now 🤞 

## Other information
STENCIL-1191


rwaskiewicz:
@christian-bromann It looks like that proposed change causes tests to fail with:
```
[chrome-headless-shell 122.0.6261.128 linux #0-0] Expect $(``) to have text

Expected: "default slot 1"
Received: "default slot 2"
[chrome-headless-shell 122.0.6261.128 linux #0-0] Error: Expect $(``) to have text
```

Do you have any suggestions here? Or should we revert that patch?

test/wdio/slot-html/cmp.test.tsx
```diff
@@ -0,0 +1,186 @@
...
+            <content-end slot="end">end slot 2</content-end>
+            <content-default>default slot 1</content-default>
+            <content-default>default slot 2</content-default>
+            <content-start slot="start">start slot 2</content-start>
+          </slot-html>
+        </>
+      ),
+    });
+
+    await $('.results1').waitForExist();
```
christian-bromann, line 106: Since our first assertion is an async one we don't need this. WebdriverIO validates element existence as part of the text assertion
```suggestion
```

test/wdio/slot-html/cmp.test.tsx
```diff
@@ -0,0 +1,186 @@
...
+    expect(results11.childNodes[6].childNodes[2].textContent.trim()).toBe('end slot 2');
+
+    const results12 = document.querySelector('.results12 div');
+    expect(results12.childNodes[1].childNodes[0].childNodes[1].textContent.trim()).toBe('start slot 1');
+    expect(results12.childNodes[1].childNodes[0].childNodes[2].textContent.trim()).toBe('start slot 2');
+    expect(results12.childNodes[3].textContent.trim()).toBe('default slot text node');
+    expect(results12.childNodes[4].textContent.trim()).toBe('default slot 1');
+    expect(results12.childNodes[5].textContent.trim()).toBe('default slot 2');
+    expect(results12.childNodes[6].childNodes[1].textContent.trim()).toBe('end slot 1');
+    expect(results12.childNodes[6].childNodes[2].textContent.trim()).toBe('end slot 2');
```
christian-bromann, line 182: How about this (imho more simplified) approach of validating what text content is visible. It removes the need of jumping through child nodes and makes the test more readable. WebDriver's `getText` command ensures that only visible is being returned which is I think exactly what we are trying to validate here: checking the order and visibility of child nodes.

```suggestion
    await expect($('.results2 div')).toHaveText(
      'default slot text node'
    )

    await expect($('.results3 div')).toHaveText(
      'default slot element 1default slot element 2default slot element 3'
    )

    await expect($('.results4 div')).toHaveText(
      'start slot 1start slot 2\n' +
      'default slot'
    )

    await expect($('.results5 div')).toHaveText(
      'start slot 1start slot 2\n' +
      'default text node'
    )

    await expect($('.results6 div')).toHaveText(
      'start slot 1start slot 2\n' +
      'default slot 1default slot 2'
    )

    await expect($('.results7 div')).toHaveText(
      'start slot 1start slot 2\n' +
      'default slot 1default slot 2'
    )

    await expect($('.results8 div')).toHaveText(
      'end slot 1end slot 2'
    )

    await expect($('.results9 div')).toHaveText(
      'default slot 1default slot 2\n' +
      'end slot 1end slot 2'
    )

    await expect($('.results10 div')).toHaveText(
      'default slot 1default slot 2default slot text node\n' +
      'end slot 1end slot 2'
    )

    await expect($('.results11 div')).toHaveText(
      'start slot 1start slot 2\n' +
      'default slot 1default slot 2default slot text node\n' +
      'end slot 1end slot 2'
    )

    await expect($('.results12 div')).toHaveText(
      'start slot 1start slot 2\n' +
      'default slot text nodedefault slot 1default slot 2\n' +
      'end slot 1end slot 2'
    )
```

rwaskiewicz: > How about this (imho more simplified) approach of validating what text content is visible. It removes the need of jumping through child nodes and makes the test more readable. WebDriver's getText command ensures that only visible is being returned which is I think exactly what we are trying to validate here: checking the order and visibility of child nodes.

I think the hierarchy is also important here too though - since we're emulating slots in a non-shadow dom environment, if I understand this test correctly, it's also important that we assert that the text nodes are children of the correct parents here

rwaskiewicz:
@christian-bromann I ended up reverting https://github.com/ionic-team/stencil/pull/5572/commits/a4966b570ba1a0e559bc18029cdc2cd0e842ab79, as it caused some test instability. I responded to https://github.com/ionic-team/stencil/pull/5572#discussion_r1536131604 as well - PTAL when you get a chance and LMK what you think!

christian-bromann (approved):
I still see usage of `await` for synchronous matchers, e.g. `await expect(results10Children[3].textContent.trim()).toBe('default slot 1');` .. while it is technically not needed it doesn't change much about the outcome, so I don't want this to be blocking our progress.

LGTM 👍

rwaskiewicz:
@christian-bromann Removed those unneeded `await`s - Thanks!

PR #5573, opened 2024-03-22
chore(wdio): migrate slot-no-default to wdio

rwaskiewicz:

GitHub Issue Number: N/A

## What is the new behavior?

port the `test/karma/test-app/slot-no-default` test suite to webdriverio

- [ ] Yes
- [x] No

## Testing

All tests were made "red" to intentionally fail them before fixing them and making them "green"

## Other information
STENCIL-1191


tanner-reits (approved):

test/wdio/slot-no-default/cmp.test.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+    await $('slot-no-default').waitForExist();
+  });
+
+  it('only renders slots that have a location', async () => {
+    await expect(await $('slot-no-default a').getAttribute('hidden')).toBe(null);
+    await expect(await $('slot-no-default header').getAttribute('hidden')).toBe('true');
+    await expect(await $('slot-no-default main').getAttribute('hidden')).toBe('true');
+    await expect(await $('slot-no-default nav').getAttribute('hidden')).toBe(null);
+    await expect(await $('slot-no-default footer').getAttribute('hidden')).toBe(null);
```
christian-bromann, line 26:
```suggestion
    await expect($('slot-no-default a')).not.toHaveAttribute('hidden');
    await expect($('slot-no-default header').toHaveAttribute('hidden', 'true');
    await expect($('slot-no-default main').toHaveAttribute('hidden', 'true');
    await expect($('slot-no-default nav').not.toHaveAttribute('hidden');
    await expect($('slot-no-default footer').not.toHaveAttribute('hidden');
```

christian-bromann (changes requested):

christian-bromann (approved):

PR #5582, opened 2024-03-22
chore(test): migrate `slot-replace-wrapper` to wdio

tanner-reits:

GitHub Issue Number: N/A

## What is the new behavior?

Migrates `test/karma/test-app/slot-replace-wrapper` to wdio

- [ ] Yes
- [x] No


test/wdio/slot-replace-wrapper/cmp.test.tsx
```diff
@@ -1,79 +1,81 @@
...
-  let app: HTMLElement;
-
   beforeEach(async () => {
-    app = await setupDom('/slot-replace-wrapper/index.html');
+    render({
+      template: () => <slot-replace-wrapper-root></slot-replace-wrapper-root>,
+    });
+
+    await $('slot-replace-wrapper-root').waitForExist();
+    await browser.pause(100);
```
christian-bromann, line 11: I would suggest this, it's more explicit towards what we are actually waiting for. Using `pause` command is rather discouraged as it is vulnerable to race conditions.

```suggestion
    await $('.results1 a').waitForExist();
```

christian-bromann (approved):
One suggestion, I don't have a strong opinion on using `pause` or not, it is not very likely that it causes any side effects since 100ms are enough for the browser to settle.

rwaskiewicz (approved):

PR #5590, opened 2024-03-25
chore(wdio): migrate stencil-sibling test suite to wdio

alicewriteswrongs:
This is another slightly more complicated test suite sort of like the `global-scripts` one ported in #5542.

In this one we have a 'sibling' Stencil project which was formerly located in `test/karma/test-sibling`. That separate project is now copied over to `test/wdio/test-sibling` with minimal changes, and is built separately from the main wdio project. The main wdio project then imports it in `global.ts` and the spec file in `test/wdio/stencil-sibling` tests that a component defined in the `test-sibling` project can be successfully built.

for #5458 

- [ ] Yes
- [x] No


test/wdio/stencil-sibling/cmp.test.tsx
```diff
@@ -0,0 +1,25 @@
...
+describe('stencil-sibling', function () {
+  beforeEach(() => {
+    render({
+      template: () => <stencil-sibling></stencil-sibling>,
+    });
+  });
+
+  it('loads sibling root', async () => {
+    const stencilSibling = await $('stencil-sibling');
+    await stencilSibling.waitForStable();
```
rwaskiewicz, line 13: Non-blocking - does this test animate? Could this be
```suggestion
    await stencilSibling.waitForExist();
```

rwaskiewicz (changes requested):
Looks good - only blocking comment is that `test/wdio/test-sibling/stencil.config.ts` has an invalid import path for `Config` after the move

test/wdio/stencil-sibling/cmp.test.tsx
alicewriteswrongs: oh yup, looks like that works 👍

rwaskiewicz (approved):

christian-bromann (approved):

PR #5594, opened 2024-03-26
feat(compiler): perform automatic key insertion in more situations

alicewriteswrongs:
This expands the scope of the automatic key insertion feature that was added in #5143. Previously the feature stopped at the border of any `JsxExpression` syntax node (a bit of arbitrary JavaScript wrapped in curly braces and then inserted into JSX) so that if you were doing something like:

```tsx
<div>{ someCondition && <span>when condition is true</span>}</div>
```

the compiler would insert a key into the `<div>` syntax tree node but _not_ the node for the `<span>`. We did this to just be a bit cautious with the feature because there are a lot of different things that could be going on within a `JsxExpression` node, and some things which could be written within curly braces which would not be safe to transform, such as the following:

```tsx
<div>{ xs.map(x => <span>{ x }</span>) }</div>
```

We wouldn't want to insert a key into the `<span>` syntax tree node because that would result in the dynamically generated `<span>` tags always having the same key attributes.

That said, there are some situations where it is safe to insert a key, such as the `condition || <some-tag />` case shown above. This change adds support for inserting keys in those situations.

STENCIL-1120

GitHub Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/src/compiler/transformers/automatic-key-insertion/index.ts
+++ b/src/compiler/transformers/automatic-key-insertion/index.ts
@@ -71,11 +71,33 @@ export const performAutomaticKeyInsertion = (transformCtx: ts.TransformationCont
    * @returns the result of handling the node
    */
   function findRenderMethodVisitor(node: ts.Node): ts.VisitResult<ts.Node> {
-    // we want to keep going (to drill down into JSX nodes and transform them) only in particular circumstances:
+    // we want to keep going (to drill down into JSX nodes and transform them)
+    // only in particular circumstances:
     //
     // 1. the syntax tree node is a method declaration
     // 2. this method's name is 'render'
     // 3. the method only has a single return statement
+    //
+    // We want to only keep going if there's a single return statement because
+    // if there are multiple return statements inserting keys could cause
+    // needless re-renders. If a `render` method looked like this, for
+    // instance:
+    //
+    // ```tsx
+    // render() {
+    //   if (foo) {
+    //     return <div>hey!</div>;
+    //   } else {
+    //     return <div>hay!</div>;
+    //   }
+    // }
+    // ```
+    //
+    // Since the `<div>` tags don't have `key` attributes the Stencil vdom will
+    // re-use the same div element between re-renders, and will just swap out
+    // the children (the text nodes in this case). If our key insertion
+    // transformer put unique keys onto each tag then this wouldn't happen any
+    // longer.
     if (ts.isMethodDeclaration(node) && node.name.getText() === 'render' && numReturnStatements(node) === 1) {
       return ts.visitEachChild(node, jsxElementVisitor, transformCtx);
     } else {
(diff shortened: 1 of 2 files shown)


christian-bromann (approved):

src/compiler/transformers/automatic-key-insertion/index.ts
```diff
@@ -91,16 +91,10 @@ export const performAutomaticKeyInsertion = (transformCtx: ts.TransformationCont
...
-      // (arbitrary JS code delimited by curly braces in JSX) in order to
-      // determine whether it's safe to add keys to JSX nodes within it, so we
-      // bail here instead.
-      return node;
-    } else if (ts.isConditionalExpression(node)) {
-      // we're going to encounter the same problem here that we encounter with
-      // multiple return statements, so we don't try to transform the arms of
-      // the conditional.
+    if (ts.isCallExpression(node) || ts.isPropertyAccessExpression(node) || ts.isConditionalExpression(node)) {
+      // we're going to encounter the same problems here that we encounter with
```
rwaskiewicz, line 95: Since we've change this conditional to remove `isJsxExpression` and add `isCallExpression`+`isPropertyAccessExpression`, I think it might make sense to update this comment. I remember you doing this original work, but I think some of the context of this comment is a little lost on me now (TBH, I don't remember what the multiple return statement problem is 😆).

rwaskiewicz (changes requested):
All in all, looks/works good! Just a couple minor asks

src/compiler/transformers/automatic-key-insertion/index.ts
alicewriteswrongs: yeah, I think that actually the `isPropertyAccessExpression` check isn't necessary overall, I was thinking of situations like

```ts
<div>{ foo.bar(<div />) }</div>
```

where we might not want to transform the `<div />` because we don't know what will happen to it (`.bar` could do anything!) but actually in this case the whole expression `foo.bar(<div />)` is a `CallExpression` where the callee is a `PropertyAccessExpression`. So just checking for `isPropertyAccessExpression` would only guard against situations like

```ts
<div>{ foo.bar }</div>
```

where the property access expression can't have any children that are JSX nodes anyway...

so anyway yeah I think I will remove it and then fix up the comments

alicewriteswrongs: I updated the code here and also fleshed out a comment in the `findRenderMethodVisitor` helper function to explain a bit more the 'multiple return' problem, lmk what you think!

src/compiler/transformers/automatic-key-insertion/index.ts
```diff
@@ -71,11 +71,32 @@ export const performAutomaticKeyInsertion = (transformCtx: ts.TransformationCont
...
+    //   if (foo) {
+    //     return <div>hey!</div>;
+    //   } else {
+    //     return <div>hay!</div>;
+    //   }
+    // }
+    // ```
+    //
+    // Since the `<div>` tags don't have `key` attributes the Stencil vdom will
+    // re-used the same div element between re-renders, and will just swap out
```
rwaskiewicz, line 97:
```suggestion
    // re-use the same div element between re-renders, and will just swap out
```

rwaskiewicz (approved):
LGTM - one non-blocking nit. Comments make sense 🙂

rwaskiewicz:
Actually, @alicewriteswrongs can we hold off on merging this for today? I wanna see if I can get a patch release out to unblock the Lisbon team

rwaskiewicz:
v4.14.1 went out this morning, once this has another ✔️ it should be good to merge. Thanks for holding off on the merge!

PR #5597, opened 2024-03-27
chore(test): migrate `invisible-prehydration-false` to wdio

tanner-reits:

GitHub Issue Number: N/A

## What is the new behavior?

Migrates `test/karma/test-app/invisible-prehydration-false` to wdio. This involves adding some additional project files for a separate Stencil project build (Stencil config, tsconfig, `components.d.ts`). The test will inject an iframe into the browser DOM so the test can run "in isolation" (same pattern as #5591)

- [ ] Yes
- [x] No


christian-bromann (approved):
👍

christian-bromann:
I think the BrowserStack pipeline will continue to fail since we don't run any tests in Karma anymore which usually causes a testrunner to fail.

tanner-reits:
@christian-bromann Ryan Disabled the browserstack CI requirement so even if it fails this should still land

tanner-reits:
I stand corrected 😂

rwaskiewicz:
Us to this PR:
9Hb10vJh3t9GcmQ_

christian-bromann:
🙌

PR #5601, opened 2024-03-28
test(wdio): add a browser matrix for our wdio tests

alicewriteswrongs:
This refactors a bit how we control what browsers are used in our webdriverio tests. Instead of it being chrome or all browsers, where all browsers are activated by setting the `CI` environment variable, this adds a check for a `BROWSER` environment variable. How this works is that the webdriverio configuration file reads the environmen variable and if it's one of `FIREFOX`, `CHROME`, or `EDGE` it will set the tests to use that browser. If it's _not_ set (or set to an unrecognized value) we'll default to using chrome (as we do now), and if it's set to `ALL` it will run all of the browsers.

This also makes it easier for us to separate the test runs in separate browsers into separate jobs, so a failure only on Edge will show up more easily (and we'll get some more parallelism).

You need to edit a file if you want to run tests in e.g. just Edge locally.

- [ ] Yes
- [x] No

### DIFF
--- a/.github/workflows/test-wdio.yml
+++ b/.github/workflows/test-wdio.yml
@@ -13,9 +13,12 @@ jobs:
     uses: ./.github/workflows/build.yml
 
   run_wdio:
-    name: Run WebdriverIO Component Tests
+    name: Run WebdriverIO Component Tests (${{ matrix.browser }})
     runs-on: ubuntu-22.04
     needs: [build_core]
+    strategy:
+      matrix:
+        browser: [CHROME, FIREFOX, EDGE]
 
     steps:
       - name: Checkout Code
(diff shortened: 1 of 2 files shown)


rwaskiewicz (approved):
LGTM 👍 I'll defer to @christian-bromann in the event he thinks there's another way we could be doing this, but in general I'm in favor of this change.

We'll have to rename the checks when this is ready to merge, LMK when it's time and we'll do the dance of disabling the current one, merging this, and updating the names of the checks in GH's settings

.github/workflows/test-wdio.yml
```diff
@@ -39,7 +42,7 @@ jobs:
           filename: stencil-core-build.zip
 
       - name: Run WebdriverIO Component Tests
-        run: npm run test.wdio
+        run: BROWSER=${{ matrix.browser }} npm run test.wdio
```
christian-bromann, line 45: Alternatively:
```suggestion
        run: npm run test.wdio
        env:
          BROWSER: ${{ matrix.browser }}
```

christian-bromann (approved):
LGTM 👍

.github/workflows/test-wdio.yml
alicewriteswrongs: true! a little less implicit that way

PR #5614, opened 2024-04-01
fix(runtime): prevent ref callbacks from being called too early

### LINKED ISSUE #4074: bug: ref called in specified order and not after initializing element with all its attributes/properties
### Prerequisites

### Stencil Version

2.20.0/3.0.1

### Current Behavior

The order of `ref` matters and can lead to props/attributes being outdated by the time the callback is invoked.

### Expected Behavior

`ref` is called right after the element is fully initialized with all props/attributes. This aligns with other libraries that implement a similar `ref` mechanism (e.g., React, preact, maquette).

### System Info

```shell
OS X 13.2.1, npm 8.3.1, node 16.13.2, Chrome 110.0.5481.100
```

### Steps to Reproduce

Follow the steps to the repro case below.

### Code Reproduction URL

https://github.com/jcfranco/stencil-ref-called-in-prop-order-instead-of-after-node-being-initialized

### Additional Information

For comparison, I also created the following repro cases:

* [React](https://codepen.io/jcfranco/pen/bGxdQOK?editors=1111)
* [Preact](https://codepen.io/jcfranco/pen/BaONGMx?editors=1111)
* [Maquette](https://codepen.io/jcfranco/pen/QWVbzwY?editors=1111)

alicewriteswrongs:
Sort attribute names before calling `setAccessor` with their changed values in order to ensure that the `ref` attribute is handled by `setAccessor` after all other attributes have been handled. This prevents the execution order for `ref` callbacks from being dependency on the order in which attributes are declared in the JSX.

STENCIL-737
fixes #4074

At present `ref` callbacks are, like other attributes, handled in declaration-order, i.e. they are called based on the order in which they are declared in the JSX for a component.

This means that in a component like this:

```tsx

@Component({
  tag: 'ref-attr-order',
  shadow: true,
})
export class RefAttrOrder {
  render() {
    return (
      <div
        ref={(el) => {
          console.log(el.tabIndex);
        }}
        tabIndex={0}
      >
        my tabIndex: {this.index}
      </div>
    );
  }
}
```

The `tabIndex` attribute won't be set on the element when the `ref` is called. This leads to unexpected behavior when trying to get a reference to an element. There is a workaround, namely ensuring that the `ref` is the last attribute declared in the JSX, but we can do better by sorting the attributes when we're updating them to ensure that we handle `ref` after the others.

- [ ] Yes
- [x] No

Not in my view, rather this fixes a bug (#4074).

## Testing

I added a webdriverio regression test and confirmed that the test fails without the fix. I also built and then installed this in Framework, where it didn't cause any test failures.

### DIFF
--- a/src/runtime/vdom/update-element.ts
+++ b/src/runtime/vdom/update-element.ts
@@ -23,7 +23,7 @@ export const updateElement = (
 
   if (BUILD.updatable) {
     // remove attributes no longer present on the vnode by setting them to undefined
-    for (memberName in oldVnodeAttrs) {
+    for (memberName of sortedAttrNames(Object.keys(oldVnodeAttrs))) {
       if (!(memberName in newVnodeAttrs)) {
         setAccessor(elm, memberName, oldVnodeAttrs[memberName], undefined, isSvgMode, newVnode.$flags$);
       }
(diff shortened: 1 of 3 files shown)


src/runtime/vdom/update-element.ts
```diff
@@ -23,15 +23,36 @@ export const updateElement = (
...
+ * @returns a list of attribute names, sorted if they include `"ref"`
+ */
+function sortedAttrNames(attrNames: string[]): string[] {
+  if (attrNames.includes('ref')) {
+    // we need to sort these to ensure that `'ref'` is the last attr
+    return [...attrNames.filter((attr) => attr !== 'ref'), 'ref'];
+  } else {
+    // no need to sort, return the original array
+    return attrNames;
+  }
```
christian-bromann, line 57: This is totally just a personal preference but wonder what you think about this little tweak:
```suggestion
  return attrNames.includes('ref')
    // we need to sort these to ensure that `'ref'` is the last attr
    ? [...attrNames.filter((attr) => attr !== 'ref'), 'ref']
    // no need to sort, return the original array
    : attrNames;
```

christian-bromann (approved):
One comment but overall LGTM 👍

src/runtime/vdom/update-element.ts
alicewriteswrongs: sure, why not!

tanner-reits (approved):
LGTM

PR #5616, opened 2024-04-02
chore(build): remove obsolete rollup code

christian-bromann:
We have some duplicated code for creating build files in rollup. This is the first PR of many to remove these obsolete artifacts.

## What is the new behavior?
Don't bundle these dev server assets anymore as have migrated this to Rollup.

- [ ] Yes
- [x] No

## Testing

n/a

## Other information

n/a


rwaskiewicz:
I think for us to take this on, we should wait until the rollup -> esbuild migration is complete. That way in case we have to roll something back, it's one less commit to revert - unless there's something we're changing here we should do before/during the migration?

christian-bromann:
> unless there's something we're changing here we should do before/during the migration?

Not really, it is just duplicated code which will get removed anyway if we removed the `bundles` directory.

rwaskiewicz:
@christian-bromann Sounds good, let's hold off on this then until we have a few esbuild-based production builds out the door

christian-bromann:
Closing in favor of #5698

PR #5624, opened 2024-04-05
chore(build): port d.ts bundling for mock-doc from rollup to esbuild

alicewriteswrongs:
In #5012 when we put together a script to build `mock-doc/` with Esbuild we refactored a custom, hand-built `.d.ts` bundler that was used in the Rollup build to instead use `dts-bundle-generator`, a library that we depend on elsewhere. Although in theory a better option for avoiding NIH syndrome the output is just too different, so we want to switch to using the old rollup approach in our new, Esbuild-based build script for `mock-doc/`.

To accomplish this we can just export the existing function from the Rollup script and then call it in the Esbuild script.

STENCIL-1246

GitHub Issue Number: N/A

- [ ] Yes
- [x] No

## Testing

To test this you can do the following:

```sh
# build Stencil with Rollup
npm run build.rollup
# copy mock-doc/index.d.ts
cp mock-doc/index.d.ts /tmp/rollup-mock-dock-index.d.ts
# build with esbuild
npm run build
# copy again!
co mock-doc/index.d.ts /tmp/esbuild-mock-doc-index.d.ts
```

then diff those two files however you like. For instance if you have the excellent [`delta`](https://github.com/dandavison/delta) installed on your system you can do

```sh
delta /tmp/rollup-mock-dock-index.d.ts /tmp/esbuild-mock-doc-index.d.ts
```

they should be the exact same now.

### DIFF
--- a/scripts/esbuild/mock-doc.ts
+++ b/scripts/esbuild/mock-doc.ts
@@ -2,9 +2,9 @@ import type { BuildOptions as ESBuildOptions } from 'esbuild';
 import fs from 'fs-extra';
 import { join } from 'path';
 
+import { bundleMockDocDts } from '../bundles/mock-doc';
 import { bundleParse5 } from '../bundles/plugins/parse5-plugin';
 import { getBanner } from '../utils/banner';
-import { bundleDts } from '../utils/bundle-dts';
 import { BuildOptions, createReplaceData } from '../utils/options';
 import { writePkgJson } from '../utils/write-pkg-json';
 import { getBaseEsbuildOptions, getEsbuildAliases, runBuilds } from './util';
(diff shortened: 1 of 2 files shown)


scripts/esbuild/mock-doc.ts
```diff
@@ -23,8 +23,7 @@ export async function buildMockDoc(opts: BuildOptions) {
   // clear out rollup stuff and ensure directory exists
   await fs.emptyDir(outputDir);
 
-  // bundle d.ts
-  await bundleMockDocDts(opts, inputDir, outputDir);
+  const bundleDtsPromise = await bundleMockDocDts(inputDir, outputDir);
```
rwaskiewicz, line 26: I don't think we need the variable assignment here:
```suggestion
  await bundleMockDocDts(inputDir, outputDir);
```
In the rollup version, we wouldn't `await` the call to `bundleMockDocDts`, but would `await` the variable pointing to the associated promise

rwaskiewicz (approved):
Output looks good! One non-blocking comment related to the linter

tanner-reits (approved):

PR #5625, opened 2024-04-05
chore(esbuild): rm inline graceful-fs in dev-server

rwaskiewicz:
update the building of the dev-server such that we no longer inline graceful-fs in the dev-server. this saves us approximately 10kb in total - 20kb get removed from dev-server/server-process.js, and 10kb are added to `sys/node/index.ts`.

this is accomplished by removing graceful-fs from the shared alias list, as it's only needed in `sys` and `dev-server`. we them mark the library as "external" so that esbuild resolves to the version held on to by the `sys` submodule

GitHub Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/scripts/esbuild/dev-server.ts
+++ b/scripts/esbuild/dev-server.ts
@@ -70,9 +70,17 @@ export async function buildDevServer(opts: BuildOptions) {
     bundleExternal(opts, opts.output.devServerDir, cachedDir, 'open-in-editor-api.js'),
   ]);
 
-  const devServerAliases = getEsbuildAliases();
-  const external = [...builtinModules, './ws.js', './open-in-editor-api'];
+  const external = [
+    ...builtinModules,
+    // ws.js is externally bundled
+    './ws.js',
+    // open-in-editor-api is externally bundled
+    './open-in-editor-api',
+    // prevent graceful-fs from being inlined in the bundle, saving us ~20kb
+    'graceful-fs',
+  ];
 
+  const devServerAliases = getEsbuildAliases();
   const devServerIndexEsbuildOptions = {
     ...getBaseEsbuildOptions(),
     alias: devServerAliases,
--- a/scripts/esbuild/util.ts
+++ b/scripts/esbuild/util.ts
@@ -30,7 +30,6 @@ export function getEsbuildAliases(): Record<string, string> {
     // reference the same file
     prompts: './sys/node/prompts.js',
     glob: './sys/node/glob.js',
-    'graceful-fs': './sys/node/graceful-fs.js',
 
     // dev server related aliases
     ws: './ws.js',


alicewriteswrongs (changes requested):
I tried this out locally and I don't think we can make this exact change. In my testing I did a build locally and then checked out the `dev-server/server-process.js` file and noticed that there is now a `require('graceful-fs')` in there. If you're in e.g. the Stencil starter this will work fine, because `graceful-fs` is present (pulled in as a dependency of puppeteer and jest) but if you don't have those packages installed then the require will fail and starting the dev server will just hang like this:

https://github.com/ionic-team/stencil/assets/6207644/5d8b5f86-014d-47eb-b7a5-e9235a0e443b

(I think some error messages from the server process aren't getting piped through correctly)

so I think instead of just externalizing we'll need to use the `externalAlias` plugin we have which externalizes an import and also aliases it at the same time - I think that should ensure that the `graceful-fs` code isn't bundled but also ensure that we can resolve the module when it doesn't happen to be installed

gh won't let me make a suggestions on the relevant lines, but here's how you could do that:

```diff
diff --git a/scripts/esbuild/dev-server.ts b/scripts/esbuild/dev-server.ts
index 7c2b00750..4329321b2 100644
--- a/scripts/esbuild/dev-server.ts
+++ b/scripts/esbuild/dev-server.ts
@@ -11,7 +11,7 @@ import { bundleExternal, sysNodeBundleCacheDir } from '../bundles/sys-node';
 import { getBanner } from '../utils/banner';
 import { type BuildOptions, createReplaceData } from '../utils/options';
 import { writePkgJson } from '../utils/write-pkg-json';
-import { getBaseEsbuildOptions, getEsbuildAliases, getFirstOutputFile, runBuilds } from './util';
+import { externalAlias, getBaseEsbuildOptions, getEsbuildAliases, getFirstOutputFile, runBuilds } from './util';
 
 const CONNECTOR_NAME = 'connector.html';
 
@@ -76,8 +76,6 @@ export async function buildDevServer(opts: BuildOptions) {
     './ws.js',
     // open-in-editor-api is externally bundled
     './open-in-editor-api',
-    // prevent graceful-fs from being inlined in the bundle, saving us ~20kb
-    'graceful-fs',
   ];
 
   const devServerAliases = getEsbuildAliases();
@@ -108,7 +106,12 @@ export async function buildDevServer(opts: BuildOptions) {
     format: 'cjs',
     platform: 'node',
     write: false,
-    plugins: [esm2CJSPlugin(), contentTypesPlugin(opts), replace(createReplaceData(opts))],
+    plugins: [
+      esm2CJSPlugin(),
+      contentTypesPlugin(opts),
+      replace(createReplaceData(opts)),
+      externalAlias('graceful-fs', '../sys/node/graceful-fs.js'),
+    ],
     banner: {
       js: getBanner(opts, `Stencil Dev Server Process`, true),
     },
```

it might be a good idea to review other bundles to make sure they're functional without `graceful-fs` being installed in a test project since it's also being removed from the thing we return in `getEsbuildAliases`

rwaskiewicz:
@alicewriteswrongs great catch, thanks!

I added your suggestion in https://github.com/ionic-team/stencil/pull/5625/commits/d5f12aa571fc84972711579ee1a9c2fd6b3e788a, tested the with/without puppeteer case, and verified this didn't mess with the size of the compiler in some other way:
Screenshot 2024-04-10 at 11 27 37 AM

tanner-reits (approved):

alicewriteswrongs (approved):
lgtm! bundle size going ⬇️

Issue #5639, opened 2024-04-08
bug: `E2EPage.find()` throws error when using piercing selector targeting a non-existing element


jcfranco:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

4.14.0

### Current Behavior

Calling `find` with a piercing selector where the target does not exist throws an error instead of returning `null` as the doc states.

### Expected Behavior

For `find()` to return `null` for shadow DOM elements that are not found.

### System Info

```shell
System: node 20.10.0
    Platform: darwin (23.3.0)
   CPU Model: Apple M1 Pro (10 cpus)
    Compiler: /Users/juan6600/dev/projects/stencil-4.14.0-breaks-piercing-selector-for-non-existent-elements/node_modules/@stencil/core/compiler/stencil.js
       Build: 1711981251
     Stencil: 4.14.0 🚡
  TypeScript: 5.4.3
      Rollup: 2.56.3
      Parse5: 7.1.2
      jQuery: 4.0.0-pre
      Terser: 5.30.0
```

### Steps to Reproduce

See repro case readme.

### Code Reproduction URL

https://github.com/jcfranco/stencil-4.14.0-breaks-piercing-selector-for-non-existent-elements

### Additional Information

4.13.0 does not throw an error and works as expected.

This was previously fixed for another version in https://github.com/ionic-team/stencil/issues/5113.

tanner-reits:
Hey @jcfranco Thanks for reporting this (again 😅). I've identified the issue and will have a PR up to fix this shortly!

jcfranco:
@tanner-reits Thanks for the speedy fix! When is the next `@stencil/core` release planned?

PR #5641, opened 2024-04-08
fix(testing): prevent `find` from throwing error when query has no match

tanner-reits:

An error will get thrown when using `page.find()` for Puppeteer e2e tests if a query returns no elements.

Fixes: #5639

## What is the new behavior?

This commit fixes an issue w/ our Puppeteer testing implementation where the `find()` method would throw an error if the query had no matching elements.

We do not need to make any changes to `findAll` since the logic there will only attempt to use any returned handles if they exist in an array. Otherwise, no entries will be pushed to the returned array so the result will be an empty array as expected.

- [ ] Yes
- [x] No

## Testing

Tested changes in the linked issue reproduction case.

Also added tests to our e2e suite so we can avoid regressions related to these functions moving forward.

### DIFF
--- a/test/end-to-end/src/non-existent-element/empty-cmp.tsx
+++ b/test/end-to-end/src/non-existent-element/empty-cmp.tsx
@@ -1,0 +1,11 @@
+import { Component, h, Host } from '@stencil/core';
+
+@Component({
+  tag: 'empty-cmp',
+  shadow: true,
+})
+export class EmptyComponent {
+  render() {
+    return <Host>I have no children!</Host>;
+  }
+}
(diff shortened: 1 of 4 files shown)


test/end-to-end/src/non-existent-element/empty-cmp.tsx
```diff
@@ -0,0 +1,11 @@
...
+
+@Component({
+  tag: 'empty-cmp',
+  shadow: true,
+})
+export class EmptyComponent {
+  render() {
+    return <Host>I have no children!</Host>;
+  }
+}
```
rwaskiewicz, line 11: Can we add a similar component (and matching suite of tests) where it doesn't use `shadow`? I don't think it matters if it's totally non-encapsulated or used `scoped` - I just want to make sure that in the case there's not an actual shadow root, we handle that case as well.

rwaskiewicz (changes requested):
All in all looks good! I have an ask for an additional test suite with a non-encapsulated component, but nothing too serious. LMK if you have any questions

kensodemann:
I have also tested this PR using the Esri sample project and it passes there without issue.

alicewriteswrongs (approved):

PR #5645, opened 2024-04-09
chore(build): remove sourcemaps, minified compiler build

alicewriteswrongs:
This removes sourcemaps from the cli, compiler, and dev-server bundles and also removes the unused minified compiler build. All these changes are made to the esbuild-based build.

We realized recently that the minified compiler is no longer necessary (it was previously used for in-browser compilation support, and was removed in #4317) and the sourcemaps are similarly not useful, but do take up a significant amount of space.

Without this change a packed Stencil tarball is around 12MB, and with this change it drops to 3.5MB.

STENCIL-1242

- [ ] Yes
- [x] No

## Testing

Test that you can build and use Stencil with this change.

Also try out the 'debug' build, which is building like so:

```sh
DEBUG=true npm run build
```

that should produce sourcemaps for all the files bundled with Esbuild.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,6 +40,7 @@ dist/
 /screenshot/index.js.map
 /screenshot/package.json
 /screenshot/pixel-match.js
+/screenshot/pixel-match.js.map
 /screenshot/*.d.ts
 
 test/**/www/*
--- a/scripts/esbuild/util.ts
+++ b/scripts/esbuild/util.ts
@@ -109,12 +109,23 @@ export function runBuilds(builds: ESBuildOptions[], opts: BuildOptions): Promise
  * @returns a base set of options
  */
 export function getBaseEsbuildOptions(): ESBuildOptions {
-  return {
+  const options: ESBuildOptions = {
     bundle: true,
     legalComments: 'inline',
     logLevel: 'info',
     target: getEsbuildTargets(),
   };
+
+  // if the `build` sub-command is called with the `DEBUG` env var, like
+  //
+  // DEBUG=true npm run build
+  //
+  // then we should produce sourcemaps.
+  if (process.env.DEBUG) {
+    options.sourcemap = 'linked';
+  }
+
+  return options;
 }
 
 /**
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -131,6 +131,17 @@ stuck, your debugger may not have switched to a worker process that has halted o
 avoid this altogether by setting `--max-workers=1` when you launch Stencil (with the possibility of not being able to
 reproduce timing issues between workers as a side effect).
 
+> [!NOTE]
+> If you want to have access to sourcemaps when debugging the compiler locally
+> you can run the `build` script in `package.json` with the `DEBUG` environment
+> variable set to `true` like so:
+>
+> ```sh
+> DEBUG=true npm run build
+> ```
+>
+> this will write a `.js.map` file to disk next to each JavaScript file.
+
 #### Debugging the Compiler in VSCode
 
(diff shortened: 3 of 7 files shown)


.gitignore
```diff
@@ -40,6 +40,7 @@ dist/
 /screenshot/index.js.map
 /screenshot/package.json
 /screenshot/pixel-match.js
+/screenshot/pixel-match.js.map
```
rwaskiewicz, line 43: For my own knowledge, what was the cause of adding this to the `.gitignore`?

scripts/esbuild/util.ts
```diff
@@ -109,12 +109,23 @@ export function runBuilds(builds: ESBuildOptions[], opts: BuildOptions): Promise
...
   };
+
+  // if the `build` subcomment is called with the `DEBUG` env var, like
+  //
+  // DEBUG=true npm run build
+  //
+  // then we should produce sourcemaps.
+  if (process.env.DEBUG) {
+    options.sourcemap = 'linked';
+  }
```
rwaskiewicz, line 126: Can we document this in `CONTRIBUTING.md` and potentially add a new top level `package.json` script to perform a `npm run build.dev` (or something like that) where we automatically set that env var?

rwaskiewicz (changes requested):
I did a search for `stencil.min.js` across the entire Ionic Org in GitHub, and found a couple of places I think we should also look into as a part of this effort:

1. [This section of the Stencil Site](https://github.com/ionic-team/stencil-site/blob/2d98088f1e55cbc454db1f73742e654e6cd73830/versioned_docs/version-v4.11/core/compiler-api.md?plain=1#L12). I think that we can safely remove the offending sentence in the same files that Christian updated in https://github.com/ionic-team/stencil-site/commit/999d8312e154debe8c02fa6caeb9a4f0b40b1e20

2. We still have a reference to this file in the compiler itself, at https://github.com/ionic-team/stencil/blob/b3886aa928c1025e636aee1466f26f15fc4dd3eb/src/compiler/sys/stencil-sys.ts#L146. I wonder if we overlooked the possibility of pulling in Stencil from a CDN in a way that _isn't_ to use the deprecated compiler APIs 🤔 [jsdelivr reports 4/5 of the most versions are for v4.0.0 and lower](https://www.jsdelivr.com/package/npm/@stencil/core?tab=stats). Either way, it looks like this is a part of the public API (we export this `createSystem` fn) - can you swap this out with `stencil.js` and validate that works in a small Node project that calls `createSystem`?

alicewriteswrongs:
@rwaskiewicz good catch, opened a docs PR here: https://github.com/ionic-team/stencil-site/pull/1397

.gitignore
alicewriteswrongs: Just if you build with the `DEBUG=true` set then this `.js.map` file is written to disk, it wasn't previously so it needs to be added

scripts/esbuild/util.ts
alicewriteswrongs: oh I meant to comment on this but didn't submit the review! I think I'd lean towards putting some docs in `CONTRIBUTING.md` but `package.json` is already quite crowded imo

alicewriteswrongs: I added a little note to `CONTRIBUTING.md` in the section on debugging

alicewriteswrongs:
yeah was just looking at that, I made that change and then installed in a project locally and created a small node script to import and use the compiler API and it works fine

tanner-reits (approved):
LGTM once the change to `src/compiler/sys/stencil-sys.ts` is made

alicewriteswrongs:
just pushed the commit that changes that lil function over

rwaskiewicz (approved):

PR #5648, opened 2024-04-10
fix(docs): respect custom README content when writing to a custom path

### LINKED ISSUE #5400: bug: custom docs-readme directory
### Prerequisites

### Stencil Version

4.12.3

### Current Behavior

When using docs-readme with the `dir` property in stencil.config.js, all my component readme.md files keep getting overwritten, including the text above the `<!-- Auto Generated Below -->` comment. According to this documentation <https://stenciljs.com/docs/docs-readme#custom-markdown-content>, I understand that text placed above the comment should be preserved.

```
   type: 'docs-readme',
   dir: "output"
```

I just noticed this as I'm creating this issue: If I have a `readme.md` with the `<!-- Auto Generated Below -->` comment and some text above it in the default output folder, this content above it is preserved.

Is this the intended feature? If so, I think it might just need a little tweak in the documentation.

### Expected Behavior

Even when using the `dir` property to customize the output directory the content above `<!-- Auto Generated Below -->` should be preserved when generating another `readme.md` on build time.

### System Info

```shell
System: node 20.11.0
    Platform: linux (5.15.133.1-microsoft-standard-WSL2)
   CPU Model: AMD Ryzen 7 5800H with Radeon Graphics (16 cpus)
    Compiler: /home/rafael/Projetos/stenciljs-custom-dir/node_modules/@stencil/core/compiler/stencil.js
       Build: 1708455660
     Stencil: 4.12.3 🐍
  TypeScript: 5.3.3
      Rollup: 2.56.3
      Parse5: 7.1.2
      jQuery: 4.0.0-pre
      Terser: 5.27.1
```

### Steps to Reproduce

I created a repo to reproduce it. Just install the dependencies and run build.

https://github.com/oliveirarafa/stenciljs-custom-dir

* Without 'dir' (default output directory)

    Regenerating the readme.md keeps all the changes before the comment.

* With 'dir' (custom output directory)

    All the content before or after the comment is cleared.

    Note: just noticed that placing a readme.md in the default folder with the comment and a text before it makes this text persistent. In my opinion it makes more a little mor difficult to maintain because ir splits my docs in 2 but it works. 

### Code Reproduction URL

https://github.com/oliveirarafa/stenciljs-custom-dir

### Additional Information

_No response_

alicewriteswrongs:
On the `docs-readme` output target it's possible to set a custom output location with the `.dir` property and the README files generation for components will then be output to relative paths (like `my-component/readme.md`) within that directory.

This fixes a bug where that behavior didn't properly respect any manually-entered content in those readme files, so that if, for instance, you set the output to `custom-readme-output` and had `"My Custom Text"` at the top of
`custom-readme-output/components/my-component/readme.md` then running a build would overwrite your custom text.

This changes things so that we read the content of the custom readme and use that as the basis for the new text that we're going to write to disk. This has the effect of preserving the custom text that a user might have input.

fixes #5400

See #5400 for a description of the issue

## What is the new behavior?

The content of README files in a custom output location is now respected.

- [ ] Yes
- [x] No

## Testing

Try out the reproduction case in #5400, and make sure you can reproduce the issue. Then build and install this branch and make sure it fixes the issue!

### DIFF
--- a/src/compiler/docs/readme/output-docs.ts
+++ b/src/compiler/docs/readme/output-docs.ts
@@ -1,7 +1,8 @@
-import { join, relative } from '@utils';
+import { join, normalizePath, relative } from '@utils';
 
 import type * as d from '../../../declarations';
 import { AUTO_GENERATE_COMMENT } from '../constants';
+import { getUserReadmeContent } from '../generate-doc-data';
 import { stylesToMarkdown } from './markdown-css-props';
 import { depsToMarkdown } from './markdown-dependencies';
 import { eventsToMarkdown } from './markdown-events';
@@ -12,6 +13,23 @@ import { propsToMarkdown } from './markdown-props';
 import { slotsToMarkdown } from './markdown-slots';
 import { usageToMarkdown } from './markdown-usage';
 
+/**
+ * Generate a README for a given component and write it to disk.
+ *
+ * Typically the README is going to be a 'sibling' to the component's source
+ * code (i.e. written to the same directory) but the user may also configure a
+ * custom output directory by setting {@link d.OutputTargetDocsReadme.dir}.
+ *
+ * Output readme files also include {@link AUTO_GENERATE_COMMENT}, and any
+ * text located _above_ that comment is preserved when the new readme is written
+ * to disk.
+ *
+ * @param config a validated Stencil config
+ * @param compilerCtx the current compiler context
+ * @param readmeOutputs docs-readme output targets
+ * @param docsData documentation data for the component of interest
+ * @param cmps metadata for all the components in the project
+ */
 export const generateReadme = async (
   config: d.ValidatedConfig,
   compilerCtx: d.CompilerCtx,
@@ -25,10 +43,20 @@ export const generateReadme = async (
   await Promise.all(
     readmeOutputs.map(async (readmeOutput) => {
       if (readmeOutput.dir) {
-        const readmeContent = generateMarkdown(userContent, docsData, cmps, readmeOutput);
-        const relPath = relative(config.srcDir, docsData.readmePath);
-        const absPath = join(readmeOutput.dir, relPath);
-        const results = await compilerCtx.fs.writeFile(absPath, readmeContent);
+        const relativeReadmePath = relative(config.srcDir, docsData.readmePath);
+        const readmeOutputPath = join(readmeOutput.dir, relativeReadmePath);
+
+        const currentReadmeContent =
+          normalizePath(readmeOutput.dir) !== normalizePath(config.srcDir)
+            ? // The user set a custom `.dir` property, which is where we're going
+              // to write the updated README. We need to read the non-automatically
+              // generated content from that file and preserve that.
+              await getUserReadmeContent(compilerCtx, readmeOutputPath)
+            : userContent;
+
+        const readmeContent = generateMarkdown(currentReadmeContent, docsData, cmps, readmeOutput);
+
+        const results = await compilerCtx.fs.writeFile(readmeOutputPath, readmeContent);
         if (results.changedContent) {
           if (isUpdate) {
             config.logger.info(`updated readme docs: ${docsData.tag}`);
@@ -42,7 +70,7 @@ export const generateReadme = async (
 };
 
 export const generateMarkdown = (
-  userContent: string,
+  userContent: string | undefined,
   cmp: d.JsonDocsComponent,
   cmps: d.JsonDocsComponent[],
   readmeOutput: d.OutputTargetDocsReadme,
@@ -50,7 +78,7 @@ export const generateMarkdown = (
   //If the readmeOutput.dependencies is true or undefined the dependencies will be generated.
   const dependencies = readmeOutput.dependencies !== false ? depsToMarkdown(cmp, cmps) : [];
   return [
-    userContent,
+    userContent || '',
     AUTO_GENERATE_COMMENT,
     '',
     '',
@@ -78,6 +106,12 @@ const getDocsDeprecation = (cmp: d.JsonDocsComponent) => {
   return [];
 };
 
+/**
+ * Get a minimal default README for a Stencil component
+ *
+ * @param docsData documentation data for the component of interest
+ * @returns a minimal README template for that component
+ */
 const getDefaultReadme = (docsData: d.JsonDocsComponent) => {
   return [`# ${docsData.tag}`, '', '', ''].join('\n');
 };
--- a/src/compiler/docs/generate-doc-data.ts
+++ b/src/compiler/docs/generate-doc-data.ts
@@ -377,7 +377,10 @@ export const getNameText = (name: string, tags: d.JsonDocsTag[]) => {
  * @returns the user generated content that occurs before {@link AUTO_GENERATE_COMMENT}. If no user generated content
  * exists, or if there was an issue reading the file, return `undefined`
  */
-const getUserReadmeContent = async (compilerCtx: d.CompilerCtx, readmePath: string): Promise<string | undefined> => {
+export const getUserReadmeContent = async (
+  compilerCtx: d.CompilerCtx,
+  readmePath: string,
+): Promise<string | undefined> => {
   try {
     const existingContent = await compilerCtx.fs.readFile(readmePath);
     // subtract one to get everything up to, but not including the auto generated comment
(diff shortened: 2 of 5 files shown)


test/docs-json/custom-readme-output/components/my-component/readme.md
```diff
@@ -0,0 +1,31 @@
+This file is in a custom location, set with `.dir` on the `docs-readme` OT.
```
rwaskiewicz, line 1: I feel like I may have asked this in the past, but does anything in CI run these tests in `test/docs-json/`? I'm not seeing anything other than `build.docs-json` and `build` (that calls `build.docs-json`) in `test/package.json`, but that's not always the easiest thing to track down between `cd` calls and searching for `npm run build`. I would have thought I've asked this in https://github.com/ionic-team/stencil/pull/4212/ where we create most (all?) of files in this `/test/docs-json/` dir, but couldn't find anything

src/compiler/docs/readme/output-docs.ts
```diff
@@ -25,10 +43,20 @@ export const generateReadme = async (
...
       if (readmeOutput.dir) {
-        const readmeContent = generateMarkdown(userContent, docsData, cmps, readmeOutput);
-        const relPath = relative(config.srcDir, docsData.readmePath);
-        const absPath = join(readmeOutput.dir, relPath);
-        const results = await compilerCtx.fs.writeFile(absPath, readmeContent);
+        const relativeReadmePath = relative(config.srcDir, docsData.readmePath);
+        const readmeOutputPath = join(readmeOutput.dir, relativeReadmePath);
+
+        const currentReadmeContent =
+          readmeOutput.dir !== config.srcDir
```
rwaskiewicz, line 50: Double checking, only because the bug I fixed for the Lisbon team looked at lot like this - are these paths normalized if the user is:
- on Windows
- Just decided to write something like `C://Documents\ryan/why\ oh \why` in their README OT config

rwaskiewicz (commented):
Can you also add the Jira ticket number to the commit msg?

src/compiler/docs/readme/output-docs.ts
```diff
@@ -25,10 +43,20 @@ export const generateReadme = async (
...
+            ? // The user set a custom `.dir` property, which is where we're going
+              // to write the updated README. We need to read the non-automatically
+              // generated content from that file and preserve that.
+              await getUserReadmeContent(compilerCtx, readmeOutputPath)
```
rwaskiewicz, line 54: Would determining the README path at https://github.com/ionic-team/stencil/blob/b569ea7a9ad5d3b43eba3cf5c556cfab2c5e9e5f/src/compiler/docs/generate-doc-data.ts#L81 make it such that we don't need to read the README file from disk again?

test/docs-json/custom-readme-output/components/my-component/readme.md
alicewriteswrongs: there's no tests _per se_, but the `build.docs-json` command is called by the `build` command in `test/` which I think is in turn called by the `test-analysis` command which then in turn runs in the `test-analysis.yml` workflow

https://github.com/ionic-team/stencil/blob/b569ea7a9ad5d3b43eba3cf5c556cfab2c5e9e5f/.github/workflows/test-analysis.yml#L42-L47

the 'test' is basically that things are written to the disk and will fail the dirty git context check if the output has differed

src/compiler/docs/readme/output-docs.ts
alicewriteswrongs: I think there may be some circumstances where one or the other isn't normalized, so I'll just throw `normalizePath` on both to be safe

src/compiler/docs/readme/output-docs.ts
alicewriteswrongs: I think we can't really decide a 'canonical' README path at that point, because we're not operating on `docs-readme` output targets at that point

Or in other words, if I set a custom readme output dir on the `docs-readme` OT I don't think that should affect how the `docs-json` OT is handled

if we want that behavior then the readme location configuration should not be output target specific, but something in e.g. the `@Component` decorator

test/docs-json/custom-readme-output/components/my-component/readme.md
rwaskiewicz: That kind of tracks - can you adding something to the backlog to make this it's own workflow? I'm scared if we remove/change the analysis stuff, we won't realize what we've done there

src/compiler/docs/readme/output-docs.ts
rwaskiewicz: > I think we can't really decide a 'canonical' README path at that point, because we're not operating on docs-readme output targets at that point

But hasn't a user specified the canonical README at that point (and we've defaulted to the directory of the component if they haven't)? If the `README.md` is the source of truth, doesn't that always affect the docs-json OT, even today?

Maybe the question here is, how many README files should exist, and which ones influence 'docs-json'? 

The Stencil component starter gives us a `README` by default for `my-component`, but on `main` today, that doesn't get regenerated/updated if we specify `dir` on the `'docs-readme'` OT. With that in mind, I'd expect the README in the custom dir to be the be the 'canonical' one (and where we pull README data from for 'docs-json').

For example:

1. `npm init stencil@latest component readme-out-test && cd $_ && npm i` to generate a component starter
2. Modifying `stencil.config.ts` before running a build:
```diff
diff --git a/stencil.config.ts b/stencil.config.ts
index 59cd15a..8167fb3 100644
--- a/stencil.config.ts
+++ b/stencil.config.ts
@@ -14,6 +14,7 @@ export const config: Config = {
     },
     {
       type: 'docs-readme',
+      dir: 'my-custom-dir'
     },
     {
       type: 'www',
```
3. Deleting `src/components/my-component/readme.md` that was a part of the starter
4. Running a build with `npm run build`

Would all result in the following dir tree:
```
my-custom-dir
└── components
    └── my-component
        └── readme.md
src
├── components
│   └── my-component
│       ├── my-component.css
│       ├── my-component.e2e.ts
│       ├── my-component.spec.ts
│       └── my-component.tsx
├── components.d.ts
├── index.html
├── index.ts
└── utils
    ├── utils.spec.ts
    └── utils.ts
```
Where `README.md` doesn't exist in `src`, because we don't generate it

Does that make sense? I've been staring at Jira for far too long today, and my brain feels a little static-y 😆 

As an aside, the behavior today on `main` and this branch creates a bit of a conundrum - we've left a README file next to the component in the filesystem, with no real way to clean it up ourselves (users will just have to do that). We have the same behavior on this branch, but I think that's out of scope here.

christian-bromann (approved):
Looking at the reproduction case I was able to verify that given patch changes the behavior as following:

__Before:__
Content within the `src/components/my-component/readme.md` file was applied to the output file (e.g. `output/components/my-component/readme.md`) if a `.dir` property was set in the `stencil.config.ts`.

__After:__
None of what is being written in `src/components/my-component/readme.md` impacts the output file (e.g. `output/components/my-component/readme.md`) if a `.dir` property was set in the `stencil.config.ts`.

src/compiler/docs/readme/output-docs.ts
rwaskiewicz: > In particular, nothing prevents me from adding multiple different docs-readme output targets, each with different dir properties.

Ahhh that's what I was missing - I was only thinking about there being 0-1, and multiple READMEs scattered around the project directory as a result of multiple runs of `stencil build --docs` (and the like) with difference configurations.

I agree with your assessment there - thanks for the explanation!

rwaskiewicz (approved):

christian-bromann:
Released in Stencil :hotsprings: v4.17.0

smorrisods:
Hi there,

I'm looking into why our readme content has been getting erased in our Stencil updates and it looks like this PR is the reason.  We use our Stencil component readmes as the source of truth for our component readmes and we used the `docs-readme` output target with the `dir` path of our Docusuarus application to copy the component readmes over so they can be published.

We've noticed after upgrade that all the custom content we added to our component readmes has been removed upon their output, which has been breaking out workflow.  Upon investigation it looks like the output readmes, which either don't exist or were clones upon building our component library aren't being written with any custom content anymore.

If I add some custom content to any of the readmes in our output `dir` then that custom content is kept in place with the auto-generated content updated, but if we don't have any files in that location yet the custom content becomes/stays blank.

This change doesn't appear to add any options for allowing the readmes to be copied over unmodified or the destination file overwritten.  I'm wondering if you have any suggestions of how to handle this case so we can continue to use our component readmes as our source of truth for our documentation.

I suppose I could add a build step to copy the component library readmes over once they are built, but this adds additional complexity to our build process that was once handled by Stencil's output target.

christian-bromann:
@smorrisods thanks for the feedback, mind raising an issue for this with a reproduction case using a brand new Stencil project?

smorrisods:
@christian-bromann Thanks for the quick response, I'll try and set one up and open an issue with it.  Cheers!

PR #5653, opened 2024-04-11
fix(docs): merge together style docs from multiple CSS files

alicewriteswrongs:
This fixes a bug where although multiple stylesheets were being processed by the `ext-transforms-plugin` we were not properly merging the style documentation for all of those stylesheets, leading to a race condition where whichever file was processed last would set the style docs for the whole component. Not good!

To fix the issue we just need to merge the style docs for the various stylesheets together (de-duping on the `name` property).

The last stylesheet to be processed 'wins' and defines the `styleDocs` property on the `ComponentCompilerMetadata` for a given component.

## What is the new behavior?

As new `StyleDocs` come in we merge them into one array, to ensure we're collecting them all.

- [ ] Yes
- [x] No

## Testing

This functionality is a bit difficult to unit-test, so I stood up a test project similar to the one we already have for testing the `docs-json` output target. It includes a reproduction of the issue, where a component has two css files associated with it in `styleUrls` which both have CSS props documented in them. Both of these should show up in the README for the file.


tanner-reits (approved):
LGTM!

christian-bromann (approved):
👍

christian-bromann:
Released in Stencil :hotsprings: v4.17.0

PR #5672, opened 2024-04-15
fix(testing): perform string -> boolean type coercion for Jest config

### LINKED ISSUE #5640: bug: Unit test --watchAll argument parsed incorrectly
### Prerequisites

### Stencil Version

4.15.0

### Current Behavior

Regardless of the argument value, if the argument is present, Stencil sets `watch=true`:

https://github.com/ionic-team/stencil/blob/76c5d5410068cdb43296911a33b20d47531383a3/src/testing/testing.ts#L235-L237

This causes this command to be incorrectly interpreted as if `watchAll` is true:

```bash
npx stencil test --spec -- --watchAll=false
```

Besides confusion, this causes third-party tools to fail or freeze: https://github.com/jest-community/vscode-jest/issues/1124#issuecomment-2000596099

### Expected Behavior

If `--watchAll` is false, it should not be treated as a `true`

Jest's documentation explicitly mentions this:

> Use `--no-watchAll` (or `--watchAll=false`) to explicitly disable the watch mode if it was enabled using `--watchAll`. In most CI environments, this is automatically handled for you.

https://github.com/jestjs/jest/blob/726ca20752e38c18e20aa21740cec7aba7891946/docs/CLI.md?plain=1#L545

For reference, Jest defines it as a boolean argument using the `typebox` library:
https://github.com/jestjs/jest/blob/726ca20752e38c18e20aa21740cec7aba7891946/packages/jest-schemas/src/raw-types.ts#L346
Which would parse the value as false if it's anything other than `true`: https://github.com/sinclairzx81/typebox/blob/480ba12942cf908fab6e2ad16b3276b035ddcb54/test/runtime/value/cast/boolean.ts#L5

### System Info

```shell
System: node 20.11.0
    Platform: darwin (23.3.0)
   CPU Model: Apple M1 Pro (10 cpus)
    Compiler: /Users/mak13180/site/esri/arcgis-web-components/node_modules/@stencil/core/compiler/stencil.js
       Build: 1702922611
     Stencil: 4.15.0 🐏
  TypeScript: 5.2.2
      Rollup: 2.42.3
      Parse5: 7.1.2
      jQuery: 4.0.0-pre
      Terser: 5.26.0
```

### Steps to Reproduce

Run `npx stencil test --spec -- --watchAll=false`

### Code Reproduction URL

https://github.com/maxpatiiuk/vscode-jest-bug-reproduction

### Additional Information

_No response_

alicewriteswrongs:
This adds some manual string -> boolean type coercion to the code that we use to convert CLI flags to a Jest config (after parsing them with yargs).

This fixes an issue documented in https://github.com/ionic-team/stencil/issues/5640 where boolean CLI flags like `--watchAll` were being passed to Jest as strings, instead of as booleans, so that it was not possible to _disable_ a flag by doing `--flagName=false` (the value of the flag would be set to the string `"false"` which is a truthy value).

There was also a somewhat related issue with not properly parsing boolean flags in our CLI parser which was recently fixed in https://github.com/ionic-team/stencil/pull/5646.

fixes #5640
STENCIL-1259

We don't properly coerce strings like `'true'`, `'false'` to booleans when constructing Jest CLI arguments. This results in a situation where if you pass something like `--watchAll=false` you'll actually end up running tests in watch mode, which isn't what you want!

This fixes that behavior by explicitly coercing strings to booleans when they're in our list of defined boolean CLI flags.

- [ ] Yes
- [x] No

## Testing

First I think it's good to confirm you can reproduce the error.

Basically, start a test project w/ `@stencil/core@latest` installed and do

```sh
npx stencil test --spec -- --watchAll=false
```

you should observe that this starts things in watch mode, contrary to what you might expect.

then build, pack, and install this branch in the same project and run the same command. It should now run the tests without starting watch mode.


scripts/bundles/plugins/alias-plugin.ts
```diff
@@ -59,6 +59,9 @@ export function aliasPlugin(opts: BuildOptions): Plugin {
       if (id === '@sys-api-node') {
         return join(opts.buildDir, 'sys', 'node', 'index.js');
       }
+      if (id === '@stencil/core/cli') {
```
rwaskiewicz, line 62: For my own understanding, why do we make changes to the alias plugin here for string->boolean coercion?

rwaskiewicz (changes requested):
The functionality here LGTM - one question/comment regarding changing the build infra here and why we need to do it as a part of this PR

scripts/bundles/plugins/alias-plugin.ts
alicewriteswrongs: since we're adding an import of the `BOOLEAN_CLI_ARGS` in the code which is bundled into `testing/` we need to make sure that module can be resolved, I was only testing locally with esbuild but got a rollup build error on CI so went and added this

the esbuild equivalent change is here:

https://github.com/ionic-team/stencil/pull/5672/files#diff-b593e897dd3eb55a2f6dbab40eec549d1bd820ba3dddbd5c6e6340f1c50ef0c2

rwaskiewicz: Makes sense - can you update the commit msg with this context before you merge?

rwaskiewicz (approved):
One non-blocking ask - https://github.com/ionic-team/stencil/pull/5672#discussion_r1568880630

src/testing/jest/jest-27-and-under/jest-config.ts
```diff
@@ -81,8 +82,10 @@ export function buildJestArgv(config: d.ValidatedConfig): Config.Argv {
     } catch (e) {}
   }
 
-  if (typeof jestArgv.ci === 'string') {
-    jestArgv.ci = jestArgv.ci === 'true' || jestArgv.ci === '';
+  for (const flag of BOOLEAN_CLI_FLAGS) {
+    if (typeof jestArgv[flag] === 'string') {
+      jestArgv[flag] = jestArgv[flag] === 'true';
```
tanner-reits, line 87: We breaking anything by removing  the `jesArgv[flag] === ''` check?

tanner-reits (commented):
One quick question

src/testing/jest/jest-27-and-under/jest-config.ts
alicewriteswrongs: You know, I dont _think_ so, but maybe to be sure I'll leave the existing special-casing here for the `ci` flag just in case

tanner-reits (approved):

Issue #5596, opened 2024-03-27
bug: instance is undefined when consuming dist-custom-elements bundle


bitflower:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

4.13.0

### Current Behavior

We get this error:
```
index.js:3989 TypeError: Cannot read properties of undefined (reading 'render')
    at callRender (index.js:2224:43)
    at updateComponent (index.js:2149:9)
    at index.js:2091:40
    at enqueue (index.js:2109:90)
    at dispatchHooks (index.js:2091:12)
    at Array.dispatch (index.js:2027:28)
    at consume (index.js:4090:21)
    at flush (index.js:4145:9) 
```

See the readme.md in the repro project.

This is pre-discussed with @rwaskiewicz on Discord

### Expected Behavior

The consumed Stencil package works without errors.

### System Info

```shell
System: node 20.3.1
    Platform: darwin (22.3.0)
   CPU Model: Apple M1 Pro (10 cpus)
    Compiler: /Users/matthiasmax/Documents/Projekte/Temp/stencil-instance-issue/consumer-app/node_modules/@stencil/core/compiler/stencil.js
       Build: 1710782619
     Stencil: 4.13.0 🚞
  TypeScript: 5.4.2
      Rollup: 2.56.3
      Parse5: 7.1.2
      jQuery: 4.0.0-pre
      Terser: 5.29.2
```

### Steps to Reproduce

# Steps

## Setup my-component-lib
cd my-componentlib
npm i
npm run build
npm link

## Setup consumer-app
cd consumer-app
npm i
npm link my-componentlib
npm start

Now open your browser and check the console

### Code Reproduction URL

https://github.com/bitflower/stencil-instance-issue

christian-bromann:
This is a duplicate of #5457

We will be starting working on a bugfix this week and will provide an update soon.

bitflower:
Hi @christian-bromann,

that's great news, thank you.

christian-bromann:
[...]

PR #5675, opened 2024-04-16
fix(runtime): throw proper error if component is loaded with invalid runtime

christian-bromann:

Given someone imports a component compiled with `dist-custom-element` into a Stencil project that uses the `dist` output target. In the case that the DCE component shares the same runtime as the `dist` elements we will run into a runtime error.

GitHub Issue Number: #5596

## What is the new behavior?

This patch proposes to throw a more meaningful error in these situations.

I wonder if there is a better solution where we detect this at compile time. One idea would be to make the DCE component import an object from the runtime library that is only exported if the runtime is dedicated to DCE components.

## Documentation

The `externalRuntime` property is [documented](https://stenciljs.com/docs/custom-elements#externalruntime) however, we don't particularly mention this use case.

- [ ] Yes
- [x] No

If the instance is `undefined` in this case an error would be thrown anyway in `callRender` as we pass `instance` along as `undefined`.

## Testing

I added an e2e test for this scenario.


src/runtime/update-component.ts
```diff
@@ -47,6 +47,24 @@ const dispatchHooks = (hostRef: d.HostRef, isInitialLoad: boolean): Promise<void
...
+   * subsequent functions.
+   *
+   * For this scenario to work the user needs to set the `externalRuntime` flag
+   * for the `dist-custom-element` component that is being imported into the `dist`
+   * Stencil project.
+   */
+  if (!instance) {
+    throw new Error(
+      `Can't render component <${elm.tagName.toLowerCase()} /> with invalid Stencil runtime! ` +
+        'Make sure this imported component is compiled with a `externalRuntime: true` flag.',
```
alicewriteswrongs, line 64: what do you think about adding the URL <https://stenciljs.com/docs/custom-elements#externalruntime> in the message here? when a tool gives me a link to click and get more info I'm always so appreciative haha

test/wdio/global-script/global-script.test.tsx
```diff
@@ -15,4 +17,29 @@ describe('global script', () => {
     const renderedDelay = parseInt(text.slice('I am rendered after '.length));
     expect(renderedDelay).toBeGreaterThanOrEqual(1000);
   });
+
+  it('logs error when component with invalid runtime is loaded', async () => {
+    /**
+     * Fetching logs like this only works in Chromium. Once WebdriverIO v9 is released there
+     * will be easier primitives to fetch logs in other browsers as well.
+     */
```
alicewriteswrongs, line 25: we might want to create a JIRA ticket for upgrading to WebdriverIO v9 and then leave a TODO here to remove the `browser.isChromium` check

test/wdio/global-script/test-cmp.tsx
```diff
@@ -1,11 +1,22 @@
...
-    return <div>I am rendered after {Date.now() - window.__testStart}</div>;
+    return (
+      <section>
+        <div>I am rendered after {Date.now() - window.__testStart}</div>
+        {/*
+        rendering this component will fail as `<attribute-basic /`> is compiled with
+        a `dist-custom-element` output target, which will break in a lazy load environment
+      */}
+        {this.renderDistCustomElementComponent && <attribute-basic></attribute-basic>}
+      </section>
```
alicewriteswrongs, line 19: should this maybe be a separate, new component so that we retain some 1:1 relationship of tests to components?

christian-bromann:
Thanks @alicewriteswrongs for the feedback, applied suggested changes.

alicewriteswrongs (approved):

alicewriteswrongs:
sorry for the delay on the review on this one!

christian-bromann:
@tanner-reits any feedback on this?

tanner-reits (approved):
Just one thought for later

test/wdio/util.ts
```diff
@@ -1,7 +1,41 @@
...
+
 /**
  * A namespace for custom type definitions used in a portion of the testing suite
  */
 export declare namespace SomeTypes {
   type Number = number;
   type String = string;
 }
+
+export async function setupIFrameTest(htmlFile: string): Promise<HTMLElement> {
```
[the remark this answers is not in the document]
christian-bromann: I will raise a PR with an update once merged

PR #5677, opened 2024-04-16
test(docs): add a gh actions workflow for testing docs output targets

alicewriteswrongs:
This adds a new GH actions workflow (`.github/workflows/test-docs-build.yml`) which runs `stencil build` in the `test/docs-json` and `test/docs-readme` directories. These two directories contain Stencil projects which exercise the `docs-json` and `docs-readme` output targets, respectively. The output for each project is checked in to the repo, so running a build on CI and then checking for a dirty git context serves as a test that the output of the output targets hasn't changed.

STENCIL-1272

GitHub Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/test/package.json
+++ b/test/package.json
@@ -5,7 +5,7 @@
     "TODO-STENCIL-389": "echo Remove the exit call below",
     "analysis": "exit 0 && node ./.scripts/analysis.js",
     "analysis.build-and-analyze": "npm run build && npm run analysis",
-    "build": "npm run build.browser-compile && npm run build.hello-world && npm run build.hello-vdom && npm run build.todo && npm run build.end-to-end && npm run build.ionic && npm run build.docs-json && npm run build.docs-readme",
+    "build": "npm run build.browser-compile && npm run build.hello-world && npm run build.hello-vdom && npm run build.todo && npm run build.end-to-end && npm run build.ionic",
     "bundlers": "cd ./bundler && npm run start",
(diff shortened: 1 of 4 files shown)


test/package.json
```diff
@@ -5,7 +5,7 @@
     "TODO-STENCIL-389": "echo Remove the exit call below",
     "analysis": "exit 0 && node ./.scripts/analysis.js",
     "analysis.build-and-analyze": "npm run build && npm run analysis",
-    "build": "npm run build.browser-compile && npm run build.hello-world && npm run build.hello-vdom && npm run build.todo && npm run build.end-to-end && npm run build.ionic && npm run build.docs-json && npm run build.docs-readme",
+    "build": "npm run build.browser-compile && npm run build.hello-world && npm run build.hello-vdom && npm run build.todo && npm run build.end-to-end && npm run build.ionic",
```
tanner-reits, line 8: Does this exclude this from analysis testing?

tanner-reits (approved):
Think this makes sense after reading through the Jira ticket

test/package.json
alicewriteswrongs: yeah, formerly these were built as part of the analysis tests but now they're only built on their own

the motivation for making the change overall was that the analysis tests are sort of in this murky area and it would be better if we know for sure that something is valuable and important to pull it out into its own separate thing

christian-bromann (approved):

PR #5678, opened 2024-04-16
chore(release): switch release scripts to use ESbuild

alicewriteswrongs:
This changes our release process for `@stencil/core` to use the ESbuild-based build pipeline instead of our now-legacy Rollup-based build setup.

This change is intentionally made as small as possible in order to make it easy to diagnose any problems and easy to revert if needed. We have a follow-up ticket for removing the Rollup-related stuff more thoroughly.

STENCIL-1017

GitHub Issue Number: N/A

- [ ] Yes
- [x] No

## Testing

It's difficult for us to test this out thoroughly without actually running a release.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
   ],
   "scripts": {
     "build.rollup": "npm run tsc.scripts && npm run tsc.prod && npm run rollup.prod.ci",
-    "build": "npm run clean && npm run tsc.scripts && npm run tsc.prod && node scripts/build/esbuild/build.js --prod --ci",
+    "build": "npm run clean && npm run tsc.scripts && npm run tsc.prod && node scripts/build --prod --ci",
     "build.watch": "npm run build -- --watch",
     "build.updateSelectorEngine": "node scripts/build/updateSelectorEngine.js",
     "clean": "rm -rf build/ cli/ compiler/ dev-server/ internal/ mock-doc/ sys/ testing/ && npm run clean.scripts && npm run clean.screenshots",
--- a/scripts/esbuild/build.ts
+++ b/scripts/esbuild/build.ts
@@ -1,4 +1,6 @@
-import { getOptions } from '../utils/options';
+import { release } from '../release';
+import { validateBuild } from '../test/validate-build';
+import { BuildOptions, getOptions } from '../utils/options';
 import { buildCli } from './cli';
 import { buildCompiler } from './compiler';
 import { buildDevServer } from './dev-server';
@@ -8,14 +10,32 @@ import { buildScreenshot } from './screenshot';
 import { buildSysNode } from './sys-node';
 import { buildTesting } from './testing';
 
-// the main entry point for the Esbuild-based build
-async function main() {
+// the main entry point for the build
+export async function run(rootDir: string, args: ReadonlyArray<string>) {
   const opts = getOptions(process.cwd(), {
-    isProd: !!process.argv.includes('--prod'),
-    isCI: !!process.argv.includes('--ci'),
-    isWatch: !!process.argv.includes('--watch'),
+    isProd: args.includes('--prod'),
+    isCI: args.includes('--ci'),
+    isWatch: args.includes('--watch'),
   });
 
+  try {
+    if (args.includes('--release')) {
+      await release(rootDir, args);
+      return;
+    }
+
+    if (args.includes('--validate-build')) {
+      await validateBuild(rootDir);
+      return;
+    }
+    await buildAll(opts);
+  } catch (e) {
+    console.error(e);
+    process.exit(1);
+  }
+}
+
+export async function buildAll(opts: BuildOptions) {
   await Promise.all([
     buildCli(opts),
     buildCompiler(opts),
@@ -27,5 +47,3 @@ async function main() {
     buildInternal(opts),
   ]);
 }
-
-main();
--- a/scripts/index.ts
+++ b/scripts/index.ts
@@ -1,6 +1,6 @@
 import { join } from 'path';
 
-import * as build from './build';
+import * as build from './esbuild/build';
 
 // This path is relative to the final location of the compiled script, not its TypeScript source
 const stencilProjectRoot = join(__dirname, '..', '..');
--- a/scripts/release-tasks.ts
+++ b/scripts/release-tasks.ts
@@ -1,7 +1,7 @@
 import color from 'ansi-colors';
 import Listr, { ListrTask } from 'listr';
 
-import { bundleBuild } from './build';
+import { buildAll } from './esbuild/build';
 import { BuildOptions } from './utils/options';
 import { isPrereleaseVersion, isValidVersionInput, SEMVER_INCREMENTS, updateChangeLog } from './utils/release-utils';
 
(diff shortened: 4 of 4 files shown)


scripts/index.ts
alicewriteswrongs, line 1: to help keep the diff as small as possible I'm basically here just re-pointing this file from `scripts/build.ts` to `scripts/build/esbuild.ts`. I added a `run` function exported from `scripts/esbuild/build.ts` which has the same call signature as `run` in `scripts/build.ts` so it can drop in here - this means that the `release.ci` scripts in `package.json` don't have to change

scripts/esbuild/build.ts
```diff
@@ -8,14 +10,32 @@ import { buildScreenshot } from './screenshot';
...
+    if (process.argv.includes('--validate-build')) {
+      await validateBuild(rootDir);
+      return;
+    }
+    await buildAll(opts);
+  } catch (e) {
+    console.error(e);
+    process.exit(1);
+  }
+}
```
alicewriteswrongs, line 36: this is added in here based on the code in `run` in `scripts/build.ts` and calls the `release` and `validateBuild` functions in the same way

package.json
```diff
@@ -24,7 +24,7 @@
   ],
   "scripts": {
     "build.rollup": "npm run tsc.scripts && npm run tsc.prod && npm run rollup.prod.ci",
-    "build": "npm run clean && npm run tsc.scripts && npm run tsc.prod && node scripts/build/esbuild/build.js --prod --ci",
+    "build": "npm run clean && npm run tsc.scripts && npm run tsc.prod && node scripts/build --prod --ci",
```
alicewriteswrongs, line 27: since `scripts/index.ts` now imports from `scripts/esbuild/build.ts` we run that here (`scripts/build` has the output of building the `scripts/` directory, so `scripts/build/index.js` is what's going to be run here, which is compiled from `scripts/index.ts`)

scripts/release-tasks.ts
```diff
@@ -1,7 +1,7 @@
 import color from 'ansi-colors';
 import Listr, { ListrTask } from 'listr';
 
-import { bundleBuild } from './build';
+import { buildAll } from './esbuild/build';
```
alicewriteswrongs, line 4: I _think_ this is the only change we need to make to the existing release scripts - just swapping out a function that runs all of the esbuild bundles for the old one that runs the rollup bundles.

alicewriteswrongs (commented):
noting a few things

scripts/esbuild/build.ts
```diff
@@ -8,14 +10,32 @@ import { buildScreenshot } from './screenshot';
...
-async function main() {
+// the main entry point for the build
+export async function run(rootDir: string, args: ReadonlyArray<string>) {
   const opts = getOptions(process.cwd(), {
-    isProd: !!process.argv.includes('--prod'),
-    isCI: !!process.argv.includes('--ci'),
-    isWatch: !!process.argv.includes('--watch'),
+    isProd: !!args.includes('--prod'),
+    isCI: !!args.includes('--ci'),
+    isWatch: !!args.includes('--watch'),
```
rwaskiewicz, line 18: Non blocking - IDK if we need the double bang args here - shouldn't `includes` return a boolean?

scripts/esbuild/build.ts
```diff
@@ -8,14 +10,32 @@ import { buildScreenshot } from './screenshot';
...
   });
 
+  try {
+    if (args.includes('--release')) {
+      await release(rootDir, process.argv);
```
rwaskiewicz, line 23: I think this needs to be `args` instead of `process.argv` to make sure we're sending the right subset of `process.argv` to this command:
```suggestion
      await release(rootDir, args);
```

scripts/esbuild/build.ts
```diff
@@ -8,14 +10,32 @@ import { buildScreenshot } from './screenshot';
...
+      return;
+    }
+
+    if (process.argv.includes('--validate-build')) {
```
rwaskiewicz, line 27: I think this needs to be `args` instead of `process.argv`. Technically this shouldn't matter too much in practice, but it'd be good to operate on the same array for all our flag testing here:
```suggestion
    if (args.includes('--validate-build')) {
```

rwaskiewicz (changes requested):
Only a couple little asks here.

I tested this locally in an air-gapped(-ish) environment, and everything looks to be working as expected.

We _may_ need to create a followup task to this in the future. I saw esbuild and listr fighting over who had control of the terminal, which led to some "choppy" output:
```
⚡ Done in 16ms

  screenshot/pixel-match.js  75.5kb

⚡ Done in 16ms

  ✔ Check for pre-release version
  ↓ Check git tag existence [skipped]
  ✔ Install npm dependencies (npm ci)
  ✔ Transpile Stencil (tsc.prod)
  ⠙ Bundle @stencil/core (1713356807)
    Publish @stencil/core to npm
    Tagging the latest git commit
    Pushing git tags

  ✔ Check for pre-release version
  ↓ Check git tag existence [skipped]
  ✔ Install npm dependencies (npm ci)
  ✔ Transpile Stencil (tsc.prod)
  ⠹ Bundle @stencil/core (1713356807)
    Publish @stencil/core to npm
    Tagging the latest git commit
    Pushing git tags

  sys/node/worker.js  1.2kb

⚡ Done in 4ms
  ✔ Check for pre-release version
  ↓ Check git tag existence [skipped]
  ✔ Install npm dependencies (npm ci)
  ✔ Transpile Stencil (tsc.prod)
  ⠸ Bundle @stencil/core (1713356807)
    Publish @stencil/core to npm
    Tagging the latest git commit
    Pushing git tags

  internal/app-data/index.js  2.5kb

⚡ Done in 9ms

  internal/app-data/index.cjs  3.5kb

⚡ Done in 10ms

  internal/client/shadow-css.js  13.4kb
```

but we can see what this looks like in practice and decide what to do from there

scripts/esbuild/build.ts
alicewriteswrongs: yup, will switch all those over

scripts/esbuild/build.ts
alicewriteswrongs: haha yeah we definitely don't need those - looks like I added them in #4187 when `scripts/esbuild/build.ts` was added but I don't remember why! going to remove

alicewriteswrongs:
@rwaskiewicz I think we could use the [`formatMessages` function provided by Esbuild](https://esbuild.github.io/api/#format-messages) to make the build output a little more neat when running in a `Listr` context - the `buildAll` function could take an interface that provides some logging capability, then when running in 'release' mode we could pass in an object that takes in log messages and prints them nicely in Listr, and then when running outside of `Listr` (like when you just do `npm run build`) it could more or less just pass in `console.log` for the same purpose.

The example shown in the ESbuild docs for `formatMessages` is confusing btw, but I found a better example of how it could be used here: https://github.com/evanw/esbuild/issues/2655#issuecomment-1303111364

I think basically we'd turn off the normal logging and then set that up. 

I created STENCIL-1290 to cover that

rwaskiewicz (approved):

christian-bromann (approved):

Issue #5692, opened 2024-04-23
cannot identified the prompts.js as a function？？


MaybeSomeone:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

4.7.0

### Current Behavior

when i execute npm generate then report a error!! uncaught cli error: TypeError: prompt is not a function, that's why????

### Expected Behavior

i hope it can be run as normal

### Steps to Reproduce

just execute npm generate

### Code Reproduction URL

https://github.com/MaybeSomeone/weathercomponent

MaybeSomeone:
Hello i have provide the url, but i just init project by export web component use, and then i execute npm run generate.

MaybeSomeone:
System: node 21.7.1
    Platform: darwin (21.6.0)
   CPU Model: Intel(R) Core(TM) i7-4770HQ CPU @ 2.20GHz (8 cpus)
    Compiler: /Users/super/Desktop/WebComponents/weathercomponent/node_modules/@stencil/core/compiler/stencil.js
       Build: 1713808785
     Stencil: 4.17.0 ♨️
  TypeScript: 5.4.5
      Rollup: 2.56.3
      Parse5: 7.1.2
      jQuery: 4.0.0-pre
      Terser: 5.30.3

rwaskiewicz:
Thanks! I wasn't able to quite replicate it, but I think I know what's happening here. I see `npm run generate` in the component starter just failing on my end (no error, no warning). I _think_ I have a solution - could you try installing this dev build and let me know if it fixes it for you?

```
npm i @stencil/core@4.17.0-dev.1713882968.4a08d0e
```

MaybeSomeone:
Ok i'll try and i change for another macbook can replicate it.

MaybeSomeone:
Hello how can i generate target component? it need me input the Component tag name, but export no relative components.

MaybeSomeone:
I don't know how to choose the target component or export all exist components.

rwaskiewicz:
You can generate a component with 
```
npm run generate new-component
```
and select any css/test files to be generated.

I'm not sure I understand this part:
[...]

PR #5693, opened 2024-04-23
fix(cli): prevent generate task from crashing 

rwaskiewicz:

See 'new behavior'

GitHub Issue Number: N/A

## What is the new behavior?

this pr is motivated by a scenario where running `npx stencil generate`
would result in either an error or an early return:
```
npm run generate

> gen-test-1@0.0.1 generate
> stencil generate

[49:23.1]  @stencil/core
[49:23.2]  v4.17.0 ♨️
```

note: the reported error was never reproduced locally, only the early
return.

technically, this was introduced in v4.17.0, when we switched over to
using esbuild for production builds. however, this has been present in
any build that is esbuild-based since we migrated the cli submodule to
esbuild (where we had a mixed rollup-esbuild build during the migration
for dev builds).

removing this alias eliminates dynamic imports in the CJS output of the
cli submodule:
```
const {
    prompt: r
  } = await import('../sys/node/prompts.js')
```
which wouldn't run properly, causing the error/early return.

this change does cause an increase in bundle size, as we end up
importing more of prompts.js than we do. to mitigate this, we direct
the cli module to the export we actually use (leading to less getting
bundled in the cli module).

fixes: https://github.com/ionic-team/stencil/issues/5692

STENCIL-1294 cannot identified the prompts.js as a function? ?

## Documentation

N/A

- [ ] Yes
- [x] No

## Testing
1. Added new automated tests to the component starter
2. Manually tested by:
    1. `npm init stencil && cd $_ && npm i`
    2. `npm run generate` <- Should fail/return early
    3. `npm i @stencil/core@4.17.0-dev.1713882968.4a08d0e` 
    4. `npm run generate` <- Works!

### DIFF
--- a/.github/workflows/test-component-starter.yml
+++ b/.github/workflows/test-component-starter.yml
@@ -103,3 +103,34 @@ jobs:
         run: npm run test -- --no-build # the project was just built, don't build it again
         working-directory: ./tmp-component-starter
         shell: bash
+
+      - name: Test npx stencil generate
+        # `stencil generate` doesn't have a way to skip file generation, so we provide it with a component name and run
+        # `echo` with a newline to select "all files" to generate (and use -e to interpret that backslash for a newline)
+        run: echo -e '\n' | npm run generate -- hello-world
+        working-directory: ./tmp-component-starter
+        shell: bash
+
+      - name: Verify Files Exist
+        run: |
+          file_list=(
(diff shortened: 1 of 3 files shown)


alicewriteswrongs (approved):

christian-bromann (approved):
👍

PR #5697, opened 2024-04-24
chore(jsdoc): add JSDoc to src/runtime/styles.ts

alicewriteswrongs:
This adds explanations for what goes on in this file, since before things were a bit mysterious.

GitHub Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/src/runtime/styles.ts
+++ b/src/runtime/styles.ts
@@ -8,6 +8,17 @@ import { HYDRATED_STYLE_ID, NODE_TYPE, SLOT_FB_CSS } from './runtime-constants';
 
 const rootAppliedStyles: d.RootAppliedStyleMap = /*@__PURE__*/ new WeakMap();
 
+/**
+ * Register the styles for a component by creating a stylesheet and then
+ * registering it under the component's scope ID in a `WeakMap` for later use.
+ *
+ * If constructable stylesheet are not supported or `allowCS` is set to
+ * `false` then the styles will be registered as a string instead.
+ *
+ * @param scopeId the scope ID for the component of interest
+ * @param cssText styles for the component of interest
+ * @param allowCS whether or not to use a constructable stylesheet
+ */
 export const registerStyle = (scopeId: string, cssText: string, allowCS: boolean) => {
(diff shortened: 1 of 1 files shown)


christian-bromann (approved):
💯

src/runtime/styles.ts
```diff
@@ -8,10 +8,21 @@ import { HYDRATED_STYLE_ID, NODE_TYPE, SLOT_FB_CSS } from './runtime-constants';
...
+ *
+ * @param scopeId the scope ID for the component of interest
+ * @param cssText styles for the component of interest
+ * @param allowCS whether or not to use a constructable stylesheet
+ */
 export const registerStyle = (scopeId: string, cssText: string, allowCS: boolean) => {
   let style = styles.get(scopeId);
   if (supportsConstructableStylesheets && allowCS) {
-    style = (style || new CSSStyleSheet()) as CSSStyleSheet;
+    style ||= new CSSStyleSheet();
```
rwaskiewicz, line 25: Does this get transpiled down in any way in the build output? Technically, we still support Chrome/Edge 79-84 (inclusive) that don't support logical-or assignment - https://caniuse.com/?search=logical%20assignment

src/runtime/styles.ts
```diff
@@ -23,6 +34,19 @@ export const registerStyle = (scopeId: string, cssText: string, allowCS: boolean
   styles.set(scopeId, style);
 };
 
+/**
+ * Attach the styles for a given component to the DOM
+ *
+ * If the element uses shadow or is already attached to the DOM then we can
+ * create a stylesheet inside of it's associated document fragment, otherwise
```
rwaskiewicz, line 41: Non-blocking nit
```suggestion
 * create a stylesheet inside of its associated document fragment, otherwise
```

rwaskiewicz (approved):

src/runtime/styles.ts
alicewriteswrongs: ah I thought we were good on that now, I'll just get rid of that

PR #5698, opened 2024-04-24
chore(build): remove rollup build infrastructure

alicewriteswrongs:
this comprises:

- the rollup bundle files
- making the esbuild code now the default
- renaming the modules in `scripts/bundles/plugins` which we want to keep around
- removing the `build-rollup` CI as well as the configuration in other workflows for which build to use

GitHub Issue Number: N/A

- [ ] Yes
- [x] No

## Testing

It's a good idea for us to make sure that a dev release works properly, and probably a good idea to test that fairly extensively in framework and whatnot.

I published a dev build with the version `4.17.1-dev.1713982255.c28d937`, install it like so:

```sh
npm install @stencil/core@4.17.1-dev.1713982255.c28d937
```


rwaskiewicz (commented):
FWIW, I think we should hold off on this for a few weeks while we resolve any lingering esbuild vs rollup build differences

alicewriteswrongs:
agreed - I'm just opening it now, I'll put it as a draft though

christian-bromann (approved):
LGTM 🔪

tanner-reits (approved):
🪓 🪓 🪓

Issue #5702, opened 2024-04-25
bug: missing root element's scope id as classname in nested child


yigityuce:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

4.17.1

### Current Behavior

think about having a component structure like below:

```
my-timepicker -------> my-input ---------> input (native)
             |
             | ------> my-menu --------> my-popover ---------> whatever


my-popover (can be used as standalone)

```

given that component structure I have a style for popover (eg.: `padding: 8;`) and I want to customize popover style from the `timepicker` component's style file like:

```scss
:host {
  my-menu {
     my-popover {
        padding: 32px;
     }
  }
}
```

Current behavior: nested element selector is **NOT** working since during the scss compilation phase it adds `.sc-my-timepicker` class to the `my-popover` selector and the compiled CSS file becomes sth like this:

```css
.sc-my-timepicker-h my-menu.sc-my-timepicker my-popover.sc-my-timepicker {
  padding: 32px;
}
```

the problem is root component's scope id (`sc-my-timepicker`) is **NOT** being added to the nested component's classNames. So the selector which is defined in root component is **NOT** working.

### Expected Behavior

works as expected

current:

<img width="712" alt="current" src="https://github.com/ionic-team/stencil/assets/5856230/eabcb0ee-57b4-42e6-a1de-2385f7f33bdf">

expected:

<img width="711" alt="expected" src="https://github.com/ionic-team/stencil/assets/5856230/8e5c3fdd-5c78-48f9-b259-e04836ab77d2">

### System Info

```shell
Your current version of Node is v16.18.0, however Stencil's recommendation is v18.16.0 or greater. Note that future versions of Stencil will eventually remove support for older Node versions and an Active LTS version is recommended (https://nodejs.org/en/about/releases/).


      System: node 16.18.0
    Platform: darwin (23.4.0)
   CPU Model: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz (12 cpus)
[...]

PR #5704, opened 2024-04-25
fix(runtime): add root scope id to the nested child as classname

yigityuce:
sass compiler adds root scope id to the nested child elements too, therefore nested element selectors are not working from the root component since the root scope id is not adding as classname to the nested children elements

#5702

GitHub Issue Number: #5702 

- [ ] Yes
- [x] No

### DIFF
--- a/src/runtime/vdom/vdom-render.ts
+++ b/src/runtime/vdom/vdom-render.ts
@@ -111,7 +111,7 @@ const createElm = (oldParentVNode: d.VNode, newParentVNode: d.VNode, childIndex:
       elm.classList.add((elm['s-si'] = scopeId));
     }
 
-    if (BUILD.shadowDom || BUILD.scoped) {
+    if (BUILD.scoped) {
       // to be able to style the deep nested scoped component from the root component,
       // root component's scope id needs to be added to the child nodes since sass compiler
       // adds scope id to the nested selectors during compilation phase
--- a/src/runtime/vdom/vdom-render.ts
+++ b/src/runtime/vdom/vdom-render.ts
@@ -112,6 +112,9 @@ const createElm = (oldParentVNode: d.VNode, newParentVNode: d.VNode, childIndex:
     }
 
     if (BUILD.shadowDom || BUILD.scoped) {
+      // to be able to style the deep nested scoped component from the root component,
+      // root component's scope id needs to be added to the child nodes since sass compiler
+      // adds scope id to the nested selectors during compilation phase
       const rootScopeId =
         newParentVNode.$elm$?.['s-rsc'] || newParentVNode.$elm$?.['s-si'] || newParentVNode.$elm$?.['s-sc'];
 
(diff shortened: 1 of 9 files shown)


test/wdio/scoped-id-in-nested-classname/cmp.test.tsx
```diff
@@ -0,0 +1,14 @@
...
+
+describe('scope-id-in-nested-classname', function () {
+  beforeEach(() => {
+    render({
+      template: () => <cmp-level-1></cmp-level-1>,
+    });
+  });
+
+  it('should have root scope id in the nested element as classname', async () => {
+    await expect($('cmp-level-3')).toHaveElementClass('sc-cmp-level-1');
```
christian-bromann, line 12: Can we also assert the applied CSS directly via `getCSSProperty`? I think this would give us a bit more end to end confidence, wdyt?

src/runtime/vdom/vdom-render.ts
```diff
@@ -111,6 +111,16 @@ const createElm = (oldParentVNode: d.VNode, newParentVNode: d.VNode, childIndex:
       elm.classList.add((elm['s-si'] = scopeId));
     }
 
+    if (BUILD.shadowDom || BUILD.scoped) {
```
christian-bromann, line 114: Can we maybe add a comment here to document what this section is doing?

christian-bromann (commented):
Two minor additional suggestions, I was able to verify that this solves described bug, though I am not fully aware how these css classes are generated in detail.

test/wdio/scoped-id-in-nested-classname/cmp.test.tsx
yigityuce: tbh, I can't see the value of adding this check since it's gonna be a test case-specific check, not the functionality. IMHO, checking the classname would be sufficient, checking one step further would be falling into browser/jsdom behavior testing area

src/runtime/vdom/vdom-render.ts
yigityuce: done 🚀

test/wdio/scoped-id-in-nested-classname/cmp.test.tsx
yigityuce: anyway, I added the test case

yigityuce:
hey @christian-bromann were you able to check the updates in the PR?

christian-bromann:
@yigityuce one little ask, it seems like this causes some unit test failures, mind taking a look and update the expectations in the tests?

yigityuce:
> @yigityuce one little ask, it seems like this causes some unit test failures, mind taking a look and update the expectations in the tests?

done 🎉

src/runtime/vdom/vdom-render.ts
```diff
@@ -111,6 +111,19 @@ const createElm = (oldParentVNode: d.VNode, newParentVNode: d.VNode, childIndex:
...
```
rwaskiewicz, line 114: For my own understanding, why do we need both `shadowDom` and `scoped` here? Could we get by with just
```suggestion
    if (BUILD.scoped) {
```

yigityuce: makes perfectly sense! done

rwaskiewicz (approved):

rwaskiewicz:
@christian-bromann can you give this one last look through? If it looks good to you, I think this should be good to merge

christian-bromann (approved):

PR #5705, opened 2024-04-25
fix(build): address dist bundle size issue

alicewriteswrongs:

GitHub Issue Number: N/A

- [ ] Yes
- [x] No

## Testing

This one is a little involved to test because the full reproduction involves ionic framework and an ionic angular project.

So! We need to do some setup.

First, you should get Stencil linked up to ionic-framework. In the Stencil repo you can do:

```sh
npm link
```

Then in Framework do:

```sh
cd core
npm link @stencil/core
cd ..
```

then for testing purposes you'll need to apply this patch:

https://gist.github.com/alicewriteswrongs/1a1967bfe4171546f5393a91bcabee95

This adds a script for building the `@ionic/core` and `@ionic/angular` packages called `./build_angular.sh` and also makes some changes to `tsconfig.json` which are necessary for the `npm link` business to work correctly.

Once you have that in place you should also clone this reproduction of the issue:

https://github.com/alicewriteswrongs/stencil-ionic-angular-repro

it's in the `test-test-test` subdirectory. If you go in there there is a script called `install_dev_versions.sh` which will install the packed `@ionic/core` and `@ionic/angular` packages into this test project.

OK! So if you get all of _that_ set up then you're ready to work on this.

First, you should confirm the issue. You can use a script like this to run through the whole repro:

```sh
#!/bin/bash

# first build stencil
npm run build

# then build angular in Ionic Framework
cd ~/Code/ionic-framework/
./build_angular.sh

# then build in the repro
cd ~/Code/scratch-stencil/stencil-4.17-issue/test-test-test/
./install_dev_versions.sh
npm run build
```

Adjust as-needed to account for where you have stencil, framework, and my reproduction cloned.

So first run that script with `main` checked out in Stencil, and confirm that the build output for the angular project should a `main.XXXXX.js` file with a size of ~700kb.

Then check out this branch and confirm that the size drops down to under 400kb.

### DIFF
--- a/scripts/esbuild/internal-platform-client.ts
+++ b/scripts/esbuild/internal-platform-client.ts
@@ -17,7 +17,7 @@ import { externalAlias, getBaseEsbuildOptions, getEsbuildAliases, getEsbuildExte
  * @param opts build options
  * @returns an array of ESBuild option objects
  */
-export async function getInternalClientBundle(opts: BuildOptions): Promise<ESBuildOptions[]> {
+export async function getInternalClientBundles(opts: BuildOptions): Promise<ESBuildOptions[]> {
   const inputClientDir = join(opts.srcDir, 'client');
   const outputInternalClientDir = join(opts.output.internalDir, 'client');
   const outputInternalClientPolyfillsDir = join(outputInternalClientDir, 'polyfills');
@@ -59,9 +59,6 @@ export async function getInternalClientBundle(opts: BuildOptions): Promise<ESBui
       replace(createReplaceData(opts)),
       externalAlias('@app-data', '@stencil/core/internal/app-data'),
       externalAlias('@utils/shadow-css', './shadow-css.js'),
-      // we want to get the esm, not the cjs, since we're creating an esm
-      // bundle here
-      externalAlias('@stencil/core/mock-doc', '../../mock-doc/index.js'),
       findAndReplaceLoadModule(),
     ],
   };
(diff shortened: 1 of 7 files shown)


test.sh
alicewriteswrongs, line 1: this is a little test script - I'm going to leave it in for the review and delete before merge

rwaskiewicz:
@alicewriteswrongs When I run `build-angular.sh` from the root of the Framework repo, I get:
```
[ ERROR ]  unhandledRejection: Error: Could not resolve '../../mock-doc/index.cjs/constants' from
           ../../stencil/internal/hydrate/index.js at error
           (/Users/ryan/workspaces/stencil/compiler/stencil.js:9734:27) at ModuleLoader.handleResolveId
           (/Users/ryan/workspaces/stencil/compiler/stencil.js:27260:20) at
           /Users/ryan/workspaces/stencil/compiler/stencil.js:27254:77

```

Is there anything obvious I might be doing wrong here?

alicewriteswrongs (commented):
just noting something

tanner-reits (approved):
LGTM!

christian-bromann (approved):
Awesome work 🎉 

Verified that this works by:

- creating a new Ionic project via `ionic start photo-gallery tabs --type=angular --capacitor` selecting `NgModules`
- building `@ionic/core` using Stencil build with this branch
- run `npm run build`
- bundle size for `main.62054bba8aec0c16.js` before: `671523`, bundle size after: `367433`

👍

Issue #5623, opened 2024-04-04
bug: generated documentation does not display custom properties when included in loud comments


duhem-s:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

v4.14.0

### Current Behavior

When I document custom CSS properties using syntax within loud comments in my CSS files, the generated documentation does not display them.

In Sass, to preserve comments we have to set them as loud comments using the following format:
```SCSS
/*!
 * @prop --mg-badge-size: Badge size
 */
```
https://sass-lang.com/documentation/syntax/comments/#in-scss

However, these loud comments do not seem to be interpreted by the documentation generation process.

### Expected Behavior

I expect the documentation to display the --mg-badge-size custom property with its associated description, indicating that it represents the badge size.

### Steps to Reproduce

Add loud comment in CSS like: https://github.com/duhem-s/stencil-css/blob/main/src/components/my-component/my-component.css

Run build command.

Custom properties are not displayed in readme.

### Code Reproduction URL

https://github.com/duhem-s/stencil-css/tree/main

rwaskiewicz:
Thanks for the issue + reproduction case! I verified that Stencil can "see" comments that begin with a `/*!`, but it's currently not doing anything with them. I'm going to label this to get it added to our backlog.

Thanks again!

rwaskiewicz:
The fix for this issue was included in today's release of [Stencil v4.17.2](https://github.com/ionic-team/stencil/releases/tag/v4.17.2)

PR #5706, opened 2024-04-25
fix(compiler): recognize loud comments when generating style docs

tanner-reits:

Stencil's doc generation does not support loud comments (comments starting with `/*!`). Loud comments prevent the comment block from being stripped from the output during the sass compilation.

Fixes: #5623 

## What is the new behavior?

The doc generation now matches against either a traditional comment block (`/**`) or a loud comment block (`/*!`).

## Documentation

N/A

- [ ] Yes
- [x] No

## Testing

Added an automated test for loud comment blocks.

Manually tested various combinations of comment blocks in a started project. This included no comment blocks, only loud comments, only traditional comments, and mixing both loud and traditional blocks.

## Other information

I tested this with Stencil's postcss library as well ([`@stencil-community/poscss`](https://www.npmjs.com/package/@stencil-community/postcss)) and didn't run into any issues.

### DIFF
--- a/src/compiler/docs/test/style-docs.spec.ts
+++ b/src/compiler/docs/test/style-docs.spec.ts
@@ -134,4 +134,17 @@ describe('style-docs', () => {
     parseStyleDocs(styleDocs, styleText);
     expect(styleDocs).toEqual([]);
   });
+
+  it('works with sass loud comments', () => {
+    const styleText = `
+      /*!
+       * @prop --max-width: Max width of the alert
+       */
+      body {
+        color: red;
+      }
+    `;
+    parseStyleDocs(styleDocs, styleText);
+    expect(styleDocs).toEqual([{ name: `--max-width`, docs: `Max width of the alert`, annotation: 'prop' }]);
+  });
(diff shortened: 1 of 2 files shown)


christian-bromann (approved):
👍

src/compiler/docs/test/style-docs.spec.ts
```diff
@@ -134,4 +134,17 @@ describe('style-docs', () => {
...
+
+  it('works with sass loud comments', () => {
+    const styleText = `
+      /*!
+       * @prop --max-width: Max width of the alert
+       */
+      body {
+        color: red;
+      }
+    `;
```
rwaskiewicz, line 146: Could we add a test for multiple comments with the same prop? I know it's a little weird, but was a part of the original repro case https://github.com/duhem-s/stencil-css/blob/main/src/components/my-component/my-component.css#L6

rwaskiewicz (changes requested):

rwaskiewicz (approved):

PR #5708, opened 2024-04-26
test(docs): unit test style doc generation

rwaskiewicz:

See 'new behavior' for the motivation here. But tl;dr - we didn't have tests, now we do 😉 

GitHub Issue Number: N/A

## What is the new behavior?

add unit tests for `src/compiler/docs/generate-doc-data.ts#getDocsStyles`. this was pulled out into a separate pull request from another effort, where we wish to add the `mode` that a stylesheet is associated with by a component's `@Component({})` decorator to the output of `docs-json`. these tests aren't strictly needed for that effort, but establish a baseline for tests taht will be added in the future.

the `getDocsStyles` function was exported in order to test it.

additional jsdoc and type annotations have been added to differentiate fields that are compiler metadata, and those that are used in our documentation intermediate representation.

- [ ] Yes
- [x] No

## Testing

Tests were made to fail/'red' during the dev process. Only then did I make them pass/'green'

## Other information

split from: STENCIL-1269 - CSS Documentation Should Account for Modes

### DIFF
--- a/src/compiler/docs/generate-doc-data.ts
+++ b/src/compiler/docs/generate-doc-data.ts
@@ -305,18 +305,26 @@ const getDocsEvents = (events: d.ComponentCompilerEvent[]): d.JsonDocsEvent[] =>
     }));
 };
 
-const getDocsStyles = (cmpMeta: d.ComponentCompilerMeta): d.JsonDocsStyle[] => {
+/**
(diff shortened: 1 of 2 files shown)


tanner-reits (approved):

christian-bromann (approved):
👍

PR #5718, opened 2024-04-29
feat(docs): add style mode to `docs-json` output

rwaskiewicz:

GitHub Issue Number: N/A

## What is the new behavior?

add a style mode to the output of the `docs-json` output target. when a stencil component is declared using the `styleUrls` property, a user may choose to apply styles depending on a "mode" that they set at runtime:
```tsx
// https://stenciljs.com/docs/component#styleurls

@Component({
  tag: 'todo-list',
  styleUrls: {
     ios: 'todo-list.ios.scss',
     md: 'todo-list.md.scss',
  }
})
export class TodoList {}
```

where the `todo-list.ios.scss` stylesheet will be applied with the mode is set to 'ios' at runtime, and `todo-list.md.scss` will be applied if the mode 'md' is set at runtime.

with this change, documented css properties will be associated with their respective modes in the output of the `docs-json` output target.

for `todo-list.md.scss`:
```sass
/**
 * @prop --button-background: Background of the button
 */
:host {}
```

the mode will now be reported in `docs-json`:
```diff
{
    "name": "--button-background",
    "annotation": "prop",
    "docs": "Background of the button",
+   "mode": "md"
},
```

if a property of the same name exists in more than one mode - e.g. if `--button-background` _also_ existed in the ios mode stylesheet, two separate entries will be generated. this is accomplished by using a composite key for deduplicating/merging arrays consisting of the name and mode of the property/stylesheet

## Documentation

https://github.com/ionic-team/stencil-site/pull/1417/files

- [ ] Yes
- [x] No

## Testing

1. Additional unit tests have been added
2. The `test/docs-json` was rerun to rebuild its output file
3. I tested this in Ionic Framework (expand section below for details)
<details>
  <summary>Details (Click Me)</summary>

  ### Changes to Framework
  1. Checkout `0873dc2ffbc623439a74926937a75f54a28b9709` in Framework, cd into `core`
  2. `npm run build` to get a baseline
  3. `npm run build` this branch, `npm pack` to tarball, `npm i <TARBALL>` in core
  4. Update `stencil.config.ts`
      ```diff
      diff --git a/core/stencil.config.ts b/core/stencil.config.ts
      index cdc5e76fe4..0415de6afa 100644
      --- a/core/stencil.config.ts
      +++ b/core/stencil.config.ts
      @@ -231,7 +231,7 @@ export const config: Config = {
           },
           {
             type: 'docs-json',
      -      file: '../packages/docs/core.json'
      +      file: '../packages/docs/core-new.json'
           },
           {
             type: 'dist-hydrate-script'
      ```
  5. `npm run build` to get a new `docs-json` file
  6. Diff the old docs-json and new docs-json file (attached)
  
[core-json-output.txt](https://github.com/ionic-team/stencil/files/15153537/core-json-output.txt)

</details>

## Other information

STENCIL-1269 CSS Documentation Should Account for Modes

### DIFF
--- a/src/compiler/bundle/ext-transforms-plugin.ts
+++ b/src/compiler/bundle/ext-transforms-plugin.ts
@@ -164,7 +164,7 @@ export const extTransformsPlugin = (
         // Set style docs
         if (cmp) {
           cmp.styleDocs ||= [];
-          mergeIntoWith(cmp.styleDocs, cssTransformResults.styleDocs, (docs) => docs.name);
+          mergeIntoWith(cmp.styleDocs, cssTransformResults.styleDocs, (docs) => `${docs.name},${docs.mode}`);
         }
 
         // Track dependencies
--- a/src/compiler/docs/generate-doc-data.ts
+++ b/src/compiler/docs/generate-doc-data.ts
@@ -1,4 +1,13 @@
-import { flatOne, isOutputTargetDocsJson, join, normalizePath, relative, sortBy, unique } from '@utils';
+import {
+  DEFAULT_STYLE_MODE,
+  flatOne,
+  isOutputTargetDocsJson,
+  join,
+  normalizePath,
+  relative,
+  sortBy,
+  unique,
+} from '@utils';
 import { basename, dirname } from 'path';
 
 import type * as d from '../../declarations';
@@ -316,15 +325,17 @@ export const getDocsStyles = (cmpMeta: d.ComponentCompilerMeta): d.JsonDocsStyle
     return [];
   }
 
-  return sortBy(cmpMeta.styleDocs, (compilerStyleDoc) => compilerStyleDoc.name.toLowerCase()).map(
-    (compilerStyleDoc) => {
-      return {
-        name: compilerStyleDoc.name,
-        annotation: compilerStyleDoc.annotation || '',
-        docs: compilerStyleDoc.docs || '',
-      };
-    },
-  );
+  return sortBy(
+    cmpMeta.styleDocs,
+    (compilerStyleDoc) => `${compilerStyleDoc.name.toLowerCase()},${compilerStyleDoc.mode.toLowerCase()}}`,
+  ).map((compilerStyleDoc) => {
+    return {
+      name: compilerStyleDoc.name,
+      annotation: compilerStyleDoc.annotation || '',
+      docs: compilerStyleDoc.docs || '',
+      mode: compilerStyleDoc.mode && compilerStyleDoc.mode !== DEFAULT_STYLE_MODE ? compilerStyleDoc.mode : undefined,
+    };
+  });
 };
 
 const getDocsListeners = (listeners: d.ComponentCompilerListener[]): d.JsonDocsListener[] => {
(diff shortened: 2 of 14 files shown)


src/compiler/docs/test/generate-doc-data.spec.ts
```diff
@@ -283,12 +290,13 @@ auto-generated content
 
   describe('default values', () => {
     it.each(['', null, undefined])(
-      'defaults the annotation to an empty string if %s is provided',
+      "defaults the annotation to an empty string if '%s' is provided",
```
rwaskiewicz, line 293: I noticed this output was previously reading as:

> defaults the annotation to an empty string if is provided

for the case of the empty string in one of my new tests, and went back and fixed it in a few places here. I can pull it out if folks wish, but it was only two or so tests affected - didn't seem worth it. LMK if you think otherwise, happy to pull it out

src/compiler/docs/generate-doc-data.ts
```diff
@@ -1,4 +1,13 @@
-import { flatOne, isOutputTargetDocsJson, join, normalizePath, relative, sortBy, unique } from '@utils';
+import {
+  DEFAULT_STYLE_MODE,
```
rwaskiewicz, line 2: This is the only new import, just adding it gets us past 120 chars

src/compiler/bundle/ext-transforms-plugin.ts
```diff
@@ -164,7 +164,7 @@ export const extTransformsPlugin = (
         // Set style docs
         if (cmp) {
           cmp.styleDocs ||= [];
-          mergeIntoWith(cmp.styleDocs, cssTransformResults.styleDocs, (docs) => docs.name);
+          mergeIntoWith(cmp.styleDocs, cssTransformResults.styleDocs, (docs) => `${docs.name},${docs.mode}`);
```
rwaskiewicz, line 167: This is _probably_ OK as far as using a comma as the joiner here. I don't think any css property nor mode would ever have a comma in it, LMK if you think this isn't safe. Regardless, it would be easy to change in the future, as it's not exposed to users in our output

alicewriteswrongs: yeah I think it should be fine because I think even if there were a comma somewhere then the string produced should still be unique (that little function is only used for identifying when items are the same item)

src/compiler/docs/test/generate-doc-data.spec.ts
alicewriteswrongs: it's all good 👍

alicewriteswrongs (commented):
I think this all looks good!

One question - would it make sense for us to add a `mode`

rwaskiewicz:
@alicewriteswrongs Good point! Added in https://github.com/ionic-team/stencil/pull/5718/commits/dd72a00248dbfa9d48398b1b769e5b6b4c546e7b

rwaskiewicz:
Hmmm....the output may not be deterministic in terms of ordering. I'll TAL at that later this afternoon (about to start a mini-meeting marathon)

alicewriteswrongs:
cool, yeah we may just need to add a sort somewhere

rwaskiewicz:
@alicewriteswrongs fixed it in https://github.com/ionic-team/stencil/pull/5718/commits/a79fb6374dccfbe03cb8b611b4aee9b44366be40

alicewriteswrongs (approved):

tanner-reits (approved):
LGTM!

PR #5719, opened 2024-04-29
chore(docs): Update contributing guidelines to use local Stencil build

christian-bromann:
It is not straight forward to connect a local Stencil build with a test project. This likely has changed with our Esbuild release.

GitHub Issue Number: N/A

## What is the new behavior?
Update docs to reflect changed conditions and allow to build projects.

- [ ] Yes
- [x] No

## Testing

n/a

## Other information

n/a

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -97,7 +97,7 @@ Afterwards, to clean up:
 2. In the directory of _stencil core_, run `npm unlink`
 
 > [!NOTE]  
-> Alternatively you can reference Stencil from a local directory in the `package.json`, e.g.
+> Instead of linking, you can reference Stencil from a local directory in the `package.json` after updating your project's `tsconfig.json` file, e.g.
 > ```patch
 > diff --git a/package.json b/package.json
 > index 1a8320a..bb1fa3a 100644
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -94,6 +96,28 @@ Afterwards, to clean up:
     2. Remove the modifications to your tsconfig.json
 2. In the directory of _stencil core_, run `npm unlink`
 
+> [!NOTE]  
+> Alternatively you can reference Stencil from a local directory in the `package.json`, e.g.
```
rwaskiewicz, line 100: I _think_ we'd still need the changes here to the `tsconfig.json` here. It might be worth rewording this so that it doesn't seem like this doesn't require changing `tsconfig.json` - unless perhaps it does an I missed something here?

```suggestion
> Instead of linking, you can reference Stencil from a local directory in the `package.json` after updating your project's `tsconfig.json` file, e.g.
```

christian-bromann: Totally makes sense, thanks!

rwaskiewicz (approved):

tanner-reits (approved):

Issue #5526, opened 2024-03-19
bug: v4.12.2 breaks E2E tests


nioe:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

4.12.2

### Current Behavior

We use an async global script, which worked totally fine for us until we updated from **v4.12.1** to **v4.12.2**. After the update, nearly all our E2E Tests started to fail. After some digging I stumbled upon this PR https://github.com/ionic-team/stencil/pull/5158. Seems that there was a bug which caused the initialization of the components not to wait for an async global script. Since this was fixed, I assume that our E2E tests try to access the components too early now, which results in errors like `Node is either not clickable or not an Element` or `shadow root does not exist for element` because the components are not actually loaded yet.

### Expected Behavior

First, even the PR was marked as **breaking**. So personally, I do not understand why it was published as a bug-fix release...

IMHO the `await newE2EPage()` call should wait for the async global script too.

### System Info

```shell
      System: node 20.10.0
    Platform: windows (10.0.19045)
   CPU Model: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz (12 cpus)
    Compiler: .\e2e-bug\node_modules\@stencil\core\compiler\stencil.js
       Build: 1707755243
     Stencil: 4.12.2
  TypeScript: 5.3.3
      Rollup: 2.56.3
      Parse5: 7.1.2
      jQuery: 4.0.0-pre
      Terser: 5.27.0
```

### Steps to Reproduce

* Create a new Stencil project with `npm init stencil@latest`
* Create async global script with some delay and add it to `stencil.config.ts`
* Run tests `npm run test`
* **Default E2E Test fails with `shadow root does not exist for element: my-component` since the component is not yet ready**
  ```shell
   my-component › renders changes to the name data

    shadow root does not exist for element: my-component

      15 |     await page.setContent('<my-component></my-component>');
      16 |     const component = await page.find('my-component');
    > 17 |     const element = await page.find('my-component >>> div');
         |                     ^
      18 |     expect(element.textContent).toEqual(`Hello, World! I'm `);
      19 |
[...]

PR #5721, opened 2024-04-30
fix(testing): wait for Stencil components to be rendered when async global script is used

christian-bromann:
When an asynchronous global script is used, e2e tests don't wait for the global script to finish and instead immediately executes the test. As long as Stencil doesn't hydrate the component under test there is no point in starting the test since the user can't interact with the component.

GitHub Issue Number: #5526

## What is the new behavior?
Add `page.waitForSelector('.hydrated');` to the `waitForStencil` function which will resolve once Stencil hydrates the component and they are ready to be tested.

- [ ] Yes
- [ ] No

## Testing

Added an end-to-end test for this.


christian-bromann:
Unfortunately this won't work since we don't know if the user has custom `hydratedFlag` properties set in the Stencil config.

PR #5725, opened 2024-05-02
fix(build): do not copy polyfills to the `dist` OT unless building es5

### LINKED ISSUE #5416: bug: `stencil build` command generates the folder `dist\esm\polyfills\` folder even if buildEs5=false in config
### Prerequisites

### Stencil Version

4.12.4

### Current Behavior

Even if set the flag `buildEs5: false` in `stencil.config.ts` the output in dist is generating the es5 files:

image

In my company they are finding vulnerabilities in this files that we don't even use 😢 

### Expected Behavior

It is possible not to generate these files that are not being used by the component?

### System Info

```shell
System: node 18.16.0
    Platform: windows (10.0.22621)
   CPU Model: 13th Gen Intel(R) Core(TM) i5-1345U (12 cpus)
    Compiler: C:\Projects\web-components\axa.be.webcomponent.abe-dropdown\node_modules\@stencil\core\compiler\stencil.js
       Build: 1707755243
     Stencil: 4.12.2
  TypeScript: 5.3.3
      Rollup: 2.56.3
      Parse5: 7.1.2
      jQuery: 4.0.0-pre
      Terser: 5.27.0
```

### Steps to Reproduce

1) Generate a new Stencil Component
2) Configure the `stencil.config.ts` with the flag `buildEs5: false`
3) Execute `npm run build`

### Code Reproduction URL

https://github.com/alvarocjunq/stencil-es5-polyfill-issue

### Additional Information

_No response_

alicewriteswrongs:
Prior to this change Stencil will copy polyfills to the `dist` output target whether or not the user has indicated they'll be necessary.

The polyfills comprise two things: copying the polyfills themselves into the 'loader' path, and adding code to the lazy-loader entry points which loads those polyfills. Instead of just assuming that the user wants this, we now gate this behavior on whether `buildEs5` is set on the Stencil configuration.

fixes #5416
STENCIL-1288

Right now we copy a bunch of polyfills which won't be necessary outside of an ES5-only environment even when `buildEs5` is not set. This adds some extra bulk to the `dist` build output.

See #5416 for more details.

## What is the new behavior?

If the user has set `buildEs5` to true then we copy the polyfills, otherwise we don't do anything.

- [ ] Yes
- [x] No

## Testing

The best way to test this is to build a sample stencil project with `@stencil/core@latest` and then with a version based on this build and then compare the output.

First do:

```sh
cd /tmp
npm init stencil@latest component test-no-polyfills
cd test-no-polyfills
npm install
npm run build
```

then edit `.gitignore` to remove `dist/` and `loader/` from the ignore. Then you can follow the setup steps in [`CONTRIBUTING.md`](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md) for local dev (e.g. the tsconfig `paths` stuff).

Then if you check out this branch in your stencil repo locally and run `npm link` you should be able to use this script to test things out in your new stencil project:

```bash
#!/bin/bash

npm unlink @stencil/core

npm install @stencil/core@latest

npm run build

git add .

git commit -m "build with @stencil/core@latest"

npm link @stencil/core

npm run build

git add .

git commit -m "build with dev build"
```

save that to like `test.sh` or something in your little project, throw a `chmod +x` on there and you're good to go.

Then you can do

```sh
git diff $(git rev-parse @~)
```

to see the diff between the `@stencil/core@latest` build and the dev build

### DIFF
--- a/src/compiler/output-targets/test/output-lazy-loader.spec.ts
+++ b/src/compiler/output-targets/test/output-lazy-loader.spec.ts
@@ -1,0 +1,73 @@
+import type * as d from '@stencil/core/declarations';
+import { mockBuildCtx, mockCompilerCtx, mockCompilerSystem, mockValidatedConfig } from '@stencil/core/testing';
+import { DIST, resolve } from '@utils';
+
+import { validateDist } from '../../config/outputs/validate-dist';
+import { outputLazyLoader } from '../output-lazy-loader';
+
+function setup(configOverrides: Partial<d.ValidatedConfig> = {}) {
+  const sys = mockCompilerSystem();
+  const config: d.ValidatedConfig = mockValidatedConfig({
+    ...configOverrides,
+    configPath: '/testing-path',
+    buildAppCore: true,
+    namespace: 'TestApp',
+    outputTargets: [
+      {
+        type: DIST,
+        dir: 'my-test-dir',
+      },
+    ],
+    srcDir: '/src',
+    sys,
+  });
+
+  config.outputTargets = validateDist(config, config.outputTargets);
+
+  const compilerCtx = mockCompilerCtx(config);
+  const writeFileSpy = jest.spyOn(compilerCtx.fs, 'writeFile');
+  const buildCtx = mockBuildCtx(config, compilerCtx);
+
+  return { config, compilerCtx, buildCtx, writeFileSpy };
+}
+
+describe('Lazy Loader Output Target', () => {
+  let config: d.ValidatedConfig;
+  let compilerCtx: d.CompilerCtx;
+  let writeFileSpy: jest.SpyInstance;
+
+  afterEach(() => {
+    writeFileSpy.mockRestore();
+  });
+
+  it('should write code for initializing polyfills when buildEs5=true', async () => {
+    ({ config, compilerCtx, writeFileSpy } = setup({ buildEs5: true }));
+    await outputLazyLoader(config, compilerCtx);
+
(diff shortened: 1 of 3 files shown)


src/compiler/output-targets/test/output-lazy-loader.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+    writeFileSpy.mockRestore();
+  });
+
+  it('should write code for initializing polyfills when buildEs5=true', async () => {
+    ({ config, compilerCtx, writeFileSpy } = setup({ buildEs5: true }));
+    await outputLazyLoader(config, compilerCtx);
+
+    const expectedIndexOutput = `export * from '../esm/polyfills/index.js';
+export * from '../esm-es5/loader.js';`;
+    expect(writeFileSpy).toHaveBeenCalledWith(resolve('/my-test-dir/loader/index.js'), expectedIndexOutput);
```
alicewriteswrongs, line 49: this is a little funny (the `resolve` call I mean) but I had to do this to get the tests passing on CI on Windows, you know, sometimes you gotta do what you gotta do

src/compiler/output-targets/test/output-lazy-loader.spec.ts
alicewriteswrongs, line 1: this is a new spec file, we didn't previously have coverage for this file in particular

alicewriteswrongs (commented):
noting two things

rwaskiewicz (approved):

christian-bromann (approved):

PR #5737, opened 2024-05-07
fix(compiler): deprecate `scriptDataOpts`

christian-bromann:
Stencil supports a way to inject options into the script tag that then will be parsed into the runtime e.g. to set the `syncQueue` flag in [`CustomElementsDefineOptions`](https://github.com/ionic-team/stencil/blob/1d0fd962ed2211afea4a7e5e42222723b0a6184a/src/declarations/stencil-public-runtime.ts#L1865). This doesn't seem to be commonly used anymore so it is fine to deprecated and remove in v5.

## What is the new behavior?
Depcreate the `scriptDataOpts` option.

## Documentation

Raised PR with docs updates in https://github.com/ionic-team/stencil-site/pull/1425

- [ ] Yes
- [x] No

## Testing

n/a

## Other information

n/a

### DIFF
--- a/src/declarations/stencil-public-compiler.ts
+++ b/src/declarations/stencil-public-compiler.ts
@@ -327,7 +327,7 @@ interface ConfigExtrasBase {
    * It is possible to assign data to the actual `<script>` element's `data-opts` property,
    * which then gets passed to Stencil's initial bootstrap. This feature is only required
    * for very special cases and rarely needed. Defaults to `false`.
-   * @deprecated This option has been deprecated and will be removed in a future version of Stencil.
+   * @deprecated This option has been deprecated and will be removed in a future major version of Stencil.
    */
   scriptDataOpts?: boolean;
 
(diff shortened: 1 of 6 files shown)


src/declarations/stencil-public-compiler.ts
```diff
@@ -327,6 +327,7 @@ interface ConfigExtrasBase {
    * It is possible to assign data to the actual `<script>` element's `data-opts` property,
    * which then gets passed to Stencil's initial bootstrap. This feature is only required
    * for very special cases and rarely needed. Defaults to `false`.
+   * @deprecated This option has been deprecated and will be removed in a future version of Stencil.
```
alicewriteswrongs, line 330: Minor suggestion, just looking at some of our other deprecation notices I think they're formatted like this:

```suggestion
   * @deprecated This option has been deprecated and will be removed in a future major version of Stencil.
```

alicewriteswrongs (approved):
looks good! just one thing that we might want to address

tanner-reits (approved):

PR #5739, opened 2024-05-08
chore(node): update support to node v22

alicewriteswrongs:
Updating to support node v22

## What is the new behavior?

it's node v22! you can see some details about what's included in the release here: https://nodejs.org/en/blog/announcements/v22-release-announce

- [ ] Yes
- [x] No

## Testing

If all of our automated tests are passing and you can build a sample project without issues then this should be good!

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -45,6 +45,7 @@
     },
     "pretty": true,
     "resolveJsonModule": true,
+    "skipLibCheck": true,
     "sourceMap": true,
     "strictBindCallApply": true,
     "strictFunctionTypes": true,
--- a/.github/workflows/test-analysis.yml
+++ b/.github/workflows/test-analysis.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node: ['16', '18', '20']
+        node: ['16', '18', '20', '22']
         os: ['ubuntu-latest', 'windows-latest']
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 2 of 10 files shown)


tsconfig.json
```diff
@@ -45,6 +45,7 @@
     },
     "pretty": true,
     "resolveJsonModule": true,
+    "skipLibCheck": true,
```
christian-bromann, line 48: Are there any TS issues introduced with this update?

alicewriteswrongs: yeah I saw an error where [path-scurry](https://www.npmjs.com/package/path-scurry) doesn't play nicely with the node typings:

```
node_modules/path-scurry/dist/commonjs/index.d.ts:117:31 - error TS2720: Class 'PathBase' incorrectly implements class 'Dirent'. Did you mean to extend 'Dirent' and inherit its members as a subclass?
  Property 'parentPath' is missing in type 'PathBase' but required in type 'Dirent'.

117 export declare abstract class PathBase implements Dirent {
                                  ~~~~~~~~

  node_modules/@types/node/fs.d.ts:247:9
    247         parentPath: string;
                ~~~~~~~~~~
    'parentPath' is declared here.


Found 1 error in node_modules/path-scurry/dist/commonjs/index.d.ts:117
```

alicewriteswrongs: I kind of like having `skipLibCheck` on because it does skip a bunch of type-checking and makes builds faster

alicewriteswrongs: just tested and I think it makes calling `tsc` around 3 seconds faster to have the option on

christian-bromann: Makes sense to me 👍

christian-bromann (approved):
👍

tanner-reits (approved):

PR #5762, opened 2024-05-13
fix(runtime): always throw if component can not be loaded

### LINKED ISSUE #5759: bug: Cannot read properties of undefined (reading 'isProxied')
### Prerequisites

### Stencil Version

master

### Current Behavior

The error 
> Cannot read properties of undefined (reading 'isProxied')

is reported a lot in our error queue.

### Expected Behavior

No error is reported.

### System Info

```shell
System: node 20.5.0
    Platform: darwin (23.4.0)
   CPU Model: Apple M2 Pro (10 cpus)
    Compiler: <private>/node_modules/@stencil/core/compiler/stencil.js
       Build: 1708455660
     Stencil: 4.12.3 🐍
  TypeScript: 5.3.3
      Rollup: 2.56.3
      Parse5: 7.1.2
      jQuery: 4.0.0-pre
      Terser: 5.27.1
```

### Steps to Reproduce

I was not able to reproduce it with a stencil project but I went down the stencil code to analyze what is happening in I think I found the issue. 

The root of the issue is located here:

` if (BUILD.member && !Cstr.isProxied) {`
https://github.com/ionic-team/stencil/blob/main/src/runtime/initialize-component.ts#L53

Cstr seems to be undefined. That variable is the result of a loadModule call

` Cstr = loadModule(cmpMeta, hostRef, hmrVersionId);`

In the error case, the call returns a promise as it should and you can see here
https://github.com/ionic-team/stencil/blob/main/src/client/client-load-module.ts#L45

Next the result is awaited here and overrides the promise with the actual content:
` Cstr = await Cstr;`
https://github.com/ionic-team/stencil/blob/main/src/runtime/initialize-component.ts#L47

But if the promise fulfills as undefined, the code will break. This can happen in case the connection is getting lost during loading of the file or the file is not there at all. 

Usually the code like this breaks and throws an error but the error case of the dynamic loading of the module is already handles which means the code will continue as normal:
https://github.com/ionic-team/stencil/blob/main/src/client/client-load-module.ts#L56

I do have a standalone example to reproduce the issue:
```
  let Cstr = import('./not-there.js').then(() => {}, console.error);
  if (Cstr.then) {
    Cstr = await tmp;
  }

  console.log('should not be undefined', Cstr);  
```

### Code Reproduction URL

https://codepen.io/wittemann/pen/xxNbeqK

### Additional Information

Might be connected to https://github.com/ionic-team/stencil/issues/2774

christian-bromann:
Due to the fact we don't have strict null checks we missed this area where `Cstr` can be null at a point where we wanted to access a property (e.g. `isProxied`).

GitHub Issue Number:
fixes #5759

## What is the new behavior?
I removed the condition that errors are only thrown if `(BUILD.isDev || BUILD.isDebug)` is truthy. It doesn't make much sense to me why we only would want this error to be elevated in these conditions.

## Documentation

n/a

- [ ] Yes
- [x] No

## Testing

n/a

## Other information

n/a

### DIFF
--- a/src/client/client-load-module.ts
+++ b/src/client/client-load-module.ts
@@ -27,7 +27,7 @@ export const loadModule = (
   cmpMeta: d.ComponentRuntimeMeta,
   hostRef: d.HostRef,
   hmrVersionId?: string,
-): Promise<d.ComponentConstructor> | d.ComponentConstructor => {
+): Promise<d.ComponentConstructor | undefined> | d.ComponentConstructor => {
   // loadModuleImport
   const exportName = cmpMeta.$tagName$.replace(/-/g, '_');
   const bundleId = cmpMeta.$lazyBundleId$;
--- a/src/hydrate/platform/index.ts
+++ b/src/hydrate/platform/index.ts
@@ -6,7 +6,7 @@ let customError: d.ErrorHandler;
 
 export const cmpModules = new Map<string, { [exportName: string]: d.ComponentConstructor }>();
 
-const getModule = (tagName: string): d.ComponentConstructor => {
+const getModule = (tagName: string): d.ComponentConstructor | null => {
   if (typeof tagName === 'string') {
     tagName = tagName.toLowerCase();
     const cmpModule = cmpModules.get(tagName);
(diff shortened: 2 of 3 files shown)


alicewriteswrongs (changes requested):
I think the change makes sense - j/w if you found any way to reproduce the failure case in a test so we can sort of make sure it fails the way we think it should?

christian-bromann:
> j/w if you found any way to reproduce the failure case in a test so we can sort of make sure it fails the way we think it should?

I looked into this and it doesn't seem to be straightforward. I am not able to import the internals in a way that would allow me to modify the runtime in order to reproduce this. The OP of the issue also mentioned that they are not able to reproduce given issue in a new Stencil project.

alicewriteswrongs (approved):
makes sense, I think we're good to go ahead then 👍

Issue #3554, opened 2022-08-25
bug: watch decorator doesn't work if we change the original tag name


urielblanco:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/.github/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

2.13.0

### Current Behavior

My problem is that the `@Watch` decorator only works if I keep the original tag name.

- If I import a custom element from my stencil library and define it as a custom element like this:

### My vanilla js app
```
import { MyComponent } from '@my-lib/dist/components/my-component.js';

customElements.define('my-component', MyComponent);
```
### My simple component

```
import { Component, Prop, h, Watch } from '@stencil/core';

@Component({
  tag: 'my-component',
  styleUrl: 'my-component.css',
  shadow: true,
})
export class MyComponent {
  /**
   * The first name
   */
  @Prop({ reflect: true, mutable: true }) checked?: boolean = false;

  @Watch('checked')
  onValueChange(): void {
    console.log('trigger from watch', this.checked);
  }

  private readonly onChange = (e: InputEvent): void => {
    const isElementChecked = (e.target as HTMLInputElement).checked;
    if (isElementChecked !== this.checked) {
      this.checked = isElementChecked;
    }
  };

  render() {
    return <input onChange={this.onChange} type="checkbox" name="test" id="" />;
  }
}
```

With this example, I can see in the console that the watch func triggered 🙌🏽

### Expected Behavior

Suppose I change the original tag name when I define my component in **customElement.define()** I hope that the ```@Watch``` decorator keeps working but it doesn't.

### Steps to Reproduce

You can recreate my example in **Current behavior** or download it and run the scripts like this:

```
npm i
npm run build
npm run start:test
```

And that's it! If you change the tag name from "my-component" to "my-component-test" the watch func stop working and you can't see in the console my message.

### Code Reproduction URL

https://github.com/urielblanco/test-watch

urielblanco:
The community in slack was help me and one guy said that the problem is here

https://github.dev/ionic-team/stencil/blob/c05cebec911ca0ef4acf7e4673e24e7801316b0d/src/runtime/initialize-component.ts#L87-L88

```
[...]

PR #5767, opened 2024-05-14
fix(runtime): support watch for components with custom tag names

christian-bromann:
If users define their component with a custom tag name, the `@Watch` decorator doesn't work.

GitHub Issue Number: #3554
STENCIL-546

## What is the new behavior?
Instead of using the component meta data we just pick the tag name for the component directly.

## Documentation

n/a

- [ ] Yes
- [x] No

## Testing

Added an unit test for this.

## Other information

n/a

### DIFF
--- a/test/wdio/setup.ts
+++ b/test/wdio/setup.ts
@@ -13,7 +13,8 @@ const testRequiresManualSetup =
   window.__wdioSpec__.includes('custom-elements-output-tag-class-different') ||
   window.__wdioSpec__.includes('custom-elements-delegates-focus') ||
   window.__wdioSpec__.includes('custom-elements-output') ||
-  window.__wdioSpec__.includes('global-script');
+  window.__wdioSpec__.includes('global-script') ||
+  window.__wdioSpec__.endsWith('custom-tag-name.test.tsx');
 
 /**
  * setup all components defined in tests except for those where we want ot manually setup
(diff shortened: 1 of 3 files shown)


test/wdio/setup.ts
```diff
@@ -13,7 +13,8 @@ const testRequiresManualSetup =
   window.__wdioSpec__.includes('custom-elements-output-tag-class-different') ||
   window.__wdioSpec__.includes('custom-elements-delegates-focus') ||
   window.__wdioSpec__.includes('custom-elements-output') ||
-  window.__wdioSpec__.includes('global-script');
+  window.__wdioSpec__.includes('global-script') ||
+  window.__wdioSpec__.endsWith('custom-tag-name.test.tsx');
```
alicewriteswrongs, line 17: we need this because we don't want the normal `.define` call to happen, right?

alicewriteswrongs (changes requested):
change makes sense to me! Just noticed I think one little typo in the comment and had one clarification

tanner-reits (approved):
LGTM

test/wdio/setup.ts
christian-bromann: Yes, so we avoid registering all the components to the registry to test the behavior where user would do this manually.

alicewriteswrongs (approved):

PR #5772, opened 2024-05-17
chore(internal): remove rollup-plugin-sourcemaps dependency

christian-bromann:

We are having `rollup-plugin-sourcemaps` listed as dependency but it doesn't seem we need it.

## What is the new behavior?

Don't have it anymore.

- [ ] Yes
- [ ] No

## Testing

n/a

## Other information

It was added when Rollup was used: https://github.com/ionic-team/stencil/pull/3108/files


tanner-reits (approved):
Confirmed I didn't see this used in the repo anywhere

alicewriteswrongs (approved):

Issue #5766, opened 2024-05-14
bug: Our build break since `v4.18.1`


deleonio:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

4.18.1

### Current Behavior

We update `stencil` in this PR: https://github.com/public-ui/kolibri/pull/6418/files#diff-2c76dc06185f93bec73660e15338433b95eac6707317564c0f778f5b9abe446c

```bash
FAIL src/components/quote/test/snapshot.spec.tsx
    ● Test suite failed to run
  
      ENOENT: no such file or directory, open 'node:path'
  
        at Runtime.readFile (../../node_modules/.pnpm/jest-runtime@26.6.3_ts-node@10.9.2/node_modules/jest-runtime/build/index.js:1987:21)
        at Object.760 (../../node_modules/.pnpm/@stencil+core@4.18.1/node_modules/@stencil/core/sys/node/glob.js:1:2995)
        at s (../../node_modules/.pnpm/@stencil+core@4.18.1/node_modules/@stencil/core/sys/node/glob.js:1:85614)
        at Object.576 (../../node_modules/.pnpm/@stencil+core@4.18.1/node_modules/@stencil/core/sys/node/glob.js:1:67906)
```

### Expected Behavior

Have no dependency issues.

### Steps to Reproduce

Sorry. I do not know, how you can reproduce this exception.

### Code Reproduction URL

https://github.com/public-ui/kolibri/pull/6418/files#diff-2c76dc06185f93bec73660e15338433b95eac6707317564c0f778f5b9abe446c

### Additional Information

Thank you!

alicewriteswrongs:
Hey @deleonio thanks for reporting this, I wasn't able to reproduce that error locally my testing. Would you possibly be able to put together a minimal reproduction? Without that we have a hard time debugging errors.

deleonio:
@alicewriteswrongs 

Here I prepare a reproduction sample: https://github.com/deleonio/kolibri-library/tree/reproduction-5766

4.18.0 works
4.18.1 does not work

christian-bromann:
[...]

PR #5784, opened 2024-05-21
fix(esbuild): remove all `node:` imports from glob script to keep support for Jest v26

christian-bromann:
With Stencil v4.18.1 Stencil broke support for Jest v26 due to the fact that transient dependencies for `glob` in `sys/node/glob.js` started to use imports with `node:` prefixes.

GitHub Issue Number: #5766

## What is the new behavior?
I enhanced the sys node build script to remove these prefixes. This won't have any impact on the script and allows us to keep support for Jest v26.

- [ ] Yes
- [ ] No

## Testing

- Check out https://github.com/deleonio/kolibri-library/tree/reproduction-5766
- run `pnpm install` and verify that test pass with `npm test`
- now update Stencil to latest, rerun test and verify that test fail due to `no such file or directory, open 'node:path'`
- now install `@stencil/core@4.18.2-dev.1716246842.b6713e1` (dev build of this branch) and rerun tests, verify that tests pass

## Other information

I was under the assumption that we are testing running tests in Jest v26, I will take a look if we could have caught this before release.

### DIFF
--- a/scripts/esbuild/sys-node.ts
+++ b/scripts/esbuild/sys-node.ts
@@ -103,7 +103,7 @@ async function sysNodeExternalBundles(opts: BuildOptions) {
    */
   const globOutputPath = join(opts.output.sysNodeDir, 'glob.js');
   const glob = fs.readFileSync(globOutputPath, 'utf8');
-  fs.writeFileSync(globOutputPath, glob.replace(/require\("node:/g, 'require("'))
+  fs.writeFileSync(globOutputPath, glob.replace(/require\("node:/g, 'require("'));
 
   // open-in-editor's visualstudio.vbs file
   // TODO(STENCIL-1052): remove once Rollup -> esbuild migration is complete
--- a/scripts/esbuild/sys-node.ts
+++ b/scripts/esbuild/sys-node.ts
@@ -96,6 +96,15 @@ async function sysNodeExternalBundles(opts: BuildOptions) {
     bundleExternal(opts, opts.output.devServerDir, cachedDir, 'ws.js'),
   ]);
 
+  /**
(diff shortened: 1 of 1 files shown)


scripts/esbuild/sys-node.ts
```diff
@@ -96,6 +96,15 @@ async function sysNodeExternalBundles(opts: BuildOptions) {
...
   ]);
 
+  /**
+   * Some of globs dependencies are using imports with a `node:` prefix which
+   * is not supported by Jest v26. This is a workaround to remove the `node:`.
+   * TODO(STENCIL-1323): remove once we deprecated Jest v26 support
+   */
+  const globOutputPath = join(opts.output.sysNodeDir, 'glob.js');
+  const glob = fs.readFileSync(globOutputPath, 'utf8');
+  fs.writeFileSync(globOutputPath, glob.replace(/require\("node:/g, 'require("'));
```
tanner-reits, line 106: Is it better to just revert glob and block that upgrade until we drop Jest 26 support (in v5)?

christian-bromann: Unfortunately that wouldn't work as `glob` itself is not the culprit here, it's `glob` dependencies, e.g. `path-scurry`, that bring in these node imports. If we would use pnpm we could use the [`overwrite`](https://pnpm.io/package_json#pnpmoverrides) or in Yarn we could use [`resolutions`](https://classic.yarnpkg.com/lang/en/docs/selective-version-resolutions/) but there is nothing in NPM to do this. Hence tweaking the build artifact seemed like the easiest approach.

alicewriteswrongs (approved):

tanner-reits (approved):

PR #5787, opened 2024-05-22
feat(runtime): support already-existing shadow roots

alicewriteswrongs:
This adds some checks when a Stencil component is booting up that will
re-use a shadow root if the host element already has one, instead of
just calling `this.attachShadow` in all cases.

Additionally, Stencil's virtual DOM will now cleanly take over any
elements which are already present in its root element when it first
runs. This will allow it to re-use elements that were already created in
the shadow root, as well as clean up any stray elements that shouldn't
be there.

Together these changes will allow the Stencil runtime to smoothly take
over a shadow root which was created via a `<template>` tag in the
page's HTML, i.e. via declarative shadow DOM (DSD).

**NOTE** this _only_ enables the Stencil runtime to re-use a shadow root
created via DSD, it does not add any support to Stencil for rendering
a suitable `<template>...` HTML string.

STENCIL-1316

GitHub Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/src/runtime/bootstrap-custom-element.ts
+++ b/src/runtime/bootstrap-custom-element.ts
@@ -94,6 +94,15 @@ export const proxyCustomElement = (Cstr: any, compactMeta: d.ComponentRuntimeMet
           } else {
             this.attachShadow({ mode: 'open' });
           }
+        } else {
+          // we want to check to make sure that the mode for the shadow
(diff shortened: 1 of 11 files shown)


alicewriteswrongs:
closing in favor of #5792

PR #5789, opened 2024-05-22
fix(emit): don't emit test files

### LINKED ISSUE #5788: bug: Test files copied to output target
### Prerequisites

### Stencil Version

4.18.0

### Current Behavior

In a new stencil project, test files are copied over to `dist`. In some cases, Jest attempts to run these files, creating errors

### Expected Behavior

Test files should be excluded of the distribution files (or, if there is a reason to copy them there, be ignored by jest)

### System Info

```shell
Tested on both Windows and MacOS, npm 10.7.0
```

### Steps to Reproduce

`npm test`
`find dist -type f -name '*.spec.*'`
`find dist -type f -name '*.e2e.*'`

### Code Reproduction URL

https://github.com/mrtnmgs/mockbug

### Additional Information

This is similar to the bug I logged a couple days ago https://github.com/ionic-team/stencil/issues/5781. I had not realized the problem was much larger than just the mocks.

alicewriteswrongs:
This fixes a bug introduced in #4315, which upgraded to TypeScript 5. In that PR we had to change the way that we prevented certain files from being emitted by the typescript compiler because in the 5.0 release our previous approach stopped working.

However, in porting over to a new approach that worked with TS 5.0 there was an oversight. I misunderstood the intent of the old code as being to merely prevent writing `.d.ts` files in the output, when actually it was about preventing the compiled JavaScript from being written for `.e2e.ts` and `.spec.ts` files.

This change ensures that we no longer emit these files.

fixes #5788
STENCIL-1325

GitHub Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/src/compiler/transpile/run-program.ts
+++ b/src/compiler/transpile/run-program.ts
@@ -43,16 +43,18 @@ export const runTsProgram = async (
   const typesOutputTarget = config.outputTargets.filter(isOutputTargetDistTypes);
   const emittedDts: string[] = [];
   const emitCallback: ts.WriteFileCallback = (emitFilePath, data, _w, _e, tsSourceFiles) => {
-    if (emitFilePath.endsWith('.js') || emitFilePath.endsWith('js.map')) {
-      updateModule(config, compilerCtx, buildCtx, tsSourceFiles[0], data, emitFilePath, tsTypeChecker, null);
-    } else if (
-      emitFilePath.endsWith('.e2e.d.ts') ||
-      emitFilePath.endsWith('.spec.d.ts') ||
-      emitFilePath === 'e2e.d.ts' ||
-      emitFilePath === 'spec.d.ts'
+    if (
+      emitFilePath.includes('e2e.') ||
+      emitFilePath.includes('spec.') ||
+      emitFilePath.endsWith('e2e.d.ts') ||
+      emitFilePath.endsWith('spec.d.ts')
     ) {
       // we don't want to write these to disk!
       return;
+    }
+
+    if (emitFilePath.endsWith('.js') || emitFilePath.endsWith('js.map')) {
+      updateModule(config, compilerCtx, buildCtx, tsSourceFiles[0], data, emitFilePath, tsTypeChecker, null);
     } else if (emitFilePath.endsWith('.d.ts')) {
       const srcDtsPath = normalizePath(tsSourceFiles[0].fileName);
       const relativeEmitFilepath = getRelativeDts(config, srcDtsPath, emitFilePath);


src/compiler/transpile/run-program.ts
```diff
@@ -43,16 +43,20 @@ export const runTsProgram = async (
...
   const emittedDts: string[] = [];
   const emitCallback: ts.WriteFileCallback = (emitFilePath, data, _w, _e, tsSourceFiles) => {
-    if (emitFilePath.endsWith('.js') || emitFilePath.endsWith('js.map')) {
-      updateModule(config, compilerCtx, buildCtx, tsSourceFiles[0], data, emitFilePath, tsTypeChecker, null);
-    } else if (
-      emitFilePath.endsWith('.e2e.d.ts') ||
-      emitFilePath.endsWith('.spec.d.ts') ||
+    if (
+      emitFilePath.includes('.e2e.') ||
+      emitFilePath.includes('/e2e.') ||
```
christian-bromann, line 48: I feel like this is a bit to generous, if for whatever reason someone calls their component `e2e.ts` , it would not be compiled right? Is there a chance we can verify if the file content has Stencil primitivis for testing?

alicewriteswrongs: hmm we could, although this change is pretty much just restoring what the behavior was before #4315. Maybe it's a question of documenting that files which are named like this for some reason wouldn't be emitted?

christian-bromann: Correct me if I am wrong but before #4315 we didn't had this whole `if` block. What is the intention of including `emitFilePath.includes('/e2e.')` and `emitFilePath.includes('/spec.')`? They don't match with common naming patterns afaics. I'ld suggest maybe only:

```ts
if (
  emitFilePath.includes('.e2e.ts') ||
  emitFilePath.includes('.spec.ts') ||
  emitFilePath === 'e2e.d.ts' ||
  emitFilePath === 'spec.d.ts'
) {
  // we don't want to write these to disk!
  return;
}
```

?

alicewriteswrongs: Yeah so we didn't have this block, but we did something similar elsewhere - here's the relevant diff: https://github.com/ionic-team/stencil/pull/4315/files#diff-641e8e3ba21b8a7166c9350fe42c34c2b33ccbc220173ecdda7b451da8a423ba

alicewriteswrongs: I just pulled in the `/spec.` thing because that was there in the old impl - I guess maybe if you had `src/components/my-component/my-component.tsx` maybe you could have `spec.ts` in the same directory?

christian-bromann: Oh now I see it. Then it would make sense to have this, no?
```ts
if (
  emitFilePath.endsWith('e2e.ts') ||
  emitFilePath.endsWith('spec.ts') ||
  emitFilePath.endsWith('e2e.d.ts') ||
  emitFilePath.endsWith('spec.d.ts')
) {
  // we don't want to write these to disk!
  return;
}
```
Above would catch both cases even though it would surprise me if folks actually used `spec.ts` as test file names and would expect Stencil to detect this as a test file. Anyway, very minor nit, feel free to move forward with what makes sense.

christian-bromann (approved):
LGTM 👍

src/compiler/transpile/run-program.ts
alicewriteswrongs: yeah, I think that makes sense to me, I'll adjust it a bit!

I may also try to get a little regression test for this using one of our end-to-end projects

alicewriteswrongs:
@christian-bromann I updated this following your feedback!

christian-bromann (approved):
👍

tanner-reits (approved):

PR #5829, opened 2024-06-13
feat(testing): add `toHaveLastReceivedEventDetail` event spy matcher

### LINKED ISSUE #2488: Missing `toHaveLastReceivedEventDetail` convenience matcher
**Stencil version:**

```
 @stencil/core@1.14.0
```

**I'm submitting a:**

[ ] bug report
[x] feature request

**Current behavior:**

There is no convenience method to assert on the last received event detail. [`toHaveNthReceivedEventDetail`](https://github.com/ionic-team/stencil/blob/22f3f23783e5bd0790c13c0e7f3720c0d7be28c0/src/declarations/stencil-private.ts#L2290) can be used in the meantime, but it would be nice to not have to manage the last event index.

**Expected behavior:**

Similar to [`toHaveFirstReceivedEventDetail`](https://github.com/ionic-team/stencil/blob/22f3f23783e5bd0790c13c0e7f3720c0d7be28c0/src/declarations/stencil-private.ts#L2284), having a `toHaveLastReceivedEventDetail` matcher would simplify assertions on the most recent event.

tanner-reits:

Stencil offers event matchers for `toHaveFirstReceivedEventDetail` and `toHaveNthReceivedEventDetail`.

GitHub Issue Number: Closes #2488 

## What is the new behavior?

Adds a matcher for `toHaveLastReceivedEventDetail` that will check the `detail` content of an event emitted from a Stencil component

- [ ] Yes
- [ ] No

## Testing

Tested the matchers (for first and last event) in a component starter.

### DIFF
--- a/src/testing/jest/jest-27-and-under/matchers/events.ts
+++ b/src/testing/jest/jest-27-and-under/matchers/events.ts
@@ -91,6 +91,33 @@ export function toHaveFirstReceivedEventDetail(eventSpy: d.EventSpy, eventDetail
 
   const pass = deepEqual(eventSpy.firstEvent.detail, eventDetail);
 
+  expect(eventSpy.firstEvent.detail).toEqual(eventDetail);
+
+  return {
+    message: () => `expected event "${eventSpy.eventName}" detail to ${pass ? 'not ' : ''}equal`,
+    pass: pass,
+  };
+}
+
+export function toHaveLastReceivedEventDetail(eventSpy: d.EventSpy, eventDetail: any) {
+  if (!eventSpy) {
+    throw new Error(`toHaveLastReceivedEventDetail event spy is null`);
+  }
+
+  if (typeof (eventSpy as any).then === 'function') {
+    throw new Error(`event spy must be a resolved value, not a promise, before it can be tested`);
+  }
+
+  if (!eventSpy.eventName) {
+    throw new Error(`toHaveLastReceivedEventDetail did not receive an event spy`);
+  }
+
+  if (!eventSpy.firstEvent) {
+    throw new Error(`event "${eventSpy.eventName}" was not received`);
+  }
+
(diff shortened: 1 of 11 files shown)


src/testing/jest/jest-27-and-under/matchers/events.ts
```diff
@@ -91,6 +91,33 @@ export function toHaveFirstReceivedEventDetail(eventSpy: d.EventSpy, eventDetail
 
   const pass = deepEqual(eventSpy.firstEvent.detail, eventDetail);
 
+  expect(eventSpy.firstEvent.detail).toEqual(eventDetail);
```
tanner-reits, line 94: Don't like how Git is showing the diff, but this line used to be: `expect(eventSpy.lastEvent.detail).toEqual(eventDetail)`

Changed it to use `firstEvent`, since, well, this is supposed to be checking what event data the spy first received. Idk why this was `lastEvent` before, but, in testing, I setup a component to emit multiple events and checked that the assertions using `toHaveFirstReceivedEventDetail` and `toHaveLastReceivedEventDetail` were working as expected with the emitted values.

I'm not sure if this is a breaking change, but seems like the original implementation was wrong in the first place 😅

christian-bromann (commented):
Do you think we can add a test for this matcher?

tanner-reits:
@christian-bromann I don't know if I'll be able to test this. The matchers use `expect` under-the-hood, so testing failed cases results in the test failing early because the assertion within the matcher fails

christian-bromann:
> The matchers use `expect` under-the-hood

I thought it would be possible somewhere in `test/end-to-end` where we call `npm run test.spec` which triggers the Stencil testrunner.

christian-bromann (approved):
LGTM 👍

PR #5830, opened 2024-06-14
fix(testing): allow to re-use pages across it blocks

### LINKED ISSUE #3720: bug: E2E beforeAll() error: The page was already closed
### Prerequisites

### Stencil Version

2.18.1

### Current Behavior

The Jest function `beforeAll()` cannot be used to set up a `page` (and possibly even other suite-wide objects) . It seems to end after the first `it()` and won't work for the second `it()`, instead it throws the error `"The page was already closed"`.

Example:

```
describe("suite", () => {
  let page

  beforeAll(async () => {
    page = await newE2EPage()
  })

  it("1", () => {
    const someEl = await page.find('my-component >>> div') // works so far
  })

  it("2", () => {
    const someEl = await page.find('my-component >>> div') // error "The page was already closed" 
  })
})
```

It works fine when using `beforeEach()` instead of `beforeAll()`, however `beforeEach()` slows down tests a lot because `newE2EPage()` takes a lot of time and isn't required in many cases.

### Expected Behavior

`beforeAll()` works just like `beforeEach()`. According to the Jest docs, `beforeAll()` and its counterpart `afterAll()` are built for exactly that: https://jestjs.io/docs/setup-teardown

### Steps to Reproduce

Clone the `stencil-component-starter` repo and alter `my-component.e2e.ts` as described above.

### Code Reproduction URL

https://github.com/ionic-team/stencil-component-starter

### Additional Information

_No response_

christian-bromann:
Currently Stencil tears down all pages in an `afterEach` hook which causes issues when you re-use a page object across multiple it blocks, e.g.:

```tsx
import { type E2EPage, newE2EPage } from '@stencil/core/testing';

describe('do not throw page already closed if page was defined in before(All) hook', () => {
  let page: E2EPage

  beforeAll(async () => {
    page = await newE2EPage()
  })

  it("first test", async () => {
    const p = await page.find('html')
    expect(p).not.toBeNull()
  })

  it("second test", async () => {
    const p = await page.find('html')
    expect(p).not.toBeNull()
  })
})
```

fixes #3720

## What is the new behavior?

Let's not remove all pages after each `it` block and instead do it after all it blocks or when the user is creating a `newE2EPage` object.

## Documentation

n/a

- [ ] Yes
- [x] No

## Testing

Added an e2e test to verify this expected behavior works.

## Other information

n/a

### DIFF
--- a/src/testing/jest/jest-27-and-under/jest-environment.ts
+++ b/src/testing/jest/jest-27-and-under/jest-environment.ts
@@ -45,10 +45,7 @@ export function createJestPuppeteerEnvironment(): JestPuppeteerEnvironmentConstr
     }
 
     async closeOpenPages() {
-      await Promise.all(this.pages
-        .filter((page) => !page.isClosed())
-        .map((page) => page.close())
-      );
+      await Promise.all(this.pages.filter((page) => !page.isClosed()).map((page) => page.close()));
       this.pages.length = 0;
     }
 
--- a/src/testing/jest/jest-27-and-under/jest-environment.ts
+++ b/src/testing/jest/jest-27-and-under/jest-environment.ts
@@ -29,6 +29,11 @@ export function createJestPuppeteerEnvironment(): JestPuppeteerEnvironmentConstr
         this.browser = await connectBrowser();
       }
 
+      /**
+       * if the user had open pages before, close them all when creating a new one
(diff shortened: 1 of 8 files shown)


tanner-reits (approved):
LGTM

Issue #2335, opened 2020-04-06
The compiler does not support import { foo as bar } from '...'


jpzwarte:
**Stencil version:**

```
 @stencil/core@1.12.2
```

**I'm submitting a:**

[x] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**

**Expected behavior:**

**Steps to reproduce:**

When you rename an import using the `as` operator, the output in `components.d.ts` is invalid:

**Related code:**

```tsx
import { IconDefinition as FaIconDefinition } from '@fortawesome/fontawesome-svg-core';
```

Output (components.d.ts):
```ts
import { FaIconDefinition, } from "@fortawesome/fontawesome-svg-core";
```

Nightmaster:
This references #2074, which did not had any answer from Stencil team for a long moment

Nightmaster:
Duplicate of #2074, which did not had any answer from Stencil team for a long moment

tanner-reits:
This was included in today's [v4.19.0 release](https://github.com/ionic-team/stencil/releases/tag/v4.19.0)!

PR #5836, opened 2024-06-17
feat(complier): support type import aliasing

tanner-reits:

Stencil does not correctly handle aliased types when generating the `components.d.ts` file. If a type import is aliased, the generated file will try to import _only_ the aliased name, which cannot be resolved.

Closes: #2335 

## What is the new behavior?

Stencil will now import and cast (`as`) from the original type name to the aliased name in the generated `components.d.ts` file.

## Documentation

N/A

- [ ] Yes
- [x] No

## Testing

Tested the behavior in a component starter example.

Added a unit test for the new behavior.


christian-bromann (approved):
LGTM 👍

PR #5842, opened 2024-06-18
chore(runtime): set exports field in client runtime package.json

### LINKED ISSUE #5275: bug: Stencil is generating invalid ESM COde
### Prerequisites

### Stencil Version

4.10.0

### Current Behavior

Im having a tsx file with the following import:

      import { Component, Prop, h, State } from '@stencil/core';

and stencil when building a custom-element is creating this import:
 
      import { proxyCustomElement, HTMLElement, h } from '@stencil/core/internal/client';

but this is invalid.

### Expected Behavior

I try to use stencil s a ESM module with an importmap, so the importmap is created from the packages.json export directive, stencil has none, so the import should look like this:

      import { proxyCustomElement, HTMLElement, h } from '@stencil/core/internal/client/index.js';

or stay in this version:

      import { proxyCustomElement, HTMLElement, h } from '@stencil/core';

or the package.json should contain an export directive like this:

    "exports": {
        ".": "./index.js",
        "./internal/client": {
          "default": "./internal/client/index.js"
        }
      }
    

     

### System Info

_No response_

### Steps to Reproduce

Build a library with stencil

### Code Reproduction URL

--

### Additional Information

_No response_

christian-bromann:
fixes #5275

## What is the new behavior?
Set an `exports` field in the `package.json` for the client runtime so it can be imported by ESM.

- [ ] Yes
- [x] No

## Testing

n/a

## Other information

n/a


package.json
```diff
@@ -22,6 +22,34 @@
...
+    },
+    "./internal/testing": {
+      "import": "./internal/testing/index.js"
+    },
+    "./internal/testing/*": {
+      "import": "./internal/testing/*"
+    },
+    "./internal/app-data": {
+      "import": "./internal/app-data/index.js",
+      "require": "./internal/app-data/index.cjs"
```
tanner-reits, line 44: Still think it's so weird we expose things under "internal" 😅

tanner-reits (approved):

package.json
christian-bromann: Totally! These aren't user interfaces though , they are all just import references that Stencil is using.

jogibear9988: you could remove them, but then your imports should use relative paths, with js extension

christian-bromann: Yep, we will do more optimizations in this direction, e.g. move to ESM

PR #5853, opened 2024-06-24
fix(compiler): removed support for tagNameTransform

christian-bromann:
It seems like Stencil currently allows to transform the tag name at runtime. The feature was introduced in https://github.com/ionic-team/stencil/pull/2343/files but was never really introduced nor added to the docs.

## What is the new behavior?
I suggest to remove this entirely since we don't have any tests for this nor do we know how it suppose to behave.

- [ ] Yes
- [x] No

Technically yes but since we never documented it, it was never actually been released as a feature.

## Other information

n/a

### DIFF
--- a/src/app-data/index.ts
+++ b/src/app-data/index.ts
@@ -107,7 +107,6 @@ export const BUILD: BuildConditionals = {
   initializeNextTick: false,
   asyncLoading: false,
   asyncQueue: false,
-  transformTagName: false,
   attachStyles: true,
   // TODO(STENCIL-914): remove this option when `experimentalSlotFixes` is the default behavior
   experimentalSlotFixes: false,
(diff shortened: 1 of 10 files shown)


TRMW:
Also wanted to chime in and say my team is currently using this feature!

christian-bromann:
@johnjenkins @TRMW thanks for providing feedback. Mind sharing in what context you are using this? What is the problem you are trying to solve?

ryuran:
I confirm, we also use it and struggle with angular react output target.

TRMW:
We are also using it as an alternative to scoped custom element registries. We use Stencil for a design system component library, and there are sometimes times when multiple versions of that library end up on the same page. We use this to allow teams to prefix/suffix their custom element names to ensure they always get the version they expect.

christian-bromann:
Closing as we plan to use this feature to support scoped components.

Issue #5860, opened 2024-06-27
bug: Missing export of mock-doc in package.json


saschanowak:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

4.19.0

### Current Behavior

Since the #5275 the exports in the package.json where changed.
We are using the `parseHtmlToDocument` from the `mock-doc` for our custom output target for an ssr server.
Now we are getting an rollup compile error:
```
(node-resolve plugin) Could not resolve import "@stencil/core/mock-doc" in /.stencil/nlx-hydrate-server/server.js using exports defined in /node_modules/@stencil/core/package.json.
(node-resolve plugin) Could not resolve import "@stencil/core/mock-doc" in undefined using exports defined in /node_modules/@stencil/core/package.json.
(node-resolve plugin) Could not resolve import "@stencil/core/mock-doc" in undefined using exports defined in /node_modules/@stencil/core/package.json.
```

### Expected Behavior

Exporting the mock-doc in the `package.json` again.

### System Info

```shell
System: node 20.14.0
    Platform: darwin (23.5.0)
   CPU Model: Apple M2 Pro (12 cpus)
    Compiler: node_modules/@stencil/core/compiler/stencil.js
       Build: 1719433468
     Stencil: 4.19.0 🏄
  TypeScript: 5.4.5
      Rollup: 2.56.3
      Parse5: 7.1.2
      jQuery: 4.0.0-pre
      Terser: 5.31.1
```

### Steps to Reproduce

* Clone the repository
* Install node dependencies
* Run build `npm run build`

{code}
> demo@0.0.1 build
> stencil build

[03:13.4]  @stencil/core
[03:13.6]  v4.19.0 🏄
[03:14.3]  build, demo, prod mode, started ...
[03:14.3]  transpile started ...
[03:15.5]  transpile finished in 1.21 s
[03:15.5]  copy started ...
[03:15.5]  generate custom elements + source maps started ...
[03:15.5]  generate hydrate app started ...
[03:15.5]  generate lazy + source maps started ...
[03:15.5]  copy finished (0 files) in 35 ms
[...]

PR #5861, opened 2024-06-27
chore(runtime): add mock-doc export in client runtime package.json (#5860)

saschanowak:

GitHub Issue Number: #5860

## What is the new behavior?

Set mock-doc export in the package.json for the client runtime so it can be imported by ESM.

- [ ] Yes
- [x] No

## Testing

n/a

## Other information

n/a


christian-bromann (approved):
👍

PR #5863, opened 2024-06-27
fix(hydrate): change type resolve order

christian-bromann:
We have received feedback that people were running into type issues with the `hydrateDocument` function resolving in returning a `Readable` by default. This goes against our objective of supporting DSD without a breaking change.

## What is the new behavior?
Change the order of function overwrites to have TS resolve into a Promise first.

- [ ] Yes
- [x] No

## Testing

Added an e2e test for this.

## Other information

Thanks @sean-perkins for the hint!


gnbm (approved):
LGTM

PR #5838, opened 2024-06-18
fix(runtime): properly assign style declarations

christian-bromann:
## What is the current behavior?

Fixes ionic-team/ionic-framework#29593

## What is the new behavior?

The type fix I initially did suggested a different type ob the style object so this fix makes this more clear that `Cstr.style` can be either a `string` or a `Record<string, string>`.

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Testing

n/a

## Other information

n/a

PR #5876, opened 2024-07-01
fix(hydrate): partially revert #5838

christian-bromann:
We have introduced a regression in #5838 which causes the hydration script to fail.

## What is the new behavior?
Partially revery the change and put in a test to ensure we won't run into this again.

- [ ] Yes
- [x] No

## Testing

Added test.

## Other information

n/a


tanner-reits (approved):
LGTM!

PR #5892, opened 2024-07-11
fix(compiler): no need for commenting selectors anymore

### LINKED ISSUE #5880: feat: hydrate - Remove scoped css classes if serializeShadowRoot is true.
### Prerequisites

Currently Stencil scopes the CSS inside the Declarative Shadow DOM, which is not necessary anymore, as they are already isolated.

The bigger problem is that this bloats the size of the HTML massively. 

### Describe the Use Case

A simple css class like 

```css
.left-0 {left:var(--spacing-0)}
```

is bloated to

```css
/*!@.left-0*/.left-0.sc-my-maybe-long-component-name{left:var(--spacing-0)}
```

This round about triples the size of css classes.

Here another example:

```css
/*!@.button.primary:focus-within.focus-wcag-none:not(:hover,:active,.suppressFocusStyle)*/
.button.primary.sc-my-maybe-long-component-name:focus-within.focus-wcag-none:not(:hover, .sc-my-maybe-long-component-name:active, .suppressFocusStyle).sc-my-maybe-long-component-name {
    outline: 2px solid var(--button-stroke-primary-focus)
}
```

No additional classname CSS selector necessary. No CSS comments necessary.

### Describe Alternatives

_No response_

### Related Code

_No response_

### Additional Information

_No response_

christian-bromann:
With previous versions of Stencil where we didn't not had support for DSD, we used to comment selectors and transform them so that shadow components can by hydrated as scoped components. To have the styles applied properly, we would transform them and keep the original selector commented, e.g.:

```diff
- div {font-size:50px;}}';
+ /*!@div*/div.a {font-size:50px;}}';
```

GitHub Issue Number: fixes #5880

Since component styles are handled at compile time there is no easy way for us to just render define this behavior when calling `renderToString`. Hence this PR will change the behavior of how shadow components are being serialized. It now __doesn't__ render the shadow DOM anymore if the `serializeShadowRoot` is set to false. Before Stencil would transform a shadow component into a scoped component. This behavior will be removed with this patch.

To keep Stencil somewhat backwards compatible, I switched the default value for `serializeShadowRoot` to `true`.

## What is the new behavior?
CSS style text is not transformed into scoped styles anymore, nor there is any selector commenting.

- [ ] Yes
- [x] No

## Testing

Updated tests and added new ones.

## Other information

n/a

### DIFF
--- a/src/compiler/bundle/bundle-output.ts
+++ b/src/compiler/bundle/bundle-output.ts
@@ -121,7 +121,7 @@ export const getRollupOptions = (
       userIndexPlugin(config, compilerCtx),
       typescriptPlugin(compilerCtx, bundleOpts, config),
       extFormatPlugin(config),
-      extTransformsPlugin(config, compilerCtx, buildCtx, bundleOpts),
+      extTransformsPlugin(config, compilerCtx, buildCtx),
       workerPlugin(config, compilerCtx, buildCtx, bundleOpts.platform, !!bundleOpts.inlineWorkers),
       serverPlugin(config, bundleOpts.platform),
       ...beforePlugins,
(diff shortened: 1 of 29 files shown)


tanner-reits (approved):
Love killing code 💪 🪓

PR #5897, opened 2024-07-15
chore(release): update execa and fix incompatibility with Listr

christian-bromann:
In https://github.com/ionic-team/stencil/commit/e7f8129827ee279b8c3cbf07e749f581dd8b3c5a we upgraded `execa` which we are using for the release script. In the next release we discovered that it would cause the pipeline to [stale](https://github.com/ionic-team/stencil/actions/runs/9066025905/job/24907994505) which lead us to revert the [change](https://github.com/ionic-team/stencil/pull/5761).

## What is the new behavior?
This patch resolves the issue we've faced which is likely a result of the incompatibility of rxjs dealing with Buffer streams. `execa` has replaced usage of Buffer in favor of [Uint8Array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array).

- [ ] Yes
- [x] No

## Testing

n/a

## Other information

We may want to update the release script and move off `Listr` which hasn't seen released in 6 years.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -47,7 +47,7 @@
         "eslint-plugin-jsdoc": "^48.0.0",
         "eslint-plugin-simple-import-sort": "^12.0.0",
         "eslint-plugin-wdio": "^8.24.12",
-        "execa": "8.0.1",
+        "execa": "9.3.0",
         "exit": "^0.1.2",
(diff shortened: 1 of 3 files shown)


tanner-reits (approved):
Let's give it a go

PR #5910, opened 2024-07-25
fix(testing): update Jest types

### LINKED ISSUE #5908: bug: incorrect TestingConfig/JestConfig type
### Prerequisites

- [x] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already include this feature request, without success.

### Stencil version

`4.19.2`

### Current behavior

When setting certain properties in Stencil's testing config, such as `restoreMocks: true`, I get TypeScript errors.

Full list of properties in the `JestConfig` interface that are incompatible with Jest's `Config` type:
- `bail`
- `globalSetup`
- `globalTeardown`
- `preset`
- `prettierPath`
- `resolver`
- `restoreMocks`
- `testRegex`
- `transform`

### Expected behavior

These errors should not occur. Setting this property to a boolean is correct, why does it expect a string?

### Steps to reproduce

Set the `testing.restoreMocks` property in `stencil.config.ts`

```ts
  // ...
  testing: {
        // ...
        restoreMocks: true,
    },
};
```

TypeScript will throw the following error:
```
Types of property 'restoreMocks' are incompatible.
  Type 'boolean' is not assignable to type 'string'.
```

### Additional information

Looks like the `JestConfig` interface contains incorrect types.
https://github.com/ionic-team/stencil/blob/7ae36aa5caae278787ff79e63093d4337bfd13c0/src/declarations/stencil-public-compiler.ts#L1790

Could this be updated to use the `Config` type directly from Jest, instead of defining your own interface?
```ts
import type { Config as JestConfig } from 'jest';

export interface TestingConfig extends JestConfig {
  // ...
}
```

christian-bromann:

GitHub Issue Number: fixes #5908

## What is the new behavior?
Adding/updating missing not correct Jest types.

- [ ] Yes
- [x] No

## Testing

n/a

## Other information

n/a

### DIFF
--- a/src/testing/jest/jest-28/jest-config.ts
+++ b/src/testing/jest/jest-28/jest-config.ts
@@ -117,6 +117,16 @@ export function buildJestConfig(config: d.ValidatedConfig): string {
   if (stencilConfigTesting.verbose) {
     jestConfig.verbose = stencilConfigTesting.verbose;
   }
+  if (typeof stencilConfigTesting.bail !== 'undefined') {
+    jestConfig.bail =
+      typeof stencilConfigTesting.bail === 'number' ? stencilConfigTesting.bail : stencilConfigTesting.bail ? 1 : 0;
+  }
+  if (stencilConfigTesting.prettierPath) {
+    jestConfig.prettierPath = stencilConfigTesting.prettierPath;
+  }
+  if (stencilConfigTesting.restoreMocks) {
+    jestConfig.restoreMocks = stencilConfigTesting.restoreMocks;
+  }
 
   jestConfig.testRunner = new Jest28Stencil().getDefaultJestRunner();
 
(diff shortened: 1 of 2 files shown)


src/testing/jest/jest-27-and-under/jest-config.ts
```diff
@@ -153,6 +153,16 @@ export function buildJestConfig(config: d.ValidatedConfig): string {
...
+  if (typeof stencilConfigTesting.bail !== 'undefined') {
+    jestConfig.bail =
+      typeof stencilConfigTesting.bail === 'number' ? stencilConfigTesting.bail : stencilConfigTesting.bail ? 1 : 0;
+  }
+  if (stencilConfigTesting.prettierPath) {
+    jestConfig.prettierPath = stencilConfigTesting.prettierPath;
+  }
+  if (stencilConfigTesting.restoreMocks) {
+    jestConfig.restoreMocks = stencilConfigTesting.restoreMocks;
+  }
```
tanner-reits, line 165: Do we need to apply these to the other adapter versions as well?

PR #5912, opened 2024-07-27
fix(compiler): don't allow shadowRoot getter to avoid hydration issues

christian-bromann:
We have discovered an issue where a user is running into hydration issues due to the fact that they have defined a component getter like so:

```ts
  /**
   * Gets shadow root of nearest host element if one exists
   */
  private get shadowRoot() {
    return this.host.getRootNode() as ShadowRoot;
  }
```

## What is the new behavior?
The compiler now detects this class member in a Stencil components and throws a nice error making users aware that this is not allowed.

- [ ] Yes
- [x] No

This can potentially break for people that do this but don't do any client side hydration.

## Testing

Added sufficient unit tests.

## Other information

n/a

### DIFF
--- a/src/compiler/transformers/static-to-meta/component.ts
+++ b/src/compiler/transformers/static-to-meta/component.ts
@@ -21,6 +21,14 @@ import { parseStringLiteral } from './string-literal';
 import { parseStaticStyles } from './styles';
 import { parseStaticWatchers } from './watchers';
 
+const BLACKLISTED_COMPONENT_METHODS = [
+  /**
+   * If someone would define a getter called "shadowRoot" on a component
+   * this would cause issues when Stencil tries to hydrate the component.
+   */
+  'shadowRoot'
+];
+
 /**
  * Given a {@see ts.ClassDeclaration} which represents a Stencil component
  * class declaration, parse and format various pieces of data about static class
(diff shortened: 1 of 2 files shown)


src/compiler/transformers/static-to-meta/component.ts
```diff
@@ -176,6 +178,42 @@ export const parseStaticComponentMeta = (
...
+  /**
+   * validate if:
+   */
+  if (
+    /**
+     * the component has a getter called "shadowRoot"
+     */
+    ts.isGetAccessorDeclaration(node) &&
+    ts.isIdentifier(node.name) &&
+    node.name.escapedText === 'shadowRoot' &&
```
tanner-reits, line 191: Can we get rid of the "shadowRoot" magic string?

src/compiler/transformers/static-to-meta/component.ts
```diff
@@ -176,6 +178,42 @@ export const parseStaticComponentMeta = (
...
+    /**
+     * the class is actually a Stencil component, has a decorator with a property named "tag"
+     */
+    if (
+      ts.isCallExpression(decorator.expression) &&
+      decorator.expression.arguments.length === 1 &&
+      ts.isObjectLiteralExpression(decorator.expression.arguments[0]) &&
+      decorator.expression.arguments[0].properties.some(
+        (prop) => ts.isPropertyAssignment(prop) && prop.name.getText() === 'tag',
+      )
```
tanner-reits, line 207: Is it easier to just check the decorator name?

tanner-reits (changes requested):

src/compiler/transformers/static-to-meta/component.ts
christian-bromann: Yes, but you could potentially import it with any arbitrary name which is why I did this extra step of checking if the decorator has 1 argument which contains a `tag` property since this is the same everywhere.

tanner-reits (approved):

src/compiler/transformers/static-to-meta/component.ts
tanner-reits: Ah yeah, good point!

PR #5930, opened 2024-08-06
fix(scripts): fix Esbuild scripts to allow to run on Windows

christian-bromann:
Stencil fails when trying to build it on a Windows machine.

STENCIL-1369

## What is the new behavior?
It turns out that the alias path needed to get normalized allowing Esbuild to pick up the right file and acknowledge it as external dependency.

This patch will also make some tweaks to the pipeline to run the `build` and `linter` tasks first and then run everything else. Before we were running WebdriverIO immediately.

- [ ] Yes
- [x] No

## Testing

Updated our pipeline to have the project being build on Windows so we ensure we don't introduce regressions.

## Other information

n/a

### DIFF
--- a/.github/workflows/test-wdio.yml
+++ b/.github/workflows/test-wdio.yml
@@ -1,21 +1,14 @@
 name: WebdriverIO Tests
 
 on:
-  merge_group:
-  pull_request:
-  push:
-    branches:
-      - 'main'
+  workflow_call:
+    # Make this a reusable workflow, no value needed
+    # https://docs.github.com/en/actions/using-workflows/reusing-workflows
 
 jobs:
-  build_core:
-    name: Build
-    uses: ./.github/workflows/build.yml
-
   run_wdio:
     name: Run WebdriverIO Component Tests (${{ matrix.browser }})
     runs-on: ubuntu-22.04
-    needs: [build_core]
     strategy:
       matrix:
         browser: [CHROME, FIREFOX, EDGE]
@@ -23,6 +16,9 @@ jobs:
     steps:
       - name: Checkout Code
         uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+
+      - name: Get Core Dependencies
+        uses: ./.github/workflows/actions/get-core-dependencies
 
       - name: Use Node Version from Volta
         uses: actions/setup-node@1e60f620b9541d16bece96c5465dc8ee9832be0b # v4.0.3
--- a/scripts/esbuild/sys-node.ts
+++ b/scripts/esbuild/sys-node.ts
@@ -1,6 +1,6 @@
 import type { BuildOptions as ESBuildOptions } from 'esbuild';
 import fs from 'fs-extra';
-import { join } from 'path';
+import path from 'path';
 import resolve from 'resolve';
 import webpack, { Configuration } from 'webpack';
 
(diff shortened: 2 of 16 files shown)


christian-bromann:
Note: the pending jobs should properly resolve once merged into `main` and whenever we raise new PRs afterwards.

scripts/esbuild/utils/index.ts
```diff
@@ -16,20 +16,11 @@ export function getEsbuildAliases(): Record<string, string> {
...
-    '@stencil/core/internal/testing': './internal/testing/index.js',
-    '@stencil/core/cli': './cli/index.js',
-    '@sys-api-node': './sys/node/index.js',
-
-    // mapping node.js module names to `sys/node` "cache"
-    //
-    // these allow us to bundle and ship a dependency (like `glob`) as part
-    // of the Stencil distributable but also have our separate bundles
-    // reference the same file
-    glob: './sys/node/glob.js',
```
tanner-reits, line 32: What's the reason for dropping these? Just not needed in all instances of `getEsbuildAliases()` being called?

scripts/esbuild/sys-node.ts
```diff
@@ -36,10 +36,15 @@ export async function buildSysNode(opts: BuildOptions) {
...
     // externalize each other we need to do so
-    join(opts.output.sysNodeDir, '*'),
+    '../../compiler/stencil.js',
+    '../../sys/node/*',
   ];
 
-  const sysNodeAliases = getEsbuildAliases();
+  const sysNodeAliases = {
+    ...getEsbuildAliases(),
+    '@stencil/core/compiler': '../../compiler/stencil.js',
```
tanner-reits, line 45: If this path is different than what's spit out by `getEsbuildAliases()`, does it make sense to have that function generate the path based on some directory? Why are they different in the first place?

.github/workflows/test-wdio.yml
```diff
@@ -1,21 +1,14 @@
 name: WebdriverIO Tests
 
 on:
-  merge_group:
-  pull_request:
-  push:
-    branches:
-      - 'main'
+  workflow_call:
```
tanner-reits, line 4: Is this action getting called from anywhere now?

tanner-reits (changes requested):

.github/workflows/test-wdio.yml
christian-bromann: good point, it is missing in the main test workflow

christian-bromann: added in https://github.com/ionic-team/stencil/pull/5930/commits/663118dddbe7425dba404eb8fee4ca382c371417

scripts/esbuild/sys-node.ts
christian-bromann: there are some build artifacts that located multiple folders deep that require a different alias path, e.g. for `/testing/index.js` the alias for `@stencil/core/compiler` has to be `../compiler/stencil.js` while for `/sys/node/index.js` it has to be `../../compiler/stencil.js`.

scripts/esbuild/utils/index.ts
christian-bromann: Correct, some aliases are only needed in certain components e.g. `glob` which is why I removed it from here. The alias really depends on where the bundled file is located, e.g. how many levels they are nested within Stencil. Most artifacts are located within a directory, e.g. `/testing/index.js` or `/compiler/stencil.js` so I kept some global ones but made sure the packages that are located in a more deeper directory structure receive a custom alias.

scripts/esbuild/sys-node.ts
tanner-reits: Right, so I was thinking we could figure out how deeply nested we are based on the relative path from `__filename` to `src` or something and use that info to generate the relative paths.

Or, alternatively, remove the helper altogether. What you have works, but feels a bit odd that we use it and then override the returned aliases in quite a few spots.

Open to thoughts!

christian-bromann: Most Stencil components can be successfully compiled using the common defined aliases as implemented in [scripts/esbuild/utils/index.ts](https://github.com/ionic-team/stencil/pull/5930/files#diff-ca5c0dbb4baece2d3fbe79900e692152217277778ef3679274a74c5c0ee4057d), there are only two artifacts that require modifications to these aliases which is:

- one artifact in `scripts/esbuild/sys-node.ts`
- and `scripts/esbuild/internal-platform-testing.ts`

If you want I can get rid of the `getEsbuildAliases` function and have this defined per component but I think this will create more code than we have right now.

Wdyt?

tanner-reits: I think it's fine for now if it's only a couple spots. Thanks for the pushback, no need to overcomplicate

christian-bromann:
Released in Stencil `v4.22.0`

PR #5933, opened 2024-08-08
fix(runtime): provide second arg to `insertBefore`

tanner-reits:

GitHub Issue Number: N/A

The runtime will throw an error when trying to apply styles to a Shadow component

## What is the new behavior?

Styles will be applied without throwing an error

## Documentation

N/A

- [ ] Yes
- [X] No


christian-bromann (approved):
👍 

Thanks a lot!

bkjohnson:
I think this might have caused a regression. When I upgrade stencil to `4.21.0` and do a build, I get a runtime error which causes tests to fail:

```

          Ferrum::JavaScriptError:
            NotFoundError: Failed to execute 'insertBefore' on 'Node': The node before which the new node is to be inserted is not a child of this node.
                at addStyle
```

bkjohnson:
I couldn't isolate the issue enough to create a reproduction, but I've just upgraded to 4.22.1 successfully.

PR #5926, opened 2024-08-05
fix(runtime): render component styles at the end of the head tag

christian-bromann:
## What is the current behavior?
The `<style />` tag for scoped components is added before any link tag. The behavior was implemented in https://github.com/ionic-team/stencil/commit/05d242d5d6ba558896495df3f7b923ca109c70d9 without any concrete description as to why. As reported in below linked issue it can cause links with `preconnect` attribute to be loaded later hence negatively impacting the performance of an application.

GitHub Issue Number: fixes #5915

## What is the new behavior?
Attach styles at the end of the `<head />` so that the browser can already load assets marked with `preconnect` before all styles are streamed over the wire.

An alternative would be to make this configurable but I don't think there is a viable reason why someone would want to have styles injected first. If someone reports the requirement for this we can add it to the backlog.

## Documentation

n/a

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Testing

Add an e2e test for this behavior.

## Other information

n/a

PR #5938, opened 2024-08-12
fix(runtime): place scoped component styles after preconnect links but before custom styles

christian-bromann:
This is an enhancement to a fix we made in in #5926 which caused our Nightly build to fail due to the inability to place custom styles in the HTML template as all component styles were injected last in the `<head />` tag.

STENCIL-1372

## What is the new behavior?
Place scoped component styles after existing preconnect links but allow custom styles to be set at the end of the `<head />` tag.

## Documentation

I will raise a separate PR to the docs page mentioning this behavior.

- [ ] Yes
- [ ] No

## Testing

Enhanced existing e2e test.

## Other information

n/a

### DIFF
--- a/src/runtime/styles.ts
+++ b/src/runtime/styles.ts
@@ -101,7 +101,8 @@ export const addStyle = (styleContainerNode: any, cmpMeta: d.ComponentRuntimeMet
                * after the last preconnect link to ensure the styles are preloaded
                */
               const preconnectLinks = styleContainerNode.querySelectorAll('link[rel=preconnect]');
-              const referenceNode = preconnectLinks.length > 0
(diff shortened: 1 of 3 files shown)


tanner-reits (changes requested):
Tested this out with the failing Framework test and still having the same issue. If there are no preconnect links, can we insert the styles where they were going before (i.e. before the first `link` tag)?

christian-bromann:
Good catch @tanner-reits , I totally missed that use case but added a change with a test.

tanner-reits (approved):
Looks good in the Framework test case 👍

christian-bromann:

PR #5946, opened 2024-08-14
fix(runtime): update call to `prepend` to remove `null` node

tanner-reits:

Inserting styles inside shadow components calls to `prepend` with a `null` node (i.e. `prepend(styles, null)`). This causes a text node with "null" to be rendered.

GitHub Issue Number: N/A

## What is the new behavior?

Removes the `null` node from the call to `prepend`.

## Documentation

N/A

- [ ] Yes
- [x] No


christian-bromann (approved):

christian-bromann:

Issue #2703, opened 2020-10-14
Button form attribute does not render/missing after compiled in DOM


liho98:
**Stencil version:**

```
 @stencil/core@ <^1.17.3>
```

**I'm submitting a:**

[x] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://stencil-worldwide.herokuapp.com/ or https://forum.ionicframework.com/

**Current behavior:**
![image](https://user-images.githubusercontent.com/44250267/96020028-06f11e00-0e80-11eb-9887-c8e9be6a9d28.png)
![image](https://user-images.githubusercontent.com/44250267/96020101-1d977500-0e80-11eb-92c9-576cb26f08fa.png)
you can see the form attribute is missing after compile in DOM.

**Expected behavior:**

I want the form attribute to work, so i can handle multiple form submission.

**Steps to reproduce:**
just add form attribute in <button> tag

**Related code:**

```tsx
      <Host style={{ all: "unset" }}>
        <button
          type={this.type}
          style={this.buttonStyle}
          class={`rm-button--button ${this.class}`}
          value="Submit"
          form="form1"
        >
          <slot></slot>
        </button>
      </Host>
```

**Other information:**

claviska:
Confirmed with a minimal component:

```tsx
import { Component, h } from '@stencil/core';

@Component({
  tag: 'my-button',
  shadow: true,
})
export class AppHome {
  render() {
    return <button form="form-id-here">Submit</button>;
  }
}
```

The resulting markup looks like this:

<img width="369" alt="CleanShot 2020-11-17 at 10 55 21@2x" src="https://user-images.githubusercontent.com/55639/99413093-69946a00-28c3-11eb-85ed-99a780e162b3.png">

claviska:
@adamdbradley any idea what might be causing this? I'm happy to take a stab at it if you can point me in the right direction.

thccorni:
As a workaround I managed to use `ref` to get it to work:
```tsx
<button
  ref={(btn) => btn?.setAttribute('form', 'form-id-here')}
>
  Submit
</button>
```
But it is a little ugly, I know.

KevinCarnaille2:
FYI, still having the issue in Stencil `2.17.3` :) 

@thccorni workaround is working ;)

DavideMininni-Fincons:
Hello guys! 
[...]

PR #5947, opened 2024-08-14
fix(runtime): only use setter if existing

christian-bromann:
Trying to set properties without a setter fails silently in Stencil.

GitHub Issue Number: #2703

## What is the new behavior?
Check if property has a setter and use `setAttribute` otherwise.

- [ ] Yes
- [x] No

## Testing

Added unit test.

## Other information

n/a


tanner-reits (approved):

christian-bromann:

PR #5948, opened 2024-08-16
fix(runtime): have fallback for style setting

christian-bromann:
The Ionic Framework nightly build is failing. After some investigations it seems that the page had a missing style tag in the header which caused the component to be rendered without padding:

Screenshot 2024-08-16 at 3 11 10 PM

_> Left side is the working version, right side is the Playwright trace_

I don't know why the style tag with the styles for `ion-list` isn't rendered only in CI. I have to assume that it has to do with Playwright behaving funky here. I can't reproduce it locally at all.

## What is the new behavior?
Added an `else` statement in cases none of the previous if statement match which should never happen but may have happened here. Worth a try.

- [ ] Yes
- [x] No

## Testing

n/a

## Other information

n/a


tanner-reits (approved):

christian-bromann:

Issue #5618, opened 2024-04-03
bug: Attr missing from docs where prop define it as global type like HTMLElement


venkatesh-nagineni:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

4.5.0

### Current Behavior

When i have prop defined with global type then i don't see attr present in custom-docs

```
  @Prop() testElem: HTMLElement;
```

![Screenshot 2024-04-03 at 12 42 08](https://github.com/ionic-team/stencil/assets/21661817/14329db3-0498-4a78-878b-5195daad4b0c)

### Expected Behavior

expecting to present ```test-elem``` in attr section in custom docs inside prop

### Steps to Reproduce

have a look at sandbox link to reproduce https://codesandbox.io/p/devbox/stencil-web-component-forked-wnqp67?file=%2Fsrc%2Fcomponents%2Fmy-component%2Fmy-component.tsx%3A31%2C1-32%2C1

### Code Reproduction URL

https://codesandbox.io/p/devbox/stencil-web-component-forked-wnqp67?file=%2Fsrc%2Fcomponents%2Fmy-component%2Fmy-component.tsx%3A31%2C1-32%2C1

rwaskiewicz:
Thanks!

I was able to verify that this is an issue with how Stencil records attribute types during compilation, resulting in `attr` being set to `undefined` in earlier in the compiler (and not being written to `docs-json` output as a result). I'm going to label this to be ingested into our backlog - Thanks again!

johnjenkins:
hi! I think the behaviour of Stencil makes sense in this case - how could an HTMLElement be represented as a string / attribute? 
Unless misunderstood, I'm going to close this out

PR #5958, opened 2024-08-28
fix(compiler): allow attribute & reflect values on props w/ non JS primitives

tanner-reits:

Stencil ignores setting the `attribute` and `reflect` values when transforming `@Prop()` decorators if their resolved types don't resolve to a JS primitive (string, number, etc.). 

GitHub Issue Number: #5618

## What is the new behavior?

Stencil will now set the `attribute` and `reflect` values regardless of the resolved type. This was they are always present in the generated `docs-json`. 

- [ ] Yes
- [x] No

## Testing

Verified the expected behavior with the reproduction case, and updated unit test cases where the previous behavior was checked

## Other information

There may be some repercussions to this change beyond what we can foresee, so we'll keep an eye on feedback once this is released if there are unwanted consequences


christian-bromann (approved):
LGTM 👍

Issue #3443, opened 2022-06-28
bug: renaming a a component file whilst being watched causes an error


johnjenkins:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/.github/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

2~ (happens in 2.17.0 but has been present for a long time)

### Current Behavior

using the latest version of stencil, whilst within `--watch` mode, change any component file name (e.g. `my-component.tsx` to `my-new-component.tsx`) and stencil fails with an error:

```
[ ERROR ]  TypeError: Cannot read property 'fileName' of undefined at flattenDiagnosticMessageText
           (.../node_modules/@stencil/core/compiler/stencil.js:1559:45) at
           loadTypeScriptDiagnostic
```

### Expected Behavior

not to fail :)

### Steps to Reproduce

` npm init stencil` > select the component variant. 
`cd NEW_COMPONENT_DIR`
`npm i`
`npm start`
rename `src/components/my-component.tsx` to `my-new-component.tsx`
 

### Code Reproduction URL

just use `npm init stencil`

rwaskiewicz:
Thanks @johnjenkins! I was able to confirm this has been around since at least v2.7.0.  Labelling this so that it gets ingested into the backlog

linuxonrails:
Hi @johnjenkins!  

I think it is a valid error and that it only warns you that something is wrong in the test file. 

If you correct the path in `src/components/my-component/my-component.spec.ts`, the error disappears and the browser reloads without problems:

```
2c2
< import { MyComponent } from './my-component';
---
> import { MyComponent } from './my-new-component';
```

I think this issue could be closed or, perhaps, renamed to "Test files should be ignored in browser transpilation with `npm start`". But personally I prefer the build to fail if there is a test failure.

rwaskiewicz:
hey @linuxonrails,

thanks for your input! however, even when no tests exist in a project, I am able to reproduce this issue. the root cause has to do with a race condition where the dev server isn't able to understand that the file has changed it's name

alicewriteswrongs:
[...]

PR #5968, opened 2024-09-04
fix(compiler): handle file rename in watch mode

tanner-reits:

Fixes: #3443

When renaming a file in watch mode, it is possible to run into two errors:

1. A build error resulting from the compiler believing that two components exist with the same tag name (one from the old file name, and one from the new)
2. An HMR issue where the new file name is not correctly registered with the file watcher. So, changes to the renamed file did not trigger an HMR reload

## What is the new behavior?

Addresses both of the issues mentioned above by updating the watched files and compiler context whenever a change is detected that triggers the watch callback

## Documentation

N/A

- [ ] Yes
- [x] No

## Testing

This is a difficult issue to test (especially the HMR since that was not always consistent), so I was only able to perform manual tests

### DIFF
--- a/src/compiler/build/watch-build.ts
+++ b/src/compiler/build/watch-build.ts
@@ -84,6 +84,26 @@ export const createWatchBuild = async (
       );
     }
 
+    // Make sure all files in the module map are still in the fs
+    // Otherwise, we can run into build errors because the compiler can think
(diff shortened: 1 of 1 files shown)


christian-bromann (approved):
👍

Issue #5969, opened 2024-09-05
bug: newly added slots are reordered when the container component's attribute set dynamically


yigityuce:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

4.21.0

### Current Behavior

Requirements:
- scoped components
- no-shadow
- experimentalSlotFixes: true
- experimentalScopedSlotChanges: true

Steps:
- container element (such as list) with default items in it
- append() new items to that container element
- click one of the items in the list
- newly added items are randomly reordered

Rationale:
- some flags are missing for the newly added elements
  - for prepend patch: `s-sh`
  - for append patch: `s-sh` and `s-ol`
- therefore, on any action that triggers the relocation algorithm causes the reordering

prepend:
<img width="994" alt="image" src="https://github.com/user-attachments/assets/edd3ddc5-fa0f-4c63-8cf0-75f2e16f41f0">

append:
<img width="827" alt="image" src="https://github.com/user-attachments/assets/1ba884b0-0dde-43bb-84fd-b5f3c61d8f32">

### Expected Behavior

Current:
<img width="1092" alt="image" src="https://github.com/user-attachments/assets/8daf359b-c261-472f-9500-3707ed1ae4ec">

Expected:
<img width="667" alt="image" src="https://github.com/user-attachments/assets/2596e471-8299-469f-a474-b71a5366854c">

### System Info

```shell
System: node 18.20.2
    Platform: darwin (23.6.0)
   CPU Model: Apple M3 Pro (12 cpus)
    Compiler: /Users/yigit.yuce/Documents/Projects/personal/yigityuce.github/stencil-v4-sc-new-child-relocation/node_modules/@stencil/core/compiler/stencil.js
       Build: 1724698030
     Stencil: 4.21.0 🐷
  TypeScript: 5.5.3
      Rollup: 2.56.3
      Parse5: 7.1.2
      jQuery: 4.0.0-pre
      Terser: 5.31.1
```

### Steps to Reproduce

stated above

### Code Reproduction URL

https://github.com/yigityuce/stencil-v4-sc-new-child-relocation

[...]

PR #5970, opened 2024-09-05
fix(runtime): scoped slot append/prepend correct order after interaction

yigityuce:

GitHub Issue Number: #5969

- [ ] Yes
- [x] No

### DIFF
--- a/src/runtime/dom-extras.ts
+++ b/src/runtime/dom-extras.ts
@@ -83,6 +83,12 @@ export const patchSlotAppendChild = (HostElementPrototype: any) => {
     const slotName = (newChild['s-sn'] = getSlotName(newChild));
     const slotNode = getHostSlotNode(this.childNodes, slotName, this.tagName);
     if (slotNode) {
+      const slotPlaceholder: d.RenderNode = document.createTextNode('') as any;
+      slotPlaceholder['s-nr'] = newChild;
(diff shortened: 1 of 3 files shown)


yigityuce:
to summarize, I added `s-sh` flag to the existing `prepend` patch and copied that block (s-ol and s-sh updating block) to the `appendChild` patch as well. 

the others (`InsertAdjacentHTML`, `InsertAdjacentText`, `InsertAdjacentElement`, `append`, etc) are using the `prepend` or `appendChild` internally, so they don't need to be updated

christian-bromann:
@yigityuce do you think we can add a unit test for this change?

tanner-reits (changes requested):
@yigityuce Changes look good! Only request would be to add a test like @christian-bromann suggested. You could probably add a Webdriver test to our existing test suite for these methods [here](https://github.com/ionic-team/stencil/blob/main/test/wdio/scoped-slot-append-and-prepend/cmp.test.tsx). Let me know if you have any questions!

yigityuce:
hey @christian-bromann & @tanner-reits tests are added as a separate test file since to reproduce the issue, I need to remove the default slot wrapper and I didn't want to break the existing test case.

tanner-reits (approved):
@yigityuce Thanks! Everything looks good. I'll get this merged and try to get a release out this week!

christian-bromann:
Released in Stencil `v4.22.0`

PR #5988, opened 2024-09-17
feat(compiler): move readme diagram color configs to output target config

tanner-reits:

GitHub Issue Number: N/A

Stencil has static color values for the diagrams generated by the `docs-readme` output target for component dependencies.

## What is the new behavior?

Adds an option to the `docs-readme` output target to customize diagram colors. These values are validated before generation occurs.

## Documentation

https://github.com/ionic-team/stencil-site/pull/1483

- [ ] Yes
- [x] No

## Testing

Unit tests added for validation and tested behavior in a component starter

## Other information

There is a limitation with this approach where the `--docs` CLI flag would not be customizable as it is in https://github.com/ionic-team/stencil/pull/5980. But this would approach would allow each target instance to have it's own color config.


christian-bromann (approved):
LGTM 👍 

Let's make sure we clearly document this change in the changelogs as it would be a potential breaking change in the interface.

Issue #4229, opened 2023-04-05
bug: `moduleResolution` is not respected


jgroth:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/.github/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

3.2.0

### Current Behavior

I have a library that is used by my component and the library has sub modules that is using the new [`exports`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-7.html#packagejson-exports-imports-and-self-referencing) syntax in `package.json`. My component is importing it something like this
```ts
import { foo } from 'my-lib/foo`;
```

To get it to work I have set `moduleResolution` in `tsconfig.json` to `node16`/`nodenext`, but when I build my project I get an error saying 
```
[ ERROR ]  TypeScript: ./src/components/my-component/my-component.tsx:3:21
           Cannot find module 'my-lib/foo' or its corresponding type declarations.

      L2:  import { myLib } from 'my-lib';
      L3:  import { foo } from 'my-lib/foo';

[28:31.4]  build failed, watching for changes... in 3 ms

``` 

VSCode can find the module when I set `moduleResolution` to `node16`, and I can run the TypeScript compiler manually and it will compile, but it seems Stencil is not respecting the `moduleResolution` setting in `tsconfig.json`

### Expected Behavior

No build errors. Stencil should use the `moduleResolution` specified in `tsconfig.json`

### System Info

```shell
System: node 16.10.0
    Platform: linux (5.19.0-38-generic)
   CPU Model: 11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz (8 cpus)
    Compiler: /home/foo/node_modules/@stencil/core/compiler/stencil.js
       Build: 1678807683
     Stencil: 3.2.0 🌷
  TypeScript: 4.9.5
      Rollup: 2.42.3
      Parse5: 7.1.2
      Sizzle: 2.42.3
      Terser: 5.16.6
```

### Steps to Reproduce

Build the included Stencil project

### Code Reproduction URL

https://github.com/jgroth/stencil-module-resolution

tanner-reits:
Hey @jgroth 👋 

[...]

PR #5274, opened 2024-01-18
feat(deps): upgrade rollup, commonjs plugin

rwaskiewicz:
**DO NOT MERGE!** I want to squash this as a `feat()` to surface the changes in the changelog. While I've tested this pretty well (subjective, I know 😉), I want folks to see this core piece of infra is being updated, as it impacts them as well. 

## What is the current behavior?

Our rollup infrastructure is lagging behind the latest. We're currently working on a few different modernization efforts for handling newer JS paradigms + infra; this is one of them.

GitHub Issue Number: N/A

## What is the new behavior?

upgrade rollup to v2.56.3 and `@rollup/plugin-commonjs` to v21.1.

this commit raises these libraries to the highest version as possible
without introducing breaking changes. v2.57 has a bug in it that
requires `@rollup/plugin-commonjs@22` for handling CJS-ESM interop.
that version of `@rollup/plugin-commonjs` requires
`@rollup/plugin-node-resolve@13`. however,
`@rollup/plugin-node-resolve@11` has changes that would be breaking
for existing users (see https://github.com/ionic-team/stencil/pull/5269)

future upgrades will be slated to stencil v5.0.0.

## Does this introduce a breaking change?

- [ ] Yes
- [x] No - We do not believe this is going to introduce any breaking changes. However, folks upgrading to the version of Stencil this upgrade lands in are encouraged to file bugs early/often.

## Testing

I created a dev build of Stencil, `4.10.0-dev.1705594858.56d0fba` and used that in a CI run of Framework, which passed https://github.com/ionic-team/ionic-framework/actions/runs/7573347378

I did a pretty comprehensive diff of the changes that are going to be applied to the Ionic Framework as a part of this. For every minor version of rollup of `@rollup/plugin-commonjs`, I:
- Built Stencil (`npm run clean && npm ci && npm run build && npm pack`)
- Installed the tarball into Ionic Framework Core
- Built Ionic Framework Core (`npm run build`)
- Diffed the `components`, `css`, `dist`, `hydrate`, `loader` directories of the latest version I just built with the previous

In doing so, I noticed minimal changes. Happy to discuss them in a jam session or otherwise. 

## Other information

My notes for this are an attachment on the JIRA ticket.

PR #5996, opened 2024-09-24
fix: support modern package exports syntax

louis-bompart:
Currently, you cannot use packages that use modern module resolution based on subpath-exports in your Stencil Project[^1].
This is caused by the hard coding of Node10 for moduleResolution in Stencil.

GitHub Issue Number: #4229 (mitigation, not 💯 fix) 

## What is the new behavior?
Replace Node10 with Bundler. While it's not a "trivial" change, the impact for the end-user should be minimal (see Breaking change section)

## Documentation

N.A.

- [ ] Yes
- [x] No

I ran some 'health-check' (running stencil compiler with & without my changes on packages that use various ways of defining sub-packages), and all projects that worked without my changes worked with it as well, without changes.

Moreover, from the TypeScript reference documentation, I reckon it is safe to say that Bundler is akin to a "superset" of Node10:
https://www.typescriptlang.org/docs/handbook/modules/reference.html#bundler
https://www.typescriptlang.org/docs/handbook/modules/reference.html#node10-formerly-known-as-node

## Testing

In addition to the existing test (`npm test`), I ran some san-check. We have a package that we are migrating from 'hackish subpackages' (using subdirectory and package.json in them) to subpath exports in a major version, I used this package as an example and did the following tests:
 - [Stencil 4.21.0 'vanilla' + hackish subpackages](https://github.com/louis-bompart/stencil-bugfixes-exports-testing/tree/legacy-package): ✅ compiling, behavior nominal
 - [Stencil 4.21.0 'patched' (i.e with my changes) + hackish subpackages](https://github.com/louis-bompart/stencil-bugfixes-exports-testing/tree/legacy-package%2Bpatched-stencil): ✅  compiling, behavior nominal
 - [Stencil 4.21.0 'vanilla' + subpath exports](https://github.com/louis-bompart/stencil-bugfixes-exports-testing/tree/next-package): ❌ failing when importing a subpath, as expected.
 - [Stencil 4.21.0 'patched' + subpath exports](https://github.com/louis-bompart/stencil-bugfixes-exports-testing/tree/next-package%2Bstencil-patch): ✅ compiling, behavior nominal.

To validate, check the linked repo (pay attention to the branch), `npm i ; npm start`

I'm not necessarily expecting this to be an 'LGTM & merge' but more a base to discuss how to improve Stencil so that it's not a blocker for good practices adoption (pure incompatibility with modern package practice is quite a sore point tbh), which is paramount IMHO for developer-oriented products.

[^1]: https://nodejs.org/api/packages.html#subpath-exports

### DIFF
--- a/src/compiler/transpile/ts-config.ts
+++ b/src/compiler/transpile/ts-config.ts
@@ -27,7 +27,7 @@ export const getTsOptionsToExtend = (config: d.ValidatedConfig): ts.CompilerOpti
     // declaration files
     declaration: config.outputTargets.some(isOutputTargetDistTypes),
     module: ts.ModuleKind.ESNext,
-    moduleResolution: ts.ModuleResolutionKind.NodeJs,
+    moduleResolution: ts.ModuleResolutionKind.Bundler,
     noEmitOnError: false,
     outDir: config.cacheDir || config.sys.tmpDirSync(),
     sourceMap: config.sourceMap,


christian-bromann:
@louis-bompart thanks for your contribution 🙏 

Let us verify this with some of the upstream packages that rely on Stencil to check compatibility. From my point of view: if the breaking change is reasonable and doesn't impact the majority of users (as far as we can expect it) I would be in favor to move forward with this.

louis-bompart:
Hey @christian-bromann 👋 Thanks for checking this out 🙏 

I checked the ❌ in the CI and followed the trail.

I reckon this use case of rollup in the browser like in `test/browser-compile/src/components/app-root/app-root.tsx` is quite... unique 😄, and would not be something the "majority of users" would do. Moreover, users (well, outside Stencil itself) should be able to update.

I understand Stencil is in a tricky position regarding Rollup & the version they can use (#5274), so updating is not a solution in the present case.

From what I gather from [v4.17.0 release notes](https://github.com/ionic-team/stencil/releases/tag/v4.17.0), I reckon Stencil's goal is to replace Rollup by esbuild in the future.

With all that in mind, I'd propose a ~solution~ hack: using [`patch-package`](https://www.npmjs.com/package/patch-package) to tweak ever so slightly the package.json of rollup in your node_modules by applying the following commit https://github.com/rollup/rollup/commit/2cca505684100c04f3c29d6c792a7865ea2affbe.
This is the path of 'least resistance,' which I think could be acceptable, given where the project is headed (from what I understand).
This method's boon is that you can still use the same rollup version in production and in your tests.

If this last point is of no concern, you could leverage subpath imports to have 2 versions of rollup side-by-side in the project.
Just need to add something like that in the `devDependencies`:
`"rollup-test": "npm:rollup@^2.79.0"

And replace the import from "rollup" in the test with an import from "rollup-test."

What do you think?

louis-bompart:
> > I reckon Stencil's goal is to replace Rollup by esbuild in the future.
> 
> We do not plan to internally replace Rollup with Esbuild. We have replaced our internal bundler but right now Rollup is too integrated into Stencil that updating it is a better move forward than completely replacing it.
> 
> > What do you think?
> 
> I think we should focus on the Rollup update to enable this feature. It is essential for a lot of other features in Stencil and we have put this on our roadmap to tackle it eventually. Any support and help would be much appreciated though.
> 
> Do you think we can close this or would you like to try a different approach?

That makes sense :+1: Let's close this for now, thanks for your answer!

PR #6025, opened 2024-10-10
fix(docs): escape backticks in type or default value columns

### LINKED ISSUE #6024: Template literal type props are not escaped properly in generated markdown
**Stencil version:**

```
 @stencil/core@4.22.1
```

**Current behavior:**

When markdown docs are generated for a prop that is a [Template Literal Type](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html), e.g. 

```
@Prop() width: `${number}px` | `${number}%`;
```

The markdown generator takes the type as-is without escaping the backticks, which has incorrect formatting (and breaks in an MDX parser)

```
| Property | Attribute | Description     | Type                                  | Default     |
| -------- | --------- | --------------- | ------------------------------------- | ----------- |
| `width`  | `width`   |                 | ``${number}%` \| `${number}px``       | `undefined` |
```

| Property | Attribute | Description     | Type                                  | Default     |
| -------- | --------- | --------------- | ------------------------------------- | ----------- |
| `width`  | `width`   |                 | ``${number}%` \| `${number}px``       | `undefined` |

**Expected behavior:**

Since the markdown generator adds its own backticks to style the type as a "code" value, it should escape values that contain backticks. I think you should be able to use two backticks and a space around each side of the type to safely escape the inner backticks.

```
| Property | Attribute | Description     | Type                                      | Default     |
| -------- | --------- | --------------- | ----------------------------------------- | ----------- |
| `width`  | `width`   |                 | `` `${number}%` \| `${number}px` ``       | `undefined` |
```

| Property | Attribute | Description     | Type                                      | Default     |
| -------- | --------- | --------------- | ----------------------------------------- | ----------- |
| `width`  | `width`   |                 | `` `${number}%` \| `${number}px` ``       | `undefined` |

**GitHub Reproduction Link:**

See the minimal reproduction repo here: https://github.com/brandongregoryscott/stencil-prop-escape-reproduction

Editing the props and running `npm run build` should update the generated readme.

**Other information:**

I can probably workaround this by writing a [Custom Docs Generator](https://stenciljs.com/docs/docs-custom), but it seems like overkill to have to reimplement generation for events, methods, etc.

brandongregoryscott:

fixes #6024 

## What is the new behavior?

Types or default values that include a backtick are escaped with two backticks and a space.

- [ ] Yes
- [x] No

## Testing

See the new unit tests in `markdown-props.spec.ts`. If you think the logic for this bug fix should exist somewhere else, let me know!

### DIFF
--- a/src/compiler/docs/readme/markdown-props.ts
+++ b/src/compiler/docs/readme/markdown-props.ts
@@ -19,8 +19,8 @@ export const propsToMarkdown = (props: d.JsonDocsProp[]) => {
       getPropertyField(prop),
       getAttributeField(prop),
       getDocsField(prop),
-      `\`${prop.type}\``,
-      `\`${prop.default}\``,
+      getTypeField(prop),
+      getDefaultValueField(prop),
     ]);
   });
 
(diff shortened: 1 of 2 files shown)


christian-bromann (approved):
Nice 👍 LGTM

brandongregoryscott:
Ah, looks like some tests are failing. Let me fix those up, running the whole suite was tanking my laptop.

Issue #6015, opened 2024-10-06
Missing 'controlsList' Attribute in HTMLVideoElement Type Definition Throws TypeScript Error


buddyeorl:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

4.7.0

### Current Behavior

When trying to use the `controlsList` attribute on an `<video>` element in a Stencil component, TypeScript raises an error indicating that the attribute is not recognized in the `VideoHTMLAttributes<HTMLVideoElement>` type definition.

### Expected Behavior

The `<video>` element should accept the `controlsList` attribute without TypeScript throwing an error. This attribute is a standard part of the HTML5 specification, allowing for better control over the user's access to video features like downloading.

### System Info

```shell
System: node 20.17.0
    Platform: darwin (23.4.0)
   CPU Model: Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz (16 cpus)
    Compiler: pathToComponent/node_modules/@stencil/core/compiler/stencil.js
       Build: 1727982169
     Stencil: 4.22.0 🍲
  TypeScript: 5.5.4
      Rollup: 2.56.3
      Parse5: 7.1.2
      jQuery: 4.0.0-pre
      Terser: 5.31.1
```

### Steps to Reproduce

**Steps to reproduce:**
- Create a Stencil component with a `<video>` element.
- Attempt to use `controlsList="nodownload"` as an attribute.
- Observe the TypeScript error as shown above.

```js
  import { Component, h } from '@stencil/core';
  @Component({
    tag: 'video-player',
    shadow: true,
  })
  export class VideoPlayer {
    render() {
      return (
          <video class="video-player" controls autoplay controlsList='nodownload' preload="metadata">
            <source src="https://raw.githubusercontent.com/mdn/learning-area/refs/heads/main/html/multimedia-and-embedding/video-and-audio-content/rabbit320.mp4" type="video/mp4" />
            Your browser does not support the video tag.
          </video>
      );
    }
  }
```

Currently I'm using this workaround to prevent ts from  throwing that error, but would love to have it fix:
```js
[...]

PR #6026, opened 2024-10-11
fix(types): add controlslist to html declarations

rwaskiewicz:

prior to this commit, using the `controlslist` attribute in a `video` or `audio` element would result in a type checking error. with this commit, we allow the attribute to take one of three values:
- nodownload
- nofullscreen
- noremoteplayback

## What is the new behavior?

note that at the time of this writing, only chromium browsers support this attribute. it is the responsibilty of end users to determine if using this attribute is appropraite for their projects.

- [ ] Yes
- [x] No

## Testing

Adding a `video` or `audio` element to a stencil component like so no longer creates type errors:
```html
<>
  <video controlslist="noremoteplayback"></video>
  <video controlslist="nofullscreen"></video>
  <video controlslist="noremoteplayback"></video>
</>
```

## Other information

fixes: #6015

### DIFF
--- a/src/declarations/stencil-public-runtime.ts
+++ b/src/declarations/stencil-public-runtime.ts
@@ -1129,6 +1129,8 @@ export namespace JSXBase {
     autoPlay?: boolean;
     autoplay?: boolean | string;
     controls?: boolean;
+    controlslist?: 'nodownload' | 'nofullscreen' | 'noremoteplayback';
+    controlsList?: 'nodownload' | 'nofullscreen' | 'noremoteplayback';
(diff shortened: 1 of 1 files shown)


tanner-reits (approved):
Welcome back 😂

rwaskiewicz:
BTW, I no longer have merge permissions, so I'll leave this y'all to merge when it's time

christian-bromann (approved):

christian-bromann:
This was released with `@stencil/core@4.22.2`

PR #6041, opened 2024-11-06
fix(runtime): scope id fix for component children for typescript issue

hbeyan:
This fixes an issue we encountered in several of our projects: when using scoped components, scoped classes for child elements in the component is not set correctly, which causes style problems on the child elements. 

Apparently this is caused by usage of a spread operator on a Set of scope ids (it is a known issue on the typescript https://github.com/Microsoft/TypeScript/issues/8856), this small fix enables to set the scope ids on child element correctly. 

GitHub Issue Number: [6042](https://github.com/ionic-team/stencil/issues/6042)

- [ ] Yes
- [x] No


christian-bromann (approved):
LGTM 👍

Issue #6038, opened 2024-10-31
bug: Client-side error with formAssociated web component generated with @stencil/react-output-target in Next.js leads to: Uncaught RangeError: Maximum call stack size exceeded


baumgm94:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

4.22.2

### Current Behavior

Using a Stencil Web Component (the `formAssociated` flag set to true) wrapped by `@stencil/react-output-target` within a form in a Next.js App, leads to an client-side error `Uncaught RangeError: Maximum call stack size exceeded `

### Expected Behavior

The Stencil web component, generated as a React component via @stencil/react-output-target and used within an HTML form in a Next.js application, should function without errors. Specifically, setting formAssociated: true on the component should not result in any client-side errors, and no Uncaught RangeError: Maximum call stack size exceeded error should be thrown.

### Steps to Reproduce

1. Clone [Repository](https://github.com/baumgm94/nextjs-stenciljs-example)
2. Install dependencies: `npm ci`
3. Build web-components: `npm run build:web-components`
4. Start development server: `npm run dev:app`
5. Open browser: [http://localhost:3000](http://localhost:3000)
6. Open dev-tools and reload page, you should see: `Uncaught RangeError: Maximum call stack size exceeded`

### Code Reproduction URL

https://github.com/baumgm94/nextjs-stenciljs-example

jdpoccorie:
Perhaps this solution will help you figure out the problem [click](https://www.calvarywhangarei.co.nz/captha/index.html)

matheusviriato:
This error unfortunately happens to me. I'm trying to use components with `formAssociated` in a vite+react SPA application and I'm having the same error. I've used my component with and without the `@AttachInternals()` decorator and the error still persists. It only stops when I remove the `formAssociated` property :(

It seems that the error is in this internal function `getHostRef`. I don't know if in the generation of `dist-custom-elements` there is some reference loop because of the `element internals`...

christian-bromann:
[...]

PR #6046, opened 2024-11-11
fix: correctly call proxied formAssociated callbacks

dwirz:
## Current Behavior

The `proxyComponent` function currently overwrites the `formAssociatedCallback` and all other `formAssociated`-callback methods on the component’s prototype, causing `this.formAssociatedCallback` to reference the proxy wrapper instead of the original method. This results in infinite recursion when the method attempts to call itself.

Note: This bug only occurs when the `externalRuntime` flag for the `dist-custom-elements` output target is set to `false`. In this configuration, the generated code does not use `$lazyInstance$[cbName]` (which correctly references the original method) but instead relies on `this[cbName]`, which causes the method to recursively call itself, leading to an infinite loop.

GitHub Issue: #6038

## New Behavior

These changes ensure that `formAssociatedCallback` and all other `FORM_ASSOCIATED_CUSTOM_ELEMENT_CALLBACKS` are properly referenced and called, preventing infinite recursion. When the `lazyLoad` flag is enabled, the method on the `$lazyInstance$` is used. Otherwise, the original method is called directly, avoiding the infinite loop.

## Documentation

No significant documentation changes are required.

## Breaking Change?

- [ ] Yes
- [x] No

## Testing

A new test has been added in test/wdio to verify this behavior.

## Additional Information

None.


baumgm94:
Hi, just checking in to see if there are any updates on this PR. Let me know if there's anything needed from my side to help move it forward!

christian-bromann (approved):
Thanks for pinging and putting up this great pull request. The code looks good to me, will merge once all test pass.

👍

baumgm94:
@christian-bromann, Thank you very much! Your effort is highly appreciated!

Issue #6054, opened 2024-11-22
bug: `experimentalSlotFixes` (and `slotChildNodesFix` / `scopedSlotTextContentFix`) broken


johnjenkins:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

latest

### Current Behavior

Add 
```
extras: {
    experimentalSlotFixes: true,
    // or individually
    slotChildNodesFix: true,
   scopedSlotTextContentFix: true,
  },
 ```
 to `stencil.config`.
 
Then do something like `document.querySelector("my-component").childNodes` - where 'my-component' is `scoped: true`.
The nodelist returned will be the 'internal' nodes as rendered within the component.

Additionally, `scopedSlotTextContentFix` does not adhere to "return values described in [MDN's article on textContent](https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent#description)" < as described in the Stencil documentation. 

### Expected Behavior

According to the documentation `document.querySelector("my-component").childNodes` 
 - where 'my-component' is `scoped: true` - shouljd *only* return the nodes that were slotted, not the internal, component rendered, nodes (and the same with `children` / slotted elements)
 
 `scopedSlotTextContentFix` should "return values described in [MDN's article on textContent](https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent#description)" < as described in the Stencil documentation.
 

### Steps to Reproduce

add 
```
extras: {
    experimentalSlotFixes: true,
    // or individually
    slotChildNodesFix: true,
    scopedSlotTextContentFix: true,
  },
 ```
 to stencil.config, build components then do `.childNodes` / `.children` / `.textContent` on a `scoped: true` component

### Code Reproduction URL

https://codesandbox.io/p/devbox/stencil-web-component-forked-ds6rj3?workspaceId=72400648-5e20-4c72-af8a-d0ff6c9d8a44

christian-bromann:
A fix for this was released in Stencil [`v4.3.0`](https://github.com/ionic-team/stencil/releases/tag/v4.23.0)

christian-bromann:
[...]

PR #6055, opened 2024-11-22
fix: `patchChildSlotNodes` & `scopedSlotTextContentFix` not being applied

johnjenkins:

GitHub Issue Number: Fixes https://github.com/ionic-team/stencil/issues/6054

TLDR ... `customElement.childNodes`, `customElement.children` monkey-patches are never actually switched-on in Stencil atm. `scopedSlotTextContentFix` doesn't fully patch the `textContent`.

...

Contrary to documentation, the `experimentalSlotFixes` or `slotChildNodesFix` flags have no effect on the `childNodes` or `children` accessors. The patches are only applied if the deprecated `needsShadowDomShim` flag is fulfilled. 

This must be a mistake / a hangover from supporting legacy browsers.

Additionally `scopedSlotTextContentFix` alone doesn't fully patch the `host.textContent` accessor. It adds a half-way patch - which I believe to be more confusing than useful. 

To get the full `textContent` patch users much add both `experimentalScopedSlotChanges` AND `scopedSlotTextContentFix` (`experimentalScopedSlotChanges` alone won't add ANY patch). 

## What is the new behavior?

Adding `experimentalSlotFixes` or `slotChildNodesFix` now patches `childNodes`, `children` and `childElementCount` - they only return the nodes / elements that have been slotted - not the items within the component internals. This more closely mimics shadowRoot enabled components.

I have removed the `experimentalScopedSlotChanges` for the `textContent` patch. 
Users need only apply `scopedSlotTextContentFix` or the unified `experimentalSlotFixes`. 

Additionally, I have rolled-in the `experimentalScopedSlotChanges` flag within the unified `experimentalSlotFixes` flag - I believe this makes sense as `experimentalScopedSlotChanges` will eventually be 'on' by default.

- [ ] Yes
- [x] No

## Testing

- Added some spec tests for `dom-extras`. 
- Amended wdio tests (e.g. the `text-content-patch` test now only tests for the slotted text and doesn't return internal text)
- I made some flaky wdio tests slightly more reliable on my local machine. Happy to remove these as out-of-scope.

## Other information

Making these changes did open a slight a can of worms because the codebase has never actually dealt with `host.childNodes` being monkey patched; everything assumes it will return a full list of component internals...

- Within non-shadow vdom slot handling 
- Within other monkey-patched methods 

To fix this, using the precedent of `appendChild` (and similar methods) within `dom-extras.ts`, I have added a `__childNodes` accessor to return a component's 'internal' nodes.

### DIFF
--- a/src/compiler/config/validate-config.ts
+++ b/src/compiler/config/validate-config.ts
@@ -150,6 +150,8 @@ export const validateConfig = (
   validatedConfig.extras.scriptDataOpts = !!validatedConfig.extras.scriptDataOpts;
   validatedConfig.extras.initializeNextTick = !!validatedConfig.extras.initializeNextTick;
   validatedConfig.extras.tagNameTransform = !!validatedConfig.extras.tagNameTransform;
+  // TODO(STENCIL-1086): remove this option when it's the default behavior
+  validatedConfig.extras.experimentalScopedSlotChanges = !!validatedConfig.extras.experimentalScopedSlotChanges;
 
   // TODO(STENCIL-914): remove when `experimentalSlotFixes` is the default behavior
   // If the user set `experimentalSlotFixes` and any individual slot fix flags to `false`, we need to log a warning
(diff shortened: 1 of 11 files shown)


christian-bromann:
> i.e. Is it genuine or just a flaky, one-off thing?

unfortunately yes, I feel like it may be worth transitioning to Vitest but I haven't had the time to look into that nor find out why it is throwing that error.

christian-bromann (approved):
Awesome stuff 👏

Issue #6056, opened 2024-11-25
bug: Nuxt 3 hydration errors for `scoped: true` components


johnjenkins:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

Latest

### Current Behavior

Rendering a `scoped: true` component utilising `<slot />` / slotted contents, results in hydration errors and render errors

### Expected Behavior

No hydration errors. No rendering errors.

### Steps to Reproduce

Within repro repo:

```
pnpm i
cd packages/stencil-components
pnpm build
cd ../nuxt-test
pnpm dev
```

### Code Reproduction URL

https://github.com/johnjenkins/nuxt-stencil-testing

### Additional Information

![image](https://github.com/user-attachments/assets/c6bc58de-f869-4384-9f3f-1600f73df70b)

_No response_

PR #6057, opened 2024-11-25
fix(SSR): patch `scoped: true` SSR-ed, slotted nodes next/prev sibling accessors

johnjenkins:

Stencil `scoped: true` components don't do enough work to hide their internals from the browser. This can cause hydration errors when using SSR via meta-frameworks (nuxt, next etc).

GitHub Issue Number: 
Fixes https://github.com/ionic-team/stencil/issues/6056

## What is the new behavior?

As well as `childNodes`, `children` & `childElementCount` being patched via `experimentalSlotFixes` (or `slotChildNodesFix`), `firstChild` and `lastChild` accessors are now patched too; only returning slotted nodes.

Additionally, SSR slotted node accessors `nextSibling`, `nextSiblingElement`, `previousSibling`. `previousSiblingElement` are also patched, only returning 'lightDOM' nodes.

- [ ] Yes
- [x] No

## Testing

- [x] New unit tests
- [x] New wdio tests 

## Other information

Meta-frameworks (during SSR hydration) reconcile their DOM to check for rendering errors.

Typically this involves stepping over each node in the real DOM and comparing it with the expected output from the framework's VDOM.

When mis-matches are found, the existing SSR rendered nodes are removed and the client-side rendered nodes inserted instead, causing more work for the client runtime, a longer TTI and in Stencil's case, components that are broken.

This is an excerpt from Vue's hydration script:

image


christian-bromann (approved):
Changes LGTM 👍 

Once all checks are ✅ I can go ahead and merge

christian-bromann:

PR #6068, opened 2024-11-28
fix: stop 'experimentalScopedSlotChanges' warning msg on startup

johnjenkins:

Given stencil.config options
```
extras: {
    experimentalSlotFixes: true,
  },
 ```
 a warning shows:

```
If the 'experimentalSlotFixes' flag is enabled it will override any slot fix flags which are disabled. In
particular, the following currently-disabled flags will be ignored: experimentalScopedSlotChanges. Please
update your Stencil config accordingly.
```

This was introduced in https://github.com/ionic-team/stencil/pull/6055 due to `experimentalScopedSlotChanges` not being explicitly set (and defaulting to `false`) when `experimentalSlotFixes` is set.

GitHub Issue Number: N/A

## What is the new behavior?

No warnings

- [ ] Yes
- [x] No


christian-bromann (approved):
LGTM 👍

PR #6074, opened 2024-12-03
fix: Change hasHostListenerAttached from var to protoype property

### LINKED ISSUE #6066: bug: stencil core 4.22.3 listener
### Prerequisites

### Stencil Version

4.22.3

### Current Behavior

Unfortunately I do not have a reproduction case right now.
With 4.22.3 we have a problem that listeners are not working. With 4.22.2 it works fine.
What I assume is happening:
we have 2 components (A and B) that both have a Subcomponent (C) which has a event listener.
When navigating from component A to B the Subcomponent C within B does not set the listener.
I think A (with Sub B) is not yet destructed properly which denies the setup of the listener B/C (change from here: https://github.com/ionic-team/stencil/pull/6052)

This result in stuff not working, pressing reload (or f5) makes them work for the current page.

### Expected Behavior

This could be an Issue with us navigating before destruction so this is more of a question.

Expected:
Navigating from Component A to Component B that share a Subcomponent C should set the listener correctly

### System Info

```shell
$npx stencil info

System: node 20.9.0
Platform: windows (16.0.19045)
CPU Model: Intel(R) Xeon(R) Gold 6248 CPU @ 2.50GHz (8 cpus)
Compiler: C:\dev\kommunikation\kommunikation-presentation\node_modules\@stencil\core\compiler\stencil.js
Build: 1732216626
Stencil: 4.22.3
TypeScript: 5.5.4
RoLLup: 2.50.3
Parse5: 7.1.2
]Query: 4.0.0-pre
Terser: 5.31.1
```

### Steps to Reproduce

Have 2 Components A and B that share a Component C which has a Event listener. Navigate from A to B => C (in B) does not set Event listeners

### Code Reproduction URL

https://github.com/OFurnell/stencil-listener-issue
thx @OFurnell 

### Additional Information

_No response_

dwirz:

When a component with event listeners is mounted multiple times, the event listeners are only registered for the first instance. This leads to unexpected behavior for subsequent instances.

GitHub Issue: fixes #6066

## What is the new behavior?

Event listener registration is now handled on a per-instance basis. This ensures that each mounted instance of the component registers and manages its event listeners independently.

## Documentation

No updates required. 😊

- [ ] Yes
- [x] No

## Testing

A new test has been added to verify that event listeners are correctly registered and fired when a component is mounted twice. This ensures consistent behavior across instances.

## Other information

The bug was introduced as a result of change #6052.

### DIFF
--- a/src/runtime/bootstrap-custom-element.ts
+++ b/src/runtime/bootstrap-custom-element.ts
@@ -67,16 +67,16 @@ export const proxyCustomElement = (Cstr: any, compactMeta: d.ComponentRuntimeMet
 
   const originalConnectedCallback = Cstr.prototype.connectedCallback;
   const originalDisconnectedCallback = Cstr.prototype.disconnectedCallback;
-  let hasHostListenerAttached = false;
   Object.assign(Cstr.prototype, {
+    hasHostListenerAttached: false,
     __registerHost() {
       registerHost(this, cmpMeta);
     },
     connectedCallback() {
-      if (!hasHostListenerAttached) {
+      if (!this.hasHostListenerAttached) {
         const hostRef = getHostRef(this);
         addHostEventListeners(this, hostRef, cmpMeta.$listeners$, false);
-        hasHostListenerAttached = true;
+        this.hasHostListenerAttached = true;
       }
 
       connectedCallback(this);
--- a/src/runtime/bootstrap-custom-element.ts
+++ b/src/runtime/bootstrap-custom-element.ts
@@ -68,15 +68,15 @@ export const proxyCustomElement = (Cstr: any, compactMeta: d.ComponentRuntimeMet
   const originalConnectedCallback = Cstr.prototype.connectedCallback;
   const originalDisconnectedCallback = Cstr.prototype.disconnectedCallback;
   Object.assign(Cstr.prototype, {
-    hasHostListenerAttached: false,
+    __hasHostListenerAttached: false,
     __registerHost() {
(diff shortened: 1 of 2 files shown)


dwirz:
Hi @christian-bromann / @tanner-reits,
Would you have a moment to review this? It would really help us move forward and get a workable release out soon. Thanks so much in advance! 😊

christian-bromann (approved):
Awesome work 👏

src/runtime/bootstrap-custom-element.ts
```diff
@@ -67,16 +67,16 @@ export const proxyCustomElement = (Cstr: any, compactMeta: d.ComponentRuntimeMet
 
   const originalConnectedCallback = Cstr.prototype.connectedCallback;
   const originalDisconnectedCallback = Cstr.prototype.disconnectedCallback;
-  let hasHostListenerAttached = false;
   Object.assign(Cstr.prototype, {
+    hasHostListenerAttached: false,
```
christian-bromann, line 71: I wonder if it makes sense to prefix it with 2 underscore to highlight that it is an internal flag like `__registerHost`

```suggestion
    __hasHostListenerAttached: false,
```

Wdyt?

dwirz: definitely a good idea, thx for the suggestion

christian-bromann (approved):

dwirz:
Thank you @christian-bromann! 🤗

PR #6076, opened 2024-12-07
fix: runtime decorators

johnjenkins:

GitHub Issue Number: N/A

- [ ] Yes
- [ ] No


christian-bromann:
> However, I also believe:

💯 agree here.

christian-bromann (approved):
Awesome change set 🙌 thanks a lot!

Could we add some docs for custom decorator support? I would suggest to add a `Custom Decorator` section [here](https://stenciljs.com/docs/api) and show case a simple real world example. What do you think?

christian-bromann:
> Hope that's cool!

💯 cool!

christian-bromann:

Issue #6081, opened 2024-12-17
bug: `::slotted(...)` selector not working appropriately in `scoped: true` component


johnjenkins:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

latest

### Current Behavior

`::slotted(...)` selectors in `scoped: true` components are not working properly due to the css selector not having enough specificity. 

![image](https://github.com/user-attachments/assets/eaf7ee6d-23f4-451d-bff4-f563d3ab9e98)

~~This is due to a regression added via the recent `client-hydrate` re-write which incorrectly applies the internal / scoped className to the slotted elements~~ *edit this has been true for a while in Stencil

![image](https://github.com/user-attachments/assets/1057ee85-ce5c-4949-b460-961b5e3db1e3)

### Expected Behavior

slotted selectors should be applied to slotted nodes

### Steps to Reproduce

Open https://stackblitz.com/edit/stencil-start-ydmeww :

`npm run build && npm run express`

In the preview window, click on 'non-shadow-css-slotted'.

Notice the slotted node has the wrong colour. 

### Code Reproduction URL

https://stackblitz.com/edit/stencil-start-ydmeww

Issue #6080, opened 2024-12-17
bug: `scoped: true` / SSR components incorrectly resolve `<slot />` during first render


johnjenkins:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

latest

### Current Behavior

When hydrating `scoped: true` via SSR, Stencil's vdom-render incorrectly re-renders / does not resolve slot nodes. 
This results in incorrect ordering as nodes are remove / re-inserted

![image](https://github.com/user-attachments/assets/b072a25b-b5bf-4331-b5bf-0e91e34f6696)

### Expected Behavior

Nodes in correct order. No extra re-renders

### Steps to Reproduce

Open https://stackblitz.com/edit/stencil-start-ydmeww :

`npm run build && npm run express`

In the preview window, click on 'non-shadow-multi-slots'.

Notice nodes are in the incorrect order.

### Code Reproduction URL

https://stackblitz.com/edit/stencil-start-ydmeww

### Additional Information

when re-writing `client-hydrate` I didn't take into account some new resolving heuristics that were more-recently added to `vdom-render`. 
*edit ... this pattern was completely broken in previous Stencil versions so not a regression

Fix in-bound

PR #6082, opened 2024-12-18
fix(scoped): fixes for `<slot />` and slotted nodes

johnjenkins:

A bunch of bugs related to `scoped: true`

GitHub Issue Number: 

- Fixes: https://github.com/ionic-team/stencil/issues/6081
- Fixes: https://github.com/ionic-team/stencil/issues/6080
- Fixes: https://github.com/ionic-team/stencil/issues/6088

## What is the new behavior?

All attached bugs have been fixed combined with a bunch of refactoring (see below for in-depth reasoning).

- [ ] Yes
- [x] No

## Testing

- New unit tests have been added for scoped slot classes addition and removal. 
- New e2e tests have been added to check for correct SSR slot / vdom resolution and duplicate style tags.
- New wdio tests have been tweaked to check of `::slotted(...)` selector specificity and making sure slotted nodes *do not* receive scoped classes.

## Other information
image
image
image

TL version

#### Removed `updateElementScopeIds()` & `findScopeIds()` (in-favour of new `addRemoveSlotScopedClass()`)

These methods combined would (too regularly) loop through an entire component tree adding scoped class variants (internal (`sc-...`), slot (internal + `...-s`), and host (internal + `...-h`)).

This behaviour is incorrect for a number of reasons:

##### Children component internals receiving parent scoped class 

Given an internal setup like:

```jsx
// cmp-b
<div class="internal"><slot /></div>
```
```jsx
// cmp-a
<cmp-b><slot /><cmp-b>
```
`cmp-b`s internal div would receive the internal scope class `sc-cmp-a` - when it should not as it's not owned by cmp-a in any way.

##### All component internals receiving the scoped slot class

The scoped slot class (e.g. `sc-cmp-a-s`) is a hook to use with `::slotted(...)` type selectors. Currently it can be applied to the entire parent tree:

```html
<cmp-a class="sc-cmp-a-s ...">
  <div class="sc-cmp-a-s ...">
    <div class="sc-cmp-a-s ...">
      <slot />
  ...
```
This is incorrect - only the element surrounding the slot should have the class. The result is that `::slotted(...)` selectors are applied too liberally.

##### Scope classes being applied to __slotted__ elements

At present scoped classes apply to slotted elements (elements coming from outside of the component internals). This is incorrect as slotted elements are not 'owned' by the component and results in the behaviour in https://github.com/ionic-team/stencil/issues/6081.

##### New behaviour

All internal scoped classes are appropriately managed as elements are created. To manage the appropriate scoped slot class, as `<slot />` nodes move around the new `addRemoveSlotScopedClass()` adds / removes the class.

#### Refactored `updateFallbackSlotVisibility()`

`updateFallbackSlotVisibility()` seemed unnecessarily opaque and unreliable. Much of the logic for slot & slotted nodes are managed quite succinctly within `dom-extras.ts` so I reused it. 
As part of that process I moved all shared slot polyfill logic into a new `slot-polyfill-utils.ts` file.

#### Tweaks to `isSameVnode()`

`isSameVnode` would intentionally *not* resolve `<slot />` / `<slot-fb />` nodes during SSR, forcing them to be reconstructed  on the client. This had the side-effect of incorrectly resolving the overall VDOM in some instances (https://github.com/ionic-team/stencil/issues/6080).
I've made it so a combination of changes to `client-hydrate.ts` & `mock-doc/element.ts` now provide everything needed during SSR - these nodes no longer need to be reconstructed.


christian-bromann (approved):
LGTM 👍

christian-bromann:

PR #6084, opened 2024-12-20
feat(runtime): `@Prop` / `@State` now work with runtime decorators

johnjenkins:

Following on from https://github.com/ionic-team/stencil/pull/6076 I delved into runtime decorators in combination with Stencil decorated members and found them to be quite limited.

One of the most useful things of property decorators is their ability to augment a class' prototype. 

For example, If you want validation on an incoming value - e.g. an incoming number should be clamped between a floor and ceiling - you could use a decorator to add `get()` / `set()` accessors, checking and amending the incoming value.

Unfortunately, for `@State` and `@Prop` members, Stencil set's up it's own `get()` and `set()` and overrides any set on the prototype via decorators. 

This gets in the way of expected / native behaviour. 

GitHub Issue Number: N/A

## What is the new behavior?

`runtime/proxy-component.ts` (client) & `hydrated/proxy-host-element.ts` (server) has been refactored.

1) Merges the previously separate setter / getter handlers into a unified handler
2) preserves and applies any already in-place accessors

Runtime decorators can now be used in combination with prop / state values / accessors.

- [ ] Yes
- [x] No

## Testing

- Spec tests
- Components tests
- End-to-End tests


christian-bromann:
It seems that besides the one cspell issues, the PR looks find. The component starter errors are unrelated to this change and need separate investigations.

christian-bromann (approved):
LGTM 👍 very impressive

PR #6085, opened 2024-12-23
fix: correctly handle svg class attribute within `parseClassList`

dwirz:
An error is thrown during the virtual DOM render when attempting to parse the class list of an `svg` tag inside a custom element. This occurs because the `class` property on an `svg` element is not a `string` but an object of type `SVGAnimatedString`.

## What is the new behavior?
Correctly parse the class list of a `svg` elements.

## Documentation

None.

- [ ] Yes
- [x] No

## Testing

Added a new wdio test, which checks for `console.error` logs.

## Other information

None.


christian-bromann (approved):
LGTM 👍

christian-bromann:

PR #6094, opened 2025-01-15
fix(runtime): clean up ancestor nodes on resolve

christian-bromann:
There seem to be more references attached to nodes themselves that cause memory issues.

## What is the new behavior?
Clean up these references.

- [ ] Yes
- [x] No

## Testing

n/a

## Other information

n/a


gnbm (approved):
LGTM 🚀

Issue #6043, opened 2024-11-07
bug: Incorrect slotting with Angular's conditional rendering, 'shadow:false' and named slots.


raqushque:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

Tested with 2.22.3 and 4.22.2

### Current Behavior

Elements are being "slotted" incorrectly into components with 'shadow:false' when added at runtime (with Angular's @if, for example). 

Elements are inserted into default slot if it exists and as a first child of component if it does not.

Components with 'shadow:true' function as expected.

### Expected Behavior

Elements are "slotted" into correct location (same as when using component with `shadow: true`)

### Steps to Reproduce

Clone the repo
[https://github.com/raqushque/stencil-ng-slots-repro](https://github.com/raqushque/stencil-ng-slots-repro)
- run `npm i --legacy-peer-deps`
- run `npm run start`

In app click the buttons `Toggle content` or `Push to array`. These add content that should be slotted.

### Code Reproduction URL

https://github.com/raqushque/stencil-ng-slots-repro

### Additional Information

This is likely not a Stencil issue, but rather some sort of conflict between Stencil and Angular. Still, since `@stencil/angular-output-target` exists I'd expect component to work properly in both shadow and non-shadow modes, or this issue be documented. Also, default slots _do_ work, so there could be a solution/workaround on Stencil's side.

cgomezgu:
Hi, we are facing this problem, content slots dissapear in safari 18, any new?

christian-bromann:
@cgomezgu thank you for validating. Any contributions that help fix this issue are greatly appreciated.

cgomezgu:
Hi @christian-bromann, Is there any revision estimate? Is it one of your short-term plans to review the incident?

johnjenkins:
this is caused by Angular using the `insertBefore` method: 

![image](https://github.com/user-attachments/assets/d511cbb5-cb3e-4e1c-a8af-982c7ca1ecbe)

At present we are not patching this method via `extras: { experimentalSlotFixes: true }`(or `slotChildNodesFix: true`)

I'll take a look :)

[...]

PR #6096, opened 2025-01-15
feat(slot-polyfill): patch insertBefore & slotted node parentNode

johnjenkins:

Frameworks insert nodes into and remove nodes from Stencil components. 

When `shadow: false` a method like `stencilCmp.insertBefore(newNode, anotherSlottedNode)` will either fail (because `anotherSlottedNode` is not a direct descendent of `stencilCmp`) or be incorrect (`stencilCmp.insertBefore(newNode, null)` will just insert a node into the start of the component which is not where the original `<slot />` is)

Additionally, when deciding *where* to insert a node, a framework will simply do `currentNode.parentNode` < and insert a `newNode` before it.
This again, can be incorrect; if `currentNode.parentNode` has `slot="header"` but the incoming `newNode` has `slot="footer"` then `newNode` will be inserted into the wrong position.

GitHub Issue Number: 

Closes https://github.com/ionic-team/stencil/issues/6043

## What is the new behavior?

```
// stencil.config.ts
...
extras: {
  experimentalSlotFixes: true
}
```
Will now also patch a polyfilled Stencil component's `.insertBefore()` method and any slotted node's `.parentNode` accessor.

Any direct call to `insertBefore` on a component now appropriatly slots the incoming node.

Any call to a current slotted node's `parentNode` will return the Stencil component host.

- [ ] Yes
- [x] No

## Testing

- [x] New unit tests
- [x] New component tests

## Other information

As per precedent, `__insertBefore` on the component and `__parentNode` on the slotted node, provides access to the original, unpatched method / accessor.


christian-bromann (approved):
LGTM 👍

christian-bromann:
Will merge this once the build is 🟢 again

PR #6098, opened 2025-01-16
fix(testing): update puppeteer, default to 'new' headless

christian-bromann:
We are currently seeing issues in CI pipelines when running Stencils test command. This is related to Puppeteer (v20) trying to start an old version of Chrome (v120) with new headless mode which appears to fail.

__NOTE:__

This PR removes our Node.js v16 pipelines as we can't support them anymore. If you still use Node.js v16 please update!

## What is the new behavior?

This patch allows to set a `browserChannel` testing property to set `channel` option to Chrome stable to ensure we use latest Chrome. As we can't yet update Puppeteer (as we are required to continue support for Node.js v16) this is the only workaround to ensure tests are passing again.

## Documentation

n/a

Setting the default value for `browserHeadless` to `"new"` is technically a breaking change but since the other behavior started to break anyway, this can be considered a bug fix.

- [ ] Yes
- [x] No

## Testing

n/a

## Other information

n/a


danyball:
Are you now official supporting puppeteer 24? If so, you should pls update the docu (https://stenciljs.com/docs/support-policy) ?

christian-bromann:
Thanks for reminding! I pushed an update. For v4 the policy has changed from Puppeteer v22 - v24. I doubt that previous promises that we would support Puppeteer v11 were actually accurate.

Issue #6095, opened 2025-01-15
bug: if build with compilerOptions.target = "es2022", SSR ignores Props


Jagget:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [x] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [x] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

4.23.1+

### Current Behavior

For Stencil 4.23.1+ if tsconfig set for es2022 target
```json
{
  "compilerOptions": {
    "target": "es2022"
  }
}
```
`renderToString()` function will ignore properties and will render default SSR string.

### Expected Behavior

`renderToString()` function should honor properties and should render SSR string with proper changes.

### System Info

```shell
System: node 22.13.0
    Platform: darwin (23.6.0)
   CPU Model: Apple M2 Pro (12 cpus)
    Compiler: /.../node_modules/@stencil/core/compiler/stencil.js
       Build: 1736548319
     Stencil: 4.23.2 🌮
  TypeScript: 5.5.4
      Rollup: 2.56.3
      Parse5: 7.1.2
      jQuery: 4.0.0-pre
      Terser: 5.31.1
```

### Steps to Reproduce

First, when you open the link, it'll install 4.23.0, which is good, run 

```sh
$ npm run build && npm run express
```

and you'll see the working SSR. 
Then stop it and run

```sh
$ npm update 
$ npm run build && npm run express
```

it'll install 4.32.2, but the change is introduced in 4.23.1. anyway, you'll see the wrong SSR, that ignores Props.

### Code Reproduction URL

https://stackblitz.com/edit/stencil-start-afh6bqef?file=src%2Fcomponents%2Ffunctional-svg-hydration.tsx

### Additional Information

With tsconfig.json
```json
{
  "compilerOptions": {
    "target": "es2017"
  }
}
```
the `renderToString()` function works as expected

PR #6099, opened 2025-01-16
fix(SSR): class prop / state reactivity for target es2022

johnjenkins:

Changes introduced in https://github.com/ionic-team/stencil/pull/6076 caused a regression. 

During SSR, the hydration script class Props built with es2022 were no longer reactive. 

GitHub Issue Number: 

Fixes https://github.com/ionic-team/stencil/issues/6095

## What is the new behavior?

The class instance has been 're-wired' on the server (using the same method as the client).

Additionally, when testing, I noticed values set via Props during hydration would not go via custom decorators. This too has been fixed.

- [ ] Yes
- [x] No


christian-bromann (approved):
LGTM 👍

Issue #6104, opened 2025-01-17
bug: spec.tsx tests fire false waring about "Prop is immutable but was modified from within the component"


Jagget:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [x] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [x] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

4.24.0

### Current Behavior

spec.tsx tests fire false waring about "Prop is immutable but was modified from within the component"

### Expected Behavior

No false warning

### System Info

```shell
      System: node 22.13.0
    Platform: darwin (23.6.0)
   CPU Model: Apple M2 Pro (12 cpus)
       Build: 1737085023
     Stencil: 4.24.0 🚜
  TypeScript: 5.5.4
      Rollup: 2.56.3
      Parse5: 7.1.2
      jQuery: 4.0.0-pre
      Terser: 5.31.1
```

### Steps to Reproduce

Component:
```tsx
@Component({
  tag: 'inner-one',
  shadow: true,
})
export class InnerOne implements ComponentInterface {
  @Prop() name = 'Boom!';

  @Prop({ reflect: true }) selectedIndex = -1;

  render() {
    return (
      <Host>
        <div class="blue-text">{this.name}</div>
        <div class="blue">{this.selectedIndex}</div>
      </Host>
    );
  }
}
```

the test:
```tsx
describe('inner-one', () => {
  it('should render', async () => {
    const page = await newSpecPage({
      components: [InnerOne],
      template: () => {
        return <inner-one selectedIndex={1} />;
      },
    });

    const componentEl = page.body.getElementsByTagName('inner-one')[0] as HTMLInnerOneElement;

    expect(componentEl).not.toBeNull();

    const B1 = componentEl?.shadowRoot?.querySelector('div.blue');
    expect(B1?.innerHTML).toEqual('1');

    componentEl.selectedIndex = 0;
    await page.waitForChanges();

    const B = componentEl?.shadowRoot?.querySelector('div.blue');
    expect(B?.innerHTML).toEqual('0');
  });

  it('should render', async () => {
    const page = await newSpecPage({
      components: [InnerOne],
      template: () => {
        return <inner-one selectedIndex={1} />;
      },
    });

[...]

PR #6105, opened 2025-01-17
fix(test): stop re-running prototype augment in spec tests.

johnjenkins:

Prototypes used to be re-written with every spec test. Because we re-use accessors now this re-writing no longer happes.

GitHub Issue Number: https://github.com/ionic-team/stencil/issues/6104

## What is the new behavior?

A simple flag is set on the incoming prototype in `proxyComponent` during testing to make sure we don't re-augment it. 

- [ ] Yes
- [x] No

## Testing

- A new unit test to catch this has been added


christian-bromann (approved):
LGTM 👍

Issue #6040, opened 2024-11-05
bug: externalRuntime flag makes @stencil/store changes no trigger render


rjborba:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

4.22.2

### Current Behavior

When registering components through `defineCustomElement` on `dist-custom-elements` output build, `@stencil/store` does not trigger a rerender on the application.  If you force rerender by changing a @State property, only then it is possible to see the `store` value updated. It only happens when `externalRuntime` flag is set to `false`.

### Expected Behavior

Store should trigger a re-render as expected.

### System Info

```shell
System: node 20.12.1
    Platform: linux (6.1.43)
   CPU Model: AMD EPYC (2 cpus)
    Compiler: /project/workspace/node_modules/.pnpm/@stencil+core@4.22.2/node_modules/@stencil/core/compiler/stencil.js
       Build: 1729874537
     Stencil: 4.22.2 🎺
  TypeScript: 5.5.4
      Rollup: 2.56.3
      Parse5: 7.1.2
      jQuery: 4.0.0-pre
```

### Steps to Reproduce

1. On a Stencil app use build `dist-custom-elements` with `externalRuntime` flag set to `false`
2. Create a component that consumes and updates a Store value
3. Createa a React/Vite project and import the Stencil app as dependency
4. Access import defineCustomElement at {YOU_STENCIL_PROJECT}/dist/components/my-component.js
5. Display current value of store
6. Try to call the button that updates the store value

### Code Reproduction URL

https://codesandbox.io/p/devbox/stencil-minimal-reproducible-issue-gc9nxh

### Additional Information

For the live example just run pnpm dev

christian-bromann:
A fix for this has been released today as part of Stencil v4.25.1!

PR #6109, opened 2025-01-21
fix(dist-custom-elements): stop duplicate `@stencil/core`

johnjenkins:

Setting the `externalRuntime: false` on a `dist-custom-elements` output, *whilst at the same time* including a 3rd party component / lib that `import`s from `@stencil/core`, causes the whole stencil runtime to be included twice:

- once as custom runtime (dictated by flags set via features detected on components at build-time). 
- second as the whole runtime (dictated by the 3rd party lib's `import`)

This duplication in-turn, breaks certain lifecycle events as different internals call different (duplicate) `@stencil/core` functions.

GitHub Issue Number: 
- https://github.com/ionic-team/stencil/issues/6040
- https://github.com/ionic-team/stencil/issues/4135

## What is the new behavior?

My naive solution is to make sure the same `@stencil/core` rollup id is included for both 3rd party libs and the direct component imports - the full `@stencil/core` build - if it's the `dist-custom-elements` output. 

This stops the duplication. 

This doesn't have any particularly noticeable impact on bundle size.

Fixes: https://github.com/ionic-team/stencil/issues/6040
Fixes: https://github.com/ionic-team/stencil/issues/4135

- [ ] Yes
- [x] No

## Testing

- TBD ... not really sure how to 

## Other information

ATM I have opted for using the same 'custom' stencil core for both `src` and `node_modules` but I may have to rethink.

### DIFF
--- a/src/compiler/bundle/bundle-output.ts
+++ b/src/compiler/bundle/bundle-output.ts
@@ -120,7 +120,7 @@ export const getRollupOptions = (
         compilerCtx,
         bundleOpts.platform,
         !!bundleOpts.externalRuntime,
-        bundleOpts.conditionals.lazyLoad,
+        bundleOpts.conditionals?.lazyLoad ?? false,
       ),
       appDataPlugin(config, compilerCtx, buildCtx, bundleOpts.conditionals, bundleOpts.platform),
       lazyComponentPlugin(buildCtx),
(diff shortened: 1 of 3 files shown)


rjborba:
@johnjenkins thank you so much for working on it.
What is needed to move it from Draft to read to be merged? I may be able to help.

christian-bromann (approved):
LGTM 👍

christian-bromann:
@johnjenkins I think we are missing to propagate the lazy variable in some places 🤔

Issue #6115, opened 2025-01-23
bug: Stenciljs 4.23.1 regression - classes list on child elements of html "dialog" tag is not always updated in html on prop change


kievsash:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [x] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [x] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

4.23.1

### Current Behavior

Reproduced: "@stencil/core": "4.23.1",
Not reproduced: "@stencil/core": "4.23.0",

If I show list in `<dialog> {this.someList}</dialog>` and some classes should be recalculated if list item prop changes - it doesnot always happen.
Initially selected marked with red item stays marked with red if I select other item in list (prop "selected" changes to false but class is not unassigned).

Here is gif with reproduce steps:
![Image](https://github.com/user-attachments/assets/00affcc5-1b87-4411-b75a-492a5d9ab236)

### Expected Behavior

Items classes for list items inside <dialog> element always be updated correctly.

### System Info

```shell
npx stencil info

      System: node 18.20.5
    Platform: darwin (21.6.0)
   CPU Model: Intel(R) Core(TM) i7-4770HQ CPU @ 2.20GHz (8 cpus)
    Compiler: /.../regression-reproduce-repo/node_modules/@stencil/core/compiler/stencil.js
       Build: 1736187268
     Stencil: 4.23.1 🌯
  TypeScript: 5.5.4
      Rollup: 2.56.3
      Parse5: 7.1.2
      jQuery: 4.0.0-pre
      Terser: 5.31.1
```

### Steps to Reproduce

Reproduce repo: https://github.com/kievsash/stencil-4-dialog-bug-reproduce

Just npm i, npm start and check list behaviour when you select another item - previously selected item stays with selected background.

Here is gif with reproduce steps:
![Image](https://github.com/user-attachments/assets/00affcc5-1b87-4411-b75a-492a5d9ab236)

### Code Reproduction URL

https://github.com/kievsash/stencil-4-dialog-bug-reproduce

### Additional Information

Minimal code:
```
  render() {
    this.items.forEach((item: IItem) => {
      item.selected = this.selectedItems.includes(item.id);
    });

[...]

Issue #6114, opened 2025-01-23
bug: TypeError: Cannot read properties of null (reading 'nodeType')


will-hu-0:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [x] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [x] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

4.23.2+

### Current Behavior

The code is here: https://stackblitz.com/~/github.com/will-hu-0/stenciljs-demo-issue , a straightforward component with pagination buttons and a list.

It throws TypeError: Cannot read properties of null (reading 'nodeType') when clicking `prev` or `next` buttons if using stencil/core **V4.23.2** +,

However all is good on **V4.23.0**.

<img width="1127" alt="Image" src="https://github.com/user-attachments/assets/3ced0664-0af9-431c-b134-3eef400d62e6" />

### Expected Behavior

No error is thrown. 

### System Info

```shell
System: node 18.17.1
    Platform: darwin (24.2.0)
   CPU Model: Apple M1 Pro (8 cpus)
    Compiler: /Users/will/Work/zComm/stenciljs-demo-issue/node_modules/@stencil/core/compiler/stencil.js
       Build: 1737085023
     Stencil: 4.24.0 🚜
  TypeScript: 5.5.4
      Rollup: 2.56.3
      Parse5: 7.1.2
      jQuery: 4.0.0-pre
      Terser: 5.31.1
```

### Steps to Reproduce

After loading the list, click "Previous" or "Next" button.

### Code Reproduction URL

https://github.com/will-hu-0/stenciljs-demo-issue

johnjenkins:
Thanks for the repro! 
Seems to be an issue with the async re-render. Will look into it asap.
In the meantime, this issue can be fixed by setting `taskQueue: 'immediate'` in `stencil.config.ts`

christian-bromann:
A fix for this has been released today as part of Stencil v4.25.1!

will-hu-0:
By using the latest version, it looks good! Thanks for the quick fix.

PR #6116, opened 2025-01-23
fix(runtime): SSR class handling breaks normal class handling

johnjenkins:

https://github.com/ionic-team/stencil/pull/6082 caused a regression. 

SSR logic, handling classes during the initial render / hydration of nodes would run all the time which would sometimes fail (https://github.com/ionic-team/stencil/issues/6115) and other times leave classes behind (https://github.com/ionic-team/stencil/issues/6114)

## What is the new behavior?

SSR logic relating to classes during hydration now only runs if a) SSR hydration is a feature of the build & b) the component is being rendered for the first time.

Fixes: https://github.com/ionic-team/stencil/issues/6115
Fixes: https://github.com/ionic-team/stencil/issues/6114

- [ ] Yes
- [x] No

## Testing

- New unit test
- 2 new wdio tests which cover both of the linked issues to stop regressions


christian-bromann (approved):
LGTM 👍

Issue #6120, opened 2025-01-24
bug: nuxt / vue hydration errors with SSR `shadow: true` components during `for` loops


johnjenkins:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [x] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [x] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

Latest

### Current Behavior

Vue places comment nodes (`<!-- [ -->` / `<!-- ] -->`) to mark the start and end position of it's for loops.

Stencil `shadow: true` components strip out comment nodes during client side hydration after SSR, breaking Vue's hydration.

### Expected Behavior

Like `scoped: true` components, Stencil should retain comment nodes (in the same order they were originally placed).

### System Info

```shell

```

### Steps to Reproduce

https://github.com/johnjenkins/nuxt-stencil-testing

```
pnpm i
cd packages/stencil-components
pnpm build
cd ../nuxt-test
pnpm dev
```

There are 2 components used, one `scoped: true` and one `shadow: true`, both utilising vue's for loops. 

The shadow component will throw a hydration error. 

### Code Reproduction URL

https://github.com/johnjenkins/nuxt-stencil-testing

PR #6123, opened 2025-01-27
fix(ssr): stop stripping comment nodes 

johnjenkins:

`client-hydrate.ts` is currently stripping out extra comment nodes of `shadow: true` components (this was originally to remove stencil's "internal" comments, which could pollute the component), but in doing so it removes slotted comment nodes which can be used by external tools / libs.

After testing, I see stencil's internal comments are removed during other stages of client-side hydration so this extra comment removal is unnecessary (the only time extra comments are seen is during unit tests due to a limitation / difference in how components are rendered on 'server')

GitHub Issue Number: https://github.com/ionic-team/stencil/issues/6120

## What is the new behavior?

Fixes https://github.com/ionic-team/stencil/issues/6120

- [ ] Yes
- [x] No

## Testing

- New unit test
- New wdio test

### DIFF
--- a/test/wdio/ssr-hydration/cmp.test.tsx
+++ b/test/wdio/ssr-hydration/cmp.test.tsx
@@ -43,8 +43,6 @@ describe('ssr-shadow-cmp', () => {
       document.body.appendChild(stage);
     }
 
-    // expect(typeof customElements.get('ssr-shadow-cmp')).toBe('undefined');
-
     // @ts-expect-error resolved through WDIO
     const { defineCustomElements } = await import('/dist/loader/index.js');
     defineCustomElements().catch(console.error);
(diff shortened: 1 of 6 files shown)


test/wdio/ssr-hydration/cmp.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+        },
+      );
+      const stage = document.createElement('div');
+      stage.setAttribute('id', 'stage');
+      stage.setHTMLUnsafe(html);
+      document.body.appendChild(stage);
+    }
+
+    // expect(typeof customElements.get('ssr-shadow-cmp')).toBe('undefined');
+
```
christian-bromann, line 47: Do we still need this?
```suggestion
```

christian-bromann (approved):
🙌

Issue #6125, opened 2025-01-27
bug: multiple of the same custom element - during ssr - will load the last element's property value for all


johnjenkins:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [x] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [x] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

Latest

### Current Behavior

During SSR, given a component:

```
@Component({
  tag: 'my-component',
  shadow: true,
})
export class MyApp {
  @Prop() aProp;

  render() {
    return <Host>My Prop: {this.aProp}</Host>;
  }
}
```

Used like:

```html
  <my-component></my-component>
  <my-component a-prop="second-component"></my-component>
```

The outputted HTML will be something like: 

```html
<my-component class="hydrated" s-id="3">
...
My Prop: second-component
</my-component>

<my-component a-prop="second-component" class="hydrated" s-id="4">
...
My Prop: second-component
</my-component>
```

^^ the last set `a-prop` will be rendered in all components.

After the client script takes-over, the correct state will then be set.

### Expected Behavior

Component should not 'share' their initial props in anyway.

### System Info

```shell

```

### Steps to Reproduce

https://stackblitz.com/edit/stencil-start-6at4nont?file=views%2Fmultiple-component.html,src%2Fcomponents%2Fmulti-component.tsx

Click on 'mulitple-component' < the client side script has intentionally *not* been included.

### Code Reproduction URL

https://stackblitz.com/edit/stencil-start-6at4nont?file=views%2Fmultiple-component.html,src%2Fcomponents%2Fmulti-component.tsx

### Additional Information

![Image](https://github.com/user-attachments/assets/5f029a93-92ba-4ff6-a455-cedef5bc2474)

christian-bromann:
This has been released in `@stencil/core@4.25.2`

PR #6126, opened 2025-01-27
fix(ssr): multiple component instances sharing initial properties

johnjenkins:

A regression added (https://github.com/ionic-team/stencil/pull/6084) during hydration, caused the value to be set on / got from the enclosure to the last component instance.

GitHub Issue Number: https://github.com/ionic-team/stencil/issues/6125

## What is the new behavior?

Fixes: https://github.com/ionic-team/stencil/issues/6125

The prototype 'get' now uses the current instance ref to get the initial value

- [ ] Yes
- [x] No

## Testing

- additional e2e test


christian-bromann (approved):
LGTM 👍

Issue #6127, opened 2025-01-28
bug: rendering > 10 of the same component during SSR significantly affects page load time


johnjenkins:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [x] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [x] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

`> 4.25.0`

### Current Behavior

Rendering more of the same Stencil component (which include `@Prop()`) during SSR, exponentially slows down the page load.

- Render 1 component ~20ms
- Render 5 components ~40ms
- Render 10 components ~150ms
- Render 20 components ~6s !

### Expected Behavior

Rendering the same component should not exponentially affect page load times

### System Info

```shell

```

### Steps to Reproduce

https://stackblitz.com/edit/stencil-start-6at4nont?file=views%2Fslow-ssr.html,src%2Fcomponents%2Fshadow.tsx

With dev tools open, network > Doc. 

Click on 'slow-ssr' and wait around 6-10s

### Code Reproduction URL

https://stackblitz.com/edit/stencil-start-6at4nont?file=views%2Fslow-ssr.html,src%2Fcomponents%2Fshadow.tsx

### Additional Information

![Image](https://github.com/user-attachments/assets/348e7da0-cd12-42e8-bb62-7298669ba116)

NoelLH:
Hey @johnjenkins, thanks for the work on this.

I know it's closed, but I'm just replying to see if you have any high levels ideas you could share about what has changed recently in this area of Stencil?

We're running [an app](https://github.com/thebiggive/donate-frontend) with Angular and [these Stencil components](https://github.com/thebiggive/components) and found that some requests caused major SSR CPU spikes and slowdown in Stencil 4.25.3, which should contain the fix for this issue. Going back to Stencil 4.23.2 seems to make things normal again.

Unfortunately our logs don't really shed enough light to implicate a specific page or component yet. It looks like a request is so slow that the infrastructure tooling gives up on the web container and kills it (after 29 seconds), before it can log anything helpful. We then only get specific logs about knock-on effects like other requests getting an HTTP 504 because all the web containers were being swapped out of service.

[...]

PR #6128, opened 2025-01-28
fix(ssr): exponential perf slow down

johnjenkins:

A regression added in `hydrate/proxy-host-element.ts` < the more of the same component rendered on the page, the more times the prototype would be augmented - causing an ever deepening chain of getters.

GitHub Issue Number: https://github.com/ionic-team/stencil/issues/6127

## What is the new behavior?

A simple property is added to the incoming prototype, making sure we only add augment it once.

Fixes https://github.com/ionic-team/stencil/issues/6127

- [ ] Yes
- [x] No

## Testing

- A new wdio perf test has been added

## Other information

image

### DIFF
--- a/src/hydrate/platform/proxy-host-element.ts
+++ b/src/hydrate/platform/proxy-host-element.ts
@@ -97,27 +97,30 @@ export function proxyHostElement(elm: d.HostElement, cstr: d.ComponentConstructo
           enumerable: true,
         });
 
-        // instance
-        Object.defineProperty((cstr as any).prototype, memberName, {
-          get: function (this: any) {
-            if (origGetter && attrPropVal === undefined && !getValue(this, memberName)) {
-              // if the initial value comes from an instance getter
-              // the element will never have the value set. So let's do that now.
-              setValue(this, memberName, origGetter.apply(this), cmpMeta);
-            }
+        if (!(cstr as any).prototype.done) {
+          // instance prototype.
+          // This only needs setting up once.
+          Object.defineProperty((cstr as any).prototype, memberName, {
+            get: function (this: any) {
+              if (origGetter && attrPropVal === undefined && !getValue(this, memberName)) {
+                // if the initial value comes from an instance getter
+                // the element will never have the value set. So let's do that now.
+                setValue(this, memberName, origGetter.apply(this), cmpMeta);
+              }
 
-            // if we have a parsed value from an attribute / or userland prop use that first.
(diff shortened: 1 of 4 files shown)


src/hydrate/platform/proxy-host-element.ts
```diff
@@ -131,6 +134,7 @@ export function proxyHostElement(elm: d.HostElement, cstr: d.ComponentConstructo
         });
       }
     });
+    (cstr as any).prototype.done = true;
```
christian-bromann, line 137: Should we make the property name more clear, e.g. `__stencilAugmented`? `done` sounds more like an official web interface.

christian-bromann (commented):
One suggestion

src/hydrate/platform/proxy-host-element.ts
johnjenkins: good suggestion - thanks.
I didn't like 'done' was just a placeholder. I think it could have caused name clashes too.

I was also trying to see if there's some better place to set this, but I think it is in-fact here... as it deals with setup / initial values etc.

src/hydrate/platform/proxy-host-element.ts
```diff
@@ -97,27 +97,32 @@ export function proxyHostElement(elm: d.HostElement, cstr: d.ComponentConstructo
...
-            return ref.$instanceValues$?.get(memberName) !== undefined
-              ? ref.$instanceValues$?.get(memberName)
-              : origGetter
-                ? origGetter.apply(this)
-                : getValue(this, memberName);
-          },
-          configurable: true,
-          enumerable: true,
-        });
+        if (!(cstr as any).prototype.__stencilAugmented) {
```
christian-bromann, line 100: Should we move the section:
```ts
    // instance prototype should only be processed once
    (cstr as any).prototype.__stencilAugmented = true;
```
into the if block?

johnjenkins: I don't think so because the if block is within a loop running over all the class' props.
Setting the flag within the loop will mean only the first class member will get augmented.

christian-bromann (approved):
LGTM 👍

src/hydrate/platform/proxy-host-element.ts
christian-bromann: Thanks for pointing that out.

christian-bromann:
This has been released in `@stencil/core@4.25.2`

Issue #6140, opened 2025-02-04
bug:  Dynamically rendered slot elements causes NotFoundError on insertBefore call


itsjustaplant:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [x] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [x] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

4.25.1

### Current Behavior

Dynamically rendered slot elements causes issue while patched insertBefore function tries to insert the child.

![Image](https://github.com/user-attachments/assets/a607b733-3f73-4281-9f33-8ada63a15e21)

This happens because currentChild element is encapsulated by another element therefore it's parent element is not same with the patched node.

### Expected Behavior

It should insert elements without an error.

### System Info

```shell
System: node 23.4.0
    Platform: windows (10.0.26100)
       Build: 1737767058
     Stencil: 4.25.1
  TypeScript: 5.5.4
      Rollup: 2.56.3
      Parse5: 7.1.2
      jQuery: 4.0.0-pre
      Terser: 5.31.1
```

### Steps to Reproduce

reproduction url: https://github.com/itsjustaplant/stencil-mono/tree/issue/insertBefore

### Code Reproduction URL

https://github.com/itsjustaplant/stencil-mono/tree/issue/insertBefore

christian-bromann:
A patch for this has been released in [`v4.27.0`](https://github.com/ionic-team/stencil/releases/tag/v4.27.0).

glafontaine:
The bug is still present when using Angular's control flow blocks to dynamically render elements, should we open a new issue ?

johnjenkins:
@glafontaine - yes please!🙏 
(A new minimum reproduction helps us to see the specific context for the error)

PR #6141, opened 2025-02-04
fix(runtime): append newChild if parent node doesn't match with patched node

itsjustaplant:

GitHub Issue Number: #6140 

## What is the new behavior?

If patched node (this) is not the same node of currentChild, it uses appendChild to insert the newChild to patched node

- [ ] Yes
- [X] No

### DIFF
--- a/src/runtime/dom-extras.ts
+++ b/src/runtime/dom-extras.ts
@@ -299,6 +299,21 @@ const patchInsertBefore = (HostElementPrototype: HTMLElement) => {
       });
       if (found) return newChild;
     }
+
+    /**
+     * Fixes an issue where slotted elements are dynamically relocated in React, such as after data fetch.
+     *
+     * When a slotted element is passed to another scoped component (e.g., <A><C slot="header"/></A>),
+     * the child’s __parentNode (original parent node property) does not match this.
+     *
+     * To prevent errors, this checks if the current child's parent node differs from this.
+     * If so, appendChild(newChild) is called to ensure the child is correctly inserted,
+     * allowing Stencil to properly manage the slot placement.
+     */
+    const parentNode = (currentChild as d.PatchedSlotNode)?.__parentNode;
+    if (parentNode && !this.isSameNode(parentNode)) {
+      return this.appendChild(newChild);
+    }
     return (this as d.RenderNode).__insertBefore(newChild, currentChild);
   };
 };


christian-bromann:
@itsjustaplant thanks for raising the PR, mind creating a test to ensure we don't regress on this?

itsjustaplant:
Hi @christian-bromann , I am unable to reproduce this issue in either a plain Stencil app or WebdriverIO tests. I believe that this only happens in React apps. PS: I replaced the unpatched __insertBefore__ with __appendChild__ as a fallback method as i believe that __appendChild__ is safer.

christian-bromann:
@itsjustaplant mhm 🤔 it would be good if we could validate this fix somehow, otherwise it may become difficult at some point to understand why we've put this in place or whether it still is relevant to Stencil.

itsjustaplant:
I have a reproduction repository [here](https://github.com/itsjustaplant/stencil-mono). If you run the **my-app**, you will encounter the error in couple of seconds due to mocked **fetch**. As it only happens in react, i don't what more could i add as reproduction but this commit somehow introduced the bug on stencil side (it was previously on react-dom).

```ts
```

yigityuce:
> @itsjustaplant mhm 🤔 it would be good if we could validate this fix somehow, otherwise it may become difficult at some point to understand why we've put this in place or whether it still is relevant to Stencil.

hi @christian-bromann hope you are well since from last touchpoint between us. Alperen (@itsjustaplant) is one of our team member and as he said unfortunately we are having this issue only in react, probably react does bulk update on DOM after stencil does update on the DOM, and, while react was trying to use insertBefore, the parentNode is **not** in the previous place where react stores in its vdom. (my strong guess is react does these kind of bulk operations in nextTick, and in the nextTick the actual DOM is already changed by Stencil)

So, unfortunately WDIO test is not able to be providable. However, I understand your concern and agree with you. Maybe we can put some more detail to the comment section in the code that describes the issue better. Would it work for you?

christian-bromann:
@yigityuce @itsjustaplant sounds good to me. Let's make sure we properly document this code section and include the reasoning for this, and in what cases this could cause an issue if the code wouldn't be there.

yigityuce:
> @yigityuce @itsjustaplant sounds good to me. Let's make sure we properly document this code section and include the reasoning for this, and in what cases this could cause an issue if the code wouldn't be there.

thanks @christian-bromann ! will do it today and update the PR.

itsjustaplant:
Hi @christian-bromann , i added a comment block to explain the issue and the fix. Thanks!

christian-bromann (approved):
👍  Thanks for adding the comment.

yigityuce:
@christian-bromann thanks for the quick action!

PR #6147, opened 2025-02-05
feat(ssr): shadow dom components can render as declarative-shadow-dom or as 'scoped'

johnjenkins:

When generating a `shadow: true` component during SSR, Stencil renders a Declarative-Shadow-Dom (DSD). 
Generally, this is preferable as it's natively supported in most browsers.

There are some scenarios at present, where DSD is not ideal:
- The requirement to support older browsers
- Rendering 10s to 100s of the same component*

(* e.g. given `my-list-item` component which is rendered 20 times. DSD will render 20 `<style>` tags which can negatively impact the initial weight of the page - especially without compression.)

GitHub Issue Number: N/A

## What is the new behavior?

The ability to use Stencil's `scoped` behaviour has been extended for use during SSR. 

During SSR a `shadow: true` scoped component will be rendered without a `template` - just plain, lightDOM html and a single `<style />` tag added to the `<head />`. 

On the client, these lightDOM nodes get added to a new shadow-root and the `<style />` tag is added as a `constructible stylesheet'.

To this end, the `serializeShadowRoot` options (passed into `renderToString()`) now accepts a number of new options:

- `'declarative-shadow-dom'` - will render all `shadow: true` components as DSD
- `'scoped'` - will render all `shadow: true` components with Stencil's scoped behaviour
- `{ 'declarative-shadow-dom': ['tag-1'], default: 'scoped'; }` will render component tags `tag-1` with DSD, but all others with Stencil's scoped behaviour
- `{ 'scoped': ['tag-1'], default: 'declarative-shadow-dom'; }` will render component tags `tag-1` with Stencil's scoped behaviour, but all others with DSD
- `false` - will disable all SSR render behaviour for `shadow: true` components
- `true` - *DEPRECATED* will render all `shadow: true` components as DSD

## Documentation

https://github.com/ionic-team/stencil-site/pull/1507

- [ ] Yes
- [x] No

## Testing

- New unit test
- New wdio tests


src/declarations/stencil-public-compiler.ts
```diff
@@ -946,18 +946,30 @@ export interface SerializeDocumentOptions extends HydrateDocumentOptions {
...
+   * - If set to `scoped` Stencil will render the contents of the shadow root as a `scoped: true` component
+   * and the shadowDOM will be created during client-side hydration.
+   * - Alternatively you can mix and match the two by providing an object with `dsd` and `scoped` keys, 
+   * the value arrays containing tag names of the components that should be rendered in that mode.
+   *
+   * *NOTE* `true | false` values have been deprecated in favor of `dsd` and `scoped`
+   * @default 'dsd'
+   */
+  serializeShadowRoot?:
+    | 'dsd'
```
christian-bromann, line 959: I feel like writing this out makes it more clear, most IDEs will autocomplete this so I don't think it adds a burden to the user
```suggestion
    | 'declarative-shadow-dom'
```

src/declarations/stencil-public-compiler.ts
```diff
@@ -946,18 +946,30 @@ export interface SerializeDocumentOptions extends HydrateDocumentOptions {
...
+    | 'scoped'
+    | {
+        dsd: string[];
+        scoped: string[];
+        default: 'dsd' | 'scoped';
+      }
```
christian-bromann, line 965: Let's add 2 examples to the JSDoc, one defining the value as e.g. `scoped` , the other showing an example of having 2 components as DSD and e.g. 1 component as scoped

src/declarations/stencil-public-compiler.ts
```diff
@@ -946,18 +946,30 @@ export interface SerializeDocumentOptions extends HydrateDocumentOptions {
...
+    | boolean;
   /**
    * The `fullDocument` flag determines the format of the rendered output. Set it to true to
    * generate a complete HTML document, or false to render only the component.
    * @default true
    */
-  fullDocument?: boolean;
+  fullDocument?: true;
```
christian-bromann, line 972: Is this change intentional?

src/declarations/stencil-public-compiler.ts
```diff
@@ -946,18 +946,30 @@ export interface SerializeDocumentOptions extends HydrateDocumentOptions {
...
   /**
-   * If set to `true` the component will be rendered within a Declarative Shadow DOM.
-   * If set to `false` Stencil will ignore the contents of the shadow root and render the
-   * element as given in provided template.
-   * @default true
-   */
-  serializeShadowRoot?: boolean;
+   * - If set to `dsd` the component will be rendered within a Declarative Shadow DOM.
+   * - If set to `scoped` Stencil will render the contents of the shadow root as a `scoped: true` component
+   * and the shadowDOM will be created during client-side hydration.
```
christian-bromann, line 951:
```suggestion
   *   and the shadow DOM will be created during client-side hydration.
```

src/declarations/stencil-public-compiler.ts
```diff
@@ -946,18 +946,30 @@ export interface SerializeDocumentOptions extends HydrateDocumentOptions {
    */
   removeHtmlComments?: boolean;
   /**
-   * If set to `true` the component will be rendered within a Declarative Shadow DOM.
-   * If set to `false` Stencil will ignore the contents of the shadow root and render the
-   * element as given in provided template.
-   * @default true
-   */
-  serializeShadowRoot?: boolean;
+   * - If set to `dsd` the component will be rendered within a Declarative Shadow DOM.
```
christian-bromann, line 949:
```suggestion
   * Configure how Stencil serializes the components shadow root.
   * - If set to `dsd` the component will be rendered within a Declarative Shadow DOM.
```

src/hydrate/runner/render.ts
```diff
@@ -48,7 +48,7 @@ export function renderToString(
   /**
    * Defines whether we render the shadow root as a declarative shadow root or as scoped shadow root.
    */
-  opts.serializeShadowRoot = typeof opts.serializeShadowRoot === 'boolean' ? opts.serializeShadowRoot : true;
+  opts.serializeShadowRoot = typeof opts.serializeShadowRoot === 'undefined' ? 'dsd' : opts.serializeShadowRoot;
```
christian-bromann, line 51: Shall we put the `serializeShadowRoot` options into an enum? Makes it easier to reuse them.

christian-bromann (commented):
Some initial thoughts

src/hydrate/runner/render.ts
johnjenkins: hm I'm not sure how / where to put that as that will change our public type definitions files as having some actual exported JS ... might be a bit risky?

src/declarations/stencil-public-compiler.ts
johnjenkins: nope 😅
Good catch!

src/hydrate/runner/render.ts
christian-bromann: good point, let's keep it simple for now.

src/runtime/bootstrap-lazy.ts
```diff
@@ -49,6 +50,13 @@ export const bootstrapLazy = (lazyBundles: d.LazyBundlesRuntimeData, options: d.
     // async queue. This will improve the first input delay
     plt.$flags$ |= PLATFORM_FLAGS.appLoaded;
   }
+  if (BUILD.hydrateClientSide && BUILD.shadowDom) {
+    const styles = doc.querySelectorAll(`[${HYDRATED_STYLE_ID}]`);
+    let i = 0;
+    for (; i < styles.length; i++) {
+      registerStyle(styles[i].getAttribute(HYDRATED_STYLE_ID), convertScopedToShadow(styles[i].innerHTML), true);
+    }
```
christian-bromann, line 58: This seems to be the same as in `src/runtime/bootstrap-custom-element.ts`, worth wrapping it into a utility function?

src/runtime/client-hydrate.ts
```diff
@@ -214,15 +214,32 @@ export const initializeClientHydrate = (
...
+  if (BUILD.shadowDom && shadowRoot && !shadowRoot.childNodes.length) {
+    // For `scoped` shadowDOM rendering (not DSD);
     // Add all the root nodes in the shadowDOM (a root node can have a whole nested DOM tree)
     let rnIdex = 0;
     const rnLen = shadowRootNodes.length;
-    for (rnIdex; rnIdex < rnLen; rnIdex++) {
-      shadowRoot.appendChild(shadowRootNodes[rnIdex] as any);
+    if (rnLen) {
+      for (rnIdex; rnIdex < rnLen; rnIdex++) {
+        shadowRoot.appendChild(shadowRootNodes[rnIdex] as any);
```
christian-bromann, line 224:
```suggestion
        shadowRoot.appendChild(shadowRootNodes[rnIdex]);
```

src/runtime/client-hydrate.ts
```diff
@@ -391,7 +408,7 @@ const clientHydrate = (
       });
 
       if (childNodeType === TEXT_NODE_ID) {
-        childVNode.$elm$ = node.nextSibling as any;
+        childVNode.$elm$ = findCorrespondingNode(node, NODE_TYPE.TextNode) as any;
```
christian-bromann, line 411:
```suggestion
        childVNode.$elm$ = findCorrespondingNode(node, NODE_TYPE.TextNode) as d.RenderNode;
```

src/runtime/client-hydrate.ts
```diff
@@ -415,7 +432,7 @@ const clientHydrate = (
           }
         }
       } else if (childNodeType === COMMENT_NODE_ID) {
-        childVNode.$elm$ = node.nextSibling as any;
+        childVNode.$elm$ = findCorrespondingNode(node, NODE_TYPE.CommentNode) as any;
```
christian-bromann, line 435:
```suggestion
        childVNode.$elm$ = findCorrespondingNode(node, NODE_TYPE.CommentNode) as d.RenderNode;
```

christian-bromann (commented):
Awesome work 👏 some minor nit picks, once pipeline passes I am happy to merge.

christian-bromann (commented):
I've been running some tests and discovered that the order of elements within the light DOM differs between DSD and scoped mode, e.g.:

```js
import { renderToString } from './test/wdio/hydrate/index.mjs';

const { html } = await renderToString(
  `
  <ssr-shadow-cmp>
    <p>Default slot content</p>
    <p slot="client-only">Client-only slot content</p>
  </ssr-shadow-cmp>
`,
  {
    fullDocument: false,
    serializeShadowRoot: true,
    constrainTimeouts: false,
    prettyHTML: true,
  },
);

console.log(1, html);
```

would return:
```html
<ssr-shadow-cmp class="hydrated" s-id="1">
  <template shadowrootmode="open">
    <style>
      :host {
        display: block;
        padding: 10px;
        border: 2px solid #000;
        background: yellow;
        color: red;
      }
    </style>
    <div c-id="1.0.0.0">
      <slot name="top" c-id="1.1.1.0"></slot>
      <slot c-id="1.2.1.1"></slot>
    </div>
  </template>
  <!--r.1-->
  <p>Default slot content</p>
  <p slot="client-only">Client-only slot content</p>
</ssr-shadow-cmp>
```

But when setting `serializeShadowRoot: 'scoped'`, I get:
```html
<ssr-shadow-cmp class="sc-ssr-shadow-cmp-h hydrated" s-id="1">
  <!--r.1-->
  <!--o.0.2.-->
  <!--o.0.4.-->
  <p slot="client-only" hidden="" c-id="0.4">Client-only slot content</p>
  <div class="sc-ssr-shadow-cmp sc-ssr-shadow-cmp-s" c-id="1.0.0.0">
    <!--s.1.1.1.0.top-->
    <!--s.1.2.1.1.-->
    <p c-id="0.2" s-sn="">Default slot content</p> 
  </div>
</ssr-shadow-cmp>
```

Note: the order of the slots changing.

christian-bromann (approved):
Thanks for clarifying, that makes total sense to me 👍 

Before we release this, can we raise a PR for the docs with an update?

Issue #3782, opened 2022-10-31
bug: slotchange event not firing with scoped mode


KevinCarnaille2:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/.github/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

` @stencil/core@2.19.1`

### Current Behavior

I'm trying to update a `@State` when the textContent used as default slot is updated.
For this, I've choosen to use the `slotchange` event.

I know this event is specific to shadow DOM, but as the compiler has polyfills for scoped mode and/or for older browsers, I thought it was working as well for `slotchange` , but it's not :( 

Not any event is fired, even when using polyfills on extras config :  

```
{
    cssVarsShim: true,
    dynamicImportShim: true,
    shadowDomShim: true,
    safari10: true,
    scriptDataOpts: true,
    slotChildNodesFix: true,
    appendChildSlotFix: true,
    cloneNodeFix: true
}

```

It is problematic because playing with MutationObserver can be a solution but is kind of tricky and adds a lot of complexity

### Expected Behavior

Be able to listen to `slotchange` events when using scoped mode. 

```
@Component({
  tag: "any-component",
  scoped: true,
})
```

### Steps to Reproduce

I've created 2 _webcomponents.dev_ snippets, with a kind of "dummy" example, but I think it's enough :) 

**Here you have the working example with a shadowDOM active :** 
https://studio.webcomponents.dev/edit/ShZNrJxFPQdBI0dnj8ac
You'll notice in the tool console the `slotchange` log.

**Here you have the working example without shadowDOM, with scoped mode active:**
https://studio.webcomponents.dev/edit/uZx5OPjzaesln7uFJtWi

Same component, same logic, but nothing is returned in the logs

### Code Reproduction URL

https://studio.webcomponents.dev/edit/uZx5OPjzaesln7uFJtWi

### Additional Information

Thanks team !

tanner-reits:
[...]

PR #6151, opened 2025-02-10
feat(runtime): add `slotchange` event and `assignedNodes` / `assignedElements` methods for `scoped: true` slots

johnjenkins:

Stencil's polyfilled - 'Pretend slots' - don't act like their native counterparts:

1) There's no way to get the nodes that are slotted (`assignedNodes()` / `assignedElements()`)
2) There's no way to know when the slotted nodes have changed (`slotchange` event)

At present, on a `scoped: true` component devs have to resort to clunky mutation observers (for text, comment and element nodes on a default slot), saving node references to internal state to track when they change.

This is frustrating as Stencil, internally, *knows* what is slotted. 

Making this knowledge available to users will enable them to remove substantial LOC and mirroring the the native APIs makes swapping between native & proprietary behaviour much simpler.

GitHub Issue Number: https://github.com/ionic-team/stencil/issues/3782

## What is the new behavior?

Fixes https://github.com/ionic-team/stencil/issues/3782

`assignedNodes` and `assignedElements` methods have been added to the mock `<slot />` nodes (which in reality is an an empty text node (or a comment node in debug mode) or a `<slot-fb />` if there is fallback nodes).

- [ ] Yes
- [x] No

No - although, there is a change in behaviour .. Previously when doing something like `<slot ref={s => this.slot} />` ... the `ref` would resolve to a slotted node. I think this behaviour is 1) unintuitive & 2) not documented.

## Testing

New WDIO tests


src/runtime/slot-polyfill-utils.ts
```diff
@@ -190,4 +193,89 @@ export const addSlotRelocateNode = (
...
+                // this 'slot' is also a `<slot-fb />` so we need to drill down recursively
+                getNestedSlotNames(nestedNode);
+              }
+            } else {
+              toReturn.push(nestedNode as d.RenderNode);
+            }
+          });
+        };
+        getNestedSlotNames(this);
+      }
```
johnjenkins, line 230: not sure if we need to support `{deep: true}`? 
Quite a niche / esoteric thing perhaps and not worth the LOC... WDYT?

johnjenkins: yep - for sure

christian-bromann (approved):
As always, fantastic job 🚀

christian-bromann:
> The components in my library use `shadow: false` **and** `scoped: false`.

Can you elaborate on the reasoning for this? We have been evaluating to deprecate this mode. From the Stencil POV it is hard to maintain this and offers almost no value. So I wonder what are the blockers for your library to adopt at least `scoped: true` mode.

gustavando:
> > The components in my library use `shadow: false` **and** `scoped: false`.
> 
> Can you elaborate on the reasoning for this? We have been evaluating to deprecate this mode. From the Stencil POV it is hard to maintain this and offers almost no value. So I wonder what are the blockers for your library to adopt at least `scoped: true` mode.

I'll explain the main reason behind this decision: We would likt to define a theme ( like "light" and "dark") to be applied to the page as a whole. We provide the components and the associated stylesheets, and the application using them has a mechanism to set the theme to be used.

```html
<html dir="ltr" lang="pt-br" data-alc-theme="light">
```

So, we want to, based on this, apply the same theme for all components used in the page.

To achieve this, our component stylesheets start out like thism, defining CSS custom properties:

```css
/* file: alc-alert.css */

[data-alc-theme="dark"] alc-alert {
  --bg-color-info: var(--alc-color-info-70);
  --border-color-info: var(--alc-color-info-60);
  --text-color-info: var(--alc-color-text-1);

  --bg-color-warning: var(--alc-color-warning-60);
  --border-color-warning: var(--alc-color-warning-50);
  --text-color-warning: var(--alc-color-text-1);

  /* and so on */
}

[data-alc-theme="light"] alc-alert {
  --bg-color-info: var(--alc-color-info-10);
  --border-color-info: var(--alc-color-info-40);
  --text-color-info: var(--alc-color-info-70);

  --bg-color-warning: var(--alc-color-warning-10);
  --border-color-warning: var(--alc-color-warning-20);
  --text-color-warning: var(--alc-color-warning-60);

  /* and so on */
}

```

And, then, we use the CSS custom properties

```css
/* file: alc-alert.css */

.alc-alert--info {
  background-color: var(--bg-color-info);
  border-color: var(--border-color-info);
  color: var(--text-color-info);
}

```


Well, we can only do this with `scoped: false`, because when` scoped` is true, the CSS rendered by StencilJS, in the browser, looks like this:

```css

/* Content of <style sty-id="sc-alc-alert"> in browser */

[data-alc-theme="dark"].sc-alc-alert alc-alert.sc-alc-alert {
  --bg-color-info:var(--alc-color-info-70);
  --border-color-info:var(--alc-color-info-60);
  --text-color-info:var(--alc-color-text-1);

  --bg-color-warning:var(--alc-color-warning-60);
  --border-color-warning:var(--alc-color-warning-50);
  --text-color-warning:var(--alc-color-text-1);
}

```

That is, it is as if the scope CSS classes were applied to the `[data-alc-theme]` selector, which is not actually in the scope of the component, as it is not part of it. This selector is used to get the theme choice from the html tag.

Therefore, we found no other solution for this scenario other than `scoped: false`.

gustavando:
Hey, @johnjenkins , I'm glad you commented. I'll complement the information I gave before.

> (...) an alternative solution would be to maintain a themes’ semantic colours in separate files (eg light.css —color-danger: var(—light-red); dark.css —color-danger: var(—dark-red);)

I've included these CSS snippets as examples, but most of the time we're not dealing with "semantic colors" only, but rather component-specific colors (and other visual aspects). These values are meaningless in a more general or global scope.



> Another alternative is, I think stencil’s scoped styles actually support host-context https://developer.mozilla.org/en-US/docs/Web/CSS/:host-context

We're not using shadow DOM (`shadow: false` and `scoped: false`). So host-context would not help us.

gustavando:
> Sorry, I meant host-context is supported in scoped: true, and that’s a possible solution for you

Now I understand what you meant, @johnjenkins !

That's interesting, and could be a solution.

However, unfortunately, `:host-context()` still has "Limited availability" (according to MDN), so it doesn't meet my non-functional requirements. Or maybe `:host-context()` polyfill for scoped has larger compatibility with browser than the shadow dom native's `:host-context()`? @christian-bromann

PR #6155, opened 2025-02-11
fix(ssr): include `<style />` for scope components when `fullDocument: false`

johnjenkins:

GitHub Issue Number: N/A

- [ ] Yes
- [ ] No


christian-bromann:
> Do you foresee any issue with that?

I don't think it makes a difference as the browser will parses these styles anyway before anything else.

PR #6156, opened 2025-02-12
refactor(runtime): make WeakMap obsolete  

### LINKED ISSUE #6148: bug: regression for disconnected components
### Prerequisites

### Stencil Version

4.25.0

### Current Behavior

In https://github.com/ionic-team/stencil/pull/6087 a regression was introduced when a component is temporarily removed from the DOM.

This PR appears to introduce a new bug on this line: https://github.com/ionic-team/stencil/pull/6087/files#diff-248f811e203ca495795884ef5d47f9e83f428ffa0ce37472bc1b552195706676R112. If a component is removed from the DOM, but still being used, then any call to set an attribute of the component while it is not in the DOM will crash because getHostRef will not find the component.

This only happens for the dist-custom-elements output target.

### Expected Behavior

The host ref should not be released when a component is removed from the DOM.

### System Info

```shell

```

### Steps to Reproduce

Use the linked repo. See the README.md

### Code Reproduction URL

https://github.com/k1w1/disconnected-set-bug

### Additional Information

_No response_

danielleroux:

GitHub Issue Number: Fixes #6148 

## What is the new behavior?

Remove WeakMap and hold each `HostRef` on the actual DOM Element. Which makes the `Weakmap` itself not necessary anymore.

Issue to discuss:

- ~Circular structure of HTMLElement (element -> $hostRef$ -> $hostElement$ -> $hostRef$ -> ...)~ Solved by using `getHostRef`
- ~Add `HostRef` (containing vdom structure) to regular DOM (Performance ?) => Checked the performance haven't found any issues yet.~ Solved by using `getHostRef`
- ~WebDriver.IO cannot convert element to JSON (Circular structure of HTMLElement) which results in failing tests atm.~ Solved by using `getHostRef`

- [ ] Yes
- [x] No

## Testing

Use existing tests, not added new ones.

### DIFF
--- a/src/client/client-host-ref.ts
+++ b/src/client/client-host-ref.ts
@@ -11,8 +11,8 @@ import type * as d from '../declarations';
  * @returns — true if the element was successfully removed, or false if it was not present.
  */
 export const deleteHostRef = (ref: d.RuntimeRef): boolean => {
-  if (ref.$hostRef$) {
-    delete ref.$hostRef$;
+  if (ref.__stencil__getHostRef) {
+    delete ref.__stencil__getHostRef;
     return true;
   }
 
@@ -26,7 +26,11 @@ export const deleteHostRef = (ref: d.RuntimeRef): boolean => {
  * @returns the Host reference (if found) or undefined
  */
 export const getHostRef = (ref: d.RuntimeRef): d.HostRef | undefined => {
-  return ref.$hostRef$;
+  if (ref.__stencil__getHostRef) {
+    return ref.__stencil__getHostRef();
+  }
+
+  return undefined;
 };
 
 /**
@@ -37,7 +41,7 @@ export const getHostRef = (ref: d.RuntimeRef): d.HostRef | undefined => {
  * @param hostRef that instances `HostRef` object
  */
 export const registerInstance = (lazyInstance: any, hostRef: d.HostRef) => {
-  lazyInstance.$hostRef$ = hostRef;
+  lazyInstance.__stencil__getHostRef = () => hostRef;
   hostRef.$lazyInstance$ = lazyInstance;
 
   if (BUILD.modernPropertyDecls && (BUILD.state || BUILD.prop)) {
@@ -73,7 +77,8 @@ export const registerHost = (hostElement: d.HostElement, cmpMeta: d.ComponentRun
     hostElement['s-rc'] = [];
   }
 
-  const ref = (hostElement.$hostRef$ = hostRef);
+  const ref = hostRef;
+  hostElement.__stencil__getHostRef = () => ref;
 
   if (!BUILD.lazyLoad && BUILD.modernPropertyDecls && (BUILD.state || BUILD.prop)) {
     reWireGetterSetter(hostElement, hostRef);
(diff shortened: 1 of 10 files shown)


christian-bromann:
> Circular structure of HTMLElement 

Could we put it behind a simple `getHostRef` function to avoid this?

> * Add `HostRef` (containing vdom structure) to regular DOM (Performance ?)

Not sure about this. Again, if we put it behind a function it may just end up being a referenced stored in memory which I don't think has a big impact.

> * WebDriver.IO cannot convert element to JSON (Circular structure of HTMLElement) which results in failing tests atm.

I can take a look at it 😉 

@johnjenkins thoughts?

danielleroux:
Use function call to resolve `HostRef` solves the circular structure 👍 Also the issue regarding webdriverio is fixed with the change.

src/client/client-host-ref.ts
```diff
@@ -3,41 +3,35 @@ import { reWireGetterSetter } from '@utils/es2022-rewire-class-members';
...
-
 /**
  * Given a {@link d.RuntimeRef} remove the corresponding {@link d.HostRef} from
  * the {@link hostRefs} WeakMap.
  *
  * @param ref the runtime ref of interest
  * @returns — true if the element was successfully removed, or false if it was not present.
  */
-export const deleteHostRef = (ref: d.RuntimeRef) => hostRefs.delete(ref);
+export const deleteHostRef = (ref: d.RuntimeRef): boolean => {
```
christian-bromann, line 13: Do we still need this function? I added it to clean up hostRef entries in the weakmap

src/runtime/vdom/vdom-annotations.ts
```diff
@@ -128,7 +128,7 @@ const parseVNodeAnnotations = (
      */
     const childNodes = [...Array.from(node.childNodes), ...Array.from(node.shadowRoot?.childNodes || [])];
     childNodes.forEach((childNode) => {
-      const hostRef = getHostRef(childNode);
+      const hostRef = getHostRef(childNode as any);
```
christian-bromann, line 131:
```suggestion
      const hostRef = getHostRef(childNode as d.RuntimeRef);
```

christian-bromann (commented):
Two minor nits

danielleroux:
> Ok, I just double check the code, can we apply this as well in the following places?
> 
> * src/hydrate/platform/index.ts
> * src/testing/platform/testing-constants.ts
> * src/testing/platform/testing-host-ref.ts
> * src/testing/platform/testing-platform.ts

Sure! Makes sense to remove the weakmaps there too.

src/testing/platform/testing-host-ref.ts
```diff
@@ -44,8 +33,8 @@ export const registerInstance = (
     hostRef = getHostRef(elm);
   }
 
+  lazyInstance.__stencil__getHostRef = () => hostRef;
   hostRef.$lazyInstance$ = lazyInstance;
-  return hostRefs.set(lazyInstance, hostRef);
```
danielleroux, line 48: Not sure if the return value as weakmap is somewhere used, i tried to check the code but couldnt find any references which depends on the return value.

christian-bromann: Yeah, I think it doesn't have any impact.

christian-bromann (approved):
Thank you so much, this is awesome 👍

christian-bromann:
Oh I will make sure this PR is mentioned.

Issue #6165, opened 2025-02-18
feat: always enable all render functions


werterhalimi:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [x] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [x] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already include this feature request, without success.

### Describe the Feature Request

As discussed with @johnjenkins on the Stencil Discord (https://discord.com/channels/520266681499779082/933029266562613309/1339564884375109744), it is useful to be able to override a component's lifecycle function even if it is not manually declared on some component in the project.
In [this article](https://medium.com/stencil-tricks/stenciljs-creating-custom-decorators-d4d8e78c5717), the author creates a custom decorator that calls a method when clicking outside of a component. This is useful, but it won't work if the componentDidLoad function is not declared somewhere in the project.
In [this Gist](https://github.com/jarrvis/stencil-click-outside/blob/master/src/index.ts), the author attempts to force the BUILD.connectedCallback attribute to be true to bypass this issue, but it doesn't work.

### Describe the Use Case

There are a lot of usecases. Here one: https://medium.com/stencil-tricks/stenciljs-creating-custom-decorators-d4d8e78c5717
I think that the fact that we cannot override those function is an unexpected behavior

### Describe Preferred Solution

@johnjenkins seems to believe that this micro-optimisation is not useful  so the simplest solution would be to remove it. 
It could also be a flag that disables this optimisation so there is no change for those who don't need it.

### Related Code

On this gist the author tries to bypass the problem by overriding the BUILD reference but it doesn't work https://github.com/jarrvis/stencil-click-outside/blob/master/src/index.ts

christian-bromann:
A patch for this has been released in [`v4.27.0`](https://github.com/ionic-team/stencil/releases/tag/v4.27.0).

Issue #2217, opened 2020-02-25
Lifecycle methods cannot all be declared as field holding arrow function


Nightmaster:
**Stencil version:**

```
from (at least)
@stencil/core@1.8.6
to
@stencil/core@1.9.0-18
```

**I'm submitting a:**
[x] bug report
[ ] feature request
[ ] support request

**Current behavior:**
Lifecycle methods declared  as fields holding arrow functions are not triggered if at least one lifecycle method is not declared as plain method

**Expected behavior:**
There should be no need to declare any lifecycle method as plain function for the app/component/PWA to work

**Steps to reproduce:**
Replace all the lifecycle methods from plain function declaration to field holding arrow function in any Stencil project

**Related code:**
https://github.com/Nightmaster/stencil-big-arrow-fail-demo
you can switch between the 2 branches to see the effects.

Issue #2195, opened 2020-02-12
Keep component lifecycle after build


MatheusBordin:
**Stencil version:**
```
 @stencil/core@1.8.8
```

**I'm submitting a:**
[ ] bug report
[x] feature request
[ ] support request

**Current behavior:**
The logic of removing lifecycle scripts during compilation prevents the creation of libraries that make use of these scripts. Example: If I create a TypeScript decorator to make some validations in component properties using some lifecycle script (not used in any other component), the build will remove the runtime script that processes this code.

**Expected behavior:**
Some configuration inside `stencil.config` providing an option to enable/disable this optimization.

**Steps to reproduce:**
Create any decorator that make use of some lifecycle method.

**Related code:**
```tsx
// decorator.ts
export function Required() {
    return function(target, key) {
        const originalWillUpdate = target.componentWillUpdate;
        target.componentWillUpdate = function() {
            originalWillUpdate.apply(this, arguments);
            // Make something (like properties validation).
        }
    }
}

// my-comp.tsx
...
@Prop() @Required() myProp: string;
...
```

**Other information:**
I suggest a benchmark to validate the gains of remove runtime lifecycle scripts during the compilation.

Currently, it was resolved by adding empty lifecycle methods in some component of the project, but it is a very bad solution besides resulting in compilation warnings.

jgroth:
We have the same use case with some custom decorators we have developed and have had the same issue since before version 1 of Stencil. Our workaround  has also been to have an empty component in the project that has all the lifecycle methods, but it would be nice to have it configurable from the config

PR #6167, opened 2025-02-19
fix(runtime): always call component lifecycle hooks

johnjenkins:

As per the linked issues, Stencil makes class components less easy to reason and not behave like the dynamic JS classes they are. Adding and extending class behaviour dynamically makes DX much better.

GitHub Issue Number: 

https://github.com/ionic-team/stencil/issues/6165

## What is the new behavior?

Removed the static flags / checks around lifecycle hook use - they are always called now.

Any argument around whether to always run lifecycle hooks hinges on any potential size savings won by Stencil's current optimisations. 

The following is the result from a single, simple component without any lifecycle hooks (remember if any single component uses a lifecycle hook, any difference is diminished).

Before `latest` stencil:

`dist-custom-elements` core:
+--------------------------------------------------------------+
| Size         | 12.71 KiB                                     |
|--------------------------------------------------------------|
| Gzipped      | 5.27 KiB                                      |
|--------------------------------------------------------------|

`dist` browser core:
+--------------------------------------------------------------+
| Size         | 14.97 KiB                                     |
|--------------------------------------------------------------|
| Gzipped      | 6.2 KiB                                       |
|--------------------------------------------------------------|

`dist` esm core (no-minification):
+--------------------------------------------------------------+
| Size         | 39.54 KiB                                     |
|--------------------------------------------------------------|
| Gzipped      | 9.99 KiB                                      |
|--------------------------------------------------------------|

After (this branch / consistent lifecycle hooks):

`dist-custom-elements` core:
+--------------------------------------------------------------+
| Size         | 12.92 KiB                                     |
|--------------------------------------------------------------|
| Gzipped      | 5.34 KiB                                      |
|--------------------------------------------------------------|

`dist` browser core:
+--------------------------------------------------------------+
| Size         | 15.38 KiB                                     |
|--------------------------------------------------------------|
| Gzipped      | 6.34 KiB                                      |
|--------------------------------------------------------------|

`dist` esm core (no minification):
+--------------------------------------------------------------+
| Size         | 40.68 KiB                                     |
|--------------------------------------------------------------|
| Gzipped      | 10.19 KiB                                     |
|--------------------------------------------------------------|

^ in these, 'worst case' scenarios, stripping away the ability to dynamically use lifecycle hooks saves 0.07-.2 kb

- Fixes https://github.com/ionic-team/stencil/issues/6165
- Fixes https://github.com/ionic-team/stencil/issues/2217
- Fixes https://github.com/ionic-team/stencil/issues/2195

- [ ] Yes
- [x] No

## Testing

New wdio tests


src/compiler/app-core/app-data.ts
```diff
@@ -24,23 +24,13 @@ export const getBuildFeatures = (cmps: ComponentCompilerMeta[]): BuildFeatures =
   const slotRelocation = cmps.some((c) => c.encapsulation !== 'shadow' && c.htmlTagNames.includes('slot'));
   const f: BuildFeatures = {
     allRenderFn: cmps.every((c) => c.hasRenderFn),
-    cmpDidLoad: cmps.some((c) => c.hasComponentDidLoadFn),
-    cmpShouldUpdate: cmps.some((c) => c.hasComponentShouldUpdateFn),
     cmpDidUnload: cmps.some((c) => c.hasComponentDidUnloadFn),
```
christian-bromann, line 27: Why do we keep this one?

johnjenkins: it's legacy and has been deprecated (but not removed) since v2.. I didn't know whether to nuke it / start pulling it out as part of this effort or just do it all in one swoop. Happy to just take it out of this bit

src/declarations/stencil-private.ts
```diff
@@ -145,16 +145,7 @@ export interface BuildFeatures {
...
-  cmpShouldUpdate: boolean;
-  cmpWillLoad: boolean;
-  cmpDidUpdate: boolean;
-  cmpWillUpdate: boolean;
-  cmpWillRender: boolean;
-  cmpDidRender: boolean;
   cmpDidUnload: boolean;
-  connectedCallback: boolean;
-  disconnectedCallback: boolean;
   asyncLoading: boolean;
```
christian-bromann, line 149: Given you remove it in the file above, should we get rid of it here too?

src/compiler/app-core/app-data.ts
christian-bromann: Let's remove it, I don't see it anywhere mentioned in the docs or somewhere else.

Same for `[hasLifecycle](https://github.com/ionic-team/stencil/pull/6167/commits/916bc33cdc1fcd22eb76ed2f13080eb3fcfd3e94)`?

johnjenkins: component `hasLifecycle` flags are used to to turn on the build flag `lifecycle` which in turn is used to turn on the flag `member` which is used quite a lot for runtime stuff

test/wdio/ts-target-props/components.d.ts
```diff
@@ -38,20 +27,9 @@ declare global {
...
         "decoratedGetterSetterProp"?: number;
-        /**
-         * @deprecated use camelCase instead. Support for dash-casing will be removed in Stencil v5.
-         */
-        "decorated-getter-setter-prop"?: number;
         "decoratedProp"?: number;
-        /**
-         * @deprecated use camelCase instead. Support for dash-casing will be removed in Stencil v5.
-         */
-        "decorated-prop"?: number;
```
christian-bromann, line 54: I've added these deprecation warnings in https://github.com/ionic-team/stencil/pull/6158 .. I would suggest to revert these changes as the next test run will re-introduce them again. Wdyt?

christian-bromann (approved):
👍 LGTM

Thanks!

PR #6175, opened 2025-02-25
fix(runtime): mock querySelectorAll

christian-bromann:

Importing the `dist-custom-elements` components in Node.js is failing due to the `doc.querySelectorAll` in `hydrateScopedToShadow`. This can cause issues when SSR components in Next.js.

## What is the new behavior?

Enhance the document mock with an additional method to make the function a noop and fix the failure.

- [ ] Yes
- [x] No

## Testing

Added an e2e test for this.

## Other information

cc @johnjenkins

### DIFF
--- a/src/client/client-window.ts
+++ b/src/client/client-window.ts
@@ -6,7 +6,7 @@ export const win = typeof window !== 'undefined' ? window : ({} as Window);
 
 export const doc = win.document || ({
   head: {},
-  querySelectorAll: () => [] as unknown as NodeListOf<Element>,
+  querySelectorAll: () => [] as unknown[],
 } as unknown as Document);
 
 export const H = ((win as any).HTMLElement || (class {} as any)) as HTMLElement;
--- a/src/client/client-window.ts
+++ b/src/client/client-window.ts
@@ -2,14 +2,11 @@ import { BUILD } from '@app-data';
 
 import type * as d from '../declarations';
 
-export const win = typeof window !== 'undefined' ? window : ({} as Window);
+interface StencilWindow extends Omit<Window, 'document'> {
+  document?: Document;
+}
 
-export const doc =
-  win.document ||
-  ({
-    head: {},
-    querySelectorAll: () => [] as unknown[],
-  } as unknown as Document);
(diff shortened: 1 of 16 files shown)


src/client/client-window.ts
```diff
@@ -4,7 +4,12 @@ import type * as d from '../declarations';
 
 export const win = typeof window !== 'undefined' ? window : ({} as Window);
 
-export const doc = win.document || ({ head: {} } as Document);
+export const doc =
+  win.document ||
+  ({
+    head: {},
+    querySelectorAll: () => [] as unknown[],
+  } as unknown as Document);
```
danielleroux, line 12: I discussed this also in the discord channel: https://discord.com/channels/520266681499779082/1343523701567787029

Any opinion?

danielleroux (commented):
@christian-bromann Totally fine to close my PR 👍

src/client/client-window.ts
christian-bromann: Uff, sorry for not checking there. It is hard to keep up with all the servers 🙈 Agree on @johnjenkins comment that we should keep the check where we use `document`. Will update my PR.

danielleroux: > Uff, sorry for not checking there. It is hard to keep up with all the servers 🙈 Agree on @johnjenkins comment that we should keep the check where we use `document`. Will update my PR.

No worries just wanted to be sure not to have parallel discussions regarding this topic.

src/runtime/vdom/vdom-render.ts
```diff
@@ -87,16 +87,21 @@ const createElm = (oldParentVNode: d.VNode, newParentVNode: d.VNode, childIndex:
     if (BUILD.svg && !isSvgMode) {
       isSvgMode = newVNode.$tag$ === 'svg';
     }
+
+    if (!win.document) {
+      throw new Error('You are trying to render a component in a non-browser environment. This is not supported.');
```
johnjenkins, line 92: total nit - but, IDK if this is a completely accurate statement (so slightly misleading) because we do render components in a non-browser environment via MockDoc?

christian-bromann: Good call, fixed.

Issue #6150, opened 2025-02-10
bug: property casing in runtime vs SSR


christian-bromann:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [x] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [x] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

latest

### Current Behavior

When defining component properties in Stencil, there is currently a discrepancy between how Stencil reads property names during runtime and hydration.

Given the following component:
```ts
// my-component.tsx
import { Component, Prop, h } from '@stencil/core';
import { format } from '../../utils/utils';

@Component({
  tag: 'my-component',
  styleUrl: 'my-component.css',
  shadow: true,
})
export class MyComponent {
  /**
   * The first name
   */
  @Prop() firstName: string;

  /**
   * The middle name
   */
  @Prop() middleName: string;

  /**
   * The last name
   */
  @Prop() lastName: string;

  private getText(): string {
    return format(this.firstName, this.middleName, this.lastName);
  }

  render() {
    return <div>Hello, World! I'm {this.getText()}</div>;
  }
}
```
and
```ts
// stencil-component.tsx
import { Component, h } from '@stencil/core';

@Component({
  tag: 'stencil-component',
  shadow: true,
})
export class StencilComponent {
  render() {
    return <div>
      <code>
        &#x3C;my-component firstName=&#x22;Stencil&#x22; middleName=&#x22;&#x27;Don&#x27;t call me a framework&#x27;&#x22; lastName=&#x22;JS&#x22;&#x3E;&#x3C;/MyComponent&#x3E;
      </code>
      <my-component firstName="Stencil" middleName="'Don't call me a framework'" lastName="JS"></my-component>
      <hr />

      <code>
        &#x3C;my-component first-name=&#x22;Stencil&#x22; middle-name=&#x22;&#x27;Don&#x27;t call me a framework&#x27;&#x22; last-name=&#x22;JS&#x22;&#x3E;&#x3C;/my-component&#x3E;
      </code>
      <my-component first-name="Stencil" middle-name="'Don't call me a framework'" last-name="JS"></my-component>
    </div>
  }
}
```

Now if you render the following HTML template:

```html
<!DOCTYPE html>
[...]

PR #6179, opened 2025-02-28
feat(compiler): allow dash-case type generation to be disabled

### LINKED ISSUE #6178: feat: allow dash-case type generation to be optional
### Prerequisites

- [x] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already include this feature request, without success.

The update made in #6158 to support dash casing results in properties being added twice in the `components.d.ts`, in both camelCase and dash-case (with an added `@deprecated` comment for dash-case properties). While this solves the initial issue highlighted in #6150, it is not necessarily needed for all projects. 

I'm going to open a PR to add a config option (`noDashCaseTypes?`) that allows for the generation of dash-cased properties to be skipped.

### Describe the Use Case

Users who are not using dash-casing in JSX templates do not need dash-cased types to be added to `components.d.ts` and should be able to skip this if necessary. 

In our particular project, we have a script that uses `components.d.ts` to generate types files for React components - we could update our script to ignore duplicated types in dash-case (also not supported by React), but having the option to switch off this feature for users who were not facing the initial issue (#6150) would be useful. 

I'm planning on opening a PR that will:
- Add a `noDashCaseTypes?` option to the `StencilConfig` interface
- Use this to determine whether to skip the generation of dash-case types in `components.d.ts`
- Update tests to account for this option

### Describe Alternatives

_No response_

### Related Code

_No response_

### Additional Information

_No response_

hannahmclarke:

#6158 added support for dash-casing by generating dash-cased properties from attribute names and adding them to `components.d.ts` (marking them as deprecated). 
This is not always needed (or desired) in projects that were not affected by #6150
 
fixes #6178 

GitHub Issue Number: #6178 

## What is the new behavior?
`noDashCaseTypes` has been added as an optional config property. If set to `true`, dash-cased attribute names/types will not be added to the `components.d.ts` on build. By default this property is set to `false`.

- [ ] Yes
- [x] No

## Testing

A test case has been added to `generate-app-types.spec.ts`
This has also been tested against a Stencil project.

I haven't added a new e2e test, but can do so if needed.


christian-bromann:
@hannahmclarke thanks for the contribution. Mind taking a look at the failing pipeline?

hannahmclarke:
> @hannahmclarke thanks for the contribution. Mind taking a look at the failing pipeline?

@christian-bromann Drawing a blank here - no obvious reasons for failure, just seems to have been canceled by the server...?

christian-bromann:
@hannahmclarke this is indeed a weird behavior. Given we received overall feedback that this change was not well received, I've decided to get rid of it after all in #6184. Hence I will go ahead and close this one. Thanks for your contribution!

PR #6188, opened 2025-03-07
fix: add missing name attribute for details html element

s-gurevich:

I wanted to use the `details` html element with the `name` property in Stencil's TSX and got the error:
```
Type '{ name: string; }' is not assignable to type 'DetailsHTMLAttributes<HTMLElement>'.Property 'name' does
           not exist on type 'DetailsHTMLAttributes<HTMLElement>'.
```

The `name` property seems to be missing here: https://github.com/stenciljs/core/blob/30fc9f5985120e636eec8acbe2253145c97072d9/src/declarations/stencil-public-runtime.ts#L922

The `name` attribute is part of the web standard and allows grouping multiple `details` elements: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details#multiple_named_disclosure_boxes

GitHub Issue Number: N/A

## What is the new behavior?

The name attribute is added to the definitions and can be used in TSX.

- [ ] Yes
- [X] No

### DIFF
--- a/src/declarations/stencil-public-runtime.ts
+++ b/src/declarations/stencil-public-runtime.ts
@@ -921,6 +921,7 @@ export namespace JSXBase {
 
   export interface DetailsHTMLAttributes<T> extends HTMLAttributes<T> {
     open?: boolean;
+    name?: string;
     onToggle?: (event: Event) => void;
   }
 


christian-bromann (approved):
LGTM 👍 Thanks a lot!

Issue #6190, opened 2025-03-11
bug: File watcher sometimes does not trigger rebuild


hettiger:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/stenciljs/core/blob/main/CONTRIBUTING.md).
- [x] I agree to follow the [Code of Conduct](https://github.com/stenciljs/core/blob/main/CODE_OF_CONDUCT.md).
- [x] I have searched for [existing issues](https://github.com/stenciljs/core/issues) that already report this problem, without success.

### Stencil Version

4.27.2

### Current Behavior

The file watcher sometimes fails to trigger a rebuild:

![Image](https://github.com/user-attachments/assets/eb968941-c276-4094-bd4e-9cc122464c54)

This is caused by the "Back up files before saving" option in my editor.
When this option is disabled the file watcher works just fine.
However, the "Back up files before saving" option is enabled per default and Stencil should be able to handle this.

> The above demo was recorded on a clean install of Stencil. It should be fairly simple to reproduce the issue as long you're using "Back up files before saving". (I don't know if it's just PHPStorm or other editors as well.)

### Expected Behavior

I expect the file watcher to pick up changes reliably and trigger rebuilds accordingly with a debounce behavior.

### System Info

```shell
macOS 15.3.1

PhpStorm 2024.3.4


npx stencil info

      System: node 20.14.0
    Platform: darwin (24.3.0)
   CPU Model: Apple M4 Max (16 cpus)
    Compiler: /Users/hettiger/Developer/experiments/stencil-watcher-reproduction/node_modules/@stencil/core/compiler/stencil.js
       Build: 1741029559
     Stencil: 4.27.2 🌱
  TypeScript: 5.5.4
      Rollup: 2.56.3
      Parse5: 7.1.2
      jQuery: 4.0.0-pre
      Terser: 5.31.1
```

### Steps to Reproduce

1. Create a new Stencil components project
2. Run `stencil build --watch --verbose`
3. Make sure "Back up files before saving" option or similar option in your editor is enabled
4. Repeatedly make changes to e.g. `src/components/my-component/my-component.css` and hit save

### Code Reproduction URL

Just install a clean Stencil project and follow steps to reproduce

PR #6191, opened 2025-03-11
fix(compiler): fix file watcher sometimes doesn't trigger rebuild

hettiger:

GitHub Issue Number: #6190

The file watcher sometimes fails to trigger a rebuild:

This is caused by the "Back up files before saving" option in my editor.
When this option is disabled the file watcher works just fine.
However, the "Back up files before saving" option is enabled per default and Stencil should be able to handle this.

> The above demo was recorded on a clean install of Stencil. It should be fairly simple to reproduce the issue as long you're using "Back up files before saving". (I don't know if it's just PHPStorm or other editors as well.)

## What is the new behavior?

After applying these changes the watcher behaves beautifully with or without "Back up files before saving" option enabled.

## Documentation

–

- [ ] Yes
- [x] No

## Testing

I tested the code with a custom `@stencil/core` build in my application running `stencil build --watch --verbose`. I also made sure that the issue is not specific to my project. I installed a clean Stencil project and repeated my tests with and without the "Back up files before saving" option.

I ran the test suite via `npm run test` and made sure tests are still passing. I did not add any new tests.

## Other information

Ignoring backup files via [`watchIgnoredRegex: [/\.\w+~$/]`](https://stenciljs.com/docs/config#watchignoredregex) or using `"**/*.*~"` on the `watchOptions.excludeFiles` `tsconfig.json` option did not work.

> These options generally don't work for me. (Or I'm misinterpreting the `NODE_SYS_DEBUG` and `WATCH_BUILD` logging.)

The existing implementation was running into the `lastTsBuilder && !timeoutId` condition with a `timeoutId` being present even though there was no running build or any other good reason not to rebuild.

The change simplifies the implementation by using a recursive `setTimeout` to wait for pending builds.

This allows to:

  - remove the extra `rebuildTimer` which was arguably confusing
  - remove the clearTimeout override

Since we immediately call `clearTimeout(timeoutId)` in the `setTimeout` override, we don't stack up rebuild requests.

### DIFF
--- a/src/compiler/transpile/create-watch-program.ts
+++ b/src/compiler/transpile/create-watch-program.ts
@@ -24,7 +24,6 @@ export const createTsWatchProgram = async (
   let isRunning = false;
   let lastTsBuilder: any;
   let timeoutId: any;
-  let rebuildTimer: any;
 
   // Get the pre-baked TS options we want to use for our builder program
   const optionsToExtend = getTsOptionsToExtend(config);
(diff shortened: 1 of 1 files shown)


hettiger:
@christian-bromann Sorry about the formatting error — overlooked the prettier scripts. I fixed it in f977746e609aacd74e8adc94b37adf522d6002b6 using `npm run prettier` and checked with `npm run prettier.dry-run`. CI should be passing now.

christian-bromann:
> Sorry about the formatting error — overlooked the prettier scripts.

No worries at all, happens to me all the time.

christian-bromann (approved):
LGTM 👍

PR #6200, opened 2025-03-12
fix(loader): remove package.json as it interferes with main

johnjenkins:
The auto-generated `package.json` that Stencil inserts within the `loader` dir seems unnecessary now. 

The auto-generated `package.json`: 
- overrules settings within the main `package.json` [leading to unexpected behaviour](https://github.com/stenciljs/core/issues/5637) 
- removes developer wishes and agency - "*I* should control how & what is exported"

GitHub Issue Number: https://github.com/stenciljs/core/issues/5637 < the fix was merely a stop gap and is not a 'modern' solution (node now supports es modules natively) 

## What is the new behavior?

Stop stencil inserting the loader `package.json`

- [ ] Yes
- [x] No


christian-bromann (approved):
LGTM 👍

PR #6203, opened 2025-03-13
build: improve `nodeResolve` config

johnjenkins:

Stencil's default `nodeResolve` config doesn't pay attention to the `exports` > `require` field (which is present in the axios lib - https://github.com/axios/axios/issues/5888)

GitHub Issue Number: https://github.com/stenciljs/core/issues/6202

## What is the new behavior?

Closes https://github.com/stenciljs/core/issues/6202

- Include `require` in nodeResolve `exportConditions` (Note - I'm not 100% sure if we should as I *think* axios is in the wrong here - https://github.com/axios/axios/issues/5888#issuecomment-2721184793)
- update `nodeResolve` types to match the plugin
- allow users' `nodeResolve` config to take precedence over default - giving users more control

- [ ] Yes
- [x] No


christian-bromann (approved):
🎉

PR #6211, opened 2025-03-24
feat: tag transform 

johnjenkins:

GitHub Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/src/compiler/output-targets/dist-custom-elements/index.ts
+++ b/src/compiler/output-targets/dist-custom-elements/index.ts
@@ -267,6 +267,7 @@ export const generateEntryPoint = (
 
   // Content related to the `bundle` export behavior
   if (outputTarget.customElementsExportBehavior === 'bundle') {
+    imports.push(`import { transformTag } from '${STENCIL_INTERNAL_CLIENT_ID}';`);
     imports.push(...cmpImports);
     body.push(
       'export const defineCustomElements = (opts) => {',
@@ -274,8 +275,8 @@ export const generateEntryPoint = (
       '        [',
       ...cmpNames.map((cmp) => `            ${cmp},`),
       '        ].forEach(cmp => {',
-      '            if (!customElements.get(cmp.is)) {',
-      '                customElements.define(cmp.is, cmp, opts);',
+      '            if (!customElements.get(transformTag(cmp.is))) {',
+      '                customElements.define(transformTag(cmp.is), cmp, opts);',
       '            }',
       '        });',
       '    }',
--- a/src/compiler/transformers/component-native/add-define-custom-element-function.ts
+++ b/src/compiler/transformers/component-native/add-define-custom-element-function.ts
@@ -2,6 +2,7 @@ import { dashToPascalCase } from '@utils';
 import ts from 'typescript';
 
 import type * as d from '../../../declarations';
+import { addCoreRuntimeApi } from '../core-runtime-apis';
 import { createImportStatement, getModuleFromSourceFile } from '../transform-utils';
 
 /**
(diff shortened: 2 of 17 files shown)


elliottdaviesss:
Hi @johnjenkins, would these changes extend to the framework adapters OOTB/'as is', or would that require additional changes in the output-targets repo?

We're running a React MFE architecture and would really benefit from being able to transform the tags at runtime to avoid version registration clashes between feature teams running different versions of the stencil component lib - just wondering whether these changes could facilitate that in future. 

Thanks!

Cliffback:
> hey @elliottdaviesss - Thanks for the interest!
> 
> Yes, it would involve some tweaks in the output adapters (wrapping `tagName` references in the newly exported `transformTag(tagName)`)
> 
> On reflection, idk if what i've done here is the best approach; @Cliffback has came up with a slightly different strategy which you can use at https://github.com/Cliffback/stencil-custom-suffix-output-target (doesn't work with react output atm I think). I like the fact it's an output target, plus it does some extra bits like transforming css selectors.

@johnjenkins I could give transforming the other output targets the same way, so that it would work with react as well? 

My main focus has been on angular so far, as that is what we need at work, but would be nice to expand it now that it is on npm regardless

Cliffback:
> @Cliffback - I decided to roll your work into this, tysm ❤️

That's amazing to hear! Really happy that it has been of use! 🤩

Cliffback:
@johnjenkins btw. do you do anything with the `component.d.ts` declaration file? To get emitters to get the correct type I had to add a patch for that file: [custom-suffix-output-target.ts#L320](https://github.com/Cliffback/stencil-custom-suffix-output-target/blob/98285c446767d025d4ad0dd639b93bb75303258c/src/custom-suffix-output-target.ts#L320) 

Without this patch, all emitters lost their type and would have to be set to `any` in the consuming application.

it turns:
```ts
    interface HTMLElementTagNameMap {
        "my-component-one": HTMLMyComponentOneElement;
        "my-component-two": HTMLMyComponentTwoElement;
    }
```
into:
```ts
    interface HTMLElementTagNameMap {
        "my-component-one": HTMLMyComponentOneElement;
        [key: \`my-component-one--\${string}\`]: HTMLMyComponentOneElement;
        "my-component-two": HTMLMyComponentTwoElement;
        [key: \`my-component-two--\${string}\`]: HTMLMyComponentTwoElement;
    }
```

This made it necessary to differentiate the postfix from the tagname with `--`, since componenets with parts of the same tagname all would be funneled into the shortest tagname, like for example `my-radio-button-group` would just be caught by `my-radio-button`

All in all, I think this differentation also is a good thing, as it separates more clearly the acutal tagname from the transformed one.

Any thoughts on this?

Cliffback:
> I think it's neat @Cliffback ... idk how it would work with what I have done though seeing as I make no assumptions at either
> 
> 1. build time ... whether a transformer is being set or used and...
> 2. the transformed tag's format - people are free to transform tags as they see fit.
> 
> I can't see a way to get the run-time's tagTransformer result into a useable ts template literal type.
> 
> In reality, I think this would mainly affect 'vanilla' consuming applications(?) as JS-framework consumers when using wrappers continue to be typed correctly no-matter the transformer used.
> 
> Perhaps there would still be a need for a custom output target for those that need it :)

Yes, that makes a lot of sense, especially given that my output targets main purpose has been getting it to work with angular (and that's were the emitter type issue arose). 

However the element map patch is non-destructive, it just adds support for tagnames having a postfix as well, as long as it is distinguished with a `--`, making sure both original tagnames and tagnames with `--my-postfix-example` will be correctly typed. The same could be done with a prefix as well, I guess.

However for this to work with your solution, enforcement of `--` between the original tagname and the added pre/postfix would be needed, which I guess might fall outside the intenteded scope of the PR.

Regardless, when this is released, I'll see how much of our setup could be migrated, and a way to solve the rest could be to change `stencil-custom-suffix-output-target` from doing everything to just do the rest needed to get the angular (and hopefully the react) proxy files working.

Eagerly looking forward to the PR getting merged! 🤩

Issue #6182, opened 2025-03-02
feat: ARIA attribute JSX types are missing


trusktr:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [x] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [x] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already include this feature request, without success.

### Describe the Feature Request

JSX types for attributes like `aria-foo` are needed.

### Describe the Use Case

To be able to have type definitions for all available attributes.

### Describe Preferred Solution

Add to `DOMAttributes`

### Describe Alternatives

n/a

### Related Code

```js
return <div aria-labelledby="foo">...</div>
```

christian-bromann:
Thanks for raising this issue. Any contributions are much appreciated!

PR #6221, opened 2025-04-02
feat(jsx): add support for aria-* attributes in JSX typings

AlirezaEbrahimkhani:
Add typings for ARIA attributes like `aria-labelledby`, `aria-hidden`, etc., to the DOMAttributes interface. This allows better accessibility support and developer experience via type safety and IntelliSense.

fixes: #6182

GitHub Issue Number: #6182 
Currently, JSX attributes with ARIA prefixes such as aria-label, aria-hidden, etc., are not typed, leading to missing IntelliSense and potential type errors in TypeScript.

## What is the new behavior?

JSX ARIA attributes are now explicitly typed in the DOMAttributes interface, enabling proper IntelliSense, autocomplete, and validation in IDEs and TypeScript-aware tooling.
This includes support for all standard ARIA attributes, and optionally, a generic index signature for custom aria-* values.

- [ ] Yes
- [x] No

### DIFF
--- a/src/declarations/stencil-public-runtime.ts
+++ b/src/declarations/stencil-public-runtime.ts
@@ -1853,6 +1853,10 @@ export namespace JSXBase {
     onTransitionRunCapture?: (event: TransitionEvent) => void;
     onTransitionStart?: (event: TransitionEvent) => void;
     onTransitionStartCapture?: (event: TransitionEvent) => void;
+
+    // WAI-ARIA Attributes
+    [key: `aria-${string}`]: string | boolean | undefined;
+    [key: `aria${string}`]: string | boolean | undefined;
   }
 }
 


christian-bromann (approved):
LGTM 👍

src/declarations/stencil-public-runtime.ts
```diff
@@ -1853,6 +1853,9 @@ export namespace JSXBase {
     onTransitionRunCapture?: (event: TransitionEvent) => void;
     onTransitionStart?: (event: TransitionEvent) => void;
     onTransitionStartCapture?: (event: TransitionEvent) => void;
+
+    // WAI-ARIA Attributes
+    [key: `aria-${string}`]: string | boolean | undefined;
```
christian-bromann, line 1858: As these types are mostly relevant for a TSX environment, we should also provide a camel case option.
```suggestion
    [key: `aria-${string}`]: string | boolean | undefined;
    [key: `aria${string}`]: string | boolean | undefined;
```

christian-bromann (changes requested):

christian-bromann (approved):
👍

duhem-s:
Is there a way to ignore this ?

One of our components have a `ariaDescribedbyIDs` props, typed as `string[]` but since this change we cannot build because it expect to have `string | boolean` type on `aria*` attributes.

Maybe it would be better to list the existing `aria-*` attributes ? https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes

danyball:
@AlirezaEbrahimkhani Thanks for this fix. Do you see a possibility to type some aria attributes a bit stronger? Because TS `ARIAMixin` interface has some attributes which are not "string | boolean | null": E.g. `ariaActiveDescendantElement: Element | null;`. 
Just asking because we use this type outside of JSX: `props: JSXBase.AnchorHTMLAttributes<HTMLAnchorElement>` and when using `props` object its valid to use `props.ariaNotExisting`.

Issue #6219, opened 2025-04-01
bug: serializeShadowRoot: 'scoped' does not apply scoped classes to host element children


Armand-Lluka:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/stenciljs/core/blob/main/CONTRIBUTING.md).
- [x] I agree to follow the [Code of Conduct](https://github.com/stenciljs/core/blob/main/CODE_OF_CONDUCT.md).
- [x] I have searched for [existing issues](https://github.com/stenciljs/core/issues) that already report this problem, without success.

### Stencil Version

4.27.1

### Current Behavior

When using `renderToString` to hydrate `scoped` web components, the "scoped" classes are not being applied correctly to the children of the host element. 

<img width="1908" alt="Image" src="https://github.com/user-attachments/assets/4b473959-ddd9-481d-b6f5-e4fd951a4b07" />

### Expected Behavior

When using `renderToString` with `serializeShadowRoot: 'scoped'`, the children of the host element should properly have the `sc-web-component-name` class applied and all the defined styles that come with it. 

### System Info

```shell
System: node 20.18.3
Platform: darwin (24.3.0)
CPU Model: Apple M1 Pro (8 cpus)
Compiler: /Users/armandlluka/Documents/work/stencil-bug-reproduction/node_modules/@stencil/core/compiler/stencil.js
Build: 1741890846
Stencil: 4.28.2 🎤
TypeScript: 5.5.4
Rollup: 4.34.9
Parse5: 7.2.1
jQuery: 4.0.0-pre
Terser: 5.37.0
```

### Steps to Reproduce

Go to: https://github.com/Armand-Lluka/stencil-bug-reproductions

1. `npm i` 
2. `npm run build`
3. `npm run serve` to spin up server
4. Visit `http://localhost:8090/`
5. Disable JS on the client and notice that classes and styles are not being properly applied. 
6. If you add the class manually, `sc-my-component`, to the child of the host element, then the styles are again properly applied. 

### Code Reproduction URL

https://github.com/Armand-Lluka/stencil-bug-reproductions

Armand-Lluka:
This was not fixed with #6224 I've updated Stencil to the latest version in the example to demonstrate this.

christian-bromann:
I've raised a PR with a fix at [#6237](https://github.com/stenciljs/core/pull/6237)

christian-bromann:
Merged and released. Closing!

PR #6237, opened 2025-04-11
fix(runtime): properly set parent classes in children scope

christian-bromann:
When using `renderToString` to hydrate `scoped` web components, the "scoped" classes are not being applied correctly to the children of the host element. 

Image

GitHub Issue Number: #6219

## What is the new behavior?
When using renderToString with serializeShadowRoot: 'scoped', the children of the host element should properly have the sc-web-component-name class applied and all the defined styles that come with it.

- [ ] Yes
- [x] No

## Testing

Updated tests that were incorrectly corrected before.

## Other information

n/a


Armand-Lluka:
Thanks for the work on this and the performance issue. This seems to resolve the in the example I created, however I would like to test this out in a real-world application. Can you create a dev build for me if possible?

christian-bromann:
Closing in favor of https://github.com/stenciljs/core/pull/6238

PR #6249, opened 2025-05-02
fix(docs): add overwriteExisting option for flexible README overwrite behaviour

smorrisods:
The `docs-readme` output target has long supported writing README files to custom locations using the `dir` property, helping teams structure their documentation workflows to suit project needs.

As part of evolving those workflows, I've introduced the `overwriteExisting` property to give developers precise control over how README files are handled. This addition supports common scenarios like always overwriting, only writing when files are missing, or preserving existing content:

- Always overwrite the full README (`true`),
- Only write the full README if no file exists (`'if-missing'`), or
- Preserve existing custom content (_default_ `false`).

These options aim to better support workflows where documentation is closely tied to source code, while also offering flexibility for varied project structures.

## What is the current behaviour?

By default, the docs-readme output target preserves any manually-entered content in destination README files, but it does not offer flexible controls for different overwrite behaviours. This made it difficult to manage workflows where documentation is expected to be tightly coupled with source code, while also needing different customisation strategies for various parts of a project.

Link to relevant issues/PRs: stenciljs/core#5648 stenciljs/core#6248

## What is the new behaviour?

This PR introduces documentation for the new `overwriteExisting` property
on the `docs-readme` output target.

The documentation includes:

- Clear description of the `overwriteExisting` configuration option.
- Available modes (`true`, `false`, and `'if-missing'`).
- Example usage in a `stencil.config.ts` file.
- Real-world use case highlighting idempotent output and CI workflows.
- Clarification that the component README beside the component is treated
  as the canonical source of truth for custom content.

## Documentation

A documentation update has been made in stenciljs/site#1521.

- [ ] Yes
- [X] No

## Testing

- Verified clarity and completeness of the documentation updates.
- Added new test cases to cover the different permutations that have been introduced with the `overwriteExisting` property on `docs-readme` as described in stenciljs/core#6248.

## Other information

See stenciljs/core#6248 for a detailed explanation of the use case this PR is covering.

### DIFF
--- a/scripts/test/copy-readme.js
+++ b/scripts/test/copy-readme.js
@@ -21,8 +21,14 @@ const fs = require('fs');
 const path = require('path');
 
 // Define source and destination paths
-const src = path.resolve(__dirname, '../../test/docs-readme/custom-readme-output-overwrite/components/styleurls-component/readme-supplemental.md');
-const dest = path.resolve(__dirname, '../../test/docs-readme/custom-readme-output-overwrite/components/styleurls-component/readme.md');
+const src = path.resolve(
+  __dirname,
+  '../../test/docs-readme/custom-readme-output-overwrite/components/styleurls-component/readme-supplemental.md',
+);
+const dest = path.resolve(
+  __dirname,
+  '../../test/docs-readme/custom-readme-output-overwrite/components/styleurls-component/readme.md',
+);
 
 // Copy the file
 try {


christian-bromann (approved):
LGTM 👍 

Great change set!

smorrisods:
Thanks, @christian-bromann!  I just noticed a bunch of tests failing, any advice on looking into them?

christian-bromann:
You can ignore the failing Component Starter Tests but definitely take a look at the failing docs test which seem to cause changes when running them `npm run test.docs-build`. Can you identify if these changes are expected? How does the new behavior differ from the old?

smorrisods:
Looking a the docs tests, it looks like it is an issue with Git being dirty after the file that is expected to be overwritten is overwritten.  One of the tests replaces one of the readmes with the one that exists within the component, I'm purposefully making sure the file gets overwritten, but this breaks Git.  Any thoughts?

image

smorrisods:
One idea that comes to mind. Perhaps I can commit the intended result, and add a step before the `docs-readme` output target to overwrite it with a sample that will then be overwritten later 🤔.  If it doesn't get overwritten back to the expected version then this Git error will show up again and break the tests!

christian-bromann:
@smorrisods I think that makes sense

christian-bromann:
@smorrisods it seems like there is a remaining prettier issue with `scripts/test/copy-readme.js`, can you run `npm run prettier`?

smorrisods:
@christian-bromann, good catch! I've prettierified the file and it looks better now.  Good catch <3

christian-bromann (approved):

PR #6258, opened 2025-05-20
Changed hash algorithm in screenshot-compare to sha256

ujas-sidapara:

Screenshot compare function uses md5 hash algorithm 

GitHub Issue Number: 6241
https://csrc.nist.rip/library/NIST%20SP%20800-107%20Recommendation%20for%20Applications%20Using%20Approved%20Hash%20Algorithms,%202009-02%20(2).pdf#:~:text=hash%20algorithm%20is%20used%20to%20map%20a%20message,hash%20algorithms%3A%20SHA-1%2C%20SHA-224%2C%20SHA-256%2C%20SHA-384%2C%20and%20SHA-512.

## What is the new behavior?

Changes from md5 to sha256 as md5 is not available in environments that need FIPS compliance (or have restrictions against older hash functions with collision risks ... such as within chainguard builder images etc) 

## Documentation

https://github.com/stenciljs/core/issues/6241 

- [ ] Yes
- [X] No


christian-bromann (approved):
This change make sense, however it will be breaking change for users since the hash will be different for all images them after the update. I will put this into a `minor` release as screenshot testing is still marked "experimental".

Note: we are likely gonna deprecate this feature within the next major release as alternative testing tools like WebdriverIO or Playwright provide a better feature set for visual tests.

PR #6264, opened 2025-05-27
feat(compiler): Automatically add @default JSDoc to @Prop() declarations

yashwanth-yenugu:

Currently, Stencil does not automatically add `@default` JSDoc to `@Prop()` declarations, even when a default value is present in the component's property definition. Developers have to manually add this JSDoc.

## What is the new behavior?

Stencil will now automatically add `@default 'value'` to the JSDoc of a `@Prop()` when a default value is detected for that property and no `@default` JSDoc is already present. This improves developer quality of life by reducing redundancy and making component APIs clearer.

## Documentation

No new documentation is required, as this is an automatic behavior change.

[ ] Yes
[x] No

## Testing

Unit tests have been added to `src/compiler/types/tests/generate-prop-types.spec.ts` to verify the new behavior, including cases where a default value is present and where `@default` JSDoc is already manually set. All tests are passing.

## Other information

This feature is similar to how Stencil automatically adds `@readonly` when a `@Prop()` only has a getter method.

### DIFF
--- a/src/compiler/types/generate-prop-types.ts
+++ b/src/compiler/types/generate-prop-types.ts
@@ -18,6 +18,11 @@ export const generatePropTypes = (cmpMeta: d.ComponentCompilerMeta, typeImportDa
         cmpProp.docs.tags = [...(cmpProp.docs.tags || []), { name: 'readonly', text: '' }];
         doc = getTextDocs(cmpProp.docs);
       }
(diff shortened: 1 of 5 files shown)


yashwanth-yenugu:
Hi @christian-bromann , I’ve pushed a fix for the CI issues — the tests are now passing ✅
could you please take another look and let me know if anything else is needed? Thanks.

christian-bromann (approved):
Thank you 🙏

Issue #6270, opened 2025-05-30
bug: ReferenceError: _interopNamespace is not defined


matheus-rosa-jsm:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/stenciljs/core/blob/main/CONTRIBUTING.md).
- [x] I agree to follow the [Code of Conduct](https://github.com/stenciljs/core/blob/main/CODE_OF_CONDUCT.md).
- [x] I have searched for [existing issues](https://github.com/stenciljs/core/issues) that already report this problem, without success.

### Stencil Version

3.28.0+

### Current Behavior

After upgrade the version of stencil, from 4.27.1 to 4.32.0, I started to face this error.
CJS build is missing `_interopNamespace` rollup helper when the config `enableImportInjection` is settled to true.

![Image](https://github.com/user-attachments/assets/99537d5d-98f7-408a-9509-020f42221048)

### Expected Behavior

`_interopNamespace ` present in the helper as was when the version was 4.27.1.

![Image](https://github.com/user-attachments/assets/e9856c74-0f0d-44bf-a649-953594210a43)

### System Info

```shell
System: node 18.20.6
Platform: windows (10.0.22631)
CPU Model: Intel(R) Core(TM) Ultra 7 165U (14 cpus)
Build: 1748464182
Stencil: 4.32.0
TypeScript: 5.5.4
Rollup: 4.34.9
Parse5: 7.2.1
jQuery: 4.0.0-pre
Terser: 5.37.0
```

### Steps to Reproduce

I create an next app to use the cjs module, but you can just check the cjs file in the build. To run the project you can:
- npm i
- npm run build
- cd next-app
- npm i
- npm run dev

### Code Reproduction URL

https://github.com/matheus-rosa-jsm/stencil-bug-reproduction

christian-bromann:
Fixed in #6272

PR #6272, opened 2025-05-31
fix(compiler): disable dynamic imports in cjs when pass enableImportI…

igorwessel:

When set `enableImportInjection` this code uses `_interopNamespace` rollup helper to inject
https://github.com/stenciljs/core/blob/33363d4077728793e0c6f635a22dccbb5740be49/src/compiler/output-targets/dist-lazy/generate-lazy-module.ts#L175-L181

After Rollup upgrade currently CJS build with this flag enabled, don't have the rollup helper.

GitHub Issue Number: #6270 

## What is the new behavior?

Basically we are making sure that it is generating the _interopNamespace when the `enableImportInjection` setting is being used.

## Documentation

[Rollup Interop](https://rollupjs.org/configuration-options/#output-interop)
[Rollup Dynamic Import is the new default](https://rollupjs.org/migration/#dynamic-import-in-commonjs-output)

- [ ] Yes
- [x] No

## Testing

Generate a dev stencil build, link with npm link, and test inside the issue repo.
Check the index.js that loader.cjs uses

### DIFF
--- a/src/compiler/output-targets/dist-lazy/generate-cjs.ts
+++ b/src/compiler/output-targets/dist-lazy/generate-cjs.ts
@@ -23,6 +23,12 @@ export const generateCjs = async (
       assetFileNames: '[name]-[hash][extname]',
       sourcemap: config.sourceMap,
     };
+
+    if (!!config.extras.experimentalImportInjection || !!config.extras.enableImportInjection) {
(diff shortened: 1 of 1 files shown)


christian-bromann (approved):
Awesome, thanks a lot!

Issue #5461, opened 2024-03-12
feat: formAssociated element should treat boolean "disabled" Prop differently


danyball:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already include this feature request, without success.

### Describe the Feature Request

If the component has set `formAssociated: true` and a boolean `disabled` Prop the component should not add this as a attribute if `disabled=false` is set.

### Describe the Use Case

We have formAssociated elements which should emit the click event if the user clicks on it. This should also work if disabled=false is set. 
For formAssociated elements the value of `disabled` is not respected like described here: https://html.spec.whatwg.org/dev/form-control-infrastructure.html#attr-fe-disabled:~:text=custom%20element%2C-,and%20the,attribute%20is%20specified%20on%20this%20element%20(regardless%20of%20its%20value),-%3B%20or

### Describe Preferred Solution

If `formAssociated: true` and `disabled=false` is set, the attribute should not be present.

### Related Code

Reproducer: https://github.com/danyball/stencil-noevent.git

### Additional Information

Closed Mozilla bug ticket (not sure if related): https://bugzilla.mozilla.org/show_bug.cgi?id=1818287

alicewriteswrongs:
Hey @danyball, thanks for filing this issue and glad to hear you've been making use of our support for form-associated custom elements!

I just want to check a little bit to make sure I'm understanding the ask here. Right now Stencil and the form-associated custom elements standard interact in a bit of an undesirable way, where if you set a `disabled` prop that will be added to the element as an attribute, and since for form controls the disabled state is based on whether or not the attribute is _present_, rather than its _value_, this results in the form control being disabled whether `disabled` is set to `true` or `false`.

If I've understood that correctly, the request is then to change Stencil's behavior from something like:

```mermaid
flowchart LR
    A[Stencil component has 'disabled' prop] --> B[set value of disabled prop on element]
```

to something like this:

```mermaid
flowchart TD
    A[Stencil component has 'disabled' prop]
    B[Is component form-associated?]
    C[set disabled attr on element to prop value]
    D[What's the value for `disabled`?]
    E[Remove `disabled` attr from element]
    A ---> B
[...]

PR #6280, opened 2025-06-09
fix(runtime): correct boolean attribute handling for form-associated components

christian-bromann:

GitHub Issue Number: #5461

Currently, form-associated Stencil components incorrectly handle boolean attributes like `disabled`. When `disabled="false"` is set on a form-associated component, Stencil parses the string `"false"` as boolean `false`, which then gets reflected back to the DOM as the `disabled` attribute. According to HTML specification, for form-associated elements, the mere presence of boolean attributes like `disabled` (regardless of their value) should disable the element.

This causes form-associated components to behave incorrectly when `disabled="false"` is explicitly set, as the component remains disabled when it should be enabled.

## What is the new behavior?

For form-associated components, boolean attribute parsing now follows HTML specification:
- Any string attribute value (including `"false"`) is parsed as boolean `true`
- Empty string `""` is parsed as boolean `true` 
- Non-string values (actual boolean `false`) continue to work as expected

For non-form-associated components, the legacy behavior is preserved:
- String `"false"` continues to be parsed as boolean `false`
- Other string values are parsed as boolean `true`

This ensures form-associated components behave correctly according to HTML standards while maintaining backward compatibility for existing non-form-associated components.

## Documentation

No documentation changes required - this is a bug fix that aligns existing behavior with HTML standards.

- [ ] Yes
- [x] No

This is a bug fix that corrects incorrect behavior. Non-form-associated components maintain their existing behavior, ensuring backward compatibility.

## Testing

- Modified `parsePropertyValue()` function to accept an optional `isFormAssociated` parameter
- Updated all call sites in:
  - `src/runtime/set-value.ts`
  - `src/runtime/proxy-component.ts` 
  - `src/runtime/client-hydrate.ts`
  - `src/hydrate/platform/proxy-host-element.ts`
- Verified the build completes successfully without errors
- The fix ensures form-associated components properly handle `disabled="false"` by not setting the disabled attribute when the property value is `false`

## Other information

This fix specifically addresses the issue where form-associated custom elements with `disabled="false"` would incorrectly remain disabled due to improper boolean attribute parsing. The solution follows HTML specifications while preserving existing behavior for non-form-associated components.

**Key technical changes:**
- Enhanced `parsePropertyValue()` with form-associated component detection
- Conditional boolean parsing logic based on component type
- Maintained backward compatibility for existing components

### DIFF
--- a/src/hydrate/platform/proxy-host-element.ts
+++ b/src/hydrate/platform/proxy-host-element.ts
@@ -46,7 +46,7 @@ export function proxyHostElement(elm: d.HostElement, cstr: d.ComponentConstructo
 
         if (attrValue != null) {
           // incoming value from `an-attribute=....`. Convert from string to correct type
-          attrPropVal = parsePropertyValue(attrValue, memberFlags);
+          attrPropVal = parsePropertyValue(attrValue, memberFlags, !!(cmpMeta.$flags$ & CMP_FLAGS.formAssociated));
         }
 
         if (propValue !== undefined) {
--- a/src/runtime/client-hydrate.ts
+++ b/src/runtime/client-hydrate.ts
@@ -67,7 +67,7 @@ export const initializeClientHydrate = (
     const attrVal = hostElm.getAttribute(attributeName);
 
     if (attrVal !== null) {
-      const attrPropVal = parsePropertyValue(attrVal, memberFlags);
+      const attrPropVal = parsePropertyValue(attrVal, memberFlags, BUILD.formAssociated && !!(hostRef.$cmpMeta$?.$flags$ & CMP_FLAGS.formAssociated));
       hostRef?.$instanceValues$?.set(memberName, attrPropVal);
     }
   });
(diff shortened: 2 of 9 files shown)


christian-bromann:
I've published a dev release `@stencil/core@4.33.1-dev.1749506469.8e2d607`, any feedback would be much appreciated!

AlexanderYW:
Hi there,

I just wanted to point out that this change does appear to have introduced breaking changes, at least in some cases we've encountered. While I understand it's a bug fix and that aligning behavior for form-associated custom elements is the right direction, it may still have unexpected side effects for existing implementations.

That said, I agree with the direction, and really appreciate the time and effort you all put into this work.

duncanhunter:
> Hi there,
> 
> I just wanted to point out that this change does appear to have introduced breaking changes, at least in some cases we've encountered. While I understand it's a bug fix and that aligning behavior for form-associated custom elements is the right direction, it may still have unexpected side effects for existing implementations.
> 
> That said, I agree with the direction, and really appreciate the time and effort you all put into this work.

I agree that this is a welcome fix, but it has led to a breaking change in our design system.

Issue #6269, opened 2025-05-30
bug: TypeError: el.addEventListener is not a function in Stencil Core


joseabernardes:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/stenciljs/core/blob/main/CONTRIBUTING.md).
- [x] I agree to follow the [Code of Conduct](https://github.com/stenciljs/core/blob/main/CODE_OF_CONDUCT.md).
- [x] I have searched for [existing issues](https://github.com/stenciljs/core/issues) that already report this problem, without success.

### Stencil Version

4.31.0

### Current Behavior

We're experiencing an issue while running unit tests for **scoped** component that have a `<slot>` element and a `onSlotchange` listener attached to it.

![Image](https://github.com/user-attachments/assets/7cca3ef5-9a63-4472-ad8b-e708e6c23e7e)

### Expected Behavior

No runtime error. 

### System Info

```shell
System: node 22.14.0
Platform: darwin (24.5.0)
CPU Model: Apple M3 Max (14 cpus)
Compiler: .../node_modules/@stencil/core/compiler/stencil.js
Build: 1746556979
Stencil: 4.31.0 🐁
TypeScript: 5.5.4
Rollup: 4.34.9
Parse5: 7.2.1
jQuery: 4.0.0-pre
Terser: 5.37.0
```

### Steps to Reproduce

1. Open the reproduction project
2. `npm install`
3. `npm test`

### Code Reproduction URL

https://github.com/joseabernardes/stenciljs-issue.

### Additional Information

- Removing the `onSlotchange` listener makes the issue not happening;
- We have the `experimentalSlotFixes` flag set to `true`. Toggling it to `false` did not resolve the issue.

christian-bromann:
An fix for this was raised in [#6281](https://github.com/stenciljs/core/pull/6281)

PR #6281, opened 2025-06-09
fix(runtime): add addEventListener support for slot elements in scope components

christian-bromann:

GitHub Issue Number: #6269

When using `onSlotchange` event listeners on `<slot>` elements in **scoped components**, the application crashes with:

```
TypeError: el.addEventListener is not a function
```

This occurs because in scoped components, Stencil uses a slot polyfill that replaces `<slot>` elements with text/comment nodes (since native shadow DOM is not used). These nodes don't have native `addEventListener` methods, causing runtime errors when the framework attempts to attach event listeners.

## What is the new behavior?

- `onSlotchange` event listeners now work correctly on `<slot>` elements in scoped components
- The slot polyfill now includes custom `addEventListener`, `removeEventListener`, and `dispatchEvent` methods for polyfilled slot nodes
- Event listeners are attached without errors and function as expected
- `slotchange` events can be dispatched and handled properly
- The fix maintains backward compatibility and doesn't affect shadow DOM components

**Technical Details:**
- Enhanced `patchSlotNode()` to add event listener polyfill methods to slot reference nodes
- Fixed timing issue by patching slot nodes immediately after creation, before event listener attachment
- Added defensive checks to prevent overwriting existing methods or double-patching

## Documentation

No documentation changes required - this restores expected functionality for existing APIs.

- [ ] Yes
- [x] No

This is a bug fix that restores expected functionality. No migration is required.

## Testing

**Unit Tests Added:**
- ✅ Test that components with `onSlotchange` render without errors
- ✅ Test that `slotchange` events are properly dispatched and handled
- ✅ Test that event handlers can be called multiple times
- ✅ Test that component state updates correctly when events fire
- ✅ End-to-end verification of the event listener polyfill functionality

**Manual Testing:**
- ✅ Verified the reproduction case from issue #6269 now works
- ✅ Confirmed existing scoped component functionality remains unaffected
- ✅ Tested that shadow DOM components continue to work as expected

**Test Command:**
```bash
npm run test.jest -- src/runtime/test/scoped.spec.tsx
```

All tests pass with comprehensive coverage of the event listener polyfill.

## Other information

**Before this fix:**
```typescript
// This would crash with "addEventListener is not a function"
@Component({ tag: 'my-component', scoped: true })
class MyComponent {
  render() {
    return <slot onSlotchange={this.handleSlotChange}></slot>;
  }
}
```

**After this fix:**
```typescript
// This now works correctly
@Component({ tag: 'my-component', scoped: true })  
class MyComponent {
  handleSlotChange = (event: Event) => {
    // Event handler executes properly
    console.log('Slot content changed!', event);
  };
  
  render() {
    return <slot onSlotchange={this.handleSlotChange}></slot>;
  }
}
```

**Key Files Modified:**
- `src/runtime/slot-polyfill-utils.ts` - Added event listener polyfill
- `src/runtime/vdom/vdom-render.ts` - Fixed timing of slot patching
- `src/runtime/test/scoped.spec.tsx` - Added comprehensive tests

This fix ensures that scoped components have feature parity with shadow DOM components for slot event handling.

### DIFF
--- a/src/runtime/slot-polyfill-utils.ts
+++ b/src/runtime/slot-polyfill-utils.ts
@@ -197,12 +197,28 @@ export const getSlotName = (node: d.PatchedSlotNode) =>
     : (node.nodeType === 1 && (node as Element).getAttribute('slot')) || undefined;
 
 /**
+ * Enhanced slot node type with additional properties
+ * for event listeners, assignedElements, and assignedNodes
+ */
+type EnhancedStencilSlotElement = HTMLSlotElement & {
+  __eventListeners: { type: string; listener: EventListenerOrEventListenerObject }[];
+};
+
+/**
  * Add `assignedElements` and `assignedNodes` methods on a fake slot node
  *
  * @param node - slot node to patch
  */
 export function patchSlotNode(node: d.RenderNode) {
-  if ((node as any).assignedElements || (node as any).assignedNodes || !node['s-sr']) return;
+  const slotNode = node as HTMLSlotElement & EnhancedStencilSlotElement;
+
+  /**
+   * If the slot node already has assignedElements or assignedNodes,
+   * we don't need to patch it again.
+   */
+  if (slotNode.assignedElements || slotNode.assignedNodes || !node['s-sr']) {
+    return;
+  }
 
   const assignedFactory = (elementsOnly: boolean) =>
     function (opts?: { flatten: boolean }) {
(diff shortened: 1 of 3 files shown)


christian-bromann:
Great catch @johnjenkins , mind taking a look at https://github.com/stenciljs/core/pull/6287?

Issue #6243, opened 2025-04-18
bug: Slotted elements within a scoped element not getting correct parentNode


AndreBarr:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/stenciljs/core/blob/main/CONTRIBUTING.md).
- [x] I agree to follow the [Code of Conduct](https://github.com/stenciljs/core/blob/main/CODE_OF_CONDUCT.md).
- [x] I have searched for [existing issues](https://github.com/stenciljs/core/issues) that already report this problem, without success.

### Stencil Version

>= 4.25.0

### Current Behavior

For stencil versions >=4.25.0, if a scoped component has a wrapper around it's slot, such as:

```
<Host>
  <div class="slot-wrapper">
    <slot></slot>
  </div>
</Host>
```

then, any element slotted into this scoped element will not return the expected value when trying to get it's **parentNode**.

Example:
```
<scoped-parent>
  <slotted-content></slotted-content>
</scoped-parent>
```

In the code above the `slotted-content` element will return `scoped-parent` as it's **parentNode** and `div` as it's **parentElement**.

### Expected Behavior

The expected behavior is that both the **parentNode** and **parentElement** for the `slotted-content` in the aforementioned example, should return the slot wrapper `div` element.

When inspecting the DOM you will see:

```
<scoped-parent>
  <div class="slot-wrapper">
    <slotted-content></slotted-content>
  </div>
</scoped-parent>
```

so it's expected that the div is both the **parentNode** and **parentElement**

### System Info

```shell
System: node 21.1.0
    Platform: darwin (23.5.0)
   CPU Model: Apple M1 Pro (8 cpus)
    Compiler: /Users/andrebarrientos/Documents/repos/stenciljs-scoped-parentNode-issue/parent-node-issue/node_modules/@stencil/core/compiler/stencil.js
       Build: 1737602259
     Stencil: 4.25.0 🏅
  TypeScript: 5.5.4
      Rollup: 2.56.3
      Parse5: 7.1.2
      jQuery: 4.0.0-pre
      Terser: 5.31.1
```

### Steps to Reproduce

Any scoped element that wraps it's slot internally will have this issue.

Steps to reproduce with my code reproduction:

1. `git clone git@github.com:AndreBarr/stenciljs-scoped-parentNode-issue.git`
2. `cd stenciljs-scoped-parentNode-issue/parent-node-issue`
3. `npm i`
4. `npm run start`
5. Open the devtools in the demo environment
[...]

PR #6284, opened 2025-06-10
fix(runtime): ensure parentNode and parentElement consistency for slotted content

christian-bromann:

When `experimentalSlotFixes` is enabled, slotted content in scoped components with wrapper elements around slots exhibits inconsistent behavior between `parentNode` and `parentElement` properties:

- `parentNode` returns the component host element (e.g., `<my-component>`)
- `parentElement` returns the actual DOM parent/wrapper element (e.g., `<div class="slot-wrapper">`)

This inconsistency violates DOM specifications where both properties should reference the same element when the parent is an Element node. It causes issues for frameworks and libraries that manipulate slotted content, leading to `insertBefore` errors and other DOM manipulation failures.

**Example:**
```html
<scoped-parent>
  <div class="slot-wrapper">
    <slotted-content></slotted-content>  <!-- parentNode ≠ parentElement -->
  </div>
</scoped-parent>
```

GitHub Issue Number: #6243

## What is the new behavior?

Both `parentNode` and `parentElement` now consistently return the actual DOM parent (wrapper element) for slotted content:

- `parentNode` now returns the wrapper element instead of the component host
- `parentElement` continues to return the wrapper element (unchanged)
- Both properties are now consistent with each other and with DOM specifications

**Changes made:**
1. Updated `patchParentNode` to return `this.__parentNode` (actual DOM parent) instead of `this['s-ol']?.parentNode` (component host)
2. Added `patchParentElement` function to ensure consistent behavior
3. Added `'parentElement'` to the `validElementPatches` array for proper patching
4. Updated existing tests to reflect the correct behavior

## Documentation

No documentation changes required - this is an internal DOM behavior fix that makes Stencil conform to standard DOM specifications.

- [x] Yes
- [ ] No

**Breaking Change Details:**

`parentNode` now returns the actual DOM parent (wrapper element) instead of the component host for slotted content when `experimentalSlotFixes` is enabled.

**Impact:**
- Code that relied on `parentNode` returning the component host will need to be updated
- Most applications should benefit from this fix as it resolves DOM manipulation issues
- This change aligns Stencil with correct DOM behavior, so most code expecting standard DOM semantics will work better

**Migration Path:**
- If your code specifically needed the component host, use the component's `__parentNode` property or traverse up the DOM tree manually
- In most cases, no changes should be needed as this fix resolves existing bugs

## Testing

**Test Coverage:**
1. **Updated existing test**: Modified `scoped-slot-slotted-parentnode/cmp.test.tsx` to verify `parentNode` returns the wrapper element instead of component host
2. **Added consistency test**: New test verifies `parentNode` and `parentElement` return the same value
3. **Regression testing**: Ran all scoped slot tests (`**/scoped-slot*/**/*.test.tsx`) to ensure no functionality is broken
4. **Cross-platform verification**: Tests run successfully in Chrome browser environment via WDIO

**Test Results:**
- ✅ All 3 tests in the modified test file pass
- ✅ All 11 scoped slot test suites continue to pass (28 total tests)
- ✅ No regressions detected in existing functionality

**Manual Verification:**
- Verified debug output shows both properties now return the wrapper element
- Confirmed the fix resolves the original `insertBefore` error scenario

## Other information

**Before the fix:**
```javascript
// Debug output showing the inconsistency
parentNode tagName: CMP-SLOTTED-PARENTNODE  // ❌ Component host
parentElement tagName: LABEL                // ✅ Wrapper element
```

**After the fix:**
```javascript
// Debug output showing consistency
parentNode tagName: LABEL    // ✅ Wrapper element  
parentElement tagName: LABEL // ✅ Wrapper element
```

This fix specifically addresses issues reported by users working with Angular and other frameworks that manipulate slotted content dynamically, where the `parentNode`/`parentElement` inconsistency was causing `NotFoundError: Failed to execute 'insertBefore'` errors.

The fix only affects components with `experimentalSlotFixes: true` in the Stencil config, so it will not impact applications not using this experimental feature.

### DIFF
--- a/src/runtime/dom-extras.ts
+++ b/src/runtime/dom-extras.ts
@@ -296,7 +296,7 @@ const patchInsertBefore = (HostElementPrototype: HTMLElement) => {
      * Fixes an issue where slotted elements are dynamically relocated in React, such as after data fetch.
      *
      * When a slotted element is passed to another scoped component (e.g., <A><C slot="header"/></A>),
-     * the child’s __parentNode (original parent node property) does not match this.
+     * the child's __parentNode (original parent node property) does not match this.
      *
      * To prevent errors, this checks if the current child's parent node differs from this.
      * If so, appendChild(newChild) is called to ensure the child is correctly inserted,
@@ -433,6 +433,7 @@ export const patchSlottedNode = (node: Node) => {
   if (node.nodeType === Node.ELEMENT_NODE) {
     patchNextElementSibling(node as Element);
     patchPreviousElementSibling(node as Element);
+    patchParentElement(node as Element);
   }
 };
 
@@ -533,7 +534,8 @@ export const patchParentNode = (node: Node) => {
   patchHostOriginalAccessor('parentNode', node);
   Object.defineProperty(node, 'parentNode', {
     get: function () {
-      return this['s-ol']?.parentNode || this.__parentNode;
+      // Return the actual DOM parent (wrapper element) instead of the slot original location parent
+      return this.__parentNode;
     },
     set: function (value) {
       // mock-doc sets parentNode?
@@ -542,9 +544,30 @@ export const patchParentNode = (node: Node) => {
   });
 };
 
+/**
+ * Patches the `parentElement` accessor of a non-shadow slotted node
+ *
+ * @param element the slotted element node to be patched
+ */
+const patchParentElement = (element: Element) => {
+  if (!element || (element as any).__parentElement) return;
+
+  patchHostOriginalAccessor('parentElement', element);
+  Object.defineProperty(element, 'parentElement', {
+    get: function () {
+      // Return the actual DOM parent (wrapper element) to match parentNode behavior
+      return this.__parentElement;
+    },
+    set: function (value) {
+      // mock-doc sets parentElement?
+      this.__parentElement = value;
+    },
+  });
+};
+
 /// UTILS ///
 
-const validElementPatches = ['children', 'nextElementSibling', 'previousElementSibling'] as const;
+const validElementPatches = ['children', 'nextElementSibling', 'previousElementSibling', 'parentElement'] as const;
 const validNodesPatches = [
   'childNodes',
   'firstChild',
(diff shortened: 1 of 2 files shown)


christian-bromann:
@johnjenkins thank you for bringing this up. You are pointing out an interesting tension in the design philosophy of Stencil. You're absolutely correct that:

1. **Scoped components should ideally mimic shadow DOM behavior** where `parentNode`/`parentElement` return the component host
2. **Exposing component internals** (wrapper elements) traditionally breaks the abstraction layer
3. **Frameworks typically expect** to interact only with lightDOM nodes and component hosts

I tested this against real shadow DOM behavior, and you're right—in true shadow DOM, slotted content always maintains the component host as its parent, regardless of internal wrapper elements.

However, this change addresses a real-world framework integration issue. The inconsistency between `parentNode` and `parentElement` is causing:

```javascript
// Before the fix:
element.parentNode.tagName        // "CMP-HOST" (component)  
element.parentElement.tagName     // "LABEL" (wrapper)

// This breaks DOM manipulation:
element.parentNode.insertBefore(newEl, element) // Error: element not in parentNode
```

This violates the DOM specification where these properties should be consistent and is breaking frameworks like React/Vue that rely on standard DOM manipulation patterns.

We're facing competing goals:
1. **Shadow DOM Fidelity**: Maintain abstraction, return component host
2. **DOM Specification Compliance**: Ensure `parentNode`/`parentElement` consistency  
3. **Framework Compatibility**: Prevent real-world integration failures

There are some options we can go here:

**Option 1: Enhanced Documentation + Migration Support**
- Keep the current fix but provide clear documentation about the behavioral change
- Create a migration guide for affected frameworks
- Add TypeScript types that reflect the new behavior

**Option 2: Feature Flag with Deprecation Path**
- Add a new flag like `legacySlotParentBehavior` for backwards compatibility
- Default to the new consistent behavior for `experimentalSlotFixes`
- Provide a clear deprecation timeline

**Option 3: Framework-Specific Workarounds**
- Investigate if we can detect framework usage patterns and adjust behavior
- Provide framework-specific utilities for DOM manipulation

This issue highlights a broader architectural question: Should Stencil prioritize theoretical shadow DOM compatibility or practical web platform integration? 

Throughts?

christian-bromann:
@johnjenkins thanks for looking into this, have you tried this reproduction case https://github.com/AndreBarr/stenciljs-scoped-parentNode-issue/tree/main by @AndreBarr?

christian-bromann:
I will go ahead and close this as I don't think there is value continuing with my current approach. Thanks @johnjenkins for all your feedback on this.

Issue #6285, opened 2025-06-10
feat: Support CustomStateSet


pfteter:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/stenciljs/core/blob/main/CONTRIBUTING.md).
- [x] I agree to follow the [Code of Conduct](https://github.com/stenciljs/core/blob/main/CODE_OF_CONDUCT.md).
- [x] I have searched for [existing issues](https://github.com/stenciljs/core/issues) that already report this problem, without success.

### Stencil Version

4.33

### Current Behavior

`@AttachInternals() internals!: ElementInternals;`

cannot be used without 
`formAssociated: true` in the metadata.

because of a condition in the compiler

### Expected Behavior

`@AttachInternals() internals!: ElementInternals;`

should work without `formAssociated: true` in the metadata.

CustomStateSet can now be used without formAssociated for components that are not formAssociated.

Future improvement: Add stencil support for custom state / :state().

### System Info

```shell

```

### Steps to Reproduce

add 

`@AttachInternals() internals!: ElementInternals;`

to a component, an console error will be shown

### Code Reproduction URL

-

### Additional Information

Will commit a fix for it, the solution is quite easy:
https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/attachInternals

christian-bromann:
@pfteter looking at the code I think the original author wanted to ensure that element internals can only be accessed when the `formAssociated` flag is set. Can you point to references that state the opposite?

pfteter:
I will formulate it differently:

How can I use/ add a custom state(CustomStateSet / :state())  to a component that is not a form component without associating it with a form in stencil so that it does't throws an error in the console ?

https://developer.mozilla.org/en-US/docs/Web/API/CustomStateSet

`CustomStateSet` / CSS `:state()` is quite a new thing for custom elements where you don't need to use class or {reflect: true} to style or  a component.

forms are just one use case for the technology. I guess at the time the code was written there was only `formAssociated` that was used with `attachInternals()`.

`attachInternals()` is just a JS technology that enables `formAssociated` but it's not exclusive to it.

Here is a nice example in the html living standard for the usage of attachInternals without formAssociated:
https://html.spec.whatwg.org/multipage/custom-elements.html#exposing-custom-element-states

for custom elements state.

That is also the use case I need it for. I don't want to make my component to be a form component since it's not one.

[...]

PR #6286, opened 2025-06-10
feat(compiler): support for CustomStateSet

pfteter:
Ticket: #6285

### Stencil Version

4.33

### Current Behavior

`@AttachInternals() internals!: ElementInternals;`

cannot be used without 
`formAssociated: true` in the metadata.

because of a condition in the compiler

### Expected Behavior

`@AttachInternals() internals!: ElementInternals;`

should work without `formAssociated: true` in the metadata.

CustomStateSet can now be used without formAssociated for components that are not formAssociated.

Improvement: Add stencil support for custom state.

### System Info

```shell

```
### Steps to Reproduce

add 

`@AttachInternals() internals!: ElementInternals;`

to a component, an console error will be shown

### Code Reproduction URL

-

### Additional Information

Will commit a fix for it, the solution is quite easy:
https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/attachInternals

- [ ] Yes
- [X] No

## Testing

Manual testing / changing code in `node_modules`

### DIFF
--- a/src/compiler/transformers/test/parse-attach-internals.spec.ts
+++ b/src/compiler/transformers/test/parse-attach-internals.spec.ts
@@ -41,7 +41,7 @@ describe('parse attachInternals', function () {
     expect(t.cmp!.attachInternalsMemberName).toBe('myProp');
   });
 
-    it('should set attachInternalsMemberName even if formAssociated is false', async () => {
+  it('should set attachInternalsMemberName even if formAssociated is false', async () => {
     const t = transpileModule(`
     @Component({
       tag: 'cmp-a',


christian-bromann (changes requested):
Thanks for raising this PR.

Can we create a test for this? I suggest a WDIO e2e test that verifies that a component without the `formAssociated` flag can use element internals.

christian-bromann:
Also could we update the [docs](https://github.com/stenciljs/site/blob/main/docs/components/form-associated.md) as they should be clarify that element internals can be used without form association.

christian-bromann:
@pfteter mind running `npm run prettier` and check in the changes?

PR #6292, opened 2025-06-11
feat(global-styles): new `addGlobalStyleToComponents` extras option. Opt-out of new globalStyle behaviour

johnjenkins:

The new behaviour added via https://github.com/stenciljs/core/pull/6268 is probably desirable for a lot of users and use-cases; sharing globalStyle across all components. 
For some users however, this new behaviour might be undesirable if (for example): 

- They have designed their globalStyles around the lightDOM *alone*. For them, adding such a stylesheet to all components is completely unnecessary. An example might be globalStyles in a design-system that seek to control the marriage of potential components: doing something like `my-menu > my-link-item::part(icon)` [is not an option via traditional ::slotted(...) selectors atm](https://github.com/w3c/csswg-drafts/issues/3896)
- They are using their components during SSR. The nature of declarative-shadow-dom for SSR being declarative (😄) means that a potentially large globalStyle sheet will be added to every component as part of a server's payload is ([this is not a solved issue atm](https://github.com/whatwg/html/issues/10673))

GitHub Issue Number: N/A

## What is the new behavior?

An escape hatch for users wanting to opt-out of the behaviour: 

`config.extras.addGlobalStyleToComponents` which defaults to `true`

## Documentation

https://github.com/stenciljs/site/pull/1532

- [ ] Yes
- [x] No

### DIFF
--- a/src/compiler/bundle/app-data-plugin.ts
+++ b/src/compiler/bundle/app-data-plugin.ts
@@ -202,7 +202,10 @@ const appendGlobalScripts = (globalScripts: GlobalScript[], s: MagicString) => {
  * @param s the MagicString to append the global styles onto
  */
 const appendGlobalStyles = async (buildCtx: d.BuildCtx, s: MagicString) => {
-  const globalStyles = buildCtx.config.globalStyle ? await buildCtx.stylesPromise : '';
+  const globalStyles =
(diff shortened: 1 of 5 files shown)


christian-bromann (commented):
Makes sense to me, mind adding a PR to the docs to include it here: https://stenciljs.com/docs/config-extras

PR #6298, opened 2025-06-13
fix(ssr): expand `::part` css selectors for ssr `scoped` components

johnjenkins:

GitHub Issue Number: https://github.com/stenciljs/core/issues/6297

## What is the new behavior?

Fixes https://github.com/stenciljs/core/issues/6297

When `scoped` components are detected during SSR, expand any `::part(...)` selectors to also include the format `[part~="..."]`. This catches any components whose shadow-root is not yet attached

- [ ] Yes
- [x] No

## Other information

Follow on idea:

I *think* this behaviour could also be tweaked slightly for client-side css selectors if scoped components are in the build.

i.e. if `BUILD.scoped` extend all `::part` selectors to include `[part~="..."]` 
This would mean `scoped: true` components could use `<div part="thing">` safely; one more step closer to being interchangeable with `shadow: true`


christian-bromann (approved):
LGTM 👍

Issue #5676, opened 2024-04-16
bug: events with `dispatchEvent()` do not reach a parent's event handler that was added with `addEventListener()` in a Spec Page


tomherni:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil/issues) that already report this problem, without success.

### Stencil Version

4.16.0

### Current Behavior

*Note: this bug is for the test setup only (Spec Page).*

When a Stencil component emits an event using `dispatchEvent()`, then a parent component can listen for that event using `addEventListener()`. This works as expected in the browser. But, while running Spec tests, the event handler is not being called.

(The `@Listen` and `@Event` decorators are not being used here.)

### Expected Behavior

While running Spec tests (with `npm run test`), then I expect event handlers (added with `addEventListener()`) to be called when a child component emits an event (emitted with `dispatchEvent()`).

### System Info

```shell
System: node 20.9.0
    Platform: darwin (23.4.0)
   CPU Model: Apple M2 Max (12 cpus)
    Compiler: /stencil-event-bug/node_modules/@stencil/core/compiler/stencil.js
       Build: 1713191202
     Stencil: 4.16.0 🚛
  TypeScript: 5.4.5
      Rollup: 2.56.3
      Parse5: 7.1.2
      jQuery: 4.0.0-pre
      Terser: 5.30.3
```

### Steps to Reproduce

1. Create a "my-parent" component
2. Add an event listener using `addEventListener()` in the `constructor`. Put a `console.log` in the event handler.
3. Render a component called "my-child" in its JSX template.
4. Create the "my-child" component that emits an event using `dispatchEvent()` in the `componentDidLoad` lifecycle method.
5. Create a Spec test file that renders "my-parent".
6. Run the Spec test file. You will not see the `console.log` being called.

### Code Reproduction URL

https://github.com/tomherni/stencil-event-bug

### Additional Information

I have everything set up, including the Spec file, in my reproduction repo.

christian-bromann:
Thanks for raising the issue @tomherni 

[...]

PR #6301, opened 2025-06-15
fix(mock-doc): ensure event bubbling follows shadow DOM boundaries

stout-ni:

MockDoc bubbles DOM events to `parentElement` every time during simulation, completely ignoring the special boundary rules of the Shadow DOM. For example, with the following DOM structure:

```bash
parent-host (Shadow Host)
  └── #shadow-root (Shadow Root) 
      └── child-host
```

When the event bubbles from the `child-host`, its `parentElement` is null, but its container is the shadow-root, and `shadow-root.host` points to the `parent-host`.

GitHub Issue Number: #5676

## What is the new behavior?

According to the W3C specification, when the event is set to `composed: true`, it can cross Shadow DOM boundaries.

Therefore, for the DOM structure example above, the correct bubbling path should be: `child → shadow-root → shadow-host → parent`.

- [ ] Yes
- [x] No

## Testing

**Test Coverage:**

Added consistency test: New test `src/mock-doc/test/shadow-dom-event-bubbling.spec.ts`.

**Test Results:**

✅ All 3 tests in the modified test file pass

**Manual Verification:**

In this [minimal reproducible example](https://github.com/tomherni/stencil-event-bug), after linking `@stencil/core` and re-running `npm run test`, the previously failing test case `my-parent: logs 3 separate steps to the console` has now succeeded.

### DIFF
--- a/src/mock-doc/event.ts
+++ b/src/mock-doc/event.ts
@@ -218,8 +218,28 @@ function triggerEventListener(elm: any, ev: MockEvent) {
   } else if (elm.parentElement == null && elm.tagName === 'HTML') {
     triggerEventListener(elm.ownerDocument, ev);
   } else {
-    triggerEventListener(elm.parentElement, ev);
+    const nextTarget = getNextEventTarget(elm, ev);
+    triggerEventListener(nextTarget, ev);
   }
+}
+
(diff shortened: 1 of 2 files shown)


stout-ni:
@christian-bromann Hi Christian, could you please help review this PR? Thank you so much!

Issue #6213, opened 2025-03-26
bug: PUPPETEER_EXECUTABLE_PATH broken since 4.24.0


pierrebaronqumu:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/stenciljs/core/blob/main/CONTRIBUTING.md).
- [x] I agree to follow the [Code of Conduct](https://github.com/stenciljs/core/blob/main/CODE_OF_CONDUCT.md).
- [x] I have searched for [existing issues](https://github.com/stenciljs/core/issues) that already report this problem, without success.

### Stencil Version

4.24.0

### Current Behavior

Hi,

I am currently running Stencil via Jenkins. I have to install chromium on my Docker instance manually and I run the Stencil test command via:

```
PUPPETEER_EXECUTABLE_PATH=$(which chromium) npm run test
```

As of Stencil v4.24.0, I get the following error

```
[2025-03-26T11:14:03.614Z] + which chromium
[2025-03-26T11:14:03.615Z] + PUPPETEER_EXECUTABLE_PATH=/usr/bin/chromium npm run test
[2025-03-26T11:14:03.815Z] 
[2025-03-26T11:14:03.815Z] > lib@1.0.0 test
[2025-03-26T11:14:03.815Z] > stencil test --spec --e2e --coverage
[2025-03-26T11:14:03.815Z] 
[2025-03-26T11:14:03.815Z] [2m[14:03.8][22m  [36m@stencil/core[39m
[2025-03-26T11:14:04.317Z] [2m[14:04.2][22m  [36mv4.25.3[39m 🚁
[2025-03-26T11:14:04.617Z] [2m[14:04.5][22m  [35mtesting e2e and spec files[39m
[2025-03-26T11:14:07.624Z] [2m[14:07.5][22m  build, ui-kit, dev mode, started [2m...[22m
[2025-03-26T11:14:07.825Z] 
[2025-03-26T11:14:07.825Z] [1m[31m[ ERROR ][39m[22m  Could not find Google Chrome executable for channel 'stable' at
[2025-03-26T11:14:07.825Z]            '/opt/google/chrome/chrome'. Error: Could not find Google Chrome
[2025-03-26T11:14:07.825Z]            executable for channel 'stable' at '/opt/google/chrome/chrome'. at
[2025-03-26T11:14:07.825Z]            computeSystemExecutablePath
[...]

PR #6308, opened 2025-06-19
feat(testing): support browser executable path detection via environm…

christian-bromann:

GitHub Issue Number: #6213

The `PUPPETEER_EXECUTABLE_PATH` environment variable is not being respected when running Stencil tests with Puppeteer. Users who set this environment variable to specify a custom browser executable (e.g., in Docker environments where Chromium is installed separately) encounter errors because Stencil is not using the specified path.

The current implementation only checks `config.testing.browserExecutablePath` but ignores common environment variables that users expect to work with Puppeteer.

## What is the new behavior?

This PR implements a robust fallback chain for browser executable path detection:

1. **Environment Variables Support**: Automatically detects browser executables from:
   - `PUPPETEER_EXECUTABLE_PATH` - Standard Puppeteer environment variable
   - `CHROME_PATH` - Common Chrome/Chromium path variable
   - Falls back to `puppeteer.puppeteerExecutablePath()` if neither is set

2. **Improved Documentation**: Added comprehensive JSDoc documentation for the `browserExecutablePath` property explaining the new default behavior and fallback chain.

3. **Simplified Logic**: Removed conditional logic in favor of a unified fallback resolution approach that always provides a value.

This change ensures that users can specify browser executables through environment variables without requiring explicit configuration, which is especially important for CI/CD environments and Docker deployments.

## Documentation

The JSDoc documentation for `TestingConfig.browserExecutablePath` has been updated to reflect the new default behavior and fallback chain.

- [ ] Yes
- [x] No

This change is backward compatible. The `browserExecutablePath` configuration option still works as before, but now users have additional options through environment variables when this config is not explicitly set.

## Testing

- Verified that existing `browserExecutablePath` configuration still works
- Tested environment variable fallback chain:
  - `PUPPETEER_EXECUTABLE_PATH` takes precedence 
  - `CHROME_PATH` is used when `PUPPETEER_EXECUTABLE_PATH` is not set
  - Default Puppeteer executable path is used when neither environment variable is set
- Confirmed the fix resolves the issue described in #6213

## Other information

This addresses the regression introduced in recent versions where `PUPPETEER_EXECUTABLE_PATH` stopped working. The fix ensures that Docker users and CI/CD environments can continue to use this standard environment variable to specify custom browser executables without additional configuration.

The implementation follows Puppeteer's own environment variable conventions, making it more intuitive for users already familiar with Puppeteer's behavior.

### DIFF
--- a/src/compiler/config/test/validate-testing.spec.ts
+++ b/src/compiler/config/test/validate-testing.spec.ts
@@ -64,6 +64,15 @@ describe('validateTesting', () => {
   });
 
   describe('browserHeadless', () => {
+    const originalCI = process.env.CI;
+    beforeEach(() => {
+      delete process.env.CI;
+    });
+
+    afterEach(() => {
+      process.env.CI = originalCI;
+    });
+
     describe("using 'headless' value from cli", () => {
       it.each([false, 'shell'])('sets browserHeadless to %s', (headless) => {
         userConfig.flags = { ...flags, e2e: true, headless };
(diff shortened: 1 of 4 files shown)


ReneWerner87:
testing of the fix
https://github.com/stenciljs/core/issues/6213#issuecomment-2990052340

PR #6311, opened 2025-06-20
fix(ssr): fixes for `scoped: true` components during SSR

johnjenkins:

GitHub Issue Number: https://github.com/stenciljs/core/issues/6313

## What is the new behavior?

Fixes https://github.com/stenciljs/core/issues/6313

- [ ] Yes
- [ ] No

### DIFF
--- a/src/runtime/client-hydrate.ts
+++ b/src/runtime/client-hydrate.ts
@@ -138,7 +138,7 @@ export const initializeClientHydrate = (
         cmpMeta.$cmpMeta$,
         BUILD.mode ? childRenderNode.$elm$.getAttribute('s-mode') : undefined,
       );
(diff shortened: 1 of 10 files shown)


src/runtime/client-hydrate.ts
```diff
@@ -166,7 +180,7 @@ export const initializeClientHydrate = (
       }
     }
     // Remove the original location from the map
-    plt.$orgLocNodes$.delete(orgLocationId);
+    if (orgLocationNode && !orgLocationNode['s-id']) plt.$orgLocNodes$.delete(orgLocationId);
```
christian-bromann, line 183:
```suggestion
    if (orgLocationNode && !orgLocationNode['s-id']) {
      plt.$orgLocNodes$.delete(orgLocationId);
    }
```

PR #6320, opened 2025-06-27
fix(utils): single global stylesheet instance for performance

aeharding:

GitHub Issue Number: None

Since Ionic v8.6.0 I've been experiencing component instance creation performance issues, especially in virtual lists, where it can take items longer to appear. This was originally identified by end users: https://vger.social/post/20787019

I narrowed it down to the [stencil upgrade here](https://github.com/ionic-team/ionic-framework/pull/30450).

From there I found the regression occurring in [Stencil v4.33.0](https://github.com/stenciljs/core/releases/tag/v4.33.0) specifically. When I reverted [33363d4077728793e0c6f635a22dccbb5740be49](https://github.com/stenciljs/core/commit/33363d4077728793e0c6f635a22dccbb5740be49) the problem went away.

Once I identified the source of the issue, I found [some suspicious code](https://github.com/stenciljs/core/commit/33363d4077728793e0c6f635a22dccbb5740be49#diff-eefa0ad2088331387be8cbea3bea2bd4996fb27dbe458a6c89d10d3cc9c59853R15-R17) that was creating a new `CSSStyleSheet` every time a new component instance is created.

What is interesting is this performance issue exists even if `globalStyles` is empty. Simply creating a new `CSSStyleSheets` and calling `replaceSync` on an empty string every instantiation is enough to hurt performance.

## What is the new behavior?

I have modified the code to only create one global `CSSStyleSheet` (if necessary) on startup that is appended to `adoptedStyleSheets` in the `createShadowRoot` function.

This solves performance issues of using `globalStyles`. In addition, I added an early return to not create the global style sheet at all, if `globalStyles` is empty. This doesn't seem to affect perf, just makes the DX a bit nicer by not having a useless/empty `CSSStyleSheet` in the `adoptedStyleSheets` array.

- [ ] Yes
- [X] No

## Testing

I tested these changes against [Voyager](https://github.com/aeharding/voyager) by building stencil, building ionic, and finally installing in Voyager.

### DIFF
--- a/src/utils/shadow-root.ts
+++ b/src/utils/shadow-root.ts
@@ -5,31 +5,7 @@ import { CMP_FLAGS } from '@utils';
 import type * as d from '../declarations';
 import { createStyleSheetIfNeededAndSupported } from './style';
 
-/**
- * Create a single, shared global stylesheet for all shadow roots.
- *
- * This singleton avoids the performance and memory hit of
- * creating a new CSSStyleSheet every time a shadow root is created.
- *
- * Lazy getter due to "ReferenceError: Cannot access 'globalStyles'
- * before initialization" if setup is run at root level
- *
- * internal state:
- * - undefined: not yet computed
- * - null: computed and cached, but stylesheet not needed/supported
- * - CSSStyleSheet: computed and cached
- *
- * @returns CSSStyleSheet | null
- */
-const getLazyGlobalStyleSheet = (() => {
-  let value: CSSStyleSheet | null | undefined;
-
-  return () => {
-    if (value === undefined) value = createStyleSheetIfNeededAndSupported(globalStyles) ?? null;
-
-    return value;
-  };
-})();
+let globalStyleSheet: CSSStyleSheet | null | undefined;
 
 export function createShadowRoot(this: HTMLElement, cmpMeta: d.ComponentRuntimeMeta) {
   const shadowRoot = BUILD.shadowDelegatesFocus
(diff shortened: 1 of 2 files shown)


christian-bromann (approved):
Great patch 👌 LGTM

src/utils/shadow-root.ts
```diff
@@ -1,9 +1,17 @@
...
 import type * as d from '../declarations';
+import { createStyleSheetIfNeededAndSupported } from './style';
+
+/**
+ * Create a single, shared global stylesheet for all shadow roots.
+ *
+ * This singleton avoids the performance and memory hit of
+ * creating a new CSSStyleSheet every time a shadow root is created.
+ */
+export const globalStyleSheet = createStyleSheetIfNeededAndSupported(globalStyles);
```
christian-bromann, line 14: It seems like the way the runtime is bundled doesn't allow to initiate it at the root level? Maybe we can do something like:

```ts
let globalStyleSheet

export function createShadowRoot(this: HTMLElement, cmpMeta: d.ComponentRuntimeMeta) {
  const shadowRoot = BUILD.shadowDelegatesFocus
    ? this.attachShadow({
        mode: 'open',
        delegatesFocus: !!(cmpMeta.$flags$ & CMP_FLAGS.shadowDelegatesFocus),
      })
    : this.attachShadow({ mode: 'open' });

  /**
   * If constructable stylesheets are supported, we can use them to
   * add the global styles to the shadow root.
   */
  if (supportsConstructableStylesheets) {
    const sheet = new CSSStyleSheet();
    sheet.replaceSync(globalStyles);
    shadowRoot.adoptedStyleSheets.push(sheet);
  }
  if (!globalStyleSheet) {
    globalStyleSheet = createStyleSheetIfNeededAndSupported(globalStyles)
    shadowRoot.adoptedStyleSheets.push(globalStyleSheet)
  };
}
```

aeharding: Interesting! I wonder why it worked in Voyager but not the test environment. I just added lazy initialization. Although, if it's possible rejigger the build to avoid that error in the first place, that would probably be a cleaner solution than the lazy initializer.

Lmk what you think!

aeharding: Decided to do a simpler approach in https://github.com/stenciljs/core/pull/6320/commits/1414c08e6fd1e523f50310b40f5eeae6f39367f0. lmk if you like it, or https://github.com/stenciljs/core/pull/6320/commits/20d40461eb86b09d0e90452217575705bb015d8b better.

christian-bromann (approved):
LGTM 👍

PR #6321, opened 2025-06-27
refactor(runtime): don't recreate adoptedStyleSheets

aeharding:

GitHub Issue Number: N/A

When investigating #6320 I saw that there were places that the adoptedStyleSheets array was being recreated instead of mutated. MDN docs say you should only mutate this array: https://developer.mozilla.org/en-US/docs/Web/API/Document/adoptedStyleSheets

## What is the new behavior?

The array is mutated instead of recreated

- [ ] Yes
- [X] No

I don't think this is a breaking change, since we already mutate adoptedStyleSheets [here (as an example)](https://github.com/stenciljs/core/pull/6268). However, I believe this was originally done because adoptedStyleSheets used to be frozen: https://github.com/GoogleChrome/web.dev/issues/10294

Because there's other places in the codebase that treat adoptedStyleSheets as mutable I don't think this is breaking but I do want to highlight that.

### DIFF
--- a/src/runtime/styles.ts
+++ b/src/runtime/styles.ts
@@ -120,7 +120,12 @@ export const addStyle = (styleContainerNode: any, cmpMeta: d.ComponentRuntimeMet
                  */
                 const stylesheet = new CSSStyleSheet();
                 stylesheet.replaceSync(style);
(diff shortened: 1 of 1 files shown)


christian-bromann (approved):
LGTM 👍

PR #6348, opened 2025-07-17
fix(runtime): only patch non-shadow components with <slot>s

johnjenkins:

The new behaviour introduced by https://github.com/stenciljs/core/pull/6314/files means that non-shadow components without `<slot>`s are now patched. 

This means components which are purely used as style tags e.g.:

```
// plain-component.ts

@Component({
  tag: 'plain-component',
  styleUrl: 'plain-component.scss',
})
export class PlainComponent {}
```

Used like:

```
<plain-component>
  <div>something</div>
</plain-component>
```

And queried like `document.querySelector('plain-component').childNodes()` returns an empty array.

GitHub Issue Number: N/A

## What is the new behavior?

Only non-child component with slots are patched. Components that don't need the patches are left alone.

- [ ] Yes
- [x] No

### DIFF
--- a/test/end-to-end/src/dom-api/dom-api.e2e.ts
+++ b/test/end-to-end/src/dom-api/dom-api.e2e.ts
@@ -240,15 +240,10 @@ describe('dom api e2e tests', () => {
 
     await page.waitForChanges();
 
-    expect(elm).toEqualText(`
-      dom apiupdated text content
-    `);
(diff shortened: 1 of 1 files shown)


christian-bromann:
> Edge image location has changed?

Yes, unfortunately updating WebdriverIO that has a fix won't work due to other issues. I will disable Edge for now.

Issue #6335, opened 2025-07-09
bug: Does not import browser field from  package.json


Tallyb:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/stenciljs/core/blob/main/CONTRIBUTING.md).
- [x] I agree to follow the [Code of Conduct](https://github.com/stenciljs/core/blob/main/CODE_OF_CONDUCT.md).
- [x] I have searched for [existing issues](https://github.com/stenciljs/core/issues) that already report this problem, without success.

### Stencil Version

4.35.3

### Current Behavior

```json
// package.json   
"main": "./lib/index.js",
  "browser": {
    "./lib/index.js": "./lib/browser.js",
    "fs": false
  },
```
The main field is pointing to a server side code. what is required in order to import the browser field, and not the main field? 

### Expected Behavior

import browser side code

### System Info

```shell
System: node 22.16.0
    Platform: darwin (24.5.0)
   CPU Model: Apple M4 Max (16 cpus)
    Compiler: node_modules/@stencil/core/compiler/stencil.js
       Build: 1751490465
     Stencil: 4.35.3 🎹
  TypeScript: 5.5.4
      Rollup: 4.34.9
      Parse5: 7.2.1
      jQuery: 4.0.0-pre
      Terser: 5.37.0
```

### Steps to Reproduce

use the qrcode package as an example and try to bundle into a component. 

### Code Reproduction URL

n/a ATM

christian-bromann:
@Tallyb thanks for raising the issue. I am not quite sure how this relates to Stencil and this repository in general 🤔

Tallyb:
@christian-bromann - you are right. I was too deep into the issue to notice I did not provide enough context. I have udpated the issue with clarifications. I hope it makes more sense now.

christian-bromann:
I am still not quite sure about the issue. Stencils artifacts with exception of the hydrate module are all meant to be consumed by the browser. It is up to you whether to point to the lazy load version (e.g. loader) or `dist-custom-element` output.

Tallyb:
I am building a qr code component, which is using a 3rd party library qrcode (https://www.npmjs.com/package/qrcode), which is built as  dual package for server and browser. The above code is from their package.json. 

This is my component. 
```
import QRCode from 'qrcode';
export class QrcodeComponent {
// implement some code here using the QRCode library
}
```

[...]

PR #6353, opened 2025-07-27
fix(bundle): remove post order of node-resolve

Fa3ris:

node-resolve resolveId hook has been moved to the end of the plugin chain, so other plugins that were meant to be called after resolve some modules incorrectly

GitHub Issue Number: #6335

## What is the new behavior?

move the node-resolve module back to its original order

- [ ] Yes
- [x] No

## Testing

manually tested against the sample repo https://github.com/Fa3ris/stencil-browser-field-resolution


johnjenkins (approved):
LGTM - thanks for contributing

PR #6359, opened 2025-07-28
fix(runtime): fixed parsing of complex attributes that contains JSON strings

Jagget:

The `HasAttribute` member flag only includes `String | Number | Boolean | Any` types, which means that props with `Unknown` types (complex objects) are not considered to have attributes. This creates an inconsistency with the `Prop` flag definition, which includes `Unknown` types.

## What is the new behavior?

The `HasAttribute` member flag now includes `Unknown` types (`String | Number | Boolean | Any | Unknown`), making it consistent with the `Prop` flag definition. This allows complex object props to properly support attribute-based initialization through JSON string parsing. The added test demonstrates that complex object props can now be correctly initialized via HTML attributes using JSON strings.

- [ ] Yes
- [x] No

## Testing

Added

### DIFF
--- a/src/runtime/test/prop.spec.tsx
+++ b/src/runtime/test/prop.spec.tsx
@@ -212,14 +212,16 @@ describe('prop', () => {
   it('should demonstrate JSON parsing for complex object props', async () => {
     @Component({ tag: 'simple-demo' })
     class SimpleDemo {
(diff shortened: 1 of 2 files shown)


johnjenkins (approved):
LGTM!

PR #6362, opened 2025-07-31
feat(runtime): allow class extending

johnjenkins:

Stencil component classes currently do not allow `extends` ... mainly because it utilises static-analysis to gather meta for properties, methods, events and docs - this is hard for dynamically extending classes.  

GitHub Issue Number: 
- https://github.com/stenciljs/core/issues/2844
- https://github.com/stenciljs/core/issues/2559

## What is the new behavior?

- closes https://github.com/stenciljs/core/issues/2844
- closes https://github.com/stenciljs/core/issues/2559

This is a decent size PR to enable all kinds of `extends` within Stencil:

1) Components can `extend` from other components
2) Components can `extend` from abstract, Stencil decorated classes that are not designed to be full components
3) Components can `extend` from 3rd party / node_module stencil components (via `dist/collection`)
4) Components can `extend` from non-stencil classes
5) The extend tree / chain is preserved and works appropriately

The only caveat is that Stencil decorated classes can only be extended from *if* the `tsconfig` target is `es2022` or higher.  This is due to how Stencil initialises classes by changing their prototype, how older classes instantiate class properties within the constructor and the order of `super()` initialisation. 

- [ ] Yes
- [x] No

## Testing

- [x] lots of new spec tets
- [x] lots of new wdio tests

### DIFF
--- a/test/wdio/ts-target/components.d.ts
+++ b/test/wdio/ts-target/components.d.ts
@@ -18,6 +18,18 @@ export namespace Components {
          */
         "prop2": string;
     }
+    interface ExtendedCmpCmp {
+        "method1": () => Promise<void>;
+        "method2": () => Promise<void>;
+        /**
+          * @default 'ExtendedCmp text'
+         */
+        "prop1": string;
+        /**
+          * @default 'ExtendedCmp prop2 text'
+         */
+        "prop2": string;
+    }
     interface ExtendsCmpCmp {
         "method1": () => Promise<void>;
         "method2": () => Promise<void>;
@@ -74,6 +86,12 @@ declare global {
         prototype: HTMLExtendedCmpElement;
         new (): HTMLExtendedCmpElement;
     };
+    interface HTMLExtendedCmpCmpElement extends Components.ExtendedCmpCmp, HTMLStencilElement {
+    }
+    var HTMLExtendedCmpCmpElement: {
+        prototype: HTMLExtendedCmpCmpElement;
+        new (): HTMLExtendedCmpCmpElement;
+    };
     interface HTMLExtendsCmpCmpElement extends Components.ExtendsCmpCmp, HTMLStencilElement {
     }
     var HTMLExtendsCmpCmpElement: {
@@ -100,6 +118,7 @@ declare global {
     };
     interface HTMLElementTagNameMap {
         "extended-cmp": HTMLExtendedCmpElement;
+        "extended-cmp-cmp": HTMLExtendedCmpCmpElement;
         "extends-cmp-cmp": HTMLExtendsCmpCmpElement;
         "extends-external": HTMLExtendsExternalElement;
         "extends-mixin": HTMLExtendsMixinElement;
@@ -108,6 +127,16 @@ declare global {
 }
 declare namespace LocalJSX {
     interface ExtendedCmp {
+        /**
+          * @default 'ExtendedCmp text'
+         */
+        "prop1"?: string;
+        /**
+          * @default 'ExtendedCmp prop2 text'
+         */
+        "prop2"?: string;
+    }
+    interface ExtendedCmpCmp {
         /**
           * @default 'ExtendedCmp text'
          */
@@ -161,6 +190,7 @@ declare namespace LocalJSX {
     }
     interface IntrinsicElements {
         "extended-cmp": ExtendedCmp;
+        "extended-cmp-cmp": ExtendedCmpCmp;
         "extends-cmp-cmp": ExtendsCmpCmp;
         "extends-external": ExtendsExternal;
         "extends-mixin": ExtendsMixin;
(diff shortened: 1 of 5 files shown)


src/compiler/transformers/transform-utils.ts
```diff
@@ -971,11 +985,9 @@ export const updateConstructor = (
   if (constructorIndex < 0 && !statements?.length && !needsSuper(classNode)) return classMembers;
 
   if (constructorIndex >= 0 && ts.isConstructorDeclaration(constructorMethod)) {
-    const constructorBodyStatements: ts.NodeArray<ts.Statement> =
-      constructorMethod.body?.statements ?? ts.factory.createNodeArray();
-    const hasSuper = constructorBodyStatements.some((s) => s.kind === ts.SyntaxKind.SuperKeyword);
+    const constructorBodyStatements = constructorMethod.body?.statements;
 
-    if (!hasSuper && needsSuper(classNode)) {
+    if (!foundSuper(constructorBodyStatements) && needsSuper(classNode)) {
```
jttommy, line 990: I think you can store `superCall` above, test against that and assign. Just avoid another function eval.

johnjenkins: good call!

danielleroux:
@johnjenkins any plans also to support mixins pattern?

danielleroux:
> @danielleroux can you give an example? 🙂

```tsx
import { Component, h, Host, Prop, State } from '@stencil/core';

type StencilElement = HTMLElement & ....;
type Constructor<T> = new (...args: any[]) => T;

declare class ValidationInterface {
  isValid: boolean;
  errorMessage: string;
  handleValidation(value: string): void;
  renderValidationFeedback(): unknown;
}

declare class LoadingInterface {
  isLoading: boolean;
  setLoading(loading: boolean): void;
  renderLoading(container: unknown): unknown;
}

const Validation = <T extends Constructor<StencilElement>>(superClass: T) => {
  class ValidationElement extends superClass {
    @Prop({ mutable: true }) isValid: boolean = true;
    @Prop({ mutable: true }) errorMessage: string = '';

    handleValidation(value: string): void {
      if (value.trim() === '') {
        this.isValid = false;
        this.errorMessage = 'This field cannot be empty.';
      } else {
        this.isValid = true;
        this.errorMessage = '';
      }
    }

    renderValidationFeedback(): unknown {
      if (!this.isValid && this.errorMessage) {
        return <div class="error-message">{this.errorMessage}</div>;
      }
      return null;
    }
  }
  return ValidationElement as Constructor<ValidationInterface> & T;
};

const Loading = <T extends Constructor<StencilElement>>(superClass: T) => {
  class LoadingElement extends superClass {
    @State() isLoading: boolean = false;

    setLoading(loading: boolean) {
      this.isLoading = loading;
    }

    renderLoading(container: unknown): unknown {
      if (this.isLoading) {
        return <div class="loading-spinner">Loading...</div>;
      }
      return container;
    }
  }
  return LoadingElement as Constructor<LoadingInterface> & T;
};

@Component({
  tag: 'example-component',
  styleUrl: 'example.css',
  shadow: true,
})
export class ExampleComponent extends Loading(Validation(HTMLElement)) {
  onValueChange(event: Event) {
    const input = event.target as HTMLInputElement;
    this.setLoading(true);
    this.handleValidation(input.value);
    this.setLoading(false);
  }

  render() {
    return (
      <Host>
        {this.renderLoading(
          <input
            type="text"
            placeholder="Type something..."
            onChange={this.onValueChange.bind(this)}
          />
        )}
        {this.renderValidationFeedback()}
      </Host>
    );
  }
}

@Component({
  tag: 'example-component',
  styleUrl: 'example.css',
  shadow: true,
})
export class MyComponent extends Loading(HTMLElement) {

  componentDidLoad() {
     setLoading(true);
     setTimeout(() => {
         {this.renderValidationFeedback()}
     }, 1000);
  }

  render() {
    return (
      <Host>
        {this.renderLoading(
          <input
            type="text"
            placeholder="Type something..."
            onChange={this.onValueChange.bind(this)}
          />
        )}
      </Host>
    );
  }
}

```

Source: https://lit.dev/docs/composition/mixins/

src/compiler/bundle/typescript-plugin.ts
```diff
@@ -57,7 +57,7 @@ export const typescriptPlugin = (
       if (isAbsolute(id)) {
         const fsFilePath = normalizeFsPath(id);
         const mod = getModule(compilerCtx, fsFilePath);
-        if (mod && mod.cmps.length > 0) {
+        if (mod && mod.cmps != null) {
```
sitaramyadav, line 60: I think this can be improved 
`if (mod && mod.cmps) {`

sitaramyadav: or even we can use optional chaining

Issue #6373, opened 2025-08-22
bug: undetected initial prop value on Angular


mamillastre:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/stenciljs/core/blob/main/CONTRIBUTING.md).
- [x] I agree to follow the [Code of Conduct](https://github.com/stenciljs/core/blob/main/CODE_OF_CONDUCT.md).
- [x] I have searched for [existing issues](https://github.com/stenciljs/core/issues) that already report this problem, without success.

### Stencil Version

>=4.36.0

### Current Behavior

Since Stencil 4.36.0, an initial prop value can be undetected on a dynamically set Angular component.

For example, I use the following code to emulate the dynamic `first` prop value:

```html
<my-component [first]="'issue'"></my-component>
```

I have the following logs:

```tsx
@Prop() first!: string;
@Watch('first')
firstChanged() {
  console.log('CHANGE', this.first);
}

connectedCallback() {
  console.log('connectedCallback', this.first);
}

componentWillLoad() {
  console.log('componentWillLoad', this.first);
}

componentDidLoad() {
  console.log('componentDidLoad', this.first);
}

render() {
  console.log('render', this.first);
  ...
}
```

In runtime:

<img width="218" height="105" alt="Image" src="https://github.com/user-attachments/assets/6d42c45d-5cf8-495f-af09-ae77c1e950f5" />

The value comes only at the second rendering, the prop has not been initialized before `componentWillLoad` and the `Watch()` method has not been called.
The prop update seems to be undetected but fire a rerender.

Note: I reproduced this issue only when the component is on a child routing page and when the app is loaded on this page (See reproduction steps). 

### Expected Behavior

 The prop is available on `componentWillLoad` OR the `Watch()` method is called.

### System Info

```shell
System: node 20.18.3
Platform: darwin (24.6.0)
CPU Model: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz (12 cpus)
Compiler: /stencil-demo/packages/elements/node_modules/@stencil/core/compiler/stencil.js
Build: 1752598220
Stencil: 4.36.0 🎊
TypeScript: 5.5.4
Rollup: 4.34.9
Parse5: 7.2.1
jQuery: 4.0.0-pre
Terser: 5.37.0
```

### Steps to Reproduce

- Clone the `angular-issue` branch from https://github.com/mamillastre/stencil-demo/tree/angular-issue
- In the root folder
- `npm i`
[...]

PR #6381, opened 2025-09-12
fix(dist-custom-elements): make watchers fire immediately when element already defined

kristilw:
customElements.whenDefined always returns the answer asynchronously. Making the check synchronous when possible resolves the issue

fixes: #6373

BREAKING CHANGE: Watchers will fire earlier than before, but this is the expected behavior

GitHub Issue Number: #6373

## What is the new behavior?
Fixes so that watched methods are triggering as expected when an element is already defined.

- [x] Yes
- [ ] No

I do not think i adds a breaking change as to how it was before 3.36.0, so I belive that no migration is nessecary.

## Testing

I had planned to add a unit test to this code but found it very difficult. Looking at the initialize-component.spec.tsx file didnt give me much confidence either. Instead I tested it in a private professional project with many many components and verified that the code reproduction of the error in the github issue (https://github.com/mamillastre/stencil-demo/tree/angular-issue) also now works as expected.

### DIFF
--- a/src/runtime/initialize-component.ts
+++ b/src/runtime/initialize-component.ts
@@ -110,7 +110,14 @@ export const initializeComponent = async (
       // wait for the CustomElementRegistry to mark the component as ready before setting `isWatchReady`. Otherwise,
       // watchers may fire prematurely if `customElements.get()`/`customElements.whenDefined()` resolves _before_
       // Stencil has completed instantiating the component.
-      customElements.whenDefined(cmpTag).then(() => (hostRef.$flags$ |= HOST_FLAGS.isWatchReady));
+      // customElements.whenDefined always returns the answer asynchronously (and slower than a queueMicrotask).
+      // Checking !!customElements.get(cmpTag) instead is synchronous.
+      const setWatchIsReady = () => (hostRef.$flags$ |= HOST_FLAGS.isWatchReady);
+      if (!!customElements.get(cmpTag)) {
+        setWatchIsReady();
+      } else {
+        customElements.whenDefined(cmpTag).then(setWatchIsReady);
+      }
     }
 
     if (BUILD.style && Cstr && Cstr.style) {


kristilw:
Done 👍

exorex:
@johnjenkins This change introduces a breaking behavior in how `@Watch` works.

Previously, watchers only fired when a prop actually changed after the component had mounted. That meant lifecycle guarantees were respected — DOM refs existed, and methods like `present()` or `dismiss()` could safely access elements.

With this change, watchers now fire immediately on initial prop assignment, before componentDidLoad or even before refs are set.

This breaks a fundamental expectation in Stencil’s lifecycle and will cause regressions across many components. For example, in Ionic we have patterns like:

```
@Watch('isOpen')
onIsOpenChange(newValue: boolean, oldValue: boolean) {
  if (newValue === true && oldValue === false) {
    this.present(); // requires DOM refs
  } else if (newValue === false && oldValue === true) {
    this.dismiss();
  }
}
```

Before:
- This watcher would only run when `isOpen` was toggled after the component was mounted.
- By that time, refs existed and methods like `present()` worked reliably.

After:
- The watcher is invoked immediately on the initial prop assignment.
- If `present()` depends on refs, they are still undefined.
- This leads to runtime errors and breaks many existing components.

This essentially changes the semantics of watchers from "respond to prop changes" → to "also run eagerly at init time", which was never the contract before.

This will:
- Break Ionic components (which rely heavily on watchers for isOpen, disabled, etc.).
- Break userland components that assume DOM refs exist when a watcher fires.
- Force developers to add guards and lifecycle hacks in many places.

Suggestion: This should be reconsidered or at least gated behind a flag/major version bump, since it’s a breaking change that alters Stencil’s watcher lifecycle in a way that existing code cannot safely adapt without regressions.

exorex:
@johnjenkins done #6388 

Thank you for your efforts.

kristilw:
If postponing watchers until it has rendered is the desired behaviour, why not just move the "isWatchReady" flag check to the `updateComponent` method (which calls the renderer)? It has a parameter for `isInitialLoad`, so should be easy to do. I was wondering about it when writing the branch but dismissed it because [lazy loaded components seem to set the flag immediately](https://github.com/stenciljs/core/blob/main/src/runtime/initialize-component.ts#L88). Could it be that the code for "isWatchReady" was added to the `initializeComponent` method, as a hack, because of the async nature of `customElements.whenDefined`?

kristilw:
After doing a small test I can at least verify that a lazy loaded component does indeed fire the watchers before the first rendrer (even before componentWillLoad).

I think moving the `isWatchReady` check into `updateComponent` won't be that difficult, can also test it out in our projects to verify if it will work there. It will introduce a breaking change compare to how lazy components currently work, but I don't know Stencil well enough to predict other possible breaking changes.

exorex:
@johnjenkins You’re absolutely right — thanks for catching that. I should have clarified earlier: this issue only happens in the `dist-custom-elements` output. The same code works fine in the regular `dist` output without errors.

kristilw:
In my test of the `dist` output I did the following:
1) Created a component by putting its tag in static html, setting the watched value with js (or with attribute, made no difference)
2) Waited for it to be fully loaded (a set timeout of 1000ms)
3) Created another instance of that component in js and set the watched value immediately

The watched method is not triggered on the first instance of the component. On the second component the watch is triggered immediately, before `comoponentWillLoad`. 

Moving the `isWatchReady` flag to the end of the `updateComponent` method it behaves more consistently, only firing the watched method after it has rendered once. In our larger project (using `dist-custom-elements`), manually testing our components and wathcing for any errors in the console, it seems like every thing still works.

Do you know why the `customElements.whenDefined` check is even needed? I see that for the `dist`-output there is no such check, and for the `dist-custom-elements` output, _as far as I can tell_, the element already has to be defined before entering `initializeComponent`, beause `initializeComponent` is triggered by the `connectedCallback` set up by `proxyCustomElement`. Looking at the original [issue](https://github.com/stenciljs/core/pull/2971) doesn't give much insight. I see that `initializeComponent` can also be triggered by some HMR stuff, maybe it is because of that.

kristilw:
After moving `isWatchReady` to the end of `updateComponent` it [breaks two unit tests](https://github.com/stenciljs/core/blob/main/src/runtime/test/watch.spec.tsx#L77) that verifies that the watched method is triggered even when changing the state inside `componentWillLoad`. So either [this lifecycle overview](https://stenciljs.com/docs/component-lifecycle) is wrong, or the tests are wrong? 🫤

Issue #6368, opened 2025-08-18
bug: stencil tries to hydrate value properties on input bindings causing bugs preventing JSON entry


ShaneK:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/stenciljs/core/blob/main/CONTRIBUTING.md).
- [x] I agree to follow the [Code of Conduct](https://github.com/stenciljs/core/blob/main/CODE_OF_CONDUCT.md).
- [x] I have searched for [existing issues](https://github.com/stenciljs/core/issues) that already report this problem, without success.

### Stencil Version

4.29.0

### Current Behavior

I suspect this issue was introduced in PR #6208 

The issue is whenever you use Stencil with an input or textarea, Stencil tries to hydrate the value on change. So if the user inputs `{}`, it becomes a literal object, which we try to Stringify in Ionic Framework and it becomes `[object Object]`.

### Expected Behavior

Stencil should not turn input values into objects when the values change to contain JSON.

### System Info

```shell

```

### Steps to Reproduce

You can reproduce this issue on the ion-input docs, here: 
https://ionicframework.com/docs/api/input

Reproduce by going to the Default Input field and typing `{}` and note that it becomes `[object Object]`.

### Code Reproduction URL

https://ionicframework.com/docs/api/input

Issue #6380, opened 2025-09-10
bug: JSON string parsed to JS object when prop type is a union


craquet:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/stenciljs/core/blob/main/CONTRIBUTING.md).
- [x] I agree to follow the [Code of Conduct](https://github.com/stenciljs/core/blob/main/CODE_OF_CONDUCT.md).
- [x] I have searched for [existing issues](https://github.com/stenciljs/core/issues) that already report this problem, without success.

### Stencil Version

v4.27.1 and v4.36.3 (others untested)

### Current Behavior

When passing a string that contains valid JSON (such as `"{\"activeOptionalProperties\":[\"digitalObjectLocationAccessProtocol\"]}"`) to a Stencil component prop whose type is not just `string` (such as `string | number`), it gets parsed into an object.

### Expected Behavior

The docs clearly mention that Stencil never parses strings in this manner, and I also would not expect this to happen. A valid JSON string being passed to a prop of type e.g. `string | number` should just remain a string.

### System Info

```shell
System: node 22.11.0
    Platform: darwin (24.6.0)
   CPU Model: Intel(R) Core(TM) i9-9900K CPU @ 3.60GHz (16 cpus)
    Compiler: /Users/chris/WebstormProjects/repro-string-to-object/node_modules/@stencil/core/compiler/stencil.js
       Build: 1755687753
     Stencil: 4.36.3 🐈
  TypeScript: 5.5.4
      Rollup: 4.34.9
      Parse5: 7.2.1
      jQuery: 4.0.0-pre
      Terser: 5.37.0

Confirmed on Firefox, Chrome and Safari
```

### Steps to Reproduce

Create a component like this

```tsx
@Component({
  tag: 'rendering-component',
})
export class MyComponent {
  @Prop() value: string | number;

  render() {
    console.log(typeof this.value);

    return <div>{this.value} (type: {typeof this.value})</div>;
  }
}
```

Then integrate this component in a different Stencil component and pass a valid JSON string to the value prop:

```tsx
<rendering-component value={"{\"activeOptionalProperties\":[\"digitalObjectLocationAccessProtocol\"]}"} />
```

Then take a look at the console. It will log type `object`. It will also throw an error, because objects can't be rendered.

### Code Reproduction URL

https://github.com/craquet/repro-string-to-object

### Additional Information

[...]

PR #6384, opened 2025-09-15
fix(runtime): stop eager json parsing for unknown and any type bindings

ShaneK:
fixes: #6368, #6380

The issue is whenever you use Stencil with an input or textarea, Stencil tries to hydrate the value on change. So if the user inputs {}, it becomes a literal object, which we try to Stringify in Ionic Framework and it becomes [object Object].

GitHub Issue Number: #6368, #6380

## What is the new behavior?

We're removing the bit that tries to imitate hydration in the HTML for any and complex types for now. In the future, this might be implemented in a more thorough way that's less likely to have impact elsewhere.

Additionally, there's regression tests to prevent this case from happening again in the future.

- [ ] Yes
- [ ] No


johnjenkins (approved):
Lgtm - thanks!

PR #6387, opened 2025-09-16
feat: new core decorators `@PropSerialize` & `@AttrDeserialize`

johnjenkins:

GitHub Issue Number: N/A

- [ ] Yes
- [ ] No


exorex:
I lean toward keeping `@PropSerialize` and `@AttrDeserialize` because they’re explicit, align with serialization terminology, and fit well within a framework context like Stencil.

PR #6389, opened 2025-09-19
feat(config): allow suppressing reserved public name warning

deleonio:

GitHub Issue Number: N/A

- [ ] Yes
- [ ] No


danielleroux:
@johnjenkins @deleonio 

Hey folks, just checked the latest release a found this PR which makes it possible to define "suppressing reserved public name". In general I do like the idea but it feels a little bit strange if I have classic extending in mind.

without stencil:
```
class MyComponent extends HTMLElement {
 
    override focus(): void {
        // do your focus logic in sync!        
    }

}
```

stencil:
```
class MyComponent {
   
   @Method() // @Method must be async https://stenciljs.com/docs/methods#public-methods-must-be-async
   async focus(): Promise<void> {
         // do your focus logic async
   }
}
```

We see we have changed the signature of focus from `focus(): void` => `focus(): Promise<void>`. So we make focus async and a "user" of my-component will not know if the project is build with the suppress flag and maybe expect that focus is executed sync.

Anybody thoughts on this point?

PR #6409, opened 2025-10-03
chore(testing): export stubComponentCompilerMeta as mockComponentMeta

Cliffback:
For being able to mock the compiler for testing custom output targets, the `stubComponentCompilerMeta` is very useful. Up until now, I have just copied the object, while exporting it from Stencil would be much better.

[stencil-custom-suffix-output-target](https://github.com/Cliffback/stencil-custom-suffix-output-target)

So with this export, I can now just import the object like this directly:
image

Instead of manually hardcoding it like I currently do:
[custom-suffix-output-target.spec-utils.ts](https://github.com/Cliffback/stencil-custom-suffix-output-target/blob/main/src/test/custom-suffix-output-target.spec-utils.ts)


Cliffback:
Hmm... Stuggling to make the tests pass.

I get this currently:

image

but if i add the export to `compiler/index.ts` and i import `stubComponentCompilerMeta` from there I then get:

image

But the function is already bundled, and works fine in actual use, so seems to be just a type issue

Any suggestions, @johnjenkins?

Cliffback:
> not off the top of my head. I would think however, it probably makes most sense if this object can instead be exported from the stencil/core/testing dir (e.g from `./src/testing/mocks.ts`) - I know that'd take some re-working within Stencil though.

Yeah. Like moving the stub-file from its current location, and into the mocks-file, and rewriting internal references to it in Stencil, or working with the bundler to keep its location but then include it in `mocks.ts`, or something like that?

Cliffback:
> > Yeah. Like moving the stub-file from its current location, and into the mocks-file, and rewriting internal references to it in Stencil, or working with the bundler to keep its location but then include it in mocks.ts, or something like that?
> 
> Option 1 I think? Obvs don't know if there are any wider repercussions to that

I'll give it a go!

For the API, does it make sense to export the `stubComponentCompilerMeta` as `mockComponent` for external use? That's at least what i use it for in my code when it comes to it.

Cliffback:
@johnjenkins I think it actually was as simple as not just reexporting, but explicitly importing and exporting on different lines. I at least managed to pass the failing command locally. Perhaps restart the workflow to see if it passes now?

Cliffback:
> `mockComponentMeta`

Aggreed

> Aside from formatting / linting it looks happier!

Awesome. Will fix linting and rename to `mockCompilerMeta`

Cliffback:
@johnjenkins Should be good now. Is PR title okay?

Cliffback:
Sorry, forgot to run prettier as well, should be updated now.

PR #6413, opened 2025-10-06
fix(runtime): better boolean attribute handling

### LINKED ISSUE #6411: bug: setAttribute with empty string on boolean attribute resolves to false instead of true
### Prerequisites

### Stencil Version

4.38.0

### Current Behavior

When I call "setAttribute" with the value of an empty string on a stencil component for a boolean attribute, it resolves to false. In version 4.37.1 it resolved to true, and this would be in accordance with the HTML spec.

Seems to be caused by this change: https://github.com/stenciljs/core/pull/6404 

### Expected Behavior

It should work like before.

### System Info

```shell

```

### Steps to Reproduce

Component:
```
@Component({
  tag: 'my-component',
  shadow: true,
})
export class MyComponent {

  @Prop() checked: boolean = false;

  render() {
    return <div>I'm {this.checked ? 'checked' : 'not checked'}</div>;
  }
}
```

HTML:
```
<body>
    <my-component></my-component>
    <button>Check</button>
  </body>
  <script>
      document.querySelector('my-component').setAttribute('checked', '');
    });
  </script>
```

With Stencil version 4.37.1, clicking the button toggles the checked attribute. With version 4.38.0 it doesn't.

### Code Reproduction URL

Can be reproduced with the given example

### Additional Information

_No response_

johnjenkins:

GitHub Issue Number: #6411

## What is the new behavior?

Fixes #6411

- [ ] Yes
- [x] No

### DIFF
--- a/src/runtime/proxy-component.ts
+++ b/src/runtime/proxy-component.ts
@@ -378,10 +378,16 @@ export const proxyComponent = (
             return;
           }
 
-          const propDesc = Object.getOwnPropertyDescriptor(prototype, propName);
+          // special handling of boolean attributes. Null (removal) means false.
+          // everything else means true (including an empty string
+          const propFlags = members.find(([m]) => m === propName);
+          if (propFlags && propFlags[1][0] & MEMBER_FLAGS.Boolean) {
+            (newValue as any) = newValue === null ? false : true;
+          }
+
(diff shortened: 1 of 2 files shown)


mpesic:
I can confirm issues from #6411 in Stencil 4.38.0. This fixes handling of boolean values for us.

PR #6429, opened 2025-10-20
chore(test): migrate from Karma to Jest and add CI jobs and mappings

gnbm:

- Standardize on Jest across the repo, remove Karma, and split CI jobs for faster signal and isolation.

GitHub Issue Number: N/A

## What is the new behavior?

- Replaced Karma in test/bundler with Jest:
    - Added `test/bundler/jest-dom-utils.ts` (jsdom loader that executes built module scripts and waits for appload/componentOnReady).
    - Converted `test/bundler/vite-bundle-test/vite-bundle.spec.ts` to Jest and pointed it to the built index.html.
    - Deprecated local Karma files: `test/bundler/karma.config.ts`, `karma-stencil-utils.ts`, and `updated test/bundler/package.json`.
- Added bundler sub-Jest config: `test/bundler/jest.config.js`.
- Added scripts:
    - test.bundler`: run bundler Jest with sub-config.
    - `ci.test`: run full repo Jest + bundler sub-config.

CI:
- New workflow `.github/workflows/ci.yml` with separate jobs:
    - Core Jest (full test suite).
    - Bundler Jest (sub-config).

## Documentation

N/A

- [ ] Yes
- [X] No

## Testing

- Local run:
    - Build as usual, then:
        - Core: `npm run test.jest`
        - Bundler: `npm run test.bundler`

## Other information

- Impact:
    - No runtime/build changes; only tests and CI.
    - Bundler test now runs under Jest; CI can call `npm run ci.test` (or each job separately).

- **Notes:** The core suite is green. Occasional “worker failed to exit gracefully” warnings persist but are benign.

### DIFF
--- a/test/bundler/jest-dom-utils.ts
+++ b/test/bundler/jest-dom-utils.ts
@@ -45,13 +45,27 @@ export function setupDomTests(document: Document): DomTestUtilities {
 
     // wait for app readiness similar to Karma helper
     await new Promise<void>((resolve) => {
+      let resolved = false;
+
       const onAppLoad = () => {
-        window.removeEventListener('appload', onAppLoad);
-        resolve();
+        if (!resolved) {
+          resolved = true;
+          window.removeEventListener('appload', onAppLoad);
+          resolve();
+        }
       };
+
       window.addEventListener('appload', onAppLoad);
-      // if app already loaded synchronously, resolve on next tick
-      setTimeout(() => resolve(), 0);
+
+      // Fallback timeout in case the 'appload' event never fires
+      setTimeout(() => {
+        if (!resolved) {
+          console.warn('appload event did not fire within 5 seconds, continuing anyway');
+          resolved = true;
+          window.removeEventListener('appload', onAppLoad);
+          resolve();
+        }
+      }, 5000);
     });
 
     await allReady();


test/bundler/jest-dom-utils.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    // wait for app readiness similar to Karma helper
+    await new Promise<void>((resolve) => {
+      const onAppLoad = () => {
+        window.removeEventListener('appload', onAppLoad);
+        resolve();
+      };
+      window.addEventListener('appload', onAppLoad);
+      // if app already loaded synchronously, resolve on next tick
+      setTimeout(() => resolve(), 0);
```
johnjenkins: how does this 'know' that the app loaded synchronously here sorry? 
Won't this just resolve every time after a tick?

johnjenkins (commented):
Love the pipeline icons 😄 
Just one question

gnbm:
> just to clarify ... these tests used to run in an actual browser and now they run in ... jsdom(?)
> Is that correct / is that what we want? (I may just be confused 😅 )

Now that I'm thinking better about this, I probably should have migrated to Playwright instead, since with this approach, even though the tests would run faster, we'll lose a lot of things about real browsers. I'll close the PR and look into this from a different perspective. Thank you for the review and for raising all of this, since I still don't have enough context and was just trying to start contributing with some AI help to accelerate the process.

gnbm:
> > Now that I'm thinking better about this, I probably should have migrated to Playwright instead, since with this approach, even though the tests would run faster, we'll lose a lot of things about real browsers. I'll close the PR and look into this from a different perspective. Thank you for the review and for raising all of this, since I still don't have enough context and was just trying to start contributing with some AI help to accelerate the process.
> 
> Absolutely no problem - appreciate you trying to tidy this up! We do use jest with a puppeteer / browser runner ... I'm just not too sure how you'd use (or if you should).

Will explore that option first. Thanks for sharing and for your amazing contribution to this great project

Issue #6235, opened 2025-04-09
bug: Event payload with self-reference causes type generation error


Mitiryl:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/stenciljs/core/blob/main/CONTRIBUTING.md).
- [x] I agree to follow the [Code of Conduct](https://github.com/stenciljs/core/blob/main/CODE_OF_CONDUCT.md).
- [x] I have searched for [existing issues](https://github.com/stenciljs/core/issues) that already report this problem, without success.

### Stencil Version

4.29.2

### Current Behavior

I've encountered an issue with Stencil's type generation when a component emits an event and passes itself as the payload. The generated TypeScript types contain an error that breaks type checking. The generated .d.ts file includes an unresolved type reference to the component, causing a TypeScript error like:

```
[ ERROR ]  TypeScript: src/components.d.ts:24:18
           Cannot find name 'MyComponent'.

     L23:  interface HTMLMyComponentElementEventMap {
     L24:      "myEvent": MyComponent;
     L25:  }

```

### Expected Behavior

Type generation should work correctly even when the event payload is a reference to the component instance itself.

### System Info

```shell
System: node 20.17.0
Stencil: 4.29.2 🔋
TypeScript: 5.5.4
```

### Steps to Reproduce

1. Create a Stencil component like this:
```ts
import { Component, Event, EventEmitter, h } from '@stencil/core';

@Component({
  tag: 'my-component',
  shadow: true,
})
export class MyComponent {
  
  @Event() myEvent: EventEmitter<MyComponent>;

  render() {
    return <div>Hello</div>;
  }
}

```

2. Build the project

### Code Reproduction URL

https://github.com/Mitiryl/stencil-types

christian-bromann:
Thank you for reporting and providing a reproduction case, any contributions that may resolve this bug are much appreciated.

johnjenkins:
hey @Mitiryl,

Can you give me some idea around what you were wanting to achieve here?

```tsx
@Component({
  tag: 'my-component',
  shadow: true,
})
export class MyComponent {
  
  @Event() myEvent: EventEmitter<MyComponent>;

  render() {
    return <div>Hello</div>;
  }
}
```

Practically speaking, What's the desired payload here(?) `this.myEvent.emit( ...what?...)`
`MyComponent` in your example is a class ctor ... is that correct? (if so why?) 

A more common requirement, is to make the calling element part of the event payload .... something like:

```ts
[...]

PR #6430, opened 2025-10-20
FIX {CORE-6235} Allow self components be used as event type

fredppro:

GitHub Issue Number: #6235 

## What is the new behavior?

The generator logic has been updated so that when the event generic type exactly matches the component’s PascalCase name, it is prefixed accordingly with Components. For example:

    interface HTMLMyComponentElementEventMap {
      "myEvent": Components.MyComponent;
    }

## Documentation

N/A

- [ ] Yes
- [x] No

## Testing

Unit & Manual/Visual Testing

## Other information

No more:
```bash
[ ERROR ]  TypeScript: src/components.d.ts:24:18
Cannot find name 'MyComponent'.

L23:  interface HTMLMyComponentElementEventMap {
L24:      "myEvent": MyComponent;
L25:  }
```

### DIFF
--- a/src/compiler/types/generate-event-listener-types.ts
+++ b/src/compiler/types/generate-event-listener-types.ts
@@ -25,7 +25,13 @@ export const generateEventListenerTypes = (
     return { htmlElementEventMap: [], htmlElementEventListenerProperties: [] };
   }
   return {
-    htmlElementEventMap: getHtmlElementEventMap(cmpEvents, typeImportData, cmp.sourceFilePath, htmlElementEventMapName, tagNameAsPascal),
+    htmlElementEventMap: getHtmlElementEventMap(
+      cmpEvents,
+      typeImportData,
+      cmp.sourceFilePath,
+      htmlElementEventMapName,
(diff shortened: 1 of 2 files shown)


fredppro:
@johnjenkins sounds like the pipe is failing in some filesystem op, can u help out?

fredppro:
Just trying to find a solution for the issue mentioned. My take is that,
although that is pratically the same, having to declare the extra `@Element`
for that doesn’t seem much clean to me. Thats why the issue reported was
trying to directly use the shape of the class itself using its own name.

Either way I also agree on your take

fredppro:
yeah yeah, i was editing my response rn.

I think the goal would be to:
1. Allow the consumer to use the class name and resolve that as the shape of the event to be emitted;
2. Mitigates the fact:
2.1 You wont get the "undefined" warnings when using HTMLMyComponent;
2.2 You as a consumer don't need to be aware of the naming convention.


Whether or not this makes sense its up to you. Just tried to help that dude. Feel free to discard the PR 🫡

PR #6435, opened 2025-10-23
fix(jsdoc): Escape special chars

### LINKED ISSUE #6376: bug: Automatically added @default JSDoc to @Prop() declarations - value mistaken for comment end
### Prerequisites

### Stencil Version

4.32.0

### Current Behavior

When a prop default is set to `'*/*'`, it ends up in components.d.ts, where `*/` is seen as the end of a comment, causing a syntax error in file.

For example:

**my-input-file.tsx**
```tsx
  /**
   * Defines specific file types for upload.
   */
  @Prop({ reflect: true })
  accept?: string = '*/*';

```

**components.d.ts**
```tsx
interface AlcInputFile {
    /**
      * Defines specific file types for upload.
      * @default '*/*'
     */
    "accept"?: string;
```

Image

### Expected Behavior

This `/` should be escaped before writing to the file.

### System Info

```shell

```

### Steps to Reproduce

1. Create a @Prop() of type `string` with default value `*/*`
2. Avoid manually write the `@default` tag in a way that Stencil will auto-generate it.
3. `yarn start` or `npm start` 

### Code Reproduction URL

https://github.com/gustavando/stenciljs-auto-jsdoc-default-bug

### Additional Information

A workaround would be manually write `@default` tag, using `\` as escape character.

```
  /**
   * Defines specific file types for upload.
   * @default '*\/*'
   */
  @Prop({ reflect: true })
  accept?: string = '*/*';

```

Also, once `components.d.ts` is generated with an error, StencilJS can't overwrite it — even after applying the workaround. The file must be deleted before running `yarn start` again.

fredppro:

GitHub Issue Number: #6376

## What is the new behavior?

Fixes #6376

- [x] Handle special chars when generating JSDoc blocks, e.g.

```typescript
@Prop({ reflect: true }) accept?: string = '*/*';
```

will output this in `component.d.ts`: 

  ```typescript
  /**
   * @default '*\/*'
   */
  @Prop({ reflect: true })
  accept?: string = '*/*';
  ```
  
where the `*/*` is correctly escaped to avoid errors.

- [ ] Yes
- [x] No

## Testing

- [x] Add tests for `escapeWithPattern` under `helpers.spec.ts`
- [x] Add tests for `getTextDocs` under `utils.spec.ts` AND `generate-prop-types.spec.ts`

### DIFF
--- a/src/compiler/types/tests/generate-prop-types.spec.ts
+++ b/src/compiler/types/tests/generate-prop-types.spec.ts
@@ -245,5 +245,31 @@ describe('generate-prop-types', () => {
 
       expect(actualTypeInfo).toEqual(expectedTypeInfo);
     });
+
+    it('escapes "*/" in default values within jsdoc', () => {
+      const stubImportTypes = stubTypesImportData();
+      const componentMeta = stubComponentCompilerMeta({
+        properties: [
+          stubComponentCompilerProperty({
+            defaultValue: "'*/*'",
+          }),
+        ],
+      });
+
+      const expectedTypeInfo: d.TypeInfo = [
+        {
+          jsdoc: "@default '*\\/*'", // Editor will evaluate this as @default '*\/*'
+          internal: false,
+          name: 'propName',
+          optional: false,
+          required: false,
+          type: 'UserCustomPropType',
+        },
+      ];
+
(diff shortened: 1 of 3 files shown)


johnjenkins (approved):
LGTM! Nice work

PR #6537, opened 2026-01-09
fix(ssr): revert change that causes css animations in nextjs to play twice

OS-jacobbell:

In 4.35.2 (https://github.com/stenciljs/core/pull/6311) there was a regression that causes CSS animations to restart when an SSR page is hydrated in Next.js dev mode using runtime-based SSR.

GitHub Issue Number: N/A

## What is the new behavior?

CSS animations only play once.

- [ ] Yes
- [X] No

## Testing

Manually tested.

### DIFF
--- a/src/runtime/client-hydrate.ts
+++ b/src/runtime/client-hydrate.ts
@@ -255,7 +255,7 @@ export const initializeClientHydrate = (
     });
   }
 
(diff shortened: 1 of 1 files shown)


OS-jacobbell:
Like this on line 273?
`if (node && !node.isConnected) {`

OS-jacobbell:
Let's close this PR and merge yours.

PR #6540, opened 2026-01-09
fix(ssr): `<style>` re-attach causing unwanted animations

johnjenkins:

In 4.35.2 (https://github.com/stenciljs/core/pull/6311) there was a regression that causes CSS animations to restart when an SSR page is hydrated in Next.js dev mode using runtime-based SSR.

GitHub Issue Number: N/A

## What is the new behavior?
CSS animations only play once.

- [ ] Yes
- [x] No


OS-jacobbell (approved):
Looks good and works in my testing!

PR #6573, opened 2026-01-27
fix(runtime): use strict comparison when updating properties

### LINKED ISSUE #6572: bug: Incorrect value update when switching falsy values for a prop typed as any.
### Prerequisites

### Stencil Version

v4.32.0

### Current Behavior

When a component property changes between different falsy values, the DOM may not update, leaving the rendered output out of sync.

### Expected Behavior

The dom should be in sync the currently value of the prop.

### System Info

```shell

```

### Steps to Reproduce

Setup:
A component with a prop as any and and rendering an input with the prop binded to the value
``` 
@Component({
  tag: 'cmp-input-wrapper',
  shadow: true,
})
export class FalsyDemo {
  @State() value: any = 0;

  render() {
    return (
      <Host>
        <input value={this.value} />
      </Host>
    );
  }
}

```
Steps:
- Now set value to an empty string 
- Now set value to 0

### Code Reproduction URL

https://github.com/SaiGovardhana/stencil-starter

### Additional Information

We have a numeric input wrapper component where the value type is defined as any, allowing consumers to pass either a string or a number depending on their use case (for example, to preserve leading zeros). Currently, we are handling this situation when switching between falsey values  by performing preliminary checks and then converting the value to a string with toString().

SaiGovardhana:

Currently, when component properties are updated, the runtime may treat distinct falsy values (like 0 and '') as equivalent due to loose equality checks. This can result in the DOM not reflecting the latest state, causing inputs or other properties to remain out of sync.

GitHub Issue Number: https://github.com/stenciljs/core/issues/6572

## What is the new behavior?

Closes #6572

This PR introduces a strict comparison (!==) when updating element properties in the setAccessor function. Now, all value changes, including transitions between distinct falsy values, are properly propagated to the DOM.

- [ ] Yes
- [X] No (Most Likely)

## Testing

Added a unit test for the numeric input wrapper to verify that the internal <input> correctly updates when the component property changes between distinct falsy values (e.g., 0 -> "").

Verified that the DOM value always reflects the latest property, ensuring transitions between falsy states are handled correctly.

### DIFF
--- a/src/runtime/test/prop.spec.tsx
+++ b/src/runtime/test/prop.spec.tsx
@@ -248,11 +248,9 @@ describe('prop', () => {
   it('updates the internal input when the prop value changes between distinct falsy states', async () => {
     @Component({ tag: 'cmp-input-wrapper' })
     class CmpA {
-      @Prop() value: any
+      @Prop() value: any;
       render() {
-        return (
-          <input id="internal-input" value={this.value}></input>
-        );
+        return <input id="internal-input" value={this.value}></input>;
       }
     }
 
@@ -272,6 +270,5 @@ describe('prop', () => {
     root.value = 0;
     await waitForChanges();
     expect(root.querySelector('#internal-input').value).toBe('0');
-
   });
 });


SaiGovardhana:
> hey @SaiGovardhana - thanks for raising! This change will cause extra renders for `''`, `null` and `undefined`.
> 
> I'm not sure if that's ok(?)

Hi @johnjenkins , Are you talking about stencil web components being re-rendered?

- I believe this branch of code is only run for standard html elements.

- Also the value stored set to element would be an empty string for null and undefined

PR #6576, opened 2026-01-30
feat(runtime): Remove redundant SSR style elements after adoptedStyleSheets adoption

Jagget:
During hydration, shadow DOM components adopt styles via `adoptedStyleSheets`. The SSR-rendered `<style sty-id="...">` elements inside shadow roots become redundant and can be removed.

- Runs only for shadow DOM components during hydration
- Removal occurs after `adoptedStyleSheets` is populated
- Negligible performance impact (~1 querySelector per shadow root)

### DIFF
--- a/src/runtime/test/hydrate-shadow.spec.tsx
+++ b/src/runtime/test/hydrate-shadow.spec.tsx
@@ -2,47 +2,6 @@ import { Component, h } from '@stencil/core';
 import { newSpecPage } from '@stencil/core/testing';
 
 describe('hydrate, shadow', () => {
-  it('shadow component with styles hydrates correctly', async () => {
-    @Component({
-      tag: 'cmp-styled',
-      shadow: true,
-      styles: `
-        :host {
-          color: red;
-        }
-        div {
-          background: blue;
-        }
-      `,
-    })
-    class CmpStyled {
-      render() {
-        return <div>Styled content</div>;
-      }
-    }
-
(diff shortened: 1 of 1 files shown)


Jagget:
@johnjenkins _not removed_ stylesheets doesn't have negative impact, they just duplicate one another and basically useless. For every element browser should read and parse the same styles twice:

image

Jagget:
@johnjenkins But still, there is twice amount of styles. And we don't need 50% of that, because it's the same. But the browser still have to read it and apply it. I think you can count browser's memory consumption as an impact here. Because browser can unload duplicated styles.

Jagget:
Removing styles frees memory, especially with many shadow DOM instances. Imagine, if a page with 100 shadow DOM components, that's 100 redundant <style> nodes... 

I mean, yes, it's not a parsing optimization. It's more of a runtime cleanup.

src/runtime/test/hydrate-shadow.spec.tsx
johnjenkins: I don't think this new test *does* anything regarding the PR?

johnjenkins: Kk. So maybe remove this new one(?) - just adds noise

Jagget: Yeah... make sense :D

johnjenkins (approved):

PR #6594, opened 2026-02-07
feat(dist-custom-elements): new `autoLoader` option

johnjenkins:

GitHub Issue Number: N/A

## What is the new behavior?

Generate an auto-loader script for `dist-custom-elements` that uses MutationObserver to lazily load and define custom elements as they appear in the DOM.

```ts
  namespace: 'AutoLoader',
  outputTargets: [
    {
      type: 'dist-custom-elements',
      autoLoader: true,
      /** // or With options
      * autoLoader: {
      *   fileName: 'my-loader.js',
      *   autoStart: false
      * }
     */
    },
  ],
};
```

## Documentation

TODO

- [ ] Yes
- [x] No

## Testing

Spec and wdio

### DIFF
--- a/test/wdio/.gitignore
+++ b/test/wdio/.gitignore
@@ -6,3 +6,4 @@ www-prerender-script
 www-invisible-prehydration
 test-ts-target-output
 test-components-autoloader
+dist-autoloader-comparison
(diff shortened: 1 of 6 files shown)


theo-staizen:
@johnjenkins are there no docs for this new `autoLoader` config?
https://stenciljs.com/docs/custom-elements

Also what's the difference between that and `customElementsExportBehavior: 'auto-define-custom-elements'`?

PR #6609, opened 2026-02-23
fix: guard attachShadow call for SSR DOMs without native support

### LINKED ISSUE #6605: bug: Hydrate runtime crashes with external SSR DOMs that do not implement `attachShadow` (Angular SSR + Domino)
### Prerequisites

### Stencil Version

4.x (bundled hydrate runtime in `@ionic/core`; exact Stencil build version of shipped bundle not directly exposed)

### Current Behavior

In Stencil hydrate runtime, shadow component hydration can crash on server-side DOM implementations that do not provide native `attachShadow` (e.g. Domino).

The crash occurs in the hydrate path (as seen in runtime output / source-mapped comments): [`proxyHostElement`](https://github.com/stenciljs/core/blob/main/src/hydrate/platform/proxy-host-element.ts) -> [`createShadowRoot`](https://github.com/stenciljs/core/blob/main/src/utils/shadow-root.ts) -> `this.attachShadow(...)`.

When `elm.attachShadow` is missing, SSR throws:

```txt
TypeError: this.attachShadow is not a function
    at createShadowRoot ...
    at proxyHostElement ...
```

Concrete scenario where we reproduce this consistently: Angular SSR (`@angular/ssr`) + Ionic web components (`@ionic/core` hydrate path).

So the repro app is Angular SSR + Ionic, but the failure is in Stencil hydrate internals above.

### Expected Behavior

Stencil hydrate should not throw when running against server-side DOMs where `attachShadow` is unavailable on host elements (e.g. Domino).

Instead, runtime should follow an SSR-safe path (fallback or graceful skip for shadow-root creation) so hydration can continue and complete without crashing the SSR request.

### System Info

```shell
- OS: Windows 11 (also reported by others on non-Windows)
- Node: 22.x
- Angular SSR: 20.3.x (`@angular/ssr`, `CommonEngine`)
- Ionic: works on `8.2.2`, fails on newer 8.x (e.g. `8.7.17`)
- Regression window (observed via shipped `@ionic/core` hydrate behavior): `@ionic/core@8.2.2` works, `@ionic/core@8.7.17` fails with `attachShadow` error.
```

### Steps to Reproduce

1. Clone repro: <https://github.com/SebastianKohler/ionic-angular-ssr-attachShadow-typeError>.
2. Install dependencies: `npm install`.
3. Use failing Ionic version (default in repo, e.g. `8.7.17`).
4. Build SSR bundles: `npm run build:ssr`.
5. Start SSR server: `npm run serve:ssr`.
6. Open <http://localhost:4000/>.
7. Observe server crash with `TypeError: this.attachShadow is not a function` when Ionic components are present.
8. (Verification) Pin `@ionic/angular`, `@ionic/angular-server`, and `@ionic/core` to `8.2.2`, reinstall, repeat steps 4–6: crash disappears. Note: this is true even if the consuming app resolves a newer top-level `@stencil/core` (e.g. `4.43.0`), which indicates the behavior is tied to the hydrate bundle shipped in `@ionic/core`, not the app’s installed `@stencil/core` package version.

### Code Reproduction URL

https://github.com/SebastianKohler/ionic-angular-ssr-attachShadow-typeError

### Additional Information

The bug has also been reported to Ionic: <https://github.com/ionic-team/ionic-framework/issues/29751>

As seen in the comments there, multiple users are affected.

tysoncung:
## Problem

When using Stencil's hydrate runtime with server-side DOM implementations that don't provide `attachShadow` (e.g., Domino used by Angular SSR), the hydration process crashes with:

```
TypeError: this.attachShadow is not a function
    at createShadowRoot ...
    at proxyHostElement ...
```

This affects Angular SSR + Ionic applications where `@angular/ssr` uses Domino as its DOM implementation.

## Fix

Added a `typeof` guard in `createShadowRoot()` (`src/utils/shadow-root.ts`) that checks whether `attachShadow` is available on the host element before calling it. When it's missing (SSR environments), the function returns early, allowing hydration to continue without crashing.

This is a minimal, safe change — in browser environments `attachShadow` is always available so the guard is never hit. In SSR environments, the shadow root simply isn't created, which matches the expected SSR behavior (shadow roots are typically handled during client-side hydration).

## Call path

`proxyHostElement()` → `createShadowRoot.call(elm, cmpMeta)` → `this.attachShadow(opts)` 💥

After fix: `this.attachShadow` checked first → early return if missing → no crash.

Fixes #6605

### DIFF
--- a/src/utils/shadow-root.ts
+++ b/src/utils/shadow-root.ts
@@ -26,6 +26,13 @@ export function createShadowRoot(this: HTMLElement, cmpMeta: d.ComponentRuntimeM
     }
   }
 
+  // Guard against server-side DOM implementations (e.g. Domino) that do not
+  // provide `attachShadow` on host elements.  In those environments we fall
+  // back gracefully so hydration can continue without crashing (#6605).
+  if (typeof this.attachShadow !== 'function') {
+    return;
+  }
+
   const shadowRoot = this.attachShadow(opts);
 
   // Initialize if undefined, set to CSSStyleSheet or null


tysoncung:
Thanks for the feedback! Good point - silently skipping isn't ideal.

What would be the preferred approach here? A few options I can think of:

1. Polyfill `attachShadow` on the server side so components still get a shadow root
2. Use a declarative shadow DOM approach for SSR contexts
3. Something else entirely?

Happy to rework this based on the team's direction.

tysoncung:
Fair point on the silent failure. Closing this - happy to revisit with a better approach.

PR #6640, opened 2026-03-13
fix(mock-doc): updated node.prepend to work even if there are no existing children 

### LINKED ISSUE #6639: bug: globalStyles not prepended to SSR rendered components
### Prerequisites

### Stencil Version

4.43.2

### Current Behavior

The global stylesheet does not get appended to the adoptedStyleSheets property of components during SSR.

### Expected Behavior

Adopted stylesheets include the global styles for SSR components.

### System Info

```shell
System: node 22.14.0
    Platform: darwin (25.3.0)
   CPU Model: Apple M4 Pro (12 cpus)
    Compiler: /Users/*****/Sanity/*****/node_modules/.pnpm/@stencil+core@4.43.2/node_modules/@stencil/core/compiler/stencil.js
       Build: 1772184285
     Stencil: 4.43.2 🌙
  TypeScript: 5.8.3
      Rollup: 4.44.0
      Parse5: 7.2.1
      jQuery: 4.0.0-pre
      Terser: 5.37.0
```

### Steps to Reproduce

Create stencil package with global styles
Create stencil component
Output hydrate package & react package
Install next
Add SSR version of stencil component to page
Observe to global styles not being in the adopted stylesheets list

### Code Reproduction URL

N/A

### Additional Information

I intend to submit a PR to fix this. The fix does not appear to be specific to this issue.

kyleMessnerLillyContractor:

If you call `node.prepend` on a `MockNode` with no children then it will not prepend anything because `firstChild` is `null`

GitHub Issue Number: #6639 

## What is the new behavior?

Fixes #6639

`node.prepend` now always calls `insertBefore` regardless of whether or not `firstChild` is a `MockNode` or `null`. This is fine because the eventual `insertBefore` function accepts `null` values for this parameter. This now also more closely matches the spec for the native `node.prepend`.

- [x] Yes
- [ ] No

If there is code relying on `node.prepend` not inserting anything in elements with no existing children then the code will be broken by this change.

## Testing

`npm run test` passes

### DIFF
--- a/src/compiler/bundle/app-data-plugin.ts
+++ b/src/compiler/bundle/app-data-plugin.ts
@@ -218,18 +218,11 @@ const appendGlobalScripts = (globalScripts: GlobalScript[], s: MagicString) => {
  * @param s the MagicString to append the global styles onto
  * @param platform the platform that is being built
  */
-const appendGlobalStyles = async (
-  buildCtx: d.BuildCtx, 
-  s: MagicString, 
-  platform: BundlePlatform,
-) => {
+const appendGlobalStyles = async (buildCtx: d.BuildCtx, s: MagicString, platform: BundlePlatform) => {
   const { addGlobalStyleToComponents } = buildCtx.config.extras;
-  const shouldIncludeGlobalStyles =  
+  const shouldIncludeGlobalStyles =
     addGlobalStyleToComponents === true || (addGlobalStyleToComponents === 'client' && platform === 'client');
-  const globalStyles =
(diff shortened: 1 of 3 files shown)


kyleMessnerLillyContractor:
> If you want this to go into a non-major version, I think you'll need to change a config option ... suggest [addGlobalStyleToComponents](https://stenciljs.com/docs/config-extras) - it's default is currently `true` < suggest you add a new default to be `client` or similar to retain the current behaviour

Hopefully that is the correct way to extend that option. Let me know if I missed anything. I'll make a PR to update the site docs once this PR is ready to be merged.

PR #6673, opened 2026-04-07
fix(runtime): skip setter when removing reflected bool attr

### LINKED ISSUE #6672: bug: Reflected boolean | undefined prop coerced to false when attribute is removed while prop is undefined (crashes with taskQueue:'immediate')
### Prerequisites

### Stencil Version

4.43.3

### Current Behavior

When a component has a `@Prop({ reflect: true, mutable: true })` of type `boolean | undefined` initialized to undefined, the following sequence crashes or produces incorrect state:

1. The prop starts as undefined (no attribute on the element).
2. The prop is set to true (attribute appears on the element).
3. The prop is set back to undefined (Stencil reflects the removal by calling removeAttribute).

At step 3, attributeChangedCallback fires with newValue === null (the attribute was removed). The callback coerces null → false for boolean props and then calls this[propName] = false even though the prop was already undefined (both meaning "not set"). This triggers the setter reentrant mid-render.

When taskQueue: 'immediate' is used, this reentrant setter corrupts the ongoing vdom patch, resulting in a runtime crash:

`TypeError: Cannot read properties of null (reading 'nodeType')`

### Expected Behavior

When the prop is set back to undefined at step 3, it should not throw any errors.

### System Info

```shell
Ubuntu 22.04.5
Google Chrome

System: node 22.2.0
    Platform: linux (6.8.0-107-generic)
   CPU Model: 11th Gen Intel(R) Core(TM) i5-1135G7 @ 2.40GHz (8 cpus)
    Compiler: /home/myUser/ui/core/compiler/stencil.js
       Build: 1775566197
     Stencil: 4.43.3-dev.1775577995.c845d46 🏃
  TypeScript: 5.8.3
      Rollup: 4.44.0
      Parse5: 7.2.1
      jQuery: 4.0.0-pre
      Terser: 5.37.0
```

### Steps to Reproduce

Using `taskQueue: 'immediate'`, create a component with a `@Prop({ reflect: true, mutable: true })` of type `boolean | undefined` initialized to `undefined`.

Then set it to `true`
Then set it back to `undefined`

### Code Reproduction URL

https://github.com/AlyssonT/reflected-boolean-stencil

### Additional Information

_No response_

AlyssonT:

When a boolean prop with `reflect: true` is `undefined` and Stencil removes the reflected attribute, `attributeChangedCallback` received null and coerced it to false. With `taskQueue: 'immediate'` this fired the setter reentrant mid-render, corrupting the vdom patch and crashing with "Cannot read properties of null (reading 'nodeType')".

Guard against this by skipping the setter when the attribute was removed (newValue was null) and the prop is already undefined. Both represent the same "not set" state for a boolean prop.

fixes: #6672

## What is the new behavior?

Before the boolean coercion, the code checks if the attribute was removed (`newValue === null`) and the prop is already `undefined`. If so, the setter call is skipped — both states mean "not set" for a boolean prop. External attribute removal while the prop holds a real value continues to work correctly.

## Documentation

No documentation changes required.

- [ ] Yes
- [x] No

## Testing

New spec added to `src/runtime/test/attr.spec.tsx` inside `describe('reflect')`. It covers the full cycle: `undefined → true → undefined → true → removeAttribute` — verifying prop value and attribute presence at each step, including that external removal still updates the prop to `false`.

### DIFF
--- a/src/runtime/proxy-component.ts
+++ b/src/runtime/proxy-component.ts
@@ -379,6 +379,7 @@ export const proxyComponent = (
 
           // special handling of boolean attributes. Null (removal) means false.
           // everything else means true (including an empty string
+          const attrWasRemoved = newValue === null;
           const propFlags = members.find(([m]) => m === propName);
           if (propFlags && propFlags[1][0] & MEMBER_FLAGS.Boolean) {
             (newValue as any) = newValue === null || newValue === 'false' ? false : true;
@@ -386,8 +387,15 @@ export const proxyComponent = (
 
           // test whether this property either has no 'getter' or if it does, does it also have a 'setter'
           // before attempting to write back to component props
+          //
+          // Guard: skip when the attribute was removed (attrWasRemoved) but the current prop is
+          // already undefined. Both mean "not set" for a boolean prop, so the assignment would be
+          // spurious. Without this guard the setter fires reentrant mid-render when
+          // taskQueue:'immediate' is used, corrupting the vdom patch and crashing with
+          // "Cannot read properties of null (reading 'nodeType')".
+          const isSpuriousBooleanRemoval = attrWasRemoved && this[propName] === undefined;
           const propDesc = Object.getOwnPropertyDescriptor(prototype, propName);
-          if (newValue != this[propName] && (!propDesc.get || !!propDesc.set)) {
+          if (!isSpuriousBooleanRemoval && newValue != this[propName] && (!propDesc.get || !!propDesc.set)) {
             this[propName] = newValue;
           }
         });
--- a/src/runtime/proxy-component.ts
+++ b/src/runtime/proxy-component.ts
@@ -377,23 +377,24 @@ export const proxyComponent = (
             return;
           }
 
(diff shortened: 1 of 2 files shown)


src/runtime/proxy-component.ts
```diff
@@ -379,15 +379,23 @@ export const proxyComponent = (
 
           // special handling of boolean attributes. Null (removal) means false.
           // everything else means true (including an empty string
+          const attrWasRemoved = newValue === null;
           const propFlags = members.find(([m]) => m === propName);
```
johnjenkins: I think we could make this whole check in one place to make it a bit more explicit?


```suggestion
          const propFlags = members.find(([m]) => m === propName);
          
          const isSpuriousBooleanRemoval = 
            propFlags && 
            (propFlags[1][0] & MEMBER_FLAGS.Boolean) && 
            newValue === null && 
            this[propName] === undefined;
```
wyt?

AlyssonT: absolutely! I simplified the checks to follow this idea.

johnjenkins (approved):

PR #6674, opened 2026-04-07
fix: cache host refs in public render() method

mpesic:

GitHub Issue Number: N/A

- [ ] Yes
- [x] No

## Other information

`render()` method creates a new host ref on every call. This causes the whole DOM subtree to be replaced instead of patching only what's changed. 

This was observed in Storybook for example, if you use `useState()` hooks to manage state. State-driven re-renders cause the whole story to re-render.

This change caches the host refs, so the element state is preserved across re-renders.

### DIFF
--- a/src/runtime/render.ts
+++ b/src/runtime/render.ts
@@ -2,7 +2,19 @@ import type * as d from '../declarations';
 import { renderVdom } from './vdom/vdom-render';
 
 /**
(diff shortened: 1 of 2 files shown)


mpesic:
@johnjenkins wdio tests passing now

johnjenkins (approved):

PR #6676, opened 2026-04-09
fix(runtime): normalize legacy watcher metadata for backward compatibility

gnbm:

When a library (e.g. Ionic Framework) is compiled with Stencil pre-4.39.x
and consumed by an app using a newer Stencil runtime, setting a watched
prop throws:

`TypeError: instance[watchMethodName] is not a function`

This happens because the old compiler emits watcher handlers as string
arrays (`{ "min": ["minChanged"] }`) while the current runtime calls
`Object.entries(watcher)` expecting objects with flags
(`{ "min": [{ "minChanged": 0 }] }`).

GitHub Issue Number: Related to [#30925](https://github.com/ionic-team/ionic-framework/issues/30925) on Ionic Framework

## What is the new behavior?

A `normalizeWatchers()` helper converts legacy string-array format to the
current object-with-flags format. It is applied at all five `$watchers$`
assignment sites:

- `bootstrap-custom-element.ts` (custom elements)
- `bootstrap-lazy.ts` (lazy bundles)
- `initialize-component.ts` (lazy initialization)
- `proxy-component.ts` (defensive fallback)
- `proxy-host-element.ts` (hydrate/SSR)

Components compiled with either format now work without errors.

The helper returns `undefined` (not `{}`) for falsy or empty input, so
that `cmpMeta.$watchers$` stays falsy for components without `@Watch()`
decorators. This avoids unnecessary per-prop overhead in `setValue` where
the truthiness of `$watchers$` gates watcher dispatch.

- [ ] Yes
- [X] No

## Testing

- 8 unit tests for `normalizeWatchers()` covering undefined, empty,
  legacy, modern, mixed, and multi-prop/multi-watcher inputs (including
  verifying `undefined` is returned for empty/falsy input)
- 1 integration test simulating a component compiled with the legacy
  pre-4.39.x watcher format, verifying callbacks fire correctly


src/runtime/bootstrap-lazy.ts
```diff
@@ -88,7 +89,7 @@ export const bootstrapLazy = (lazyBundles: d.LazyBundlesRuntimeData, options: d.
         cmpMeta.$attrsToReflect$ = [];
       }
       if (BUILD.propChangeCallback) {
-        cmpMeta.$watchers$ = compactMeta[4] ?? {};
+        cmpMeta.$watchers$ = normalizeWatchers(compactMeta[4]);
```
johnjenkins, line 97: I know this is intentional, but I think it could use a comment explaining why `$watchers$` is treated differently?

src/hydrate/platform/proxy-host-element.ts
```diff
@@ -1,12 +1,12 @@
...
-import { getValue, parsePropertyValue, setValue } from '@runtime';
+import { getValue, normalizeWatchers, parsePropertyValue, setValue } from '@runtime';
 import { CMP_FLAGS, createShadowRoot, MEMBER_FLAGS } from '@utils';
 
 import type * as d from '../../declarations';
 
 export function proxyHostElement(elm: d.HostElement, cstr: d.ComponentConstructor): void {
   const cmpMeta = cstr.cmpMeta;
-  cmpMeta.$watchers$ = cmpMeta.$watchers$ || cstr.watchers;
+  cmpMeta.$watchers$ = cmpMeta.$watchers$ || normalizeWatchers(cstr.watchers);
```
johnjenkins: I don't think this is necessary? 
In-order to to be part of a `hydrate-script` components are necessarily bundled with the same stencil/core version, so there can't be a core mismatch

src/runtime/set-value.ts
```diff
@@ -8,6 +8,56 @@ import { scheduleUpdate } from './update-component';
...
+ * indices as method names, causing:
+ *   `TypeError: instance[watchMethodName] is not a function`
+ *
+ * This helper is called once at each of the five `$watchers$` assignment sites so
+ * that all downstream runtime code can safely assume the new object format.
+ *
+ * @param raw The raw watcher map from compiled metadata (new or legacy format).
+ * @returns A normalized watcher map in the `{ [methodName]: flags }[]` format, or `undefined` if `raw` is `undefined` or empty.
+ */
+export const normalizeWatchers = (
```
johnjenkins: I think I'd prefer it if this were in it's own module (much like `parse-property-value.ts`(?) Esp as it's imported on-it's-own in multiple other places

src/hydrate/platform/proxy-host-element.ts
gnbm: That makes sense. Thank you!

src/runtime/bootstrap-lazy.ts
gnbm: Totally agree! Thank you

src/runtime/test/normalize-watchers.spec.ts
```diff
@@ -0,0 +1,52 @@
+import { normalizeWatchers } from '../normalize-watchers';
+
+describe('normalizeWatchers', () => {
+  it('returns undefined when input is undefined', () => {
```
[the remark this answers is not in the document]
gnbm: @copilot apply changes based on [this feedback](https://github.com/stenciljs/core/pull/6676#discussion_r3064665545)

johnjenkins (approved):
LGTM!

PR #6728, opened 2026-05-20
fix(compiler): resolve @Prop variable defaults to literal values in docs output

### LINKED ISSUE #6727: bug: generatedproperty documentation does not resolve variables set as the default value
### Prerequisites

### Stencil Version

4.43.4

### Current Behavior

Stencil generates documentation for all Ionic Framework components under the [@ionic/docs](https://github.com/ionic-team/ionic-framework/tree/main/packages/docs) package.

The [JSON file](https://cdn.jsdelivr.net/npm/@ionic/docs) that is published contains all of the information that is used by the [component documentation](https://ionicframework.com/docs/components).

The issue:

When a property’s default value is set to a variable, such as when on [Split Pane](https://github.com/ionic-team/ionic-framework/blob/c88c0de3ade92469fa1f37e1b8220911adf113e0/core/src/components/split-pane/split-pane.tsx#L53), Stencil generates it as this:

"default": "QUERY['lg']",
Which renders in the [documentation](https://ionicframework.com/docs/api/split-pane#prop-when) as:

CleanShot 2026-05-19 at 11.00.38@2x-20260519-150041.png

### Expected Behavior

The default value should grab the value from a variable and render that value.

For example, in Split Pane, QUERY['lg'] would render as '(min-width: 992px)'

### System Info

```shell

```

### Steps to Reproduce

Generate Ionic Framework component documentation

### Code Reproduction URL

https://github.com/ionic-team/ionic-framework/blob/c88c0de3ade92469fa1f37e1b8220911adf113e0/core/src/components/split-pane/split-pane.tsx#L53

### Additional Information

_No response_

gnbm:
Closes #6727

Stencil emits the source text of a `@Prop` initializer verbatim into the generated docs JSON (`defaultValue` field). When the initializer is a reference to a const variable or an object-property access, the docs end up showing the variable expression rather than the actual default value.

Example from `ion-split-pane` (Ionic Framework):

```ts
const QUERY: { [key: string]: string } = { lg: '(min-width: 992px)', ... };

@Component({ tag: 'ion-split-pane' })
export class SplitPane {
  @Prop() when: string | boolean = QUERY['lg'];
}
```

Today the generated docs JSON contains:

```json
"default": "QUERY['lg']"
```

…which is what users see in https://ionicframework.com/docs/api/split-pane#prop-when.

GitHub Issue Number: #6727

## What is the new behavior?

The `@Prop` decorator parser now follows the initializer expression to its underlying primitive literal when it is safe to do so. With this fix, the same component produces:

```json
"default": "'(min-width: 992px)'"
```

Shapes that are now resolved:

- `IDENT` where `IDENT` is a `const` variable whose initializer chain ends in a primitive literal.
- `OBJ.key` and `OBJ['key']` (or `OBJ[0]`) against `const` object literals.
- Object shorthand properties (`const label = 'x'; const CONFIG = { label }; @Prop() v = CONFIG.label;`) — resolved via `getShorthandAssignmentValueSymbol`.
- Cross-file imported `const`s (`import { QUERY } from './queries';`) — alias symbols are unwrapped via `getAliasedSymbol` before walking declarations.
- Value-preserving wrapper expressions are unwrapped: `as` (`'x' as const`), `<T>x` type assertions, parenthesized expressions, `satisfies`, and non-null `!`.
- Wrapped object-literal `const`s (`const CONFIG = ({ label: 'x' } as const)`) — the object side of `OBJ.key` is also unwrapped.
- Chained `const`-to-`const` object aliases (`const A = CONFIG; @Prop() v = A.label;`) — bounded by `MAX_RESOLVE_DEPTH`.
- Signed numeric literals (`-1`, `+2`) — TypeScript represents these as `PrefixUnaryExpression`, now recognised as primitives.
- `const X = undefined; @Prop() v = X;` resolves to the literal `undefined`.

Shapes that intentionally still fall back to the original `getText()` output (preserving prior behavior):

- Nested property access (`OBJ.a.b`).
- Template literals with substitutions.
- Computed / dynamic indexed access (`OBJ[varKey]`).
- Function calls, arithmetic, ternaries, etc.
- Enum members (`SomeEnum.VALUE`) — out of scope to avoid silently changing existing docs strings.
- `let` / `var` bindings — never followed because they may be reassigned.

Safety properties:

- Only primitive literals (`string` / `number` / signed number / `true` / `false` / `null` / `undefined`) are emitted. Object and array literals are never serialized into `defaultValue`.
- Recursion is bounded (`MAX_RESOLVE_DEPTH = 5`) for both the literal walk and the object-literal walk.
- Source quote style is preserved (everything still flows through `getText()` at the leaf).
- Any expression that can't be resolved falls back to the previous `getText()` output via `??`, so no existing case regresses.

## Documentation

N/A — internal compiler change. The Gallery component's hardcoded `columns` / `gap` description defaults in `ionic-team/ionic-docs` can be dropped once Ionic picks up this Stencil release; that follow-up belongs in the docs repo.

- [ ] Yes
- [x] No

The change only widens the set of expressions that are resolved into a literal. Every expression shape the previous code already produced output for produces the same output now.

## Testing

Fifteen new unit tests in `src/compiler/transformers/test/parse-props.spec.ts` covering:

- `const` string / number variable resolution.
- `OBJ.key` and `OBJ['key']` resolution — the latter uses the verbatim `ion-split-pane` shape (FW-7298 reproduction).
- Object shorthand property resolution (`{ label }`).
- Cross-file imported `const` resolution — self-contained 2-file `ts.Program` inline in the spec.
- `as const`, parenthesized + non-null wrapper unwrapping.
- Wrapped object-literal `const` accessed via property access.
- Chained `const`-to-`const` object alias resolution.
- Signed numeric literal (`-1`).
- `const X = undefined; @Prop() v = X;` chain.
- Function-call initializer / dynamic-key indexed access — must fall back to raw text.
- Direct `@Prop() x = undefined;` — preserved as raw text.

```
npm run test.jest -- src/compiler/transformers/test/parse-props.spec.ts
# Tests: 40 passed, 40 total
```

Full transformers suite: **325 passed**, zero regressions.

## Other information

- The fix is localized to `src/compiler/transformers/decorators-to-static/prop-decorator.ts` and the corresponding test file; no other files are touched.

### DIFF
--- a/src/compiler/transformers/decorators-to-static/prop-decorator.ts
+++ b/src/compiler/transformers/decorators-to-static/prop-decorator.ts
@@ -411,9 +411,6 @@ const resolveLiteralText = (node: ts.Expression, typeChecker: ts.TypeChecker, de
 
   // Identifier referencing a `const` variable with a resolvable initializer.
   if (ts.isIdentifier(node)) {
-    if (node.text === 'undefined') {
-      return 'undefined';
-    }
     const init = getConstVariableInitializer(node, typeChecker);
     return init ? resolveLiteralText(init, typeChecker, depth + 1) : undefined;
   }
@@ -451,7 +448,10 @@ const isPrimitiveLiteral = (node: ts.Expression): boolean => {
     ts.isNumericLiteral(node) ||
     node.kind === ts.SyntaxKind.TrueKeyword ||
     node.kind === ts.SyntaxKind.FalseKeyword ||
-    node.kind === ts.SyntaxKind.NullKeyword
+    node.kind === ts.SyntaxKind.NullKeyword ||
+    // `undefined` is an Identifier referencing the global, not a syntax keyword,
+    // but it is treated as a primitive literal value here.
+    (ts.isIdentifier(node) && node.text === 'undefined')
   );
 };
 
@@ -461,10 +461,7 @@ const isPrimitiveLiteral = (node: ts.Expression): boolean => {
  * Only `const` declarations are followed because `let` / `var` bindings may be
  * reassigned and so are not safe to inline at compile time.
  */
-const getConstVariableInitializer = (
-  node: ts.Identifier,
-  typeChecker: ts.TypeChecker,
-): ts.Expression | undefined => {
+const getConstVariableInitializer = (node: ts.Identifier, typeChecker: ts.TypeChecker): ts.Expression | undefined => {
   const symbol = typeChecker.getSymbolAtLocation(node);
   const decl = symbol?.declarations?.find(ts.isVariableDeclaration);
   if (!decl || !decl.initializer) {
@@ -498,10 +495,7 @@ const resolveObjectLiteral = (
   return undefined;
 };
 
-const findObjectLiteralMember = (
-  obj: ts.ObjectLiteralExpression,
-  name: string,
-): ts.Expression | undefined => {
+const findObjectLiteralMember = (obj: ts.ObjectLiteralExpression, name: string): ts.Expression | undefined => {
   for (const member of obj.properties) {
     if (!ts.isPropertyAssignment(member)) {
       continue;
--- a/src/compiler/transformers/decorators-to-static/prop-decorator.ts
+++ b/src/compiler/transformers/decorators-to-static/prop-decorator.ts
@@ -152,7 +152,7 @@ const parsePropDecorator = (
 
   // extract default value
   if (ts.isPropertyDeclaration(prop) && prop.initializer) {
-    propMeta.defaultValue = prop.initializer.getText();
+    propMeta.defaultValue = resolveInitializerText(prop.initializer, typeChecker);
   } else if (ts.isGetAccessorDeclaration(prop)) {
     // shallow comb to find default value for a getter
     const returnStatement = prop.body?.statements.find((st) => ts.isReturnStatement(st)) as ts.ReturnStatement;
@@ -166,7 +166,7 @@ const parsePropDecorator = (
       const foundProp = findGetProp(nameToFind, newMembers);
 
       if (foundProp && foundProp.initializer) {
-        propMeta.defaultValue = foundProp.initializer.getText();
+        propMeta.defaultValue = resolveInitializerText(foundProp.initializer, typeChecker);
 
         if (propMeta.type === 'unknown') {
           const type = typeChecker.getTypeAtLocation(foundProp);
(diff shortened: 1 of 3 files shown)


gnbm:
> hey @gnbm thanks for raising!
> 
> Can you work through / resolve the points raised by copilot - than I'll do a proper review :)
> 
> Also nit - but can we not refer to internal tickets (`FW-*`) as Stencil is an open-source project and we have no way of seeing OutSystem's internal Jira / tickets for context 🙏

Hey @johnjenkins 
I think I addressed them all. Thank you for the heads-up

src/compiler/transformers/decorators-to-static/prop-decorator.ts
```diff
@@ -379,3 +379,226 @@ const findSetter = (propName: string, members: ts.ClassElement[]): ts.SetAccesso
...
+    // so use `getShorthandAssignmentValueSymbol` to reach the original binding
+    // and then walk to its `const` initializer.
+    if (ts.isShorthandPropertyAssignment(member) && member.name.text === name) {
+      return resolveConstSymbolInitializer(typeChecker.getShorthandAssignmentValueSymbol(member), typeChecker);
+    }
+  }
+  return undefined;
+};
+
+const getPropertyNameText = (name: ts.PropertyName): string | undefined => {
```
johnjenkins, line 611: tiny nit :) 
add a comment explaining computed property names are unsupported?

src/compiler/transformers/decorators-to-static/prop-decorator.ts
```diff
@@ -379,3 +379,226 @@ const findSetter = (propName: string, members: ts.ClassElement[]): ts.SetAccesso
...
+
+/**
+ * Maximum number of hops to follow when traversing variable references to
+ * resolve an initializer to its literal value. With the guards below
+ * (`depth > MAX_RESOLVE_DEPTH`), a chain of up to this many `const` / property
+ * follows is supported before bailing out; any deeper chain falls back to the
+ * original source text. Prevents pathological / cyclic chains from blowing the
+ * stack.
+ */
+const MAX_RESOLVE_DEPTH = 5;
```
johnjenkins, line 391: tiny nit ... 
it'd be good to find a way we can test `MAX_RESOLVE_DEPTH + 1`

i.e. 
`it("falls back to getText() at a chain depth over 5")`

johnjenkins (approved):
just some tiny things. Don't worry about them if you don't have time :)

src/compiler/transformers/decorators-to-static/prop-decorator.ts
gnbm: Tried to do that (with Claude's help..lol)..What do you think?

src/compiler/transformers/decorators-to-static/prop-decorator.ts
gnbm: Well pointed..as always ;) 
Done!

src/compiler/transformers/decorators-to-static/prop-decorator.ts
johnjenkins: haha - perfect!

johnjenkins (approved):

OS-jacobbell:
Hi @gnbm, this works for ion-split-pane's `when` and ion-gallery's `gap`, but still leaves ion-gallery's `columns` as `DEFAULT_COLUMNS`. Here's the before and after generated docs:
[ionic-docs-diff.tar.gz](https://github.com/user-attachments/files/28109457/ionic-docs-diff.tar.gz)

PR #6730, opened 2026-05-21
chore(deps): update typescript to 5.9

OS-jacobbell:

GitHub Issue Number: N/A

Stencil is on typescript 5.8 for compatibility with Ionic Framework v8. Typescript 5.9 defines `autocorrect: boolean` on `HTMLElement`, which would be a breaking change for Ionic Framework.

## What is the new behavior?

With Ionic Framework v9 in development, we can update Stencil to typescript 5.9. Ionic Framework will need to update the type of `autocorrect` properties for components that use them.

- [X] Yes
- [ ] No

Components that define `autocorrect` with different types than the standard JavaScript `autocorrect` (boolean) will need to update the type to `boolean`.

## Testing

Ran `npm run test` from repo root.


johnjenkins (approved):

gnbm (approved):
LGTM

OS-jacobbell:
Updating Typescript in Stencil v4 could be an issue for older versions of Ionic Framework that would auto-update, and since the v5 branch already has typescript updated, I'm closing this PR.

PR #6772, opened 2026-07-14
fix(runtime): retry lazy component load after a failed dynamic import

### LINKED ISSUE #6771: bug: Constructor for "<tag>#undefined" was not found — component permanently bricked after a failed lazy-chunk load
### Prerequisites

- [x] I have read the Contributing Guidelines.
- [x] I agree to follow the Code of Conduct.
- [x] I have searched for existing issues that already report this problem, without success (closest related: #3281 "Dynamic import / lazyloading randomly not working in production", which was closed as inconclusive/not reproducible — this report includes a deterministic, scriptable reproduction).

### Stencil Version

4.43.5

### Current Behavior

When the dynamic `import()` of a lazy component's `*.entry.js` chunk fails (network drop, request blocked, flaky mobile connection, stale/evicted CDN entry, ad-blocker, etc.), the custom element is **permanently and silently bricked** for the remainder of that page's lifetime:

1. `initializeComponent()` sets `HOST_FLAGS.hasInitializedComponent` on `hostRef.$flags$` **before** attempting to load the module ([`initialize-component.ts#L34-L36`](https://github.com/stenciljs/core/blob/06258c870fb58bfa59eec1473a9a7048249341ed/src/runtime/initialize-component.ts#L34-L36)):
   ```ts
   if ((hostRef.$flags$ & HOST_FLAGS.hasInitializedComponent) === 0) {
     hostRef.$flags$ |= HOST_FLAGS.hasInitializedComponent;
     const bundleId = cmpMeta.$lazyBundleId$;
     if (BUILD.lazyLoad && bundleId) {
       const CstrImport = loadModule(cmpMeta, hostRef, hmrVersionId);
   ```
2. If the import fails, `loadModule()`'s `.then(onFulfilled, onRejected)` rejection handler ([`client-load-module.ts`](https://github.com/stenciljs/core/blob/06258c870fb58bfa59eec1473a9a7048249341ed/src/client/client-load-module.ts)) logs the error but returns nothing, which **converts the rejected promise into a resolved promise with value `undefined`**:
   ```ts
   return import(`${MODULE_IMPORT_PREFIX}${bundleId}.entry.js...`).then(
     (importedModule) => { ...; return importedModule[exportName]; },
     (e: Error) => { consoleError(e, hostRef.$hostElement$); /* implicit return undefined */ },
   );
   ```
3. Back in `initializeComponent`, `Cstr` is `undefined`, which throws `Constructor for "<tag>#undefined" was not found` ([`initialize-component.ts#L56`](https://github.com/stenciljs/core/blob/06258c870fb58bfa59eec1473a9a7048249341ed/src/runtime/initialize-component.ts#L56)). This is caught by the outer `try/catch` ([L201](https://github.com/stenciljs/core/blob/06258c870fb58bfa59eec1473a9a7048249341ed/src/runtime/initialize-component.ts#L201)), which logs the error and resolves `$onReadyResolve$`/`$onRenderResolve$` — but **never clears `HOST_FLAGS.hasInitializedComponent`**.
4. Because `hostRef` (and its `$flags$`) is created exactly once per DOM node, in `registerHost()` ([`client-host-ref.ts`](https://github.com/stenciljs/core/blob/06258c870fb58bfa59eec1473a9a7048249341ed/src/client/client-host-ref.ts)), called from the custom element's constructor, the stuck flag lives for the lifetime of that DOM node. `connected-callback.ts` gates re-initialization on a *different* flag, `HOST_FLAGS.hasConnected`, which is also set unconditionally on first connect — so even disconnecting/reconnecting the *same* node never re-triggers `initializeComponent()`/`loadModule()`.
5. **Compounding factor confirmed via reproduction below**: even a **brand-new** element instance of the same tag, created later in the same document after the network has fully recovered, also fails — with **zero new network requests** reaching the server. The browser's per-document ES module map appears to cache the failed dynamic-import specifier and short-circuits any subsequent `import()` of that exact URL, so Stencil never even gets a chance to retry, regardless of `cmpModules` state.

Net result: a single transient chunk-load failure (very plausible on mobile networks, per the linked issue #3281 which reports this happening "randomly" in production) permanently breaks that tag for the rest of the page's life — no rendered content, no lifecycle callbacks, no automatic recovery — until a full page reload creates a fresh document/module map and fresh `hostRef`s.

### Expected Behavior

Either:
- On a load failure, the runtime resets enough internal state (`HOST_FLAGS.hasInitializedComponent`, and ideally a mechanism to avoid the browser-level module-map caching, e.g. a cache-busting retry parameter) so that a subsequent `connectedCallback` — or an explicit public retry API — can re-attempt loading the component; or at minimum
- The failure mode is surfaced as a rejected `componentOnReady()` promise (rather than silently resolved) so application code can reliably detect and react to it (e.g. render a fallback, retry, or reload) without having to string-match `setErrorHandler` error messages.

### System Info

```
Stencil: 4.43.5
Node: v24.18.0
npm: 11.16.0
OS: macOS (Darwin)
Browser: Chromium (Playwright)
```

### Steps to Reproduce

This is a minimal, fully scriptable reproduction — no manual DevTools "block request" step needed (though blocking the chunk URL in DevTools' Network tab reproduces the identical error and was how this was first discovered).

**1. Minimal component**, `src/components/web-button/web-button.tsx`:
```tsx

@Component({ tag: 'web-button', shadow: true })
export class WebButton {
  render() {
    return (
      <button>
        <slot />
      </button>
    );
  }
}
```

**2. `stencil.config.ts`**:
```ts
import { Config } from '@stencil/core';

  namespace: 'repro',
  outputTargets: [{ type: 'www', serviceWorker: null }],
};
```

**3. `src/index.html`**:
```html
<!DOCTYPE html>
<html>
<head>
  <script type="module" src="/build/repro.esm.js"></script>
</head>
<body>
  <web-button>click me</web-button>
</body>
</html>
```

**4. Build**: `npx stencil build` → note the generated entry chunk filename in `www/build/*.entry.js` (e.g. `p-5e3b1241.entry.js` — the hash is deterministic per build).

**5. Serve with a tiny server that fails exactly that one request** (simulates a blocked/dropped mobile request — toggle via presence of a `.block-chunk` file so the same server can demonstrate both the failure and the "recovery" attempt):
```js
// server.js
const http = require('http');
const fs = require('fs');
const path = require('path');
const ROOT = path.join(__dirname, 'www');
const BLOCK_FLAG = path.join(__dirname, '.block-chunk');

http.createServer((req, res) => {
  if (req.url.includes('ENTRY_CHUNK_FILENAME.entry.js') && fs.existsSync(BLOCK_FLAG)) {
    req.socket.destroy(); // simulate a dropped/blocked request, not a clean 404
    return;
  }
  const filePath = path.join(ROOT, req.url === '/' ? 'index.html' : req.url.split('?')[0]);
  fs.readFile(filePath, (err, data) => {
    if (err) return res.writeHead(404).end();
    res.writeHead(200).end(data);
  });
}).listen(8991);
```

**6. Reproduction sequence** (observed with Playwright, but any browser + DevTools works identically):
```
touch .block-chunk && node server.js &
# open http://localhost:8991/  ->  console shows:
#   net::ERR_EMPTY_RESPONSE  @ /build/p-5e3b1241.entry.js
#   TypeError: Failed to fetch dynamically imported module: .../p-5e3b1241.entry.js
#   Error: Constructor for "web-button#undefined" was not found
# <web-button> now has an empty (but present) shadow root — permanently un-upgraded.

rm .block-chunk   # network "recovers"

# (a) reconnect the SAME node (remove + re-append to DOM) -> still broken, 0 new network requests
# (b) create a BRAND NEW <web-button> and append it       -> still broken, 0 new network requests
# (c) full page reload (F5)                                 -> works correctly, renders as expected
```

### Actual results observed (verified with Playwright against the exact repro above)

```
Console (during initial blocked load):
[ERROR] Failed to load resource: net::ERR_EMPTY_RESPONSE @ http://localhost:8991/build/p-5e3b1241.entry.js
[ERROR] TypeError: Failed to fetch dynamically imported module: http://localhost:8991/build/p-5e3b1241.entry.js
[ERROR] Error: Constructor for "web-button#undefined" was not found
    at http://localhost:8991/build/p-DKJupDqi.js:2:114

DOM state after failure:
<web-button>click me</web-button>          <-- outerHTML, never gets `class="hydrated"`
shadowRoot.innerHTML === ""                <-- shadow root exists (created in ctor) but render() never ran

After removing the block (network healthy again), same page, no reload:
- Re-appending the SAME <web-button> node: still `shadowRoot.innerHTML === ""`, 0 new HTTP requests for the chunk.
- Appending a BRAND NEW <web-button> node: still `shadowRoot.innerHTML === ""`, 0 new HTTP requests for the chunk.
  (Network requests list shows exactly ONE request for p-5e3b1241.entry.js for the whole page session, status FAILED.)

Full page reload (network healthy):
<web-button class="hydrated">click me</web-button>
shadowRoot.innerHTML === "<button><slot></slot></button>"   <-- works correctly
```

### Code Reproduction URL

N/A — full minimal reproduction (component, config, index.html, and the network-failure-simulating server) is inlined above; happy to publish as a public repo if that's preferred for triage.

### Additional Information

Root cause, precisely:

- [`src/runtime/initialize-component.ts#L34-L56`](https://github.com/stenciljs/core/blob/06258c870fb58bfa59eec1473a9a7048249341ed/src/runtime/initialize-component.ts#L34-L56) — `HOST_FLAGS.hasInitializedComponent` is set before the async load and never reset on failure.
- [`src/runtime/connected-callback.ts`](https://github.com/stenciljs/core/blob/06258c870fb58bfa59eec1473a9a7048249341ed/src/runtime/connected-callback.ts) — `HOST_FLAGS.hasConnected` gates re-entry into the init path and is also set unconditionally on first connect, so reconnecting the same node can't recover either.
- [`src/client/client-load-module.ts`](https://github.com/stenciljs/core/blob/06258c870fb58bfa59eec1473a9a7048249341ed/src/client/client-load-module.ts) — the `.then(onFulfilled, onRejected)` pattern silently converts a rejected dynamic `import()` into a resolved `undefined`, so the failure never propagates as a rejection to callers of `componentOnReady()`.

Suggested directions for a fix (happy to submit a PR if a direction is confirmed):
1. On the `loadModule` failure path, reset `HOST_FLAGS.hasInitializedComponent` (and don't let `HOST_FLAGS.hasConnected` alone block a future retry) so a later `connectedCallback` — or an explicit `retry`/`reload` API — can re-attempt the load.
2. Let the failure reject `hostRef.$onReadyPromise$` (i.e. `componentOnReady()`) instead of resolving it, so app code has a first-class, promise-based way to detect and react to a failed load, rather than string-matching messages via `setErrorHandler`.
3. Consider a cache-busting query param on retry attempts, since testing shows the browser's own module map caches the failed specifier and won't even re-issue the network request for subsequent `import()` calls of the same URL within the page — meaning even a manual app-level retry strategy cannot succeed without changing the specifier.
4. At minimum, document this as a known limitation (currently a single transient network failure during lazy-load is an unrecoverable, silent dead end for that tag, for the rest of the page's life).

This is very plausibly the underlying mechanism behind #3281 ("Dynamic import / lazyloading randomly not working in production") and similar production reports — those were closed as inconclusive for lack of a reliable local reproduction, which this report now provides.

yasinkocak:

GitHub Issue Number: #6771

A host element is permanently "bricked" for the rest of the page's lifetime if its lazy bundle's dynamic `import()` fails a single time (dropped network request, flaky mobile connection, ad-blocker, stale CDN entry, etc.) — no rendered content, no lifecycle callbacks, and no recovery, not even for a brand-new element of the same tag, until a full page reload.

Root cause:
1. `HOST_FLAGS.hasInitializedComponent` is set before the load attempt in [`initialize-component.ts`](https://github.com/stenciljs/core/blob/main/src/runtime/initialize-component.ts) and never cleared on failure.
2. `HOST_FLAGS.hasConnected` (set unconditionally on first connect) makes [`connected-callback.ts`](https://github.com/stenciljs/core/blob/main/src/runtime/connected-callback.ts) skip re-initialization on every subsequent reconnect.
3. The browser's own per-document module map caches the failed `import()` specifier, so even a bare retry of the exact same bundle URL never reaches the network again — confirmed in a live-browser repro (see Testing below).

## What is the new behavior?

Fixes #6771

- On a failed lazy load, `HOST_FLAGS.hasInitializedComponent` is cleared instead of staying stuck.
- `connected-callback.ts` retries `initializeComponent()` the next time the host element is reconnected (disconnect + reconnect), instead of only handling the "already has a `$lazyInstance$`" case.
- `client-load-module.ts` appends a cache-busting query param (`?s-retry=N`) on retries, the same pattern already used for `?s-hmr=`, so the retry actually reaches the network instead of being short-circuited by the browser's module map.

No public API changes. No behavior change on the happy path — a retry only ever happens after a load has actually failed.

## Documentation

N/A — this is an internal runtime resilience fix, no public API or documented behavior changes.

- [ ] Yes
- [x] No

## Testing

- Added `src/runtime/test/lazy-load-retry.spec.tsx` (2 new tests): verifies `hasInitializedComponent` is cleared on a failed load, and that a disconnect+reconnect after a failure re-attempts initialization and succeeds once the module becomes available.
- Full existing suite (`npm run test.jest -- --runInBand`, 64 suites / 651 tests across `src/runtime` + `src/client`) passes with no regressions.
- End-to-end verified against a real browser (Playwright + a minimal Stencil app + a tiny Node server that drops the `*.entry.js` request to simulate a blocked/flaky network request):
  - Initial load blocked → `Constructor for "web-button#undefined" was not found`, element stays un-hydrated (empty shadow root).
  - Network "recovers", element disconnected + reconnected:
    - **Before this fix**: still un-hydrated, zero new network requests (browser module-map caching of the failed specifier).
    - **After this fix**: `GET /build/p-xxx.entry.js?s-retry=1 → 200`, element renders correctly (`class="hydrated"`, populated shadow DOM) — all without a page reload.

## Other information

Per CONTRIBUTING.md, issue #6771 was filed first with a full reproducible write-up before starting this PR.


moessie:
Thanks for working on this! I tested the proposed fix with a minimal reproduction by blocking the initial lazy-loaded *.entry.js request and then reconnecting the component after restoring the chunk.

With this change, the component successfully retries and renders instead of remaining stuck. This looks like a solid improvement for the recovery behavior

t-model00291182:
We are running into the same issue in our Stencil-based component library.

A consumer recently reported this behavior to us, and after debugging we traced it back to this issue. This confirms that the problem is not only reproducible in isolated cases but can also affect real-world consumers of Stencil components.

We really appreciate the work the Stencil team is doing. A fix for this would have a meaningful impact for us and other teams building component libraries on top of Stencil.

Thanks for looking into this!

src/client/client-load-module.ts
```diff
@@ -5,6 +5,16 @@ import { consoleDevError, consoleError } from './client-log';
...
 export const cmpModules = /*@__PURE__*/ new Map<string, { [exportName: string]: d.ComponentConstructor }>();
 
+/**
+ * Tracks how many times the dynamic `import()` for a given lazy bundle has
+ * failed. A failed dynamic import is cached by the browser's own module map
+ * for the resolved specifier URL, so a bare retry of the exact same URL
+ * would be rejected immediately without ever reaching the network again.
+ * This lets a retry (see `connected-callback.ts`) bust that cache with a
+ * query param, the same way `hmrVersionId` already does for HMR.
+ */
```
johnjenkins, line 11: quite verbose :) 

```suggestion
/**
 * Tracks how many times a dynamic `import()` for a given lazy bundle has failed.
 * Used to cache bust the retry attempt in connected-callback.ts
 */
```

src/runtime/test/lazy-load-retry.spec.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { LazyBundlesRuntimeData } from '../../internal';
+import { HOST_FLAGS } from '../../utils';
+import { bootstrapLazy } from '../bootstrap-lazy';
+
+/**
+ * Regression tests for a bug where a host element would be permanently
+ * "bricked" (never rendered, no lifecycle callbacks) if its lazy bundle
+ * failed to load a single time (e.g. a dropped network request). See:
+ * https://github.com/stenciljs/core/issues/6771
+ */
```
johnjenkins: unnecessary :)

```suggestion
```

src/runtime/connected-callback.ts
```diff
@@ -118,6 +118,23 @@ export const connectedCallback = (elm: d.HostElement) => {
...
+        //
+        // This must key off the dedicated `hasFailedLoad` flag rather than
+        // `hasInitializedComponent` being unset: that flag is also
+        // (transiently) unset while an initialization attempt is merely
+        // queued/in-flight and hasn't failed at all -- e.g. behind
+        // `nextTick`, or when `connectedCallback` fires again for the same
+        // element before that queued attempt has run, which Stencil's own
+        // server-side hydration deliberately does (see
+        // `serverSideConnected()` in `update-component.ts`). Retrying in
+        // that case double-initializes the component.
```
johnjenkins: again - very verbose and repeated in constants.ts

```suggestion
        // init for this host element failed. Retry now that we're reconnecting.
```

src/runtime/initialize-component.ts
```diff
@@ -53,6 +55,18 @@ export const initializeComponent = async (
...
+          // The lazy bundle failed to load (e.g. a dropped network request).
+          // Clear the "initialized" flag and mark the load as failed so a
+          // future `connectedCallback` (see `connected-callback.ts`) is able
+          // to retry loading this component instead of leaving the host
+          // element permanently un-upgraded. `hasFailedLoad` (rather than
+          // just the absence of `hasInitializedComponent`) is what gates the
+          // retry, since `hasInitializedComponent` is also unset while an
+          // attempt is merely queued/in-flight (e.g. behind `nextTick`, or
+          // between repeated `connectedCallback` invocations during
+          // server-side hydration) -- neither of which is a failure.
```
johnjenkins: again - unnecessary - I think it's clear from context :)

```suggestion
          // The lazy bundle failed to load (e.g. a dropped network request).
```

src/utils/constants.ts
```diff
@@ -56,10 +56,18 @@ export const enum HOST_FLAGS {
   isWatchReady = 1 << 7,
   isListenReady = 1 << 8,
   needsRerender = 1 << 9,
+  /**
+   * Set when a lazy component's dynamic `import()` fails to resolve a
+   * constructor. Distinct from `hasInitializedComponent` being unset, which
+   * is also (transiently) true while an initialization attempt is merely
+   * queued/in-flight (e.g. behind `nextTick`) and hasn't failed at all.
+   * Only this flag should gate a `connectedCallback` retry.
+   */
```
johnjenkins, line 63:
```suggestion
  /**
   * Set when a lazy component's dynamic `import()` fails to resolve a a
   * constructor. Distinct from `hasInitializedComponent` being unset, which
   * is true while an initialization attempt is merely queued/in-flight.
   */
```

src/runtime/initialize-component.ts
johnjenkins, line 42: nit .. but this only pertains to lazy bundles so move into the lazyLoad gated block

```suggestion
      const bundleId = cmpMeta.$lazyBundleId$;
      if (BUILD.lazyLoad && bundleId) {
        // lazy loaded components

        // clear any failure recorded by a previous attempt
        hostRef.$flags$ &= ~HOST_FLAGS.hasFailedLoad;
```

src/runtime/test/lazy-load-retry.spec.tsx
johnjenkins, line 1: there doesn't appear to be any coverage here on the actual network-facing fix(?) -  the `?s-retry=N` query-param logic (because you're using the testing platform's mock `loadModule`). 
We probably need a more focused unit test; something to guard against regressions on the `cacheBustParams` join logic

johnjenkins (changes requested):
great work - tsym for looking into this 🙏 

General comments:

1) Retry has no cap or backoff atm so will just retry forever. Suggest a cap of 3 with a 1s interval between retries - wdyt?
2) To complete the "component can now recover" story, can you look at `componentOnReady`. Atm `$onReadyResolve$(elm)` is called unconditionally in initializeComponent's catch block on the first failure (in initialize-component.ts 212-214), and `$onReadyPromise$` is a plain Promise created once per host element in `registerHost()` (client-host-ref.ts 62) so can only settle once. Any caller that awaited `componentOnReady()` before the retry succeeds will have already received the (un-hydrated) element and will never be notified when the retry actually works.

johnjenkins (approved):
I made those changes :) 
LGTM now - thanks for contributing to stencil 🙏

PR #6773, opened 2026-07-15
fix(compiler): escape backslashes before other escape sequences in styles

### LINKED ISSUE #6710: bug: compilation fails if stylesheet contains content: content: "`";
### Prerequisites

### Stencil Version

4.41.2

### Current Behavior

If the stylesheet used in component contains eg 
```css
p::before {
  content: "`";
}
```
Compilation fails with
```
Rollup: Parse Error
           src/components/my-component/my-component.css?tag=my-component&encapsulation=shadow (1:183): Unterminated
           string constant (Note that you need plugins to import files that are not JavaScript) 1: :host { ^ 2:
           display: block; 3: }
```

I prepared a minimal reproducer at https://codesandbox.io/p/github/johnjenkins/stencil-starter/csb-x6l59f/draft/practical-robinson

### Expected Behavior

Stencil supports all valid CSS-Syntax

### System Info

```shell
System: node 22.20.0
    Platform: linux (6.1.43)
   CPU Model: AMD EPYC (2 cpus)
    Compiler: /project/workspace/node_modules/.pnpm/@stencil+core@4.41.2/node_modules/@stencil/core/compiler/stencil.js
       Build: 1768558830
     Stencil: 4.41.2 🐝
  TypeScript: 5.8.3
      Rollup: 4.34.9
      Parse5: 7.2.1
      jQuery: 4.0.0-pre
      Terser: 5.37.0
```

### Steps to Reproduce

Add a style rule with the ` in it

### Code Reproduction URL

https://codesandbox.io/p/github/johnjenkins/stencil-starter/csb-x6l59f/draft/practical-robinson

### Additional Information

This used to work with Stencil 4.29.3 and Rollup 4.34.9

Arul1998:
Escaping backslashes last re-escaped the backslash added for backticks, leaving an unescaped backtick that terminated the generated template literal and failed the build. Escape backslashes first, and also escape dollar-brace so CSS content cannot be evaluated as a template literal interpolation.

Fixes #6710

GitHub Issue Number: #6710

A component stylesheet containing a backtick, e.g.

```css
p::before {
  content: "`";
}
```

fails the whole build with:

```
Rollup: Parse Error
src/components/my-component/my-component.css?tag=my-component&encapsulation=shadow (1:183): Unterminated string constant
```

When embedding CSS into the generated JS template literal, `css-to-esm.ts` escaped backslashes *last*, which re-escaped the backslash it had just added for backticks (`` \` `` became ``\\` ``), leaving an unescaped backtick that terminates the template literal early. This is a regression — it worked in 4.29.3, before the escape chain was introduced.

`${` in CSS content had the same class of problem: it was not escaped at all, so it was evaluated as a template literal interpolation at runtime.

## What is the new behavior?

Backslashes are escaped first, so the escape sequences added afterwards are not re-escaped, and `${` is now escaped as well. CSS containing backticks or `${` compiles and round-trips correctly.

Behavior for everything else is unchanged, including the intentional handling of control characters (`\f`, `\b`, `\v`, `\0`) and icon-font content like `content: "\f101"` — the generated source for those inputs is identical to before this change.

## Documentation

N/A — internal compiler fix, no API or documentation changes.

- [ ] Yes
- [x] No

## Testing

- Strengthened the existing `escapes backticks in CSS content` unit test: it previously only did a substring check that the broken output also satisfied; it now evaluates the generated module to prove it is valid JavaScript and round-trips the input CSS.
- Added unit tests for `${}` in CSS content and for icon-font escape sequences (`\f101`) round-tripping.
- `src/compiler/style/test/css-to-esm.spec.ts`: 31/31 pass locally; full jest suite run locally as well.
- End-to-end: built a minimal component project using the reproduction CSS from #6710 against this branch — the build succeeds and the emitted bundle contains the correctly escaped style:

```js
const myComponentCss = () => `:host{display:block}p::before{content:"\`"}`;
```

Before this change, the same input reproduced the Rollup parse error from the issue.

## Other information

N/A

### DIFF
--- a/src/compiler/style/css-to-esm.ts
+++ b/src/compiler/style/css-to-esm.ts
@@ -165,18 +165,25 @@ const generateTransformCssToEsm = (
 ): d.TransformCssToEsmOutput => {
   const s = new MagicString('');
 
-  // Replace literal newlines/tabs/returns with spaces to avoid them becoming escape sequences
-  // Then handle other special characters and escape backslashes last
+  // Escape backslashes first so the escape sequences added below are not
+  // re-escaped afterwards. Escaping backslashes last would turn the
+  // escaped backtick produced below back into an unescaped backtick,
+  // which terminates the generated template literal early.
+  // Then replace literal newlines/tabs/returns with spaces, keep control
+  // characters visible as escape sequences in the runtime CSS string,
+  // and escape the remaining characters that have a special meaning
+  // inside a template literal (backticks and dollar-brace).
   results.styleText = results.styleText
+    .replace(/\\/g, '\\\\')
     .replace(/\n/g, ' ')
     .replace(/\r/g, ' ')
     .replace(/\t/g, ' ')
+    .replace(/\u000c/g, '\\\\f')
+    .replace(/\u0008/g, '\\\\b')
+    .replace(/\u000b/g, '\\\\v')
+    .replace(/\0/g, '\\\\0')
     .replace(/`/g, '\\`')
-    .replace(/\u000c/g, '\\f')
-    .replace(/\u0008/g, '\\b')
-    .replace(/\u000b/g, '\\v')
-    .replace(/\0/g, '\\0')
-    .replace(/\\/g, '\\\\');
+    .replace(/\$\{/g, '\\${');
 
   if (input.addTagTransformers) {
     results.styleText = addTagTransformToCssString(results.styleText, input.tags);
--- a/src/compiler/style/test/css-to-esm.spec.ts
+++ b/src/compiler/style/test/css-to-esm.spec.ts
@@ -434,8 +434,9 @@ var(--theme-secondary-color) var(--theme-secondary-opacity));
   });
 
   it('escapes backticks in CSS content', () => {
+    const input = '.my-class::before { content: "`"; }';
     const result = transformCssToEsmSync({
-      input: '.my-class::before { content: "`"; }',
+      input,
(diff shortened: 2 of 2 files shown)


johnjenkins (changes requested):
TYSM 🙏 
The fix looks good I think - traced through the escape-order logic (including the `\f`/`\b`/`\v`/`\0` branches); moving backslash-escaping to the front (combined with switching those four control-char replacements to double-backslash output) cancels out the reordering so control-char output stays identical to before. 
The `${` escape is also safe against pre-existing backslashes in the source.

One test gap: none of the tests (old or new) actually test `\u000c` / `\u0008` / `\u000b` / `\0` regex branches with a raw control-character byte. The new "preserves CSS escape sequences like icon font content" test uses the JS string `'\\f101'` (I know that test was done by me 😅 ), which is a literal backslash character followed by the text `f101` - not a real `U+000C` form-feed byte.

Would you mind adding one more test with a genuine raw control-character byte (e.g. an actual `\u000c` in `input`, not its escaped text form) asserted via the same `evaluateEsmOutput` round-trip pattern used for the other cases - tysm!

Arul1998:
Thanks for the thorough review! Added a test that feeds genuine raw control-character bytes (U+000C, U+0008, U+000B, U+0000) through the transform and asserts the round-tripped runtime CSS via evaluateEsmOutput, covering all four control-character branches.

johnjenkins (approved):

PR #6776, opened 2026-07-16
fix(runtime): map camelCase SVG props to kebab-case attributes

### LINKED ISSUE #5674: feat: support prop casing for SVG attributes
### Prerequisites

It'd be useful to support prop case for SVG attributes ([similar to React](https://stackblitz.com/edit/react-basic-example-bhqowf?file=index.js,example.js)). 

### Describe the Use Case

Spreading an object with props onto an SVG will not render properly and warnings/errors will be swallowed. 

See [repro case](https://github.com/jcfranco/stencil-svg-props-are-not-mapped-to-attrs).

SVG props get mapped to attributes. 

In the following example, `strokeWidth` would be mapped and rendered as `stroke-width` 

```tsx
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 30 10">
  <circle cx={15} cy={5} r={3} stroke="green" strokeWidth={3} />
</svg>
```

### Describe Alternatives

_No response_

### Related Code

_No response_

### Additional Information

_No response_

Arul1998:
SVG attribute matching is case-sensitive, so camelCase props spread onto SVG elements (e.g. strokeWidth) were set as-is and silently ignored by the renderer. Map them to their kebab-case attribute names like React does, leaving SVG's genuinely camelCase attributes (viewBox, preserveAspectRatio, etc.), xml/xmlns prefixed names, and xlink handling unchanged. The mapping is gated behind the svg build conditional.

Fixes #5674

GitHub Issue Number: #5674

camelCase SVG props are set verbatim as attributes. Since SVG attribute matching is case-sensitive, an attribute named `strokeWidth` does not match `stroke-width` and the styling is silently ignored — no warning or error. This mainly bites when spreading an object of props onto an SVG element (React-style prop objects, shared prop factories, etc.):

```tsx
const circleProps = { cx: 15, cy: 5, r: 3, stroke: 'green', strokeWidth: 3 };

<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 30 10">
  <circle {...circleProps} />
</svg>
```

renders `<circle ... strokeWidth="3">`, which has no visual effect.

## What is the new behavior?

In an SVG context, camelCase prop names are mapped to their kebab-case SVG attribute names (`strokeWidth` → `stroke-width`, `fillOpacity` → `fill-opacity`, …) for setting, updating, and removal — matching React's behavior for SVG props.

Deliberately excluded from the mapping:

- SVG attributes that are genuinely camelCase per the SVG specification (`viewBox`, `preserveAspectRatio`, `gradientTransform`, `baseFrequency`, etc.) — passed through unchanged via an explicit exception list
- `tabIndex` and `crossOrigin`, which map to all-lowercase attribute names rather than kebab-case
- `xml`/`xmlns` prefixed names — left unmodified
- existing `xlink` handling — the mapping runs after it, so `xlinkHref` continues to work as before

The mapping only applies in SVG contexts (`isSvg`), is gated behind the `svg` build conditional so it is tree-shaken from applications that don't render SVG, and non-SVG elements are unaffected.

The kebab-case prop names Stencil's JSX typings declare today (`stroke-width={3}`) continue to work exactly as before — this is purely additive.

## Documentation

N/A — behavior now matches what the JSX allows; happy to add a docs note to the SVG guide in stenciljs/site if desired.

- [ ] Yes
- [x] No

Previously, camelCase names in SVG contexts produced attributes that SVG ignores, so no working code depended on the old behavior.

## Testing

- Six new unit tests in `set-accessor.spec.ts`: setting, updating, and removing a camelCase prop as its kebab-case attribute; camelCase-in-spec attributes (`viewBox`, `preserveAspectRatio`) preserved; `tabIndex` → `tabindex`; and no mapping outside SVG contexts. Full `set-accessor` spec: 132/132 passing.
- Full runtime jest suite run locally: 655/655 across 63 suites — no regressions, including the existing SVG patching and xlink tests.
- Manual end-to-end: built a component using the exact reproduction from #5674 against this branch and inspected the DOM in a browser — the circle renders `<circle cx="15" cy="5" r="3" stroke="green" stroke-width="3">`, computed style reports `stroke-width: 3px`, and no `strokeWidth` attribute is present.

## Other information

The exception list mirrors the camelCase attribute names from the SVG specification (the same set React preserves). It lives behind the `BUILD.svg` conditional and a `@__PURE__` annotation, so it adds no bytes to builds without SVG rendering.

### DIFF
--- a/src/runtime/test/svg-element.spec.tsx
+++ b/src/runtime/test/svg-element.spec.tsx
@@ -74,6 +74,64 @@ describe('SVG element', () => {
     `);
   });
 
+  it('should map camelCase props to kebab-case svg attributes', async () => {
+    @Component({ tag: 'cmp-a' })
+    class CmpA {
+      render() {
+        const circleProps = { cx: 15, cy: 5, r: 3, stroke: 'green', strokeWidth: 3 };
+        return (
+          <svg viewBox="0 0 30 10">
+            <circle {...circleProps} />
+            <rect width={4} height={4} fillOpacity={0.5} />
+          </svg>
+        );
+      }
+    }
+    const { root } = await newSpecPage({
+      components: [CmpA],
+      html: `<cmp-a></cmp-a>`,
+    });
+    expect(root).toEqualHtml(`
+      <cmp-a>
+        <svg viewBox=\"0 0 30 10\">
+          <circle cx=\"15\" cy=\"5\" r=\"3\" stroke=\"green\" stroke-width=\"3\"></circle>
+          <rect width=\"4\" height=\"4\" fill-opacity=\"0.5\"></rect>
+        </svg>
+      </cmp-a>
+    `);
+  });
+
+  it('should update and remove kebab-case mapped svg attributes', async () => {
+    @Component({ tag: 'cmp-a' })
+    class CmpA {
+      @Prop() strokeWidth?: number = 3;
+
+      render() {
+        return (
+          <svg viewBox="0 0 30 10">
+            <circle cx="15" cy="5" r="3" strokeWidth={this.strokeWidth} />
+          </svg>
+        );
+      }
+    }
+    const { root, waitForChanges } = await newSpecPage({
+      components: [CmpA],
+      html: `<cmp-a></cmp-a>`,
+    });
+    let circle = root.querySelector('circle');
+    expect(circle.getAttribute('stroke-width')).toBe('3');
+
+    root.strokeWidth = 5;
+    await waitForChanges();
+    circle = root.querySelector('circle');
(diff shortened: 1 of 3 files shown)


johnjenkins (changes requested):
❤️ 
Can you add a spec test to src/runtime/test/svg-element.spec.tsx 🙏

Arul1998:
Added integration specs to svg-element.spec.tsx one covering the spread-props reproduction from the issue rendered through the vdom, and one covering update/removal of a mapped attribute via a @Prop. Thanks for the review!

PR #6782, opened 2026-07-17
fix(compiler): resolve inherited type references against the declaring file

### LINKED ISSUE #6687: bug: components.d.ts resolves inherited types to wrong import paths
### Prerequisites

### Stencil Version

4.43.3

### Current Behavior

The issue:
The auto-generated components.d.ts resolves the Validator type import to the consuming component's directory instead of the shared directory where it actually lives:

// What Stencil generates (wrong):
import { Validator } from "./components/data-entry/checkbox/input.types";

The file at dist/types/components/data-entry/checkbox/input.types.d.ts doesn't exist — the type only exists at dist/types/components/shared/input/input.types.d.ts. This breaks type resolution for consumers who import type { Validator } from 'my-package'.

The individual component .d.ts files resolve correctly (they properly reference ../../shared/input/...), so the issue is specifically in how components.d.ts resolves import paths for types inherited through base classes.

It looks like this issue happens when there is more than 1 extend.

### Expected Behavior

// What Stencil generates (wrong):
import { Validator } from "./components/data-entry/checkbox/input.types";

// What it should be:
import { Validator } from "./components/shared/input/input.types";

### System Info

```shell

```

### Steps to Reproduce

- Shared base classes in src/components/shared/input/ that import a Validator type from a sibling input.types.ts file
- Concrete components (e.g. checkbox, input-text, select, etc.) in src/components/data-entry/ that extend a class that extends this base class

### Code Reproduction URL

https://codesandbox.io/p/github/johnjenkins/stencil-starter/csb-yx4ntc/draft/exciting-wave?file=%2Fdist%2Ftypes%2Fcomponents.d.ts%3A11%2C28

### Additional Information

_No response_

### LINKED ISSUE #6700: bug: Relative imports of Prop/Event types from abstract component cannot be resolved in components.d.ts for subclasses
### Prerequisites

### Stencil Version

4.43.4

### Current Behavior

When an abstract component uses imported types for Events/Props/Methods, the components.d.ts might have an error referencing said types.

### Expected Behavior

When generating the components.d.ts the correct relative imports should be generated dependent on the abstract component which imports the types.

### System Info

```shell
System: node 24.14.0
    Platform: windows (10.0.26200)
   CPU Model: 13th Gen Intel(R) Core(TM) i7-13620H (16 cpus)
    Compiler: C:\develop\test\node_modules\@stencil\core\compiler\stencil.js
       Build: 1776084048
     Stencil: 4.43.4
  TypeScript: 5.8.3
      Rollup: 4.44.0
      Parse5: 7.2.1
      jQuery: 4.0.0-pre
      Terser: 5.37.0
```

### Steps to Reproduce

- Checkout [repository](https://github.com/FabioGimmillaro/stencil-export-error)
- Run "npm i"
- Run "npm run build"
- => Incorrect import and export of UtilType in components.d.ts

### Code Reproduction URL

https://github.com/FabioGimmillaro/stencil-export-error

### Additional Information

_No response_

Arul1998:
Type references on component members are recorded relative to the file declaring them. When members are inherited from a base class in another directory, the references were merged into the component's metadata unchanged, and type generation later resolved them against the component's own file. This produced imports in `components.d.ts` pointing at nonexistent paths, broke type resolution for library consumers, and emitted duplicate aliased imports for each extending component. This affected both imported types and types declared locally in the base class's file, with any level of inheritance.

Re-anchor inherited members' type references when they are merged: relative import specifiers are rewritten to resolve from the component's directory, and base-class-local types become imports of the base class's module.

fixes #6687
fixes #6700

GitHub Issue Number: #6687

When a component inherits decorated members from a base class in a different directory, the generated `components.d.ts` resolves the members' type imports against the component's file instead of the base class's file. For the reproduction in #6687, it emits `import { Validator } from "./components/data-entry/checkbox/input.types"` — a path that doesn't exist in `dist/types` — instead of `./components/shared/input/input.types`. This happens with a single level of `extends` (not only chains), also affects types declared locally in the base class's file, and each extending component emits its own broken duplicate import under an incremented alias. A consumer type-checking against the package gets hard `TS2307`/`TS2305` errors.

## What is the new behavior?

When inherited members are merged into a component's metadata (`mergeExtendedClassMeta`), their type references are re-anchored to the component's file: relative import specifiers are rewritten relative to the component's directory, and types declared in the base class's file are converted to imports of that file. Since the metadata itself is corrected, every consumer of it (`components.d.ts`, docs outputs, collections) gets the right paths. Shared types now emit a single correct import instead of per-component broken duplicates. References for same-directory base classes, package imports, globals, and base classes shipped in external collections are left untouched.

## Documentation

N/A

- [ ] Yes
- [x] No

## Testing

Added unit tests covering: re-anchoring relative import specifiers (including ones pointing outside the base class's tree), converting base-class-local references to imports, `./`-prefixing bare specifiers, and the no-op cases (same directory, package imports, globals, `node_modules` base classes, members without references). Manually verified end-to-end with a project matching the issue's structure: pre-fix, `tsc` against the generated `dist/types` fails with 8 errors; with this fix it passes, and `components.d.ts` contains single, correct imports for both inherited-imported and inherited-local types, while non-inherited components are byte-identical.

## Other information

N/A


johnjenkins (approved):
LGTM - thanks for your work and your patience 🙏 
I took the liberty of adding a small integration test within the wider `mergeExtendedClassMeta` and re-used the `isNodeModulePath` utility so now it's good to go I think

PR #6794, opened 2026-07-27
feat(hydrate): cache platform closure and add opt-in window reuse

Armand-Lluka:
The hydrate factory wraps the entire platform (runtime, vdom and every component class) in hydrateAppClosure() so it lexically captures the per-call window, and re-executes that closure on EVERY renderToString call. For component libraries this is the dominant fixed cost per render (framework output targets call renderToString once per component instance).

Two changes:

1. The factory now caches the evaluated closure on the window object and reuses it whenever the same window is passed again. Transparent for fresh-window renders.

2. New opt-in HydrateDocumentOptions.reuseWindow: the string-input path of hydrateDocument()/renderToString() reuses a process-global MockWindow, one per serializeShadowRoot mode (scoped serialization permanently ORs shadowNeedsScopedCss into component metadata inside the cached closure). Fresh head/body ELEMENTS are swapped in per render (never innerHTML='') because rootAppliedStyles is keyed on the head node. Renders are serialized through an internal promise queue since the shared window is not concurrency-safe.

Measured 2.2x per-render speedup on a 2-component test app; scales with component count (~4.3x on a 475-component library).

GitHub Issue Number: N/A

## What is the new behavior?

Closes #6794

- [ ] Yes
- [x] No

### DIFF
--- a/src/declarations/stencil-public-compiler.ts
+++ b/src/declarations/stencil-public-compiler.ts
@@ -948,6 +948,21 @@ export interface HydrateDocumentOptions {
    */
   referrer?: string;
   /**
+   * Reuse a process-global mock window (one per `serializeShadowRoot` mode) across
+   * `renderToString()`/`hydrateDocument()` calls that receive an HTML string. This
+   * avoids re-executing the entire hydrate platform closure (runtime, vdom and all
+   * component definitions) on every call, which is the dominant fixed cost when
+   * rendering many small fragments (e.g. one `renderToString` call per component
+   * instance, as the framework output targets do).
+   *
+   * The reused window is not concurrency safe, so renders are serialized through
+   * an internal queue. Intended for rendering HTML fragments
+   * (`fullDocument: false`); one known output difference is that hydration
+   * annotation counters (e.g. `<!--r.N-->`) become unique across the process
+   * instead of restarting at 1 per call. Defaults to `false`.
+   */
+  reuseWindow?: boolean;
+  /**
    * Removes every `<script>` element found in the `document`. Defaults to `false`.
    */
   removeScripts?: boolean;
--- a/src/hydrate/runner/render.ts
+++ b/src/hydrate/runner/render.ts
@@ -22,6 +22,44 @@ import { initializeWindow } from './window-initialize';
 
 const NOOP = () => {};
 
+/**
+ * Process-global windows reused across renders when `reuseWindow` is enabled,
+ * ONE PER `serializeShadowRoot` mode: scoped serialization permanently ORs
+ * `shadowNeedsScopedCss` into each component's runtime metadata inside the
+ * cached hydrate platform, which would otherwise leak into later
+ * declarative-shadow-dom renders sharing the same window.
+ */
+const reusableWindows = new Map<string, MockWindow>();
+
+/**
+ * The reused windows are not concurrency-safe, so renders against them are
+ * serialized through this promise queue.
+ */
+let reuseRenderQueue: Promise<unknown> = Promise.resolve();
+
+function getReusableWindow(doc: string, opts: HydrateFactoryOptions): MockWindow {
+  const modeKey = JSON.stringify(opts.serializeShadowRoot ?? null);
+  let win = reusableWindows.get(modeKey);
+  if (win) {
+    const document = win.document;
+    /**
+     * Swap in fresh `<head>`/`<body>` ELEMENTS for every render (never reset via
+     * `innerHTML = ''`): the runtime's `rootAppliedStyles` WeakMap is keyed on
+     * the head node, so keeping the node identity would silently drop scoped
+     * `sc-` styles after the first render.
+     */
+    const newHead = document.createElement('head');
+    document.documentElement.replaceChild(newHead, document.head);
+    const newBody = document.createElement('body');
+    newBody.innerHTML = doc;
+    document.documentElement.replaceChild(newBody, document.body);
+  } else {
+    win = new MockWindow(doc);
+    reusableWindows.set(modeKey, win);
+  }
+  return win;
+}
+
 export function streamToString(html: string | any, option?: SerializeDocumentOptions) {
   return renderToString(html, option, true);
 }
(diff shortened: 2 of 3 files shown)


Armand-Lluka:
> hey @Armand-Lluka - thanks for raising - I think the concept is a good one. I get concerns around potential memory leaks with this kind of thing on the server, so I had an agent to test some scenarios and it found some unfortunately - https://github.com/johnjenkins/stencil-hydrate-reuse-leak-repro (excuse the long winded prologue) - please can you address those first, then I'll have a complete picture before doing a full code-review.

Updated the code, my results below. The delta has been reduced dramatically so the memory leak issue looks to be solved 👍 

@johnjenkins 

--- 

| Scenario | Renders measured | Start heap | End heap | Change | Rate per 1,000 | reported original rate |
|---|---:|---:|---:|---:|---:|---:|
| A: Fresh-window control with listeners | 2,800 | 6.66 MB | 7.22 MB | +0.56 MB | +0.201 MB | Flat/noise |
| B: Reused window with listeners | 2,800 | 7.46 MB | 8.17 MB | +0.71 MB | +0.253 MB | +8.1 MB |
| C: Reused window without listeners | 2,800 | 8.09 MB | 8.22 MB | +0.13 MB | +0.048 MB | Flat/noise |
| D: Varying `serializeShadowRoot` objects | 750 | 8.09 MB | 8.01 MB | -0.08 MB | -0.100 MB | +44 MB |

src/hydrate/runner/render.ts
```diff
@@ -18,10 +18,17 @@ import type {
...
+import { canReuseWindow, deleteReusableWindow, getReusableWindow } from './reusable-window';
 import { initializeWindow } from './window-initialize';
 
 const NOOP = () => {};
 
+/**
+ * The reused windows are not concurrency-safe, so renders against them are
+ * serialized through this promise queue.
+ */
+let reuseRenderQueue: Promise<unknown> = Promise.resolve();
```
johnjenkins: this is shared across *all* modes - both scoped / dsd (right?), but since each mode gets it's own independent `MockWindow` and shares no state - they could be rendered concurrently? 
So this could instead be `Map<modeKey, Promise>` - wdyt?

src/declarations/stencil-public-compiler.ts
```diff
@@ -947,6 +947,22 @@ export interface HydrateDocumentOptions {
...
+   * `renderToString` call per component instance, as the framework output targets do).
+   *
+   * The reused window is not concurrency safe, so renders are serialized through
+   * an internal queue. This option only applies when rendering HTML fragments
+   * (`fullDocument: false`) with a boolean or built-in string `serializeShadowRoot`
+   * value; full-document renders and per-component `serializeShadowRoot` objects use
+   * a fresh window. One known output difference is that hydration annotation counters
+   * (e.g. `<!--r.N-->`) become unique across the process instead of restarting at 1
+   * per call. Defaults to `false`.
+   */
```
johnjenkins, line 956: v verbose

```suggestion
  /**
   * Reuse a process-global mock window across `renderToString()`/`hydrateDocument()`
   * calls that receive an HTML string. This avoids re-executing the entire hydrate
   * platform closure (runtime, vdom and all component definitions) on every call.
   */
```

src/hydrate/runner/render.ts
```diff
@@ -85,6 +92,41 @@ export function hydrateDocument(
...
+        } catch (e) {
+          if (reusedWin) {
+            deleteReusableWindow(opts.serializeShadowRoot);
+            if (reusedWin.close) {
+              reusedWin.close();
+            }
+          }
+          renderCatchError(results, e);
+          return results;
+        }
```
johnjenkins, line 114: `reusedWin` is only assigned *after* `getReusableWindow()` returns. 
If the reset logic inside `getReusableWindow` throws, the catch's `if (reusedWin)` guard is false, `deleteReusableWindow` never runs, and the same broken window gets fetched and rethrown on every subsequent call for that mode.

src/hydrate/runner/reusable-window.ts
```diff
@@ -0,0 +1,86 @@
...
+  let win = reusableWindows.get(modeKey);
+  if (!win) {
+    win = new MockWindow(doc);
+    reusableWindows.set(modeKey, win);
+    return win;
+  }
+
+  const document = win.document;
+  resetReusableWindow(win);
+  const defaults = new MockWindow(false);
```
johnjenkins: behaviour nit (so not likely to happen for the use case you're wanting .. but an inconsistency for others wanting to use this independently):
Consider:
`const html = '<link rel="stylesheet" href="/foo.css"><plain-cmp></plain-cmp>'`
render 1 (fresh window) / current behaviour =  `<plain-cmp ...>...</plain-cmp>` <  link silently dropped
render 2 (reused window): `<link rel="stylesheet" href="/foo.css"><plain-cmp ...>...</plain-cmp>` < link kept

johnjenkins (changes requested):
lookin' good :) 

More generally, I'd like to see some e2e tests for this new behaviour (under test/end-to-end) 🙏

src/hydrate/runner/reusable-window.ts
```diff
@@ -0,0 +1,151 @@
...
+import { resetEventListeners } from '../../mock-doc/event';
+
+/**
+ * Process-global windows reused across renders when `reuseWindow` is enabled,
+ * one per built-in `serializeShadowRoot` mode. Scoped serialization permanently
+ * updates component metadata inside the cached hydrate platform.
+ */
+const reusableWindows = new Map<string, MockWindow>();
+const windowDefaults = new MockWindow(false);
+const documentKeyKeepers = new Set([
```
Armand-Lluka: I’m not sure the hydrate runner is the right place to define which document and window properties survive a reset; that responsibility may belong in mock-doc. For now, I’ve kept the minimal set needed to preserve the identities captured by the cached hydrate closure.

Armand-Lluka: I've demonstrated what a mock-dock refactor would look like, it can be dropped if we think it's out of scope.

Armand-Lluka:
Let me know if anything here needs adjusting 🙂 @johnjenkins

Armand-Lluka:
Hi @johnjenkins, just following up on this one 🙂 

Since your review on:
- Added e2e coverage for the reuse behaviour in `renderToString.e2e.ts`, per your request
- Rebased on main, CI green
- Memory-leak scenarios from your earlier repro are addressed with the numbers posted above (deltas now flat/noise-level vs. the original leak rates)

If the PR has grown past what you'd like to review in one pass, happy to split it into smaller pieces or walk through it async, whatever's easiest. Otherwise, is it ready for a re-review?

PR #6808, opened 2026-08-04
feat(jsx): add global `translate` attribute to HTMLAttributes

### LINKED ISSUE #6807: feat: add global `translate` attribute to JSX HTMLAttributes typings
### Prerequisites

- [x] I have read the Contributing Guidelines.
- [x] I agree to follow the Code of Conduct.
- [x] I have searched for existing issues that already include this feature request, without success.

Stencil's JSX typings (`JSXBase.HTMLAttributes` in `src/declarations/stencil-public-runtime.ts`) include the standard global HTML attributes such as `dir`, `lang`, `hidden`, `spellcheck`, `draggable`, etc., but are missing the global [`translate`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/translate) attribute.

As a result, writing `<p translate="no">…</p>` in a Stencil component's `render()` fails type-checking with an error that `translate` is not a valid attribute, even though it is a valid global HTML attribute supported by every element.

### Describe the Use Case

The `translate="no"` attribute is commonly used to tell browser/machine translation tools (e.g. Google Translate) not to translate certain content — for example brand/product names, prices, or codes that must remain verbatim. This is important for legal/branding correctness.

Add `translate` to `JSXBase.HTMLAttributes`, alongside the other global attributes:

```ts
translate?: 'yes' | 'no' | (string & {});
```

The `'yes' | 'no'` literals provide autocomplete for the common values, while `(string & {})` preserves that autocomplete without rejecting values that get widened to `string`.

### Describe Alternatives

Spreading an untyped object literal, or module-augmenting `@stencil/core/internal` in the consuming project.

### Additional Information

Branding is important for my company and we use this attribute a bunch to avoid translating silly product names. Currently we've patched stencil in our local repo to solve this.

erik-gullberg:

Stencil's JSX typings (`JSXBase.HTMLAttributes`) include the standard global HTML attributes such as `dir`, `lang`, `hidden`, `spellcheck`, and `draggable`, but omit the global [`translate`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/translate) attribute.

As a result, `<p translate="no">…</p>` fails to type-check in a Stencil component, forcing consumers to work around it by spreading an untyped object.

GitHub Issue Number: #6807

## What is the new behavior?

Closes #6807

Adds `translate` to `JSXBase.HTMLAttributes` so it is available on every native and custom element, consistent with the other global HTML attributes already present:

```ts
translate?: 'yes' | 'no' | (string & {});
```

## Documentation

N/A — this mirrors the existing global HTML attributes already documented via the type definitions.

- [ ] Yes
- [x] No

Purely additive to an interface; existing code continues to type-check.

## Testing

These JSX attribute typings are validated by TypeScript at compile time and have no dedicated runtime unit test (the same is true for the sibling global attributes like `spellcheck`, `dir`, and `lang`). Verified that:

- The change compiles.
- `prettier --check` passes for the modified file.

### DIFF
--- a/src/declarations/stencil-public-runtime.ts
+++ b/src/declarations/stencil-public-runtime.ts
@@ -1583,9 +1583,7 @@ export namespace JSXBase {
     tabIndex?: number;
     tabindex?: number | string;
     title?: string;
-    // The `'yes' | 'no'` literals provide autocomplete for the common values,
-    // while `(string & {})` keeps that autocomplete without rejecting values
-    // that are widened to `string` (e.g. spread from an object literal).
+    // `(string & {})` keeps autocomplete without rejecting vals spread from an object literal
     translate?: 'yes' | 'no' | (string & {});
     // These types don't allow you to use popover as a boolean attribute
     // so you can't write HTML like `<div popover>` and get the default value.
--- a/src/declarations/stencil-public-runtime.ts
+++ b/src/declarations/stencil-public-runtime.ts
@@ -1583,6 +1583,10 @@ export namespace JSXBase {
     tabIndex?: number;
     tabindex?: number | string;
     title?: string;
+    // The `'yes' | 'no'` literals provide autocomplete for the common values,
+    // while `(string & {})` keeps that autocomplete without rejecting values
+    // that are widened to `string` (e.g. spread from an object literal).
+    translate?: 'yes' | 'no' | (string & {});
     // These types don't allow you to use popover as a boolean attribute
     // so you can't write HTML like `<div popover>` and get the default value.
     // Developer must explicitly specify one of the valid popover values or it will fallback


johnjenkins (approved):

src/declarations/stencil-public-runtime.ts
```diff
@@ -1583,6 +1583,10 @@ export namespace JSXBase {
     tabIndex?: number;
     tabindex?: number | string;
     title?: string;
+    // The `'yes' | 'no'` literals provide autocomplete for the common values,
+    // while `(string & {})` keeps that autocomplete without rejecting values
+    // that are widened to `string` (e.g. spread from an object literal).
```
johnjenkins:
```suggestion
    // `(string & {})` keeps autocomplete without rejecting vals spread from an object literal
```

johnjenkins (changes requested):
let's just make this comment a bit less verbose :)

johnjenkins (approved):

erik-gullberg:
Thank you! @johnjenkins

PR #6811, opened 2026-08-06
fix(runtime): memory leak on document.hidden

### LINKED ISSUE #6762: bug: Memory Leak - Destroyed stencil component retained while tab is hidden
### Prerequisites

### Stencil Version

4.43 (but I suspect all versions)

### Current Behavior

Hello ! 

**Context**
I am developer on a Angular webapp where we use our custom web components built with stencil, as our in-house UI framework.

We noticed recently during our soak (basically a stress test where we let the app run overnight) that we have a memory leak on a page that receives lots of events per second. This eventually causes the page to crash out of memory, naturally.

The page simply consists of a mix of web components to display event data whichs changes as soon as we receive a new event : 
```html
@if(event$ | async; as event){
   <my-web-component1/>
   <my-web-component2/>
   ...
}
```

At first, we thought the problem was on our side : a classic case of forgotten observable / event handler unsub.
But, by digging more at this problem over the past few week, we realized the **leak only happens when the tab is hidden** `(document.hidden = true)`.

That has caught my attention, and I've captured several Chrome heap snapshots while the tab is in background.
I've confirmed without the shadow of a doubt that the only thing leaking are references to our stencil web components (host elements). It's constantly increasing in size without ever being cleared out. Node count increases linearly with heap memory. in the form of detached DOM elements.

----

**Plausible cause**
Here's the interesting part : 

I was curious about why does it only happen when the tab is masked, when the tab has the focus, everything gets cleared out periodically.

Then, I stumbled across many articles saying that the browser throttles internal functions to save energy when `document.hidden = true`, which makes sense.  `setTimeout/setInterval` are throttled in frequency, but what is the most interesting here is that `requestAnimationFrame(callback)` is completly paused !

You might see where I go with that, it seems like Stencil internal component flush mechanism uses that `requestAnimationFrame` to flush the host element data, which explains the leak and why it only happens in background.

The heap snapshots I analyzed also contained stack traces leading up to stencil component destruction hang in requestAnimationFrame call.

### Expected Behavior

Expected : No leaks appearing when tab is background (and in foreground)

**Monkey patch fix**
To further confirm my hypothesis, I have monkey patched `rAf` by injecting this script : 

```js
window.requestAnimationFrame = (cb) => {
  return setTimeout(() => cb(performance.now(), 0);
};
```
That stops the leak from happening even when the tab is hidden!
Obviously this script can have other unintended side effects and is not supposed to be a final solution, but it further proves that the problem resides in the fact `rAf` is asynchronous. And by making it synchronous, it is executed immediately, so the component can teardown normally and memory is cleaned up.

### Steps to Reproduce

**Minimal repro**
I made sure the leak happens on all browsers and is Angular / React agnostic.

To repro the same scenario,: 
1. Create a new stencil project with `create-stencil`
2. Run the project as is in a browser.
3. Inject this script in browser console to simulate rapid basic `<my-component>` creation and destruction : 
```js
setInterval(() => {
  const el = document.createElement('my-component');
  document.body.appendChild(el);

  setTimeout(() => {
    el.remove();
  }, 10);
}, 10);
```
4. Switch to another tab, minimize the browser or lock your screen (anything that triggers `document.hidden = true`
5. Monitor the memory with either Chrome DevTools or a script that periodically polls Chrome memory via remote debugging.
6. It should increase indefinitely without ever falling off.
7. When you go back to the tab, the memory is eventually cleaned up.

**Do the same experiment with native element like `<div>` and the leak never happen.**

### Code Reproduction URL

Stencil Starter (no-modification) + script to create and destroy <my-component> repeatedly (mentioned above)

### Additional Information

**Help needed**
I wanted to know if anyone of you ever experienced such issue and if it is fixable or by design as part of Stencil ?

I have glanced over `stencil-core` code (`client-task.queue.ts`) it's clear that it seems to be a deliberate choice, but I'd like some opinions on that.

Honestly, this is quite concerning for any stencil component destroyed in the background to cause a leak.
It can screw up many applications, like using a virtual-scroll that populates itself over time and frequently creates and destroys DOM elements. 

Thanks ! If stencil devs could chime in that would much appreciated.

*I can also share the script I used to monitor Chrome memory from the command line if needed.*

johnjenkins:

GitHub Issue Number: #6762

requestAnimationFrame never fires when `document.hidden` (e.g. when a tab isn't open). 
This means:

1) all flushed tasks can build up indefinitely
2) any disconnected elements are not properly cleaned-up / garbage collected

In both cases (esp disconnected cleanup) the indiscriminate use of requestAnimationFrame was never a design decision, just an oversight

## What is the new behavior?

Fixes #6762

1) flushed tasks detect if `document.hidden` - if true will use microtask instead of RaF
2) disconnected elements always run cleanup via microtask

- [ ] Yes
- [x] No

### DIFF
--- a/src/client/client-task-queue.ts
+++ b/src/client/client-task-queue.ts
@@ -12,7 +12,6 @@ const queueDomReads: d.RafCallback[] = [];
 const queueDomWrites: d.RafCallback[] = [];
 const queueDomWritesLow: d.RafCallback[] = [];
 
-
 // Fallback to microtask when `document.hidden`: `requestAnimationFrame` callbacks
 // do not fire so scheduling flush queues tasks indefinitely.
 const scheduleFlush = () => (win.document?.hidden ? nextTick(flush) : plt.raf(flush));
--- a/src/client/client-task-queue.ts
+++ b/src/client/client-task-queue.ts
@@ -12,10 +12,9 @@ const queueDomReads: d.RafCallback[] = [];
 const queueDomWrites: d.RafCallback[] = [];
 const queueDomWritesLow: d.RafCallback[] = [];
 
-/**
- * Fallback to microtask when `document.hidden`: `requestAnimationFrame` callbacks
- * do not fire so scheduling flush queues tasks indefinitely.
- */
+
+// Fallback to microtask when `document.hidden`: `requestAnimationFrame` callbacks
+// do not fire so scheduling flush queues tasks indefinitely.
 const scheduleFlush = () => (win.document?.hidden ? nextTick(flush) : plt.raf(flush));
 
 const queueTask = (queue: d.RafCallback[], write: boolean) => (cb: d.RafCallback) => {
--- a/src/client/client-task-queue.ts
+++ b/src/client/client-task-queue.ts
@@ -3,7 +3,7 @@ import { BUILD } from '@app-data';
 import type * as d from '../declarations';
 import { PLATFORM_FLAGS } from '../runtime/runtime-constants';
 import { consoleError } from './client-log';
-import { plt, promiseResolve } from './client-window';
+import { plt, promiseResolve, win } from './client-window';
 
 let queueCongestion = 0;
 let queuePending = false;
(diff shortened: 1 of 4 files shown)


mpesic (approved):
👍

PR #6825, opened 2026-08-12
fix(compiler): track type query references

### LINKED ISSUE #2482: keyof typeof with @Prop decorator not importing class into components.d.ts
**Stencil version:**
```
1.13.0
```

**I'm submitting a:**
[x] bug report

**Current behavior:**
Having a @Prop of type 'keyof typeof SomeEnumClass' will not compile properly.
if that SomeEnumClass is not otherwise referenced in the component.

when running e.g. `npm run start` 
compiler will generate a components.d.ts that does not include an import of SomeEnumClass.

Resulting Error: 

TypeScript: ./src/components.d.ts
"Cannot find name 'SomeEnumClass'.  

But if i create another prop with that class, then components.d.ts properly imports in the class.
e.g. 
@Prop propName: keyof typeof EnumClass
@Prop prop2Name: EnumClass

<WORKS!>

**Expected behavior:**

Should generate the proper components.d.ts imports if type is 
'keyof typeof Enumclass'

**Steps to reproduce:**
Create a stencil component with 

@Prop propName: keyof typeof EnumClass
and do not have any other reference to that EnumClass in the component.

`npm run start` will fail.

Boulea7:

Class and enum names used in identifier-based `keyof typeof` prop types are not collected as type references. As a result, generated `components.d.ts` files can omit required imports. Aliases and re-exports can also produce incorrect imports or colliding docs type-library IDs.

GitHub Issue Number: Fixes #2482

## What is the new behavior?

The compiler now collects identifier-based `TypeQueryNode` references when they resolve to classes or enums. It preserves local, named, default, and re-exported aliases in generated component declarations, supports multiple local names for the same default export, and derives docs type-library IDs from the resolved declaration source.

## Documentation

N/A. This is a compiler fix covered by regression tests.

- [ ] Yes
- [x] No

## Testing

- `npm run test.jest -- src/compiler/transformers/test/parse-props.spec.ts --runInBand` (48 tests)
- `npm run test.jest -- src/compiler/transformers/test --runInBand` (356 tests)
- `npm run test.jest -- src/compiler/types/tests/generate-app-types.spec.ts --runInBand` (16 tests, 16 snapshots)
- `npm run tsc.prod`
- `npm run lint`
- `npm run prettier.dry-run`
- `npm run build`

## Other information

Qualified namespace expressions are outside this change's scope; this addresses identifier-based enum and class queries such as the one reported in #2482.

### DIFF
--- a/src/compiler/transformers/test/parse-props.spec.ts
+++ b/src/compiler/transformers/test/parse-props.spec.ts
@@ -1,6 +1,6 @@
-import { mockBuildCtx, mockConfig, mockValidatedConfig } from '@stencil/core/testing';
+import { mockBuildCtx, mockCompilerSystem, mockConfig, mockValidatedConfig } from '@stencil/core/testing';
 import { normalizePath } from '@utils';
-import { join, resolve } from 'path';
+import { join, relative, resolve } from 'path';
 import * as ts from 'typescript';
 
 import { createCompiler } from '../../compiler';
@@ -156,16 +156,17 @@ describe('parse props', () => {
   });
 
   it('generates imports for default and aliased exports used by type queries', async () => {
-    const rootDir = normalizePath(process.cwd(), false);
-    const fixtureDir = join(rootDir, '.stencil-type-query-exports');
-    const srcDir = join(fixtureDir, 'src');
+    const sys = mockCompilerSystem();
+    const rootDir = '/';
+    const srcDir = join(rootDir, 'src');
     const config = mockConfig({
       buildDocs: true,
+      sys,
       rootDir,
       srcDir,
-      outputTargets: [{ type: 'docs-json', file: join('.stencil-type-query-exports', 'docs.json') }],
+      outputTargets: [{ type: 'docs-json', file: join(rootDir, 'docs.json') }],
     });
-    const tsconfigPath = join(fixtureDir, 'tsconfig.json');
+    const tsconfigPath = join(rootDir, 'tsconfig.json');
     const componentDir = join(srcDir, 'components');
 
     await config.sys.createDir(componentDir, { recursive: true });
@@ -399,7 +400,7 @@ describe('parse props', () => {
         [],
       );
 
-      const docsJson = JSON.parse(await compiler.sys.readFile(join(fixtureDir, 'docs.json')));
+      const docsJson = JSON.parse(await compiler.sys.readFile(join(rootDir, 'docs.json')));
       const barrelComponent = docsJson.components.find((component: { tag: string }) => component.tag === 'cmp-barrel');
       const getBarrelReference = (propertyName: string, referenceName: string) =>
         barrelComponent.props.find((property: { name: string }) => property.name === propertyName).complexType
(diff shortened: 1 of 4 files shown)


src/compiler/transformers/test/parse-props.spec.ts
```diff
@@ -78,6 +82,395 @@ describe('parse props', () => {
...
+      referenceLocation: 'BestEnum',
+    });
+    expect(getReferences('classValue').ImportedClass).toMatchObject({
+      location: 'import',
+      path: '../transform-utils',
+      referenceLocation: 'ObjectMap',
+    });
+  });
+
+  it('generates imports for default and aliased exports used by type queries', async () => {
```
johnjenkins, line 160: this test seems to write (and leave behind) real files. 
Please re-factor to use `mockCompilerSystem`

e.g.

```diff
--- a/src/compiler/transformers/test/parse-props.spec.ts
+++ b/src/compiler/transformers/test/parse-props.spec.ts
@@
-import { mockBuildCtx, mockConfig, mockValidatedConfig } from '@stencil/core/testing';
+import { mockBuildCtx, mockCompilerSystem, mockConfig, mockValidatedConfig } from '@stencil/core/testing';
 import { normalizePath } from '@utils';
 import { join, resolve } from 'path';
@@
   it('generates imports for default and aliased exports used by type queries', async () => {
-    const rootDir = normalizePath(process.cwd(), false);
-    const fixtureDir = join(rootDir, '.stencil-type-query-exports');
-    const srcDir = join(fixtureDir, 'src');
+    const sys = mockCompilerSystem();
+    const rootDir = '/';
+    const srcDir = join(rootDir, 'src');
     const config = mockConfig({
       buildDocs: true,
+      sys,
       rootDir,
       srcDir,
-      outputTargets: [{ type: 'docs-json', file: join('.stencil-type-query-exports', 'docs.json') }],
+      outputTargets: [{ type: 'docs-json', file: join(rootDir, 'docs.json') }],
     });
-    const tsconfigPath = join(fixtureDir, 'tsconfig.json');
+    const tsconfigPath = join(rootDir, 'tsconfig.json');
     const componentDir = join(srcDir, 'components');

     await config.sys.createDir(componentDir, { recursive: true });
     await config.sys.writeFile(tsconfigPath, JSON.stringify({ /* ...unchanged... */ }));
```

src/compiler/transformers/transform-utils.ts
```diff
@@ -572,6 +572,9 @@ interface TypeReferenceIR {
   type: ts.Type;
 }
 
+const resolveAliasedSymbol = (checker: ts.TypeChecker, symbol: ts.Symbol | undefined): ts.Symbol | undefined =>
```
johnjenkins: housekeeping 🙏  
we have a function already (`unalias`) in src/compiler/transformers/type-library.ts that does the same thing - please remove `unalias` and export `resolveAliasedSymbol` (and import / use in  type-library.ts) - then please add a jsdoc

johnjenkins (changes requested):
looks great - tysm!

Few little bits. More generally, I noticed you removed some of the (now outdated)  comments describing the branching variations etc. Can you make sure to replace with updated ones, seeing as the branching logic - if anything - is slightly more opaque / subtle now. Nothing too long - just simple pointers 🙏

src/compiler/transformers/test/parse-props.spec.ts
Boulea7: Updated the integration fixture to use `mockCompilerSystem()` with `rootDir: "/"`; its sources, generated declarations, and docs JSON now remain entirely in the in-memory filesystem. The focused transformer/type-library tests pass (58 tests).

src/compiler/transformers/transform-utils.ts
Boulea7: Removed the duplicate `unalias` helper. `type-library.ts` now imports the exported `resolveAliasedSymbol`, and the shared helper has JSDoc. I also restored concise comments for the updated export/import branches.

johnjenkins (approved):

Boulea7:
Windows was reading the docs output through an unnormalized backslash path, so the mock filesystem lookup returned `undefined`. The fixture now normalizes every joined path before it reaches the mock system and restores the original cwd before compiler teardown. The focused regression and full unit suite pass; lint, formatting, and build pass as well.

PR #6836, opened 2026-08-19
fix(runtime): guard against null parent in insertBefore

### LINKED ISSUE #6822: insertBefore crashes with "null is not an object (evaluating 'insertBefore')" when parent is null
### Prerequisites

### Stencil Version

4.43.5 (also present on main/nightly 4.44.0-dev)

### Current Behavior

We see ~20 crashes/week, exclusively on iOS Safari. We strongly suspect Safari's page-translation/Reader feature mutates light-DOM slotted nodes, detaching them so parentNode becomes null during a re-render (same class of problem as Google-Translate-breaks-React). The unguarded access then throws. Suggested fix: if (parent?.insertBefore) { … } (or parent != null && parent.insertBefore), consistent with the already null-safe fallback branch. Repro: No deterministic repro; occurs on shadow: false/scoped components under Safari-side DOM mutation.

### Expected Behavior

 In src/runtime/vdom/vdom-render.ts, the insertBefore helper guards the fallback path against a null parent, but not the insertBefore path:

```js
if (parent.__insertBefore) {                 // ← unguarded, throws if parent === null
  return parent.__insertBefore(newNode, reference);
} else {
  return parent?.insertBefore(newNode, reference);   // ← already null-safe
}
```

### Steps to Reproduce

We see ~20 crashes/week, exclusively on iOS Safari. We strongly suspect Safari's page-translation/Reader feature mutates light-DOM slotted nodes, detaching them so parentNode becomes null during a re-render (same class of problem as Google-Translate-breaks-React). The unguarded access then throws. Suggested fix: if (parent?.insertBefore) { … } (or parent != null && parent.insertBefore), consistent with the already null-safe fallback branch. Repro: No deterministic repro; occurs on shadow: false/scoped components under Safari-side DOM mutation.

### Code Reproduction URL

https://github.com/dso-toolkit/dso-toolkit/

### PR with patch
https://github.com/dso-toolkit/dso-toolkit/pull/3921

### Additional Information

_No response_

Arul1998:
The `insertBefore` helper dereferenced `parent.__insertBefore` without a null check, while the fallback branch directly below already guarded the parent with `parent?.insertBefore`. When a slotted light-DOM node is detached mid-render e.g. iOS Safari's Reader/translation features mutate slotted nodes during re-renders, setting `parentNode` to null the unguarded branch threw "Cannot read properties of null (reading '__insertBefore')", crashing the render.

Add optional chaining to the `__insertBefore` branch so a null parent falls through to the already-guarded fallback, matching the existing pattern in the same function.

Fixes #6822

GitHub Issue Number: #6822

In `src/runtime/vdom/vdom-render.ts`, the `insertBefore` helper ends with two branches:

```ts
if ((parent as d.RenderNode).__insertBefore) {
  return (parent as d.RenderNode).__insertBefore(newNode, reference);
} else {
  return parent?.insertBefore(newNode, reference);
}

### DIFF
--- a/src/runtime/vdom/vdom-render.ts
+++ b/src/runtime/vdom/vdom-render.ts
@@ -971,7 +971,7 @@ export const insertBefore = (
     }
   }
 
-  if ((parent as d.RenderNode)?.__insertBefore) {
+  if (BUILD.slotRelocation && (parent as d.RenderNode)?.__insertBefore) {
     return (parent as d.RenderNode).__insertBefore(newNode, reference) as d.RenderNode;
   } else {
     return parent?.insertBefore(newNode, reference) as d.RenderNode;
--- a/src/runtime/vdom/vdom-render.ts
+++ b/src/runtime/vdom/vdom-render.ts
@@ -971,7 +971,7 @@ export const insertBefore = (
     }
   }
 
-  if ((parent as d.RenderNode).__insertBefore) {
+  if ((parent as d.RenderNode)?.__insertBefore) {
     return (parent as d.RenderNode).__insertBefore(newNode, reference) as d.RenderNode;
   } else {
     return parent?.insertBefore(newNode, reference) as d.RenderNode;
--- a/src/runtime/vdom/test/vdom-render.spec.tsx
+++ b/src/runtime/vdom/test/vdom-render.spec.tsx
@@ -1,5 +1,5 @@
 import { h, newVNode } from '../h';
-import { isSameVnode, patch } from '../vdom-render';
+import { insertBefore, isSameVnode, patch } from '../vdom-render';
 
 describe('template elements', () => {
   it('should append children to template.content, not template directly', () => {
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: src/runtime/vdom/vdom-render.ts [hunk 1]
```diff
@@ -971,7 +971,7 @@ export const insertBefore = (
     }
   }
 
-  if ((parent as d.RenderNode).__insertBefore) {
+  if ((parent as d.RenderNode)?.__insertBefore) {
```
johnjenkins: it's odd that this one tweak has pushed the bundle size by that much.. could probably add
```suggestion
  if (BUILD.slotRelocation && (parent as d.RenderNode)?.__insertBefore) {
```
seeing as `__insertBefore` will only be set for non-shadow, patched components

Arul1998:
Good call applied. Confirmed it's behavior-preserving too: __insertBefore is only assigned via patchPseudoShadowDom for !shadow && hasSlot components, which is exactly what flips slotRelocation on in app-data, so the guard is always true when the property can exist and tree-shakes out otherwise. Runtime suite stays green.

johnjenkins (approved):

Issue #6857, opened 2026-09-02
bug: scoped hydration does not restore selectors inside @layer


wessmeister:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/stenciljs/core/blob/main/CONTRIBUTING.md).
- [x] I agree to follow the [Code of Conduct](https://github.com/stenciljs/core/blob/main/CODE_OF_CONDUCT.md).
- [x] I have searched for [existing issues](https://github.com/stenciljs/core/issues) that already report this problem, without success.

### Stencil Version

4.44.2

### Current Behavior

When a shadow component places selectors inside `@layer` and is server rendered with `serializeShadowRoot: 'scoped'`, Stencil does not preserve the original selectors in the generated scoped CSS.

For a `:host` rule, the server output contains a scoped host class without the original selector annotation:

```css
@layer defaults {
  .sc-layered-box-h { /* declarations */ }
}
```

During client hydration, Stencil creates the shadow root but cannot convert that selector back to `:host`. The scoped class remains in the shadow stylesheet, cannot match the host, and the host styles no longer apply. Selectors for elements inside an `@layer` block are also not scoped.

### Expected Behavior

Stencil should process selectors inside `@layer` in the same way as selectors inside its other supported grouping rules. Scoped server output should retain the original selector annotation so client hydration can restore `:host` in the shadow stylesheet.

### System Info

```shell
System: node 22.15.1
Platform: darwin (25.5.0)
CPU Model: Apple M4 Pro (12 cpus)
Stencil: 4.44.2
TypeScript: 5.8.3
Rollup: 4.44.0
Parse5: 7.2.1
jQuery: 4.0.0-pre
Terser: 5.37.0
Browser: Google Chrome 152
```

### Steps to Reproduce

1. Clone the reproduction repository.
2. Run `npm ci`.
3. Run `npm start`.
4. Open `http://localhost:3342/`.
5. Compare the hydrated component with the reference and inspect the component's shadow stylesheet.

The component loses its 160 by 48 pixel size, flex display, background, and text color after client hydration. The reference retains those styles.

### Code Reproduction URL

https://github.com/wessmeister/stencil-scoped-layer-repro

### Additional Information

`scopeCss` currently descends into `@media`, `@supports`, `@page`, and `@document`, but not `@layer`. Its original selector annotation pass also stops after one level of nested grouping rules. `convertScopedToShadow` relies on those annotations during client hydration.

[...]

PR #6858, opened 2026-09-02
fix(scope-css): process layer selectors

wessmeister:

`scopeCss` does not traverse `@layer` blocks. Selectors for elements inside a layer are therefore left unscoped.

This also affects shadow components rendered with `serializeShadowRoot: 'scoped'`. Their server generated CSS omits the original selector annotation inside `@layer`. During client hydration, `convertScopedToShadow` cannot restore selectors such as `:host`, so the shadow stylesheet retains a scoped class that cannot match the host and its styles stop applying.

GitHub Issue Number: #6857

## What is the new behavior?

`scopeCss` now processes `@layer` alongside its other supported grouping rules. The same predicate is used when scoping selectors and preserving original selector annotations. Annotation also recurses through nested supported grouping rules, so combinations such as `@layer` containing `@media` restore correctly during client hydration.

## Documentation

N/A

- [ ] Yes
- [x] No

## Testing

- Added focused unit coverage for layer scoping and nested original selector annotations.
- Ran the focused `scope-css` test suite, repository lint, and production build.
- Reproduced the failure with Stencil 4.44.2 and verified the locally packed fix in Chrome 152 using the standalone reproduction.

## Other information

Minimal reproduction: https://github.com/wessmeister/stencil-scoped-layer-repro

### DIFF
--- a/src/utils/shadow-css.ts
+++ b/src/utils/shadow-css.ts
@@ -471,6 +471,13 @@ const scopeSelector = (selector: string, scopeSelectorText: string, hostSelector
     .join(', ');
 };
 
+const isScopableAtRule = (selector: string) =>
+  selector.startsWith('@media') ||
+  selector.startsWith('@supports') ||
+  selector.startsWith('@page') ||
+  selector.startsWith('@document') ||
+  selector.startsWith('@layer');
+
 const scopeSelectors = (
   cssText: string,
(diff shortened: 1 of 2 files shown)


johnjenkins (approved):
lgtm - tysm!

PR #6900, opened 2026-09-21
fix(docs): derive CEM readonly from getter/setter instead of mutable

Cliffback:

GitHub Issue Number: N/A

The `docs-custom-elements-manifest` output target sets a field's `readonly` from the prop's `mutable` flag:

```ts
...(!prop.mutable && { readonly: true }),
```

These describe different things. The [CEM schema](https://github.com/webcomponents/custom-elements-manifest/blob/main/schema.d.ts) defines `readonly` as *"Whether the property is read-only."* Stencil's [`mutable`](https://stenciljs.com/docs/properties#prop-mutability-mutable) controls whether a prop *"is immutable from inside the component logic"*, the docs add that *"once a value is set by a user, the component cannot update it internally"*, i.e. a non-mutable prop is one that **only** the consumer writes.

Since props default to `mutable: false`, nearly every prop is emitted as `readonly: true`. On one real component library this is 392 of 424 fields. Tooling reading the manifest concludes those properties cannot be set.

This also contradicts Stencil's other outputs built from the same metadata:

- `dist/types/components.d.ts` declares the same props as writable, no `readonly` modifier anywhere in the file.
- The Angular output target's proxies assign to them directly (`this.el[item] = val`).

Stencil already models a read-only prop as a [`@Prop()` getter with no setter](https://stenciljs.com/docs/properties#prop-getters-and-setters), and already exposes `getter`/`setter` on `JsonDocsProp` for exactly this purpose.

## What is the new behavior?
`readonly` is derived from `getter && !setter`, matching both the CEM schema and Stencil's own definition. Virtual props have `getter: undefined` and are correctly no longer marked read-only.

The existing `prop-cmp` fixture already covers every case:

```tsx
@Prop() first: string;                               // plain       -> writable
@Prop() get fullName() { ... }                       // getter      -> readonly
@Prop() get clothes() { ... } set clothes(v) { ... } // get + set   -> writable
```

All three were previously `readonly: true`. Now only `fullName` is.

## Documentation

None needed — this aligns the output with the already-documented behaviour of `mutable` and of `@Prop()` getters/setters.

- [ ] Yes
- [x] No

A bug fix to a generated docs artifact. The emitted manifest changes, but only to stop reporting writable properties as read-only.

- [ ] Yes
- [x] No

## Testing

- Updated the unit test that asserted `readonly: true` for a non-mutable prop, and added cases for getter-only and getter+setter props.
- Regenerated `test/end-to-end/custom-elements-manifest.json`: `readonly` drops from 26 fields to 3 — `my-cmp.barProp`, `my-jsx-cmp.barProp` and `prop-cmp.fullName`, the only getter-without-setter props in the test app. Note that `prop-cmp.clothes` and `runtime-decorators.decoratedGetterSetterProp` both have explicit setters in the fixture source and were previously reported as read-only.

## Other information

Introduced in #6568 along with the output target. The unit test added there asserted the current behaviour, so this change inverts that expectation.


johnjenkins (approved):
LGTM ❤️

